PR #7, opened 2018-04-10
Feature/wiff sim support

chambm:


chambm:
I've only got one WIFF with SIM data and it's from a user. It's tiny (140kb), so I guess we could ask to use it for unit testing?

bspratt:
Sounds ideal!

PR #14, opened 2018-04-13
Fix IndexOutOfRangeException in GraphChromatogram.DisplayAllIonsSumma…

nickshulman:
…ry if your color scheme has only one color. (#12)

This cherry picks commit #12 from branch Skyline/skyline_4_1

I am still running the unit tests. I will pull this change after I get approval and they pass.


nickshulman:
No, I hadn't passed the test yet for that 4.1 commit, but luckily they ended up passing when I subsequently ran them on machine.
Still running the tests on this commit for trunk now.

PR #19, opened 2018-04-14
Bug/fix generatecode bat

nickshulman:
This fix to "generatecode.bat" needs to get ported to skyline_4_1. It adds a "readme.txt" to the GeneratedCode directory so that the directory exists.

If you switched from master to skyline_4_1, your build won't be broken, but if you started out on skyline_4_1, the build won't work because the directory never got created.


bspratt:
Excellent, thanks - I'll do the merge

PR #50, opened 2018-05-03
Improvements to Bruker PASEF support

chambm:
- added combineIonMobilitySpectra support for Bruker TIMS data; it has special behavior compared to Waters and Agilent IMS data, and on PASEF MS2 data it's even more special:
  - for non-PASEF data, all mobility scans are merged in each frame, but the mobility values are kept as a 3rd binary data array for that merged spectrum; whenever m/z values would overlap between mobility scans, a small jitter (1e-8) is added to the duplicate values so they can be preserved without violating the m/z uniqueness constraint; thus all mobility information is preserved but in a much more compact representation
  - for PASEF data, MS1 frames are merged the same way as above, but MS2 frames are merged only within PASEF precursors (i.e. each precursor gets its own merged scan); MS2 scans that do not belong to a PASEF precursor are dropped entirely
  - merged spectrum ids are like 'merged=1234'; for PASEF data, the constituent ids (frame and scan) are preserved in the scanList element (but not for full frame merges, that would be too verbose)
- added IonMobility field to SpectrumList_TitleMaker
- fixed some bugs in SpectrumList_ScanSummer and gave it the ability to combine peaks (within an individual spectrum) that occur in very close proximity (1e-7); (the intent is that SpectrumList_ScanSummer will be used to prepare PASEF data for identification, so at this point the mobility dimension for fragment peaks is collapsed)
- updated threshold filter to work on all arrays with the same size as defaultArrayLength, not just the two primary arrays (usually m/z and intensity)
- fixed SeeMS issues with showing 'merged=1234' ids, and showing IMS heatmaps for files created with combineIonMobilitySpectra
* added MSData::getArrayByCVID() convenience function for retrieving extra array identified by CVID
* added <scan> attributes to msdata::TextWriter
* removed runtime-debugging=off restriction for using Bruker API
* fixed building outside a git clone or without git available
* added a <location-prefix> for --without-compassxtract builds

Ping @hroest

### DIFF
--- a/pwiz/data/vendor_readers/Bruker/SpectrumList_Bruker.cpp
+++ b/pwiz/data/vendor_readers/Bruker/SpectrumList_Bruker.cpp
@@ -277,15 +277,18 @@ PWIZ_API_DECL SpectrumPtr SpectrumList_Bruker::spectrum(size_t index, DetailLeve
             for (auto itr = ++scanNumbers.begin(); itr != scanNumbers.end(); ++itr)
             {
                 scans.push_back(Scan());
+                scans.back().instrumentConfigurationPtr = msd_.run.defaultInstrumentConfigurationPtr;
                 generate(std::back_insert_iterator<std::string>(scans.back().spectrumID),
                          "frame=" << int_ << " scan=" << int_,
                          frameScanPair.first, *itr);
 
-                if (scanTime > 0)
-                    scans.back().set(MS_scan_start_time, scanTime, UO_second);
+                // CONSIDER: do we need this? all scan times will be the same and it's rather verbose
+                //if (scanTime > 0)
+                //    scans.back().set(MS_scan_start_time, scanTime, UO_second);
 
-                if (scanRange.first > 0 && scanRange.second > 0)
-                    scans.back().scanWindows.push_back(ScanWindow(scanRange.first, scanRange.second, MS_m_z));
+                // CONSIDER: do we need this? all scan ranges will be the same and it's very verbose!
+                //if (scanRange.first > 0 && scanRange.second > 0)
+                //    scans.back().scanWindows.push_back(ScanWindow(scanRange.first, scanRange.second, MS_m_z));
             }
             result->scanList.set(MS_sum_of_spectra);
         }


chambm:
For MGFs: added IonMobility field to SpectrumList_TitleMaker
Jamroot change: fixed building outside a git clone or without git available

I forgot about the PASEF determination from Skyline. I'll hack it in there.

bspratt:
I was naively expecting to see a literal "1/k0=" string the in the code but I suppose it makes more sense to do it more generally - though it will need to declare units (or perhaps I'm missing something and it already does). It looks like this will get us into trouble if a 3rd type of ion mobility ever is introduced.

In any event it looks like I'll need to go back into BlibBuild to handle whatever we come up with as a 4th variant for Mascot results parsing.

chambm:
TitleMaker allows maximum flexibility. The user can make the format whatever they want. Including:
`--filter "titleMaker 1/K0=<IonMobility>"`

I think it's fair to leave it up to the user to put the correct quantity type there although there's nothing difficult about adding extra fields like `<IonMobilityQuantity>` and `<IonMobilityUnits>`. Is it necessary? The MSData interface will provide units if you're reading straight from the raw data for MS1 quant.

bspratt:
I see - that should suffice. The units do matter in the context of BiblioSpec, but they're easily inferred from suitable formatted data as with your example --filter "titleMaker 1/K0=<IonMobility>".

chambm:
From Brendan email:
> Doesn’t that seem a little brittle to make this depend on users getting a format string right?
> 
> For our users, I would prefer to have them just use MSConvertGUI and choose MGF and have that work.
> 
> I think you are imagining command line users.
The idea of "prepare this PASEF data for search by Mascot", which includes merging across retention time (using the scanSummer filter), is not something that should just happen by default by selecting MGF. That's too much magic for a generic conversion tool's default behavior. But I could see that kind of capability being baked into MSConvertGUI "presets", which would set the filters and other options in a certain way. But yeah, currently the GUI doesn't have combineIonMobilitySpectra, scanSummer, or a customizable titleMaker...it only uses it for the 'TPP compatibility' mode. So I'll need to add those at the very least for this to work from the GUI.

chambm:
From Brian email:
> I agree that some kind of reasonable default for including available ion mobility in the MGF  in the absence of a titleMaker filter would be highly desirable. That's what I was initially imaging, which is why I was confused.

It's certainly do-able but doesn't solve the more significant issue with PASEF of merging the MS2s properly. But I'm also pretty sure we could ask 10 people what they think the default MGF title should be and get 10 different answers. This is why we have standards...

brendanx67:
Right and why in the absence of standard we make software popular enough to
set the standard or follow the lead of other software, like the Bruker tool.

pwiz_tools/SeeMS/Dialogs/HeatmapForm.cs
```diff
@@ -327,8 +360,24 @@ protected override void OnShown(EventArgs e)
 
                 double scanTime = (double) dgv[scanTimeColumn.Index, i].Value;
                 double ionMobility = (double) dgv[ionMobilityColumn.Index, i].Value;
+                if (ionMobility == 0)
```
chambm: @bspratt At this point, when ionMobility == 0, I check whether the spectrum has a mobility array. If it does, then I add all the mobility bins from that array. For an MS1, all heatmap data comes from a single spectrum with 3 arrays rather than 1000 spectra with 2 arrays. The number of data points is that same but the performance is much better with 1 spectrum due to data locality.

pwiz_tools/SeeMS/Dialogs/HeatmapForm.cs
```diff
@@ -250,22 +250,53 @@ private void ShowHeatmap(double scanTime, int msLevel)
...
                 {
-                    double mz = points[j].X;
-                    double intensity = points[j].Y;
-                    bounds.MinX = Math.Min(bounds.MinX, mz);
-                    bounds.MinY = Math.Min(bounds.MinY, bin.IonMobility);
-                    bounds.MaxX = Math.Max(bounds.MaxX, mz);
-                    bounds.MaxY = Math.Max(bounds.MaxY, bin.IonMobility);
-
-                    heatmapPoints.Add(new Point3D(mz, bin.IonMobility, intensity));
+                    var mobilityArray = mobilityBDA.data;
```
chambm: @bspratt Here's where I am adding each mz/mobility/intensity point to the heatmap. The heatmap has to know all the mobility bins in both cases, but with the combined representation, it populates them all from a single spectrum (for MS1; for PASEF MS2, there will still be multiple spectra, but far fewer).

chambm:
Agreed. I've got a request in for a CV change to add parent terms for ion mobility concepts so I can just look for a child of a single term. Also it'll allow putting those terms in mzIdentML files, so search engines can start putting it in structured output rather than title strings. The CV change seems unopposed so should be available soon. Title strings give me the heebie jeebies, but they're certainly a necessary evil right now because it's currently necessary from pepXML/mzIdentML to go back to the raw data (e.g. mzML) to get the ion mobility value.

chambm:
@bspratt Did your ion mobility changes cause the Skyline test failures?

bspratt:
Oh, OK. At least I know what I'm looking for, and I'm sure that it's all
related, yes.

bspratt:
I'm seeing some test failures with non-PASEF Bruker data, we're not there
yet...

chambm:
Are those failing tests not part of the Skyline's automated tests?

bspratt:
Yes, findAbbreviated seems to be the issue. Not getting from "x.y" to
"frame=x scan=y"

chambm:
Can you point me at the test? I'm not able to reproduce it in the vendor reader unit test.

bspratt:
Sure, it's MeasuredInverseK0ValuesPerfTest()

bspratt:
I guess I should say that all I know is that the call to findAbbreviated is
failing. I don't know that it needs to get to frame= etc, could even be
that the problem is that it DOES get to frame= etc...

bspratt:
I wonder if it's not just that non-PASEF TIMS data handling got dinged in
the MGF combination work.

bspratt:
Rebuilding now. I believe it's failing with actual TDF data.

bspratt:
Same effect, with an actual .d file.

>>  I'm wondering if it'll be fixed by the bug fix for non-combined ion
mobility.

I kind of think so, yes. Can you make that happen?

chambm:
```
> filepath
"D:\\test\\Bruker\\tims\\BSA_50fmol_TIMS_InfusionESI_10prec.d\\analysis.tdf"
> var msd = new MSDataFile(filepath);
> var sl = msd.run.spectrumList;
> sl.size()
464013
> sl.spectrumIdentity(300000)
{pwiz.CLI.msdata.SpectrumIdentity}
    base_: 0x0000000027d87440
    id: "frame=306 scan=796"
    index: 300000
    owner_: {pwiz.CLI.msdata.SpectrumList}
    sourceFilePosition: 18446744073709551615
    spotID: ""
> sl.findAbbreviated("306.796")
300000
```
How do I run your failing test myself?

bspratt:
You'll need to build SkylineTester, and run it from there.

My guess is that it's not about findAbbreviated per se, but about having
put off some non-PASEF TDF features while working on the MGF combination.

That's work we're going to need done anyway, your time might be better
spent there than in figuring out our rather exotic test mechanism.

chambm:
Any ideas for how I figure out what aspect of non-PASEF features is the culprit? Non-PASEF functionality  should be unaffected by the changes in this PR. The TDF file for the unit test had an out of date mzML (i.e. TDF now has a dedicated nativeID CV term, and using inverse reduced mobility CV term instead of drift time), but it's mostly the same as the mzML I generated today. It has the same number of scans and format of the native ID.

bspratt:
I'll do a pwiz debug build and see what I can figure out.

bspratt:
I'll keep chasing this down, but I don't think its a problem in the reader
- rather its with the way Skyline calls the reader using preferOnlyMsLevel.

chambm:
That's the missing detail that I didn't test with. I'm sure it's the prefer
parameter that is mucking up the frame scan pair to index math. I will fix
that tomorrow.

chambm:
OK, pull the latest changes. The find() issues should be fixed. I also updated VendorReaderTestHarness to allow testing with non-default Reader configs, so this kind of thing will be caught in the core tests in the future. I may try this technique for other vendors to see if we can expand the test coverage for them.

I've also added a PASEF file to the unit tests by trimming down a large tdf and tdf_bin. I edited the TDF SQLite file to delete frames with id > 6 (and references to them), set the number of scans in each frame to smaller number (i.e. from 985 to 350 for ms2, and 100 for ms1), then used Process Monitor to see what file offsets were being accessed when enumerating the full file. I was able to simply truncate the file after the last accessed offset (plus 100kb of buffer after the last offset).

bspratt:
>>  I've also added a PASEF file to the unit tests by trimming down a large
tdf and tdf_bin.

Impressive!

bspratt:
Not there yet - that Skyline perftest still fails, albeit in a slightly
different way. But clearly something to do with not indexing the MS2.
Investigating....

chambm:
ARGH! Of course. Because the ms1 filtered file only has 1 frame! :scream:

bspratt:
Sounds like you have a fix in mind?

chambm:
Hmm, false alarm. I can't reproduce the find() bug anymore. Either on the full HeLa file or the trimmed one (or a newer trimmed one where I included 17 frames instead of 6). Can you give more detail about the error? There are some other issues I need to take care of with the unit test, but those are related to mz5 and out of date test mzMLs.

bspratt:
It comes down to an index mismatch - Skyline asks for MS2 to not be
indexed, then asks for a frame.scan pair that the reader hasn't indexed
because its MS2, even though Skyline thinks its asking for MS1 data.

So I have to work out whether the problem is with the frame number sense of
the asker or the askee.

chambm:
Has Skyline cached the spectrum index somewhere rather than the abbreviated id? Those indexes certainly won't be the same with/without preferOnlyMsLevel set.

bspratt:
It's something like that. I'm running master alongside our branch, in both
cases the frame.scan value are the same so its not a Skyline cacheing issue
per se.

The trouble seems to be that MsDataFileImpl::GetSpectrumIndex(string id)
returns a different value depending on whether MS2 prefiltering is on, and
MsDataFileImpl:: GetSpectrum(int spectrumIndex) seems not to be dealing
what that difference. So probably one or the other is dealing or failing to
deal with the gaps in the indexing?

chambm:
> The trouble seems to be that MsDataFileImpl::GetSpectrumIndex(string id)
> returns a different value depending on whether MS2 prefiltering is on

That's correct. `SpectrumList.find(SpectrumList.spectrumIdentity(SpectrumList.size()-1).id)` will return a different number depending on how `preferOnlyMsLevel` is set, and it should always be less than `SpectrumList.size()` (which would mean the id wasn't found). 

> and MsDataFileImpl:: GetSpectrum(int spectrumIndex) seems not to be dealing
> what that difference. 

This is what I don't understand and haven't been able to reproduce since the find() fix. Is there a SpectrumList wrapper in play here?

So probably one or the other is dealing or failing to deal with the gaps in the indexing?

bspratt:
If you want to have a run at it, in Skyline.sln debug TestRunner with these
arguments::
status=on offscreen=False loop=1  perftests=on
test=MeasuredInverseK0ValuesPerfTest

bspratt:
You know, I'm pretty sure it's Skyline screwing up. It really looks like
its managing to ask for an MS2 scan after promising not to to that.  Ball
is in my court.

bspratt:
Rerunning all my tests to be sure (including perf tests), but I'm pretty
confident that what I just pushed puts us in position to get this work into
trunk.

I'm still fooling around with some performance optimizations here, but what
we have is better than what we had.

PR #59, opened 2018-05-15
Skyline41/localization/zh chs final

brendanx67:
Just integrating Chinese localization changes from reviewers


brendanx67:
Not trying to merge this into master.

PR #65, opened 2018-05-18
Skyline/work/20180515 unifi support

nickshulman:


chambm:
Next builds on TeamCity should have UNIFI_PASSWORD set (securely, so it doesn't show up in log).

PR #73, opened 2018-05-22
Speed optimizations for non-combined scan PASEF access

bspratt:
Speed optimizations for non-combined scan PASEF access, mostly in the form of grabbing an entire frame at once, and doing the index to mz conversion at one go to take advantage of threadedness inside the Bruker DLL

Passes all tests, including Skyline perf tests.


chambm:
One drawback to squash merging is that the original commits aren't technically merged, so that's why they all show up here even though they're already in master. So you have to use a fresh branch. :(

bspratt:
Will do.

bspratt:
shall we delete this branch?

PR #82, opened 2018-05-29
Skyline/work/20180529 move to net472

nickshulman:


chambm:
I'm working on this PR as the main integration point for upgrade to VS 2017 and 4.7.2.

chambm:
@bspratt @nickshulman 
This change to SkylineTester needs review. It seems to work fine for me in both bjam and running it as a GUI but I don't know what other contexts it's used in:
https://github.com/ProteoWizard/pwiz/pull/82/commits/467d14fba98f5674a6a9c7d67622dcc20f1c15dd

It certainly works better with bjam. The output doesn't get lost this way, although it mysteriously shows up before the msbuild output instead of between the `echo` statements like I would expect.

chambm:
All 2017 tests passed. Ready to merge? At that point I'll pause the 2017 configs and reconfigure the old configs to use 2017. The existing branches based on older masters will need to be updated before they'll build on the new configs.

bspratt:
Give me another day or two to convince myself that there are no performance
hits. I configured several more machines to build and test VS2013 (master)
then VS2017 (integration). Currently I have a few machines that do this but
run the perf tests on VS2013, so it's not quite apples to apples.

So far what I'm seeing were I do have apples to apples is that some tests
are faster, some are slower, total tests run is slightly higher with the
old build.

PR #83, opened 2018-05-29
Skyline: fix DocumentGrid "PrecursorNeutralFormula" to include isotop…

bspratt:
…ic labels, and fix ""PrecursorIonFormula" to include iostopic labels and ionization. Both were simply displaying the molecule neutral formula, a holdover from when that *was* the ion formula. Also, fix a problem with Library Match containing code that assumed peptides and thus wasn't working for small molecules.


pwiz_tools/Skyline/Model/Databinding/Entities/Precursor.cs
```diff
@@ -141,26 +141,41 @@ public string IonName
...
+        private void GetPrecursorFormulaAndAdduct(out Adduct adduct, out string formula)
+        {
+            if (IsSmallMolecule())
+            {
+                formula = (DocNode.CustomMolecule.Formula ?? string.Empty);
+                adduct = DocNode.PrecursorAdduct;
+            }
+            else
+            {
+                PeptideDocNode parent = DataSchema.Document.FindNode(IdentityPath.Parent) as PeptideDocNode;
```
nickshulman: You can just use "Peptide.DocNode" here.

PR #89, opened 2018-05-31
Skyline - Audit logging for settings

Tobias-Rohde:
- Adds audit logging for Peptide settings, Transition settings and Document settings
- If the users makes a settings change, the old and new SrmSettings are compared using reflection.
  Only properties with the Diff or DiffParent attribute are compared.
- This makes adding a new setting fairly easy, one only has to add the Diff/DiffParent attribute to the property corresponding to the
  setting in the UI and add the name and containing class of the property to PropertyName(s).resx (format: Class_Property)
- For now the audit logging form is hidden, however the "undo-redo level" messages are seen
  in the undo redo combo box. (Therefore replacing the generic "Settings changed" messages)
- Fixed exception in FoldChangeVolcanoPlot and small bug in


pwiz_tools/Shared/Common/SystemUtil/DiffAttribute.cs
```diff
@@ -0,0 +1,66 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+
+namespace pwiz.Common.SystemUtil
+{
+    public interface IAuditLogObject
```
nickshulman: Can we come up with a name for this interface which is more descriptive of what it does, rather than where it's being used?
Yes, the audit log needs to know how to describe things, but do we want to call the interface "IAuditLogObject", or should it be maybe "IDescribable".

pwiz_tools/Shared/Common/SystemUtil/DiffAttribute.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        string AuditLogText { get; }
+        bool IsName { get; }
```
nickshulman: What is this method "IsName" supposed to do? Can we add a comment?

pwiz_tools/Shared/Common/SystemUtil/DiffAttribute.cs
```diff
@@ -0,0 +1,66 @@
...
+        public bool IsTab { get; protected set; }
+        public bool IgnoreName { get; protected set; }
+
+        public virtual bool DiffProperties { get { return false; } }
+
+        public Type CustomLocalizer;
+    }
+
+    [AttributeUsage(AttributeTargets.Property)]
+    public class DiffAttribute : DiffAttributeBase
```
nickshulman: I feel like "Diff" is not right name here.
This attribute appears over properties that we want to appear in the audit log.
Should we maybe call this "Tracked"? And maybe the one below called "DiffParent" should actually be called "TrackChildren"?

pwiz_tools/Skyline/Model/AuditLog/LogMessage.cs
```diff
@@ -0,0 +1,267 @@
...
+                ? reader.ReadElementString()
+                : string.Empty;
+
+            while (reader.IsStartElement(EL.name))
+                names.Add(reader.ReadElementString());
+
+            if (names.Count != GetExpectedNameCount(Type))
+                throw new XmlException();
+
+            Names = ImmutableList<string>.ValueOf(names);
```
nickshulman: Just use ImmutableList.ValueOf(names).
You do not need the <string>.

pwiz_tools/Skyline/Model/AuditLog/Property.cs
```diff
@@ -0,0 +1,88 @@
...
+                name = parentObject.GetType().Name + '_' + name;
+
+            if (_diffAttribute.CustomLocalizer != null)
+            {
+                var localizer = CustomPropertyLocalizer.CreateInstance(_diffAttribute.CustomLocalizer);
+                if (localizer.Relative || rootObject != null)
+                    name = localizer.Localize(rootObject, parentObject) ?? name;
+            }
+
+            return string.Format("{{0:{0}}}", name); // Not L10N
```
nickshulman: What does this format string mean? I am thinking that:
return "{0:" + name + "}";
would be easier to understand.

pwiz_tools/Shared/ProteowizardWrapper/ProteowizardWrapper.csproj
```diff
@@ -112,6 +112,10 @@
     </NativeReference>
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\Common\Common.csproj">
```
nickshulman: I do not think that this reference needs to be here since "DiffAttribute" is in the "CommonUtil" project not "Common".

PR #92, opened 2018-06-01
Skyline: clean up some language in the UI around spectral library ion…

bspratt:
… mobility usage - some was still in terms of drift time rather than a generalized IM sense


brendanx67:
Okay. Sadly, I just updated those long tips in Japanese. Hopefully, they won't have too much trouble making the change from drift time to ion mobility in 4.2.

PR #94, opened 2018-06-07
Migrate to VS2017 (msvc-14.1) and .NET 4.7.2

chambm:
* updated Boost to 1.67
* updated Boost-Build from latest master (this includes some warnings about coming lexer changes; the new lexer isn't turned on yet but it will be more lenient about whitespace in Jamfiles, e.g. requiring a space before semicolons)
* updated most solutions and projects to VS 2017 and .NET 4.7.2
* removed Boost.Filesystem jam logic about filesystem versions: v2 isn't in the code anymore
* removed MSVC version restrictions from Waters reader
* fixed some data dependencies copied around with common.copy (which seem to have broken somehow with Boost.Build changes) by using assembly-dependency instead, but this didn't work for doctests, so I moved Embedder's doctest into EmbedderTest
* removed `using namespace std;` from BlibFilter and BlibMaker, and fully specified the `boost::log::aux::snprintf()` calls to disambiguate them from `::snprintf()` (which comes from including stdio.h in VC 2017 apparently, but not earlier VC)
* updated Agilent test tarball due to forcing fixed point Q1/Q3 masses

Probably won't build on TC yet. I doubt it has 4.7.2. But it should be able to build Skyline on the "Express" agent (which is really VS 2017 Community, not Express).


chambm:
Also, the VS2015 DLLs for Mascot Parser seem to work just fine for VS2017. But the 2013 DLLs don't work for either 2015 or 2017 - at least in release mode, haven't tested debug yet. So we seem to be mostly in the clear there.

bspratt:
Hi Matt,

Back from ASMS - I won't be able to properly test this until I can get our IT group to install 2017 for me.

So this will break the ability to build with VS2013? Not sure that's a bad thing, but certainly import to understand up front.

Thanks

Brian

chambm:
Can't you install the Community edition? That's what's on my machine and on the agent.

Almost everything will still build with 2013, but our upgrade pattern has always been that the .NET projects are only officially supported on our current VS version. Amusingly, before I added build configurations to test with VS2017, I noticed that the VS2013 configs were quite happy to build the .NET projects even though the project and solution files have been upgraded to 2017.

bspratt:
Yeah, but I have a vague dread of later trying to upgrade from that to Enterprise. I'll see if I can get IT to do me an install today.

PR #96, opened 2018-06-12
Change ms level filter for non-MS spectra

chambm:
- changed ms level filter so that non-MS spectra (e.g. EMR spectra) are considered ms level 0, so they can be included/excluded by the filter the same way other levels can


chambm:
If you're still at work when the required tests have passed, go ahead and merge.

bspratt:
Will do.  Did you see how fast I responded? Even I'm impressed.

PR #98, opened 2018-06-15
Fix problem where a mixed polarity .raw file can result in an error a…

nickshulman:
…bout the number of times not being equal. ("Times xxx and intensities xxx disagree in point count").


nickshulman:
Here's a fix for a bug that I introduced a few months ago.
We assume that so long as no scheduling is going on, all MS1 chromatograms will have the same number of points in them.
However, if some of your MS1 scans are positive, and some are negative, then not all chromatograms will have the same number of points (at least, ever since my change back in February to ignore MS1 scans with the same polarity).
I think the easiest way to fix this is to have two different ChromCollector's for positive and negative scans.

nickshulman:
Oops. Did not mean to close this pull request. I misunderstood what the "Close and comment" button did.

bspratt:
Looks like a simple enough change - no such polarity concerns in SIM?

brendanx67:
SIM is in theory more targeted MS1. Let's wait for someone to show us a data set where they target MS1 to the same range with alternating polarities. Or will positive-negative switching DIA appear before that?

PR #108, opened 2018-06-25
Skyline: Don't attempt Uniprot lookup of proteins for which we don't …

bspratt:
…have sequence - too much potential ambiguity (also, it's slow since we go one by one when all proteins have the same length - 0 in this case)


bspratt:
This passes all tests locally, feel free to merge then delete the branch if you're happy with it.

PR #114, opened 2018-06-27
Fix OverflowException when importing large files from commandline.

nickshulman:
This is the simplest fix for the problem of the Spill File getting too large, causing an OverflowException when we try to read it into memory.
If we think that the MaxRetentionTime is 0, then we just use the number 30 instead.

I am going to try to track down the actual reason that MaxRetentionTime is 0. It has something to do with \_allChromData being null at the time that the Spectra constructor gets called.


brendanx67:
Okay. A bit strange, but I guess okay. It will be interesting to know what the difference is between command-line and the UI.

brendanx67:
If you think about DIA runs, for example a standard SCIEX 64-variable window SWATH isolation scheme, that will have 65 spectra per cycle, and I think this code would overestimate the expected number of points by 65-fold, right? That is pretty huge, 65-fold.

nickshulman:
We should probably be keeping track of two numbers: the maximum size that
the chromatogram could be, and its most likely size.
Then, we will be safe if we make sure to keep the maximum size less than
2GB, but we can tweak the limit on the most likely size any way that we
think might improve performance.

The code looks like it's trying to make the spill file 200MB, but I don't
know how important it is for performance to be close to that number.
This is controlled by the constant:
        private const int MAX_SPILL_FILE_SIZE = 200 * 1024 * 1024;
That seems like a large amount of data to want to read into a byte array
all at once.
-- Nick

brendanx67:
Perhaps the code should simply be changed to use BlockedArray? That would keep it from getting the array overflow error, no matter how large the file ends up being. Though, certainly the point is to try to keep the amount of data Skyline needs in memory at one time to a reasonable size. 200 MB sounds great from that perspective. But, maybe it is just a good idea to code it so that it can never blow up, regardless of how much ends up in each file. We do eventually need to get each file into memory to get the chromatograms written to it using in-memory random access.

nickshulman:
This pull request seems to be all messed up and is not updating when I submit new commits. I am going to close this pull request and make a new one.

PR #116, opened 2018-06-28
Skyline - volcano plot formatting threading issue fix

Tobias-Rohde:
- fixed issue that caused the create match expression dialog to display peptides/proteins matching a previous filter when rapidly changing the filtering options (this caused recent TestVolcanoPlotFormatting test failures)


brendanx67:
Looks okay. Maybe a bit of description on which issue this fixes would help.

PR #132, opened 2018-07-14
Skyline/work/20180713 output parser fixes

nickshulman:
I was hitting a NullReferenceException if I had a redundant "using" directive on the very first line of my source file.

Also, currently for unlocalized strings, we just output the file without any line number information. I think we should output the line number too.


yboss:
Changes seem fine thanks for taking care of this Nick will get a new L10N plugin soon as well

PR #146, opened 2018-08-01
SkylineNightly: temporarily back out the recently added reporting of …

bspratt:
…handle leaks until we can get the test server updated to deal with that. In anticipation of that, change the SkylineNightly automation scheme such that TaskScheduler now kicks off SkylineNightlyShim.exe, which pulls the latest SkylineNightly.exe from TeamCity and runs that. So, hopefuly in future no more manually updating each of our fleet of test machines any time SkylineNightly behavior needs to change.

### DIFF
--- a/pwiz_tools/Skyline/SkylineNightlyShim/Program.cs
+++ b/pwiz_tools/Skyline/SkylineNightlyShim/Program.cs
@@ -23,18 +23,13 @@
 // Accepts same argmuments as SkylineNightly, but first updates local SkylineNightly.exe from GitHub artifacts before invoking it
 // 
 
+// ReSharper disable NonLocalizedString
+
 using System;
 using System.Diagnostics;
-using System.Drawing;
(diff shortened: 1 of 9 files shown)


pwiz_tools/Skyline/SkylineNightlyShim/Program.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Linq;
+using System.Net;
+using System.Reflection;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Xml.Linq;
+using Ionic.Zip;
+
+namespace SkylineNightlyShim
```
nickshulman: You should add:
        // ReSharper disable NonLocalizedString

bspratt: Thanks - I'd left some extra "using" statements in there as well

PR #150, opened 2018-08-08
New "modifications.xml" file from MaxQuant 1.6.2.10

nickshulman:
Also, when looking for "modifications.xml" alongside msms.txt file, also look for "modifications.local.xml", and merge the two together.


bspratt:
:shipit:

PR #153, opened 2018-08-13
Skyline: support the use of "D" and "T" (Deuterium, Tritium) in molec…

bspratt:
…ular formulas and adduct descriptions, as these are in common parlance for small molecules


bspratt:
This isn't as ready as I thought - I tabled it some months ago, but another request came up so I dusted it off thinking it was good to go. Clearly not the case, a few tests are failing. Closing the PR for now.

PR #156, opened 2018-08-15
1. Add IOModuleQTFL.dll to Skyline.csproj so that it gets included in…

nickshulman:
… Skyline's Installer.

2. Upgrade Skyline installer project to use Wix Toolset version 3.11.1.2318
3. Delete BootstrapInstaller project (which was responsible for installing .Net framework etc) since it was too hard to upgrade to the new Wix and Visual Studio, and since we never use it.


chambm:
Er, there are quite a few new DLLs for Shimadzu QTOF. Both pure .NET and platform specific. Are those being covered by this?

PR #157, opened 2018-08-16
Skyline/work/20180815 runsettings

nickshulman:
Matt,
I am not sure whether you were planning on checking in the runsettings files that you gave me, or whether I should.
This change deletes the testsettings files and adds the runsettings file.
One thing I have not been able to figure out is how to make it so that "TestSettings_x64.runsettings" is selected by default. I have no idea how "TestSettings_x64.testsettings" used to get selected by default. The only place that name appears that I can find is in "Skyline.sln", but when I replace it with "TestSettings_x64.runsettings", it does not cause this new file to be selected by default.

It's probably not that important to figure out how to set the default. As it is, even though Visual Studio usually picks TestSettings_x64.testsettings to start with, it tends to forget that setting after a while and I need to choose it again.


brendanx67:
Let's get this merged.

PR #165, opened 2018-08-20
Add sampled noise arrays from Thermo RAW files

rfellers:
This PR follows up on an email conversation I had with @chambm over a year ago about adding noise support for Thermo RAW files. Specifically, I'm populating 3 binary arrays with "sampled" noise arrays, which are substantially shorter than the m/z and intensity arrays. These arrays are consistent with the way RAW files store noise and are useful for spectral visualization and other downstream processing.

I have done some initial performance comparisons with and without the modification:

| File   |      RAW Size      |  mzML Size  | mzML Size w/ Noise | % Increase |
|----------|-------------------:|--------------:|-----------------------:|------------:|
| Internal MS/MS standard | 242MB | 290MB | 292MB | 0.7 |
| PT4530-1 [from here](https://massive.ucsd.edu/ProteoSAFe/dataset.jsp?task=d1d22e43000143118a26ff255439aa10) |    2.55GB  |   2.36GB | 2.58GB | 9.3 |

| File                                    |     Benchmark (avg. 3 runs)*  | Benchmark w/ Noise (avg. 3 runs)* | % Increase |
|-----------------------------|-------------------------------------:|-----------------------------------:|-:|
| Internal MS/MS standard | 9.8 | 13.6 | 38.8 |
| PT4530-1 [from here](https://massive.ucsd.edu/ProteoSAFe/dataset.jsp?task=d1d22e43000143118a26ff255439aa10) |  97.4  |   116.8 | 19.9 |

\* Real time in seconds from msbenchmark.

I ran the following commands to generate the data above:
`msconvert.exe .\file.raw -n`
`msbenchmark.exe spectra full-data .\file.raw`

What are your thoughts given this outcome? Thanks!


pwiz/data/msdata/MSData.hpp
```diff
@@ -439,6 +439,13 @@ struct PWIZ_API_DECL MZIntensityPair
 
 PWIZ_API_DECL std::ostream& operator<<(std::ostream& os, const MZIntensityPair& mzi);
 
+struct PWIZ_API_DECL NoiseDataInfo
+{
+    double mz;
+    float baseline;
+    float intensity;
+};
+
```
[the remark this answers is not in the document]
rfellers: I think I understand what you're saying. I was just copying the example of MZIntensityPair. Are you suggesting that I use 3 specific vectors (noise m/z, noise intensity, noise baseline) instead? Perhaps it would be even better if SpectrumList_Thermo.cpp called a generic "setBinaryArray" method on MSData 3 times instead. If other readers ever want to respect the --addNoiseData flag, they may want to set a different number or types of binary arrays for noise.

rfellers:
Thanks for your comments. No worries on updating the 64 bit support for RAW files to use RawFileReader. Actually, we've been using that internally for years, so I'm even more familiar with the API. When the time comes I'll merge your changes and fix up RawFile.cpp for noise handling.

I've added a new --addNoiseData flag for msconvert in my latest commit.

pwiz/data/msdata/MSData.hpp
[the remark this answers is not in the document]
chambm: I think a setBinaryDataArray is a good idea. Also swapBinaryDataArray. And yes, do that 3 times with 3 `vector<double>`s.

pwiz/data/msdata/MSData.hpp
```diff
@@ -560,6 +567,9 @@ struct PWIZ_API_DECL Spectrum : public SpectrumIdentity, public ParamContainer
     /// set binary data arrays 
     void setMZIntensityPairs(const MZIntensityPair* input, size_t size, CVID intensityUnits);
 
+    /// set sampled noise data arrays
+    void setNoiseInfo(const NoiseDataInfo* input, size_t size);
+
```
chambm: With set/swapBinaryDataArray you can get rid of setNoiseInfo. It'll be easy enough to do all the calls from SpectrumList_Thermo.

rfellers:
I've got this working for 64 bit, but I still need to test for 32 bit (didn't have MSFileReader installed on my development box). I'll test it tomorrow on another VM to make sure 32 bit is OK.

rfellers:
I was having trouble building x86 in early December, but now things are working for me locally. I've built msconvert in x86 and x64 and tested that it creates noise arrays when --addNoiseData is passed. A couple things:

- The Core Windows x86 CI build failed. Things work for me locally and I don't see much in the output, maybe the problem is
`[09:19:43][Step 1/1] ...failed common.mkdir Z:\pwiz\build-nt-x86\pwiz\data\vendor_readers\ABI\Reader_ABI_Test.test\msvc-14.1\rls\async-excpt-on\lnk-sttc\thrd-mlt...`
- I haven't added any unit tests. If you'd like that, could you point me in the right direction? I'd guess `Reader_Thermo_Test.cpp`, but it looks like I'd have to modify more to actually test noise reading

brendanx67:
Can we wrap this up? Either merge or close the pull request. At least it is on a branch if we want to come back to it, but it is our oldest pull request.

rfellers:
Apologies all. I got stuck on the unit test, put it down for a bit, and then lost track of it. Should we simply merge without a test to get this cleared? Or I'm glad to give the unit test one more go. Either way, I'm back up to date with your master branch.

brendanx67:
In my opinion, we should not merge without a test. Rather we should drop it from the pull request list and raise it as a PR once it is ready to go. If you can get it ready to Matt's satisfaction on this next push, then we can merge it. Otherwise, it can stay on a branch, but not as a PR.

pwiz_aux/msrc/utility/vendor_api/thermo/RawFile.cpp
```diff
@@ -1014,6 +1015,54 @@ MassListPtr RawFileImpl::getMassListFromLabelData(long scanNumber)
...
+        _variant_t noisePacket2(varNoisePacket, false);
+        long size = (long)noisePacket2.parray->rgsabound[0].cElements;
+        result->mzArray.resize(size);
+        result->baselineArray.resize(size);
+        result->intensityArray.resize(size);
+        double* pdval = (double*)noisePacket2.parray->pvData;
+        for (long i = 0; i < size; ++i)
+        {
+            result->mzArray[i] = (double)pdval[(i * 3) + 0];
+            result->intensityArray[i] = (float)pdval[(i * 3) + 1];
```
chambm: Since you're casting here you might as well cast to `double` (and change NoiseData accordingly) because that's what the BinaryDataArrays are so it'll make swapping them with `swapBinaryDataArray()` possible.

pwiz_tools/commandline/msconvert.cpp
```diff
@@ -71,6 +71,7 @@ struct Config : public Reader::Config
     {
         simAsSpectra = false;
         srmAsSpectra = false;
+        addNoiseData = false;
```
chambm: Not needed. You're writing directly into config.addNoiseData

chambm:
Do you have a small FT file to make a unit test with? If so I'll take a crack at adding the test myself.

rfellers:
Thanks for offering to create the unit test! I have a standards file [here](https://proteinaceous.blob.core.windows.net/publish/standards.raw) that is FTMS/ITMS MS1 with a Top1 MS2, 95MB. Off hand I can't think of anything smaller. If that is too big I'm happy to make as many single scan RAW files as you see fit.

chambm:
We'd like something less than 2mb. So it can be a few scans or cycles rather than just 1, but not anything like a useful-for-analytical-chemists length. If you could create such a file that'd be wonderful! If you can put some MSX and SPS in there, even better. :1st_place_medal: The FT should be profile (or at least one of them should be) but the IT can be all centroid.

rfellers:
Sadly, using Thermo's QualBrowser, I can't merge multiple scans into a single file, only export 1 scan at a time. So, the attached ZIP contains 3 RAW files (ITMS MS1, FTMS MS1, FTMS MS2), totaling under 1MB in size. Sorry, no SPS ... don't think I've ever seen a file with that. I included the ITMS because it is an example of a scan that doesn't contain noise data (and therefore shouldn't include anything in the resulting file).

[ThermoRAWScans.zip](https://github.com/ProteoWizard/pwiz/files/3821059/ThermoRAWScans.zip)

chambm:
You can create RAW files with QualBrowser?!

Mind...
blown...

Must check this out.

rfellers:
Yup, right click on bottom panel, Export ->Write to RAW file

image

chambm:
Hmph. Would be a lot more useful if I could create a cycle of scans. But this will certainly work for testing noise extraction and FT centroiding. Thanks for pointing it out!

chambm:
Superseded by #737

PR #166, opened 2018-08-21
Skyline code inspection tools: don't rely on Resharper v8 inspection,…

bspratt:
… it can't support newer .net versions. Note this doesn't actually yet restore our L10N checks, but it does restore other code inspections in our TC config.


bspratt:
Not deleting this branch yet - would still like to restore actual L10N checks if possible

PR #168, opened 2018-08-22
Skyline/work/20180821 fix test settings

nickshulman:
Here's my fix to make it so that we run out of the TestResults/xxx/ directory.

Add back x64 and x86 .testSettings files.
Add Shimadzu .dll's to .testSettings files.
Change PeptideChromDataSets.SortAndLimitPeaks to use Enumerable.OrderBy instead of List.Sort, so that "OptimizeTest" passes regardless of whether we have specified the TargetFrameworkVersion.


chambm:
@nickshulman @bspratt @brendanx67 Recent versions of VS2017 give this warning:
`TestSettings will soon be deprecated for automated unit and functional testing scenarios. It is recommended that you use RunSettings. To learn more, see http://aka.ms/runsettings`

brendanx67:
How well did we document the issues we faced moving to just .runsettings without our .testsettings files? Currently, we do use .runsettings, but it points to the .testsettings file. Is that what you are doing, Matt? I always choose the .runsettings file. Though, I know it relies on the .testsettings file. If we didn't thoroughly document why we think we need the .testsettings file and what we have tried to get rid of it, then we should at least do that now.

PR #169, opened 2018-08-22
Skyline/work/20180820 restore l10 n checks

bspratt:
continued work toward full working ReSharper code inspection under VS2107


bspratt:
Not deleting branch yet, still more work to do on code inspection for VS2017

PR #174, opened 2018-08-24
Skyline/bugfix/20180824

brendanx67:


pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -1089,47 +1087,48 @@ private void OnAuditLogEntryAdded(object sender, AuditLogEntry.AuditLogEntryAdde
...
+            if (!IsRecordAuditLogForTutorials)
+                AssertEx.NoDiff(expected, actual);
+            else
+            {
+                // Copy the just recorded file to the project for comparison or commit
+                File.Copy(recordedFile, projectFile, true);
+                if (!existsInProject)
+                    Assert.Fail("Successfully recorded tutorial audit log");
+                else
+                    AssertEx.NoDiff(expected, actual, "Successfully recorded changed tutorial audit log:");
```
Tobias-Rohde: This means that if someone records the tutorial logs and there is no difference, the tutorial test won't fail and it's much more likely that we push the code with IsRecordAuditLogForTutorials set to true. Maybe we should always fail if IsRecordAuditLogForTutorials is true?

brendanx67: Yeah. I knew that, but the other way makes it much more manual to get everything updated in SkylineTester. But, okay, I will rethink my solution to that to only skip failing in the case of a Quality pass. That will probably allow me to be even more fully-automatic for a complete update and still fail in more common cases that will keep us from committing in record mode.

Note that committing in record mode still fails if the anything changes in my current solution. So, it doesn't mean the tests will pass on bad results. It just means you can end up copying files into your project when the test does fail.

brendanx67: I actually felt that the original implementation was part of the problem that caused you to merge before everything actually passed running under SkylineTester conditions, because it required you to run under single-language conditions to record everything and then failed under multi-language conditions.

Tobias-Rohde: Yes, and it made recording logs for multiple languages much more tedious. This makes me think that there should be a way for us to still run a test in other languages even if it fails in another.

brendanx67: We could make it dependent on Program.SkylineOffscreen. That would allow you to test in record mode, as I have been doing, using the Quality tab, and getting everything recorded in the process, but Skyline would still fail enough to provide a good balance between keeping what happened to you from happening again and keeping people from committing with the flag on.

brendanx67: Or, yeah, a way to control when TestRunner stops running a test (either at first failure) or after running all languages. That would probably be better. I will leave this for you, after I merge this to get nightly tests working again.

Tobias-Rohde: Okay sounds good

PR #175, opened 2018-08-24
Fixed TestTargetedMSMSTutorialAsSmallMolecules

Tobias-Rohde:
Fixed TargetedMSMSTutorialTest when run as small molecules (88 vs 87 transitions failure; The document was converted to small molecules in between pages of the ImportPeptideSearchDlg, which didn't change the internal document copy maintained by the wizard)


brendanx67:
Thanks for looking into this and keeping the test actually testing small molecules.

PR #179, opened 2018-08-28
Skyline: handle TIMS PASEF MS2 data. Also, avoid an array out bounds …

bspratt:
…error in progress window on long loads.


chambm:
I think the behavior of excluding the MS2 scans that aren't part of any precursor should be optional. Perhaps it should be the default behavior, but those other scans are not totally devoid of peaks. For most frames, even most scans that DO have a precursor are devoid of peaks or only have a few. Some readers may also be expect a continuous scale of mobilities. Here's an illustrative example:
image
The blue highlighted cells (scans 611-636) are the scans for the only precursor in this frame. You can see a big intense band in the heatmap. You'd think that would be the isolated scans. But no, the intense scans are from 442-496. 611-636 is down in the dross (look at the peak counts in the blue highlighted scans).

chambm:
But of course we should be including precursor information, when available, in non-combined mode for PASEF MS2s. That was an oversight on my part.

bspratt:
Fair enough, I guess, but it seems kind of weird to emit a file that's
effectively a mix of DIA and DDA. The presence of the PasefFrameMsMsInfo
table seems to imply they want to be DDA (and the only example file I've
seen with this in it actually had "DDA" in its title). I have no idea what
the contents of those other precursor-less scans represent.

bspratt:
Can we go with "this is better than leaving out MS2 altogether as we
currently do" for the moment, and accept the PR?

chambm:
We don't leave out MS2 altogether. Just the precursor information. But no I don't think it's really ok to ditch all those peaks (even if it does seem like mixing DIA and DDA, as you said). There's a call with Bruker in a few minutes. I'll bring it up there.

chambm:
I can handle adding in the no-known-precursor peaks today or tomorrow if you need to move onto something else.

bspratt:
FWIW as currently coded, we do drop MS2 in noncombined mode - that's
checked in as a TODO.

chambm:
No we don't. The TODO refers to precursor metadata. Otherwise how could I have shown you that example from SeeMS (that was a file converted by stock msconvert filtered on scan time so I could actually open the damn file in non-combined mode)

bspratt:
Sounds good, thanks - but ideally Skyline will want a way to signal that it
doesn't want mixed DIA and DDA when it's looking for DDA.

chambm:
OK I added the "DIA" spectra in the proper order (which incidentally fixed the combined mode having the merged precursors in randomish order). I didn't implement an option yet to exclude the "DIA" spectra. What I really want is a fast way to exclude the spectra without peaks. That would cut down about 3/4 of the data, But filtering out "DIA" spectra will also exclude a lot of spectra too. I can add that (probably as an option when creating SpectrumList_Bruker's index: exclude MS2 spectra without precursor info).

But I do need to clarify with Markus/Sven exactly where these "DIA" spectra are coming from. Technically MS2s must have some kind of isolation. Are they basically all-ions (full scan) isolation, or is the isolation just not recorded in the TDF...

bspratt:
So where did we land on this? From what I can tell from those email conversations discarding anything not mentioned in the MS2Frames is the correct action.

chambm:
As far as I'm concerned this can be merged as is. Did you test whether the extra MS2s without precursor info caused trouble for your Skyline code?

bspratt:
I haven't no. But from an efficiency point of view we'd rather not hear about them at all, so this change is preferable.

chambm:
So we can merge it now? If Markus replies that there's no possible use to the pseudo-DIA spectra, then I will update it to make it more efficient.

bspratt:
I think so - but I think I'll run the full battery of Skyline tests against it first (including the ones that don't run on TC). If that goes well I'll do the merge.

bspratt:
Yeah, looking at the Skyline code I'm not sure it's going to do well with this implementation that passes a mix of MS2 data with and without precursors. Let's sit on it a while longer I guess.

chambm:
You mean it won't work? Or it will just not perform optimally? Because if it won't work, then there's a missing test, right? The PASEF semantics should be tested on TC on a small file.

chambm:
@bspratt Assuming this passes tests, it should be ready to merge. MSConvert will still output the MS2s without precursor info, but MsDataFileImpl will skip them.

bspratt:
Excellent! Thanks.

PR #180, opened 2018-08-28
Skyline/work/20180827 warnings

nickshulman:
I have gone through all of the Suggestions that we had and either fixed them, or turned them into hints.
Now, the only suggestion that remains in the code base is "Method was too complex to analyze".

I added a reference to "JetBrains.Annotations.dll".  This enables us to do things like "[UsedImplicitly]" on top of some of our NHibernate classes so that ReSharper won't complain about unused properties.
The JetBrains annotations all get stripped out at compile time, so we do not need to ship JetBrains.Annotations.dll with the Skyline installer.

There were some warnings that I was not sure the right way to fix (I ended up downgrading them to Hints). For instance, there are a lot of methods on IToolService that are unused. What is the correct way of telling ReSharper that this is a public API, and even though it doesn't get used in this set of projects, there are other projects out there that do use it?

The Suggestions which have to do with new CSharp features (which ended up being a lot of unused local out variables) got downgraded to Hints.


jbotaldrich:
From what I've read about public interfaces the suggestion is to use the "UsedImplicitly" Attribute in those cases.  If you have added the jetbrains .dll this should work.

https://resharper-support.jetbrains.com/hc/en-us/community/posts/206046399--method-never-used-warning-in-public-interfaces

pwiz_tools/Shared/ProteomeDb/Fasta/WebEnabledFastaImporter.cs
```diff
@@ -1313,7 +1313,7 @@ private int DoWebserviceLookup(IList<ProteinSearchInfo> proteins, char searchTyp
                 {
                     const string STATUS_REVIEWED = "reviewed"; // Uniprot reviewed status  // Not L10N
                     // now see if responses are ambiguous or not
-                    if (proteins.Count() == 1)
```
jbotaldrich: Why are you allowed to do this?  So confusing...

Tobias-Rohde: The Count method is an extension method for classes implementing IEnumerable, but since "proteins" is an IList it has a Count property that doesn't require enumeration of all items, instead it just keeps track of its size.

(If you look at the decompiled Count method you will notice that it actually tries to cast the IEnumerable to an IList to avoid enumerating all items, so there's not really a performance improvement)

bspratt:
My only remaining concern is the "too complex to analyze" suggestion, of which there are many. Though I'm not sure what we can do about that - it's probably a good suggestion but a  lot of bother to implement. The fact that it will tend to mask other new good suggestions makes it harmful, though.

brendanx67:
Yes. I think we have to keep that at Suggestion level and encourage people to avoid committing any new instances of it. Maybe we should have a suggestion count limit in Inspect Code? Cudos to anyone who can break up an existing function where this occurs and reduce the count, but probably not worth a concerted push to get rid of all instances.

bspratt:
Sadly I don't see any way to configure TeamCity to bark about anything but
Warning and Error counts.

PR #190, opened 2018-08-30
turn on checks for unlocalized elements

bspratt:
Skyline: re-running my automation to update from our old homebrewed "// Not L10N" localization inspection to standard ReSharper inspections. This first commit is just turning on the inspection as warning, to make sure TC is catching unlocalized elements.


bspratt:
It was meant to only go after files containing "// Not L10N" comments. I'll
back out any changes in those areas.

bspratt:
The only surprise I'm finding is in Bumbershoot, where there was a lone "//
Not L10N" comment that got the treatment. I'll back that one out.

The changes in pwiz_tools/Shared/BiblioSpec are intentional (no changes in
pwiz_tools/BiblioSpec -  we don't attempt L10N there).

Thanks for the review!

pwiz_tools/Skyline/Model/Results/RatioValue.cs
```diff
@@ -81,7 +81,8 @@ public static RatioValue ValueOf(double? ratio)
 
         public override string ToString()
         {
-            return Ratio + (double.IsNaN(StdDev) ? "" : (" rdotp " + DotProduct));  // Not L10N
+            // ReSharper disable once LocalizableElement
+            return Ratio + (double.IsNaN(StdDev) ? "" : (" rdotp " + DotProduct));
```
nickshulman: There's more than one string on this line. Doesn't that mean the "disable once" isn't going to work?

pwiz_tools/Skyline/Skyline.csproj.DotSettings
```diff
@@ -0,0 +1,3 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
nickshulman: What's the deal with this file? This does not look like valid XML. No end tag.

pwiz_tools/Skyline/Util/UtilUI.cs
```diff
@@ -302,7 +302,7 @@ function submitForm()
 </script>
 <form id=""my_form"" action=""{0}"" method=""post"" style=""visibility: hidden;"">
 <textarea name=""SkylineReport"">{1}</textarea>
-</form>", // Not L10N
+</form>@",
```
nickshulman: This is not right.

bspratt:
Pretty sure the entire pwiz_tools\Skyline\Executables\LocalizationHelper project goes away.

PR #191, opened 2018-08-31
- added RTime field to MS2s from BlibToMs2

chambm:
* added retentionTime to LibReader queries


pwiz_tools/BiblioSpec/tests/reference/demo.lms2
```diff
@@ -499,8 +500,9 @@ D	modified seq	FKNGFQTGSASK
 1196.20	61.7
 1197.00	14.5
 1284.50	23.4
-S	2	2	745.3
-Z	2	1471.7
+S	2	2	745.27
+I	RTime	0
```
brendanx67: Does it make sense to even write RTime 0 or should we just skip writing RTime altogether when the library has a zero retention time. Seems nonsensical. So, it must mean the library didn't get a time. Agreed it would be better to have this value be NULL, but 0 is not a valid RT.

chambm: I didn't think about it too much at the time. I agree it doesn't make sense to write 0s. That file does have one non-zero retention time. If it was all zeros I would have further updated the test.

PR #198, opened 2018-09-04
Speed improvements opening a Skyline document with 500 replicates

nickshulman:
I was sent a Skyline document that had 500 replicates in it. It took 10 minutes to open the file because of how long it takes to iterate through the tabs in the DockableFormCollection.
I have other changes in our SVN repository for the DigitalRune.Windows.Docking source code which speeds up iterating through the DockableFormCollection, and also makes it so that the DockPaneStrip pays attention to whether SuspendLayout was called.
In the changes here, I made it so that:
1. We call SuspendLayout before opening the "Chromatograms" menu item.
2. In "DockPanelLayoutLock" we not only lock the DockPanel, but also all of the DockPanes in the DockPanel, and the controls on those DockPane's. Also, every time we create a new chromatogram, we make sure that the DockPane that it's in is locked (and it's children too).


chambm:
Can you rebase this so that only the new commits are part of the PR? It seems to have confused the smartBuildTrigger script into thinking that a bunch of files were modified in the PR that weren't really. I'm working on that, but in the meantime it might be avoided by making sure the PRs are clean. (The consequence of that is that it will trigger Linux, Core, and Bumbershoot builds when you don't really need them).

nickshulman:
It seems I do not know how to rebase. I tried, and it looked like it was going to work, but then the seven commits that were here turned into 24 commits.

chambm:
Hah, been there, suffered through that (learning curve). Still get confused about git sometimes. Are you doing it with TortoiseGit?

PR #200, opened 2018-09-04
SkylineNightly/SkylineTester: if user kills a running SkylineTester w…

bspratt:
…hich is part of a SkylineNightly run, don't send any report

### DIFF
--- a/pwiz_tools/Skyline/SkylineNightly/Nightly.cs
+++ b/pwiz_tools/Skyline/SkylineNightly/Nightly.cs
@@ -136,9 +136,15 @@ namespace SkylineNightly
 
         public enum RunMode { parse, post, trunk, perf, release, stress, integration, release_perf }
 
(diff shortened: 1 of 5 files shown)


pwiz_tools/Skyline/SkylineNightly/Nightly.cs
```diff
@@ -382,6 +388,12 @@ public string Run()
                     Log(result = "SkylineTester has exceeded its " + durationSeconds +
                                  " second WaitForExit timeout.  You should investigate.");
                 }
+                else if (skylineTesterProcess.ExitCode == 0xDEAD)
+                {
+                    // User killed, don't post
+                    Log(result = SkylineTesterStoppedByUser);
```
brendanx67: This should only mention "no results will be posted" if it is a nightly run. Otherwise, it is understood that no results will be posted, and the message should just be "SkylineTester stopped by user."

PR #206, opened 2018-09-06
Skyline/work/20180904 skylinetester user stop no post

bspratt:


bspratt:
These changes fix the problem with SkylineTester going interactive when being shutdown by SkylineNightly on a timer. They also add logging for SkylineNightlyShim, and make it self-updating.

PR #213, opened 2018-09-07
Skyline: Clean-up SkylineTester Stop/Run code to make it clear when h…

brendanx67:
…uman interaction is involved

- Add support for IsNightlyRun() simply returning true for testing
- Fix nightly runs always treating results like they were stopped by a user (and not uploading logs)
- Fix TabNightly handling of ShiftKeyPressed and failing to schedule (show error instead of simply not handling the exception)


bspratt:
We went over the changes together, looks solid. :shipit:

PR #216, opened 2018-09-10
* fixed TestRunner to allow specifying log file names

chambm:
* fixed TestRunner to allow specifying log file names so multiple TestRunner processes don't try to write to the same file concurrently (perhaps making TestRunner run tests in parallel in one process would be preferable to this)


brendanx67:
I notice a fair amount of this deals with the memory log, which I was considering removing entirely, because it is just a single column of TotalMemory numbers. I don't think anyone is using this log, and I have had no problem getting this type of output from the normal log using Find and Replace with regular expressions. Just not sure we need the memory log.

chambm:
@brendanx67 @nickshulman So is it ok to just remove TestMemory.log entirely now? These changes were geared toward making parallel test running possible. Although they weren't necessary for what I ended up trying, which was making the inner run tests loop parallelized (but only between tests that were grouped by which ZIP files they used for their test data; i.e. can't run tests at the same time if they depend on the same zip file).

brendanx67:
Yes. I think we can get rid of the memory log. It is just a vestige that no one has looked at in years.

PR #228, opened 2018-09-12
Skyline: fix a problem where manually changing a molecule's chemical …

bspratt:
…formula, or its adduct, would not update the mz value of associated precursor transitions.  Also, make "f,p" (fragments and precursors) the default transition filter value for small molecules (formerly just "f").


bspratt:
found a problem in tests that convert documents from peptide to small molecule, closing PR for now

PR #232, opened 2018-09-13
Skyline/bugfix/20180912 small mol isotope dist hand edit fix

bspratt:
renewing this PR with a fix for a problem converting peptide docs to small molecule docs in tests


pwiz_tools/Skyline/Properties/Settings.cs
```diff
@@ -2900,7 +2900,7 @@ public sealed class SrmSettingsList : SerializableSettingsList<SrmSettings>
                         new[] { IonType.y }, // PeptideFragmentTypes
                         new[] { Adduct.M_PLUS_H, }, // SmallMoleculePrecursorCharges
                         new[] { Adduct.M_PLUS, }, // SmallMoleculeProductCharges
-                        Transition.MOLECULE_ION_TYPES, // SmallMoleculeFragmentTypes
```
brendanx67: Let's not change the defaults in this pull request. Let's leave it compatible with the peptide defaults, and then we can implement a second change set that adds "p" to this setting when MS1 gets turned on. I don't think people doing SRM want this default precursor targeting, but everyone extracting MS1 chromatograms does, and we have a problem of teaching them to enable MS1 and add "p" to this setting. I don't think making "p" on by default only for small molecules is the solution.

bspratt: Fair enough, leaving it at default "f" makes it consistent with peptide behavior (though adding ", p" on enabling full scan would not be consistent with peptide behavior, so best to leave it alone completely I guess).

bspratt: OK, those changes are in.

pwiz_tools/Skyline/Model/RefinementSettings.cs
```diff
@@ -1025,7 +1025,8 @@ public enum ConvertToSmallMoleculesChargesMode
                         Resources.RefinementSettings_ConvertToSmallMolecules_Converted_To_Small_Molecules,
```
brendanx67: Let's go over this section. I still see a mix of "document" and "newDoc" which doesn't look completely rationalized to me.

PR #236, opened 2018-09-17
UNIFI and MSConvertGUI improvements

chambm:
- added combineIonMobilitySpectra support for UNIFI data
- moved MS Level and Activation filters to the Subset panel in MSConvertGUI
- added Polarity, Analyzer, Charge state, and "Number of data points" (defaultArrayLength) filters to Subset panel in MSConvertGUI
- fixed flicker and interface non-responsiveness when loading sources in UNIFI browser
* updated IdentityModel.dll to v3.9 and Newtonsoft.Json.dll to v11 (which removed the need for the Newtonsoft.Json.dll `<assemblyBinding>` in several App.config files)
* added unit test for Reader_UNIFI
* added indexRange option to ReaderTestConfig
* fixed VendorReaderTestHarness to support a urls.txt file with a list of URLs rather than only reading vendor files in the directory


nickshulman:
Skyline uses the "TokenClient" class in "UnifiAccount.cs" (pwiz_tools\Skyline\Model\Results\Unifi).
(It uses it just so that there can be a button on the account dialog which tests whether the password is correct).
Since you seem to have changed which TokenClient constructor you're using everywhere else, I wonder if we need to update the code in UnifiAccount.cs.

PR #241, opened 2018-09-19
- fixed BiblioSpec with pdResult files that have multiple workflows

chambm:
* changed MSFReader to use string id (e.g. "workflowID.spectrumID") to allow for multi-workflow pdResult files
* added charge to SpecData.h (removes need for separate spectraChargeStates_ map)
* fixed incorrect signature for cvParamChildValueOrDefault()


pwiz_tools/BiblioSpec/src/MSFReader.cpp
```diff
@@ -133,15 +141,15 @@ namespace BiblioSpec
             specCount = getRowCount("MSnSpectrumInfo WHERE SpectrumID IN (SELECT DISTINCT MSnSpectrumInfoSpectrumID FROM TargetPsmsMSnSpectrumInfo)");
             hasCompensationVoltage = columnExists(msfFile_, "MSnSpectrumInfo", "CompVoltageV");
             statement = getStmt(
-                std::string("SELECT SpectrumID, MSnSpectrumInfo.RetentionTime, Mass, Charge, Spectrum") +
+                std::string("SELECT SpectrumID, MSnSpectrumInfo.RetentionTime, Mass, Charge, Spectrum, MSnSpectrumInfo.WorkflowID") +
                 (hasCompensationVoltage ? ", CompVoltageV " : " ") +
                 "FROM MSnSpectrumInfo "
                 "JOIN MassSpectrumItems ON MSnSpectrumInfo.SpectrumID = MassSpectrumItems.ID "
-                "WHERE SpectrumID");
+                "GROUP BY MSnSpectrumInfo.WorkflowID, SpectrumID");
```
nickshulman: I'm confused. Don't you want to put "MSnSpectrumInfo.WorkflowID = MassSpectrumItems.WorkflowID" in either the WHERE or ON clause?
(Also, I thought everything in the SELECT clause has to also appear in the GROUP BY clause, so I'm doubly confused-- SQLite must be weird).

chambm: You're right. I missed it there (but did add that to the TargetPsms query). SQLite does not require aggregation for non-grouped columns. It will just pick some arbitrary/randomish member from the group. It does make it easier to write queries but also makes it easier to shoot yourself in the foot.

nickshulman: I don't know what you should use to create a test. Do you want me to email the person (Paul) who originally reported this issue and see if we can get a small file for testing? (Or do you want to email him?)

chambm: Just checking if there's some secret pdResult test generator (or more likely, big-file-subsetter). ;) I'll ask Paul if I can trim his file down for testing or another one if he wants.

chambm:
It occurred to me after I switched to the `Workflow.Spectrum` string id that this id is not traceable back to the original raw file. It would make more sense to use the original scan number for SpecIdInFile (but still have to use `Workflow.Spectrum` internally as a unique id).

PR #246, opened 2018-09-19
Skyline/work/20180829 add pasef perftest

bspratt:
This adds a perf test for PASEF MS1 handling. Note the change to BiblioSpec to pick up raw file cues from various mascot .dat header fields. Note also the addition of a user control to the import peptide search wizard that asks about using ion mobility values when they have been found in libraries generated during the import.


pwiz_tools/BiblioSpec/src/MascotResultsReader.cpp
```diff
@@ -713,6 +728,18 @@ string MascotResultsReader::getFilename(ms_inputquery& spec){
...
     }
 
+    if (filename.empty())
+    {
+        // often the filename is in the comment "COM=" or uploaded file "FILE=" or uploaded file url "URL="
+        string globalFilename;
+        if (IsPlausibleRawFileName(globalFilename = ms_params_->getFILENAME()) ||
+            IsPlausibleRawFileName(globalFilename = ms_params_->getDATAURL()) ||
+            IsPlausibleRawFileName(globalFilename = ms_params_->getCOM()))
+        {
```
chambm: What about the case where the filename is in one of these fields but it's not the only (or last) thing? Like:
COM=File:blahblah.raw AcqDateTime:2018-04-02T12:34:56

But I suppose FILENAME and DATAURL are much more robust.

bspratt: Given that it's a comment it could be nearly anything of course, it's a bit of a hail mary play but I have seen an actual example where it was just the bare filename in the comment. Could be anything of course, and we could work harder to grub something out but I think this is still a step up.

chambm: Would anything break (worse than not having a filename at all) if you got something like:
COM=File:somefile.raw
Then you'd parse "File:somefile.raw" as the filename, right?

bspratt: That's a risk, yes, Is there a convenient boost function that might be employed to catch that "ends with a reasonable .ext but otherwise not really looking like a file name" case?

bspratt: As to worse than no filename at all, no, I don't think so. It's already a last ditch effort, pulling it from the COM field. It would at least result in some kind of error downstream in the workflow that might help the user locate the actual file needed.

chambm: Yes, although I didn't remember it: `boost::filesystem::windows_name()` (which is a rough superset of `portable_posix_name()`). That would at least get rid of anything with a colon or quotes, although File=filename.raw would still slip through.

bspratt: Noted. But I think what we have is sufficient, and already unlikely to actually come into play.

PR #251, opened 2018-09-20
Skyline: fix a problem with reading negative ion mode Agilent chromat…

bspratt:
…ogram files


bspratt:
Yeah, don't know how we started down that road, I feel the same way though.

chambm:
Hmm, we'd better have let the tests run on this to make sure it works on all platforms

PR #256, opened 2018-09-21
Investigate making "inputs.tar.bz2" managed by Git Large File Service.

nickshulman:
This change makes it so that the file "pwiz_tools/BiblioSpec/tests/inputs.tar.bz2" gets managed by **git lfs**. That file is 54MB and has a long revision history.
If you do not have the **git lfs** extension installed, then instead of getting the big file, you will just get a tiny text file that looks like this:
`version https://git-lfs.github.com/spec/v1
oid sha256:eaa9e8cb1ec77ee1c24e04f14e3c35b55b51fe0f97347d562a5ffa6ed20f9768
size 55884441`

I assume this change will not pass on TeamCity, since I imagine it does not have **git lfs** installed, but I wanted us to start thinking about using git lfs for our larger files.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 #  .gitattributes for pwiz
-
 # Don't change line endings on potentially indexed mass spec data
 *.mzml binary
 *.mzxml binary
+pwiz_tools/BiblioSpec/tests/inputs.tar.bz2 filter=lfs diff=lfs merge=lfs -text


chambm:
I agree to the unwritten assertion that the blib inputs tarball has become too big. :) Especially given the update frequency. But I'm not sure putting it in LFS is the right way to go. Why don't we just commit the inputs individually? I've been thinking of doing that for the vendor reader test files too. There are also several blib test files that are probably much bigger than they should be, but it would probably take more effort than it's worth to trim them down. Trimming down the pdResult file was relatively easy since there are only a few tables that blib actually uses.

In any case thank you for investigating LFS! Is it actually faster to download from LFS than from GH?

chambm:
One reason not to use LFS:

> We wanted to let you know that you’ve used 80% of your data plan for Git LFS on the organization ProteoWizard. No immediate action is necessary, but you might want to consider purchasing additional data packs to cover your bandwidth and storage usage:
> 
>   https://github.com/organizations/ProteoWizard/billing/data/upgrade
> 
> Current usage as of 25 Sep 2018 08:28AM UTC:
> 
>   Bandwidth: 0.83 GB / 1 GB (83%)
>   Storage: 0.1 GB / 1 GB (10%)

So it seems bandwidth is a huge bottleneck.

chambm:
And now we're at 100% bandwidth. Not sure what happens now. But I guess I better close the PR so TC stops building it.

nickshulman:
I think what happens when we use up our free quota is that instead of
people getting the big file that they are supposed to, they just end up
getting the small text file pointer to it.

If TeamCity is causing us to use up our bandwidth quota, then it must mean
that at least one of those machines has the Git LFS extension installed.

It looks like Git LFS bandwidth costs $0.10 per gigabyte, which is
comparable to what Amazon S3 charges for bandwidth.
That only appears expensive because the regular github hosting is free.

If we converted all of the large files that we ought to to use LFS, we
would probably end up paying $100 per month for bandwidth, but it could be
that we need to be smarter about not always cloning the entire repository.
-- Nick

PR #267, opened 2018-10-01
Add "TestCategory" attributes to tests.

nickshulman:
Here is some infrastructure work which I think we might want to check in after we branch.

Instead of disabling tests the way that we currently do, where there's code that checks "IsPerfTest" and "RunPerfTests", you can put [TestCategory] attributes on the test methods.
(Also, for convenience, I made [PerfTestAttribute] which is equivalent to [TestCategory("Perf")].

There is a new "skipcategories" argument on TestRunner, which defaults to "Perf,SmallMolecules".
Also, changed ReSharper's setting "Skip tests from these categories" to that.

If you want to filter these categories out of Visual Studio's Test Explorer, you have to type the following into the search box:
-Trait:"Perf" -Trait:"SmallMolecules"
(I wish I could figure out a way to check in that setting so that it is the default for everyone).

I removed all of the code within the tests which checks whether perf tests or small molecules versions are enabled.


bspratt:
The major reason we discriminate against PerfTest and SmallMolecule versions is to that one can get through a "Run All" in Test Explorer in a reasonable amount of time. Does this consider that need?

nickshulman:
In order to skip the perf tests etc., you would have to type the following
into the search box in the Test Explorer window:
-Trait:"Perf" -Trait:"SmallMolecules"

So far, I have not figured out a way to make it so that that's the setting
by default-- I think we might need to update the developer setup
instructions to say that you need to type that in.
Or, we could recommend that you use ReSharper to run the tests.

ReSharper currently does not run our tests because ReSharper's default
platform is "Automatic" (although you can get it to do the right thing by
choosing "Options > Platform > x64" in ReSharper's Unit Test Sessions
dialog), which means that ReSharper looks at the test dll's platform to
decide whether to be x64 or x86. However, our test dll's are compiled as
"AnyCPU". This pull request also makes it so that test dll's are explicitly
either x64 or x86, so we might now want to say that the recommended way of
running tests is through ReSharper.
-- Nick

bspratt:
If it gets tests running in ReSharper that's a win, unless it prevents it in Test Explorer, since ReSharper isn't a build requirement.

PR #276, opened 2018-10-09
- changed installers to use side-by-side MSVC runtime

chambm:
- changed installers and build system to use side-by-side MSVC runtime DLLs rather than requiring the appropriate redistributables to be installed
* removed VC10 DLLs no longer in use
* moved VC11 OpenMP DLLs to VC110 CRT directory to be consistent with other OpenMP DLLs
* added VC9 CRT/MFC DLLs that CompassXtract uses
* added VC14 MFC and concurrency runtime DLLs used by timsdata.dll
* commented out x64/x86 and .dll in .gitignore
* changed IDPicker deploy test to use non-admin install
* updated WiX templates to "Single Package Authoring" which allow per-user or per-machine installation, factored out the shared list of pwiz vendor DLLs into a standalone file (vendor-dlls.wxs-fragment), and made them check for .NET 4.7.2 as a prerequisite
* removed SharpSvn DLLs and related code


chambm:
Were you all aware there are a few MSVC DLLs used (not including CompassXtract) that aren't currently included by Skyline? Do you think we ought to centralize our DLL distribution? The Skyline Wix template could be run through a preprocessor script like the pwiz/IDPicker ones already are. In this PR I updated the preprocessing to insert the DLL list from a shared file (`vendor-dlls.wxs-fragment`), but I could have gone further: now that the Jamfiles know about all the MSVC dependencies, that list of DLLs could be auto-generated as a Jamfile target, which in turn could auto-update the Wix templates and csproj files:
Jamfiles -> vendor-dlls list
vendor-dlls list -> insert into Wix templates (ProteoWizard, Skyline, IDPicker)
vendor-dlls list -> insert into csproj files (or some file included by those files?)

chambm:
Correct I didn't add the missing DLLs (VC14 MFC and concurrency runtime), because obviously they haven't yet caused a problem (or at least it hasn't been understood). I suspect perhaps the missing DLLs have usually already been installed in a Windows update even though they are technically part of the redistributables.

chambm:
@brendanx67 Shall I commit this as is and fix any problems as they pop up?

brendanx67:
Sorry. Let me build this branch and run Skyline installation tests with it. I need to see if I can locate some clean system VMs I created for this purpose.

chambm:
@brendanx67 Did you get a chance to test this? I know it fixed issues for one user who couldn't install with merge modules (mysteriously). How to run the Skyline installation tests, btw?

PR #280, opened 2018-10-11
Fix problem reading Thermo raw files that have scans with multiple pr…

nickshulman:
There is a .raw file here which cannot be opened by current builds of ProteoWizard.
https://skyline.ms/announcements/home/support/download.view?entityId=c482aa73-aebe-1036-8cc6-e465a393e3d4&name=Skyline_share_topdown_prm.zip

The existing code does not make sense. It appears that "msOrder" is the MS level, and I would not expect that to be equal to the number of isolation windows.

This change fixes the problem but I have not run any unit tests yet.

### DIFF
--- a/pwiz_aux/msrc/utility/vendor_api/thermo/RawFile.cpp
+++ b/pwiz_aux/msrc/utility/vendor_api/thermo/RawFile.cpp
@@ -1826,8 +1826,8 @@ vector<double> RawFileImpl::getIsolationWidths(long scanNumber) const
     if ((int) msOrder == 1)
         return isolationWidths;
 
-    long numMSOrders = (int) msOrder > 0 ? msOrder-1 : 0;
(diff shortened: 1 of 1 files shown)


chambm:
The precursor count == isolation width count works for MSFileReader. Give me a couple days to review this and compare the 32-bit vs 64-bit output. Thanks for going ahead and fixing it though!

chambm:
How do we regression test this? Does Skyline have a test with MSX data?

nickshulman:
Skyline has a MSX test, but it uses mzML files (and we no longer recommend
doing deconvolution in Skyline, so we will probably be ripping that feature
out soon).

Maybe we should write a new tutorial on how to do deconvolution in Skyline
using MSConvert (and then we always write a functional test which covers
the steps of the tutorial).
-- Nick

PR #298, opened 2018-10-24
Disable audit log testing in TestPeakPickingTutorial because it fails…

nickshulman:
TestPeakPickingTutorial currently fails on 32 bit because one of the "PercentageContribution" properties on a "EnabledFeatureScores" gets formatted slightly differently.

The easiest way to fix this test is to turn off audit log checking.

Was anyone hoping for a more conscientious solution to this problem?


Tobias-Rohde:
I think we discussed this about a month ago, my guess was that the differences was due to the fact that internally 32 bit skyline uses the FPU stack for these calculations while 64 bit skyline uses SSE instructions. But it turned out that simply adding the roundtrip specifier when calling ToString fixed this issue, I'll push a fix for this later.

PR #301, opened 2018-10-26
Skyline Wine fixes

chambm:
* fixed some issues that prevented some Skyline tests from executing under Wine
* updated BuildMethod projects to .NET 4
* added bindingRedirect for log4net
* moved FormUtilTest to TestFunctional

This is a WIP; I want to see if it passes the Docker tests on TeamCity. If so I will find a way to disable the problematic BadRawFile test and UTF8/Japanese encoding only when running under Wine.


pwiz_tools/Skyline/TestA/CommandLineTest.cs
```diff
@@ -1095,7 +1095,7 @@ public static int CountErrors(string searchSpace, bool allowUnlocalized = false)
             return count;
         }
         
-        [TestMethod]
+        //[TestMethod]
```
brendanx67: Obviously, don't merge this to master.

pwiz_tools/Skyline/TestRunner/Program.cs
```diff
@@ -181,7 +181,7 @@ static int Main(string[] args)
             Application.SetUnhandledExceptionMode(UnhandledExceptionMode.CatchException);
             Application.ThreadException += ThreadExceptionEventHandler;
 
-            Console.OutputEncoding = Encoding.UTF8;  // So we can send Japanese to SkylineTester, which monitors our stdout
+            //Console.OutputEncoding = Encoding.UTF8;  // So we can send Japanese to SkylineTester, which monitors our stdout
```
brendanx67: Also not for merge to master.

pwiz_tools/Skyline/TestRunner/TestRunner.csproj
```diff
@@ -60,7 +60,7 @@
     <OutputPath>..\bin\x64\Debug\</OutputPath>
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <DebugType>full</DebugType>
-    <PlatformTarget>AnyCPU</PlatformTarget>
+    <PlatformTarget>x64</PlatformTarget>
```
brendanx67: And not sure how this would impact running 32-bit tests with TestRunner.

chambm: Shouldn't affect it at all. It only sets x64 for the 64-bit TestRunner exe. 32-bit tests will run the 32-bit TestRunner.exe.

pwiz_tools/Skyline/TestRunner/Program.cs
chambm: Is the language-dependent conditional I added for this acceptable?
```
if (commandLineArgs.ArgAsString("language") != "en" && commandLineArgs.ArgAsString("language") != "en-US")
    Console.OutputEncoding = Encoding.UTF8;  // So we can send Japanese to SkylineTester, which monitors our stdout
```

pwiz_tools/Skyline/TestA/CommandLineTest.cs
chambm: No longer commented out with the latest push.

chambm:
@brendanx67 What do you think of my latest changes to add TeamCity test decoration to TestRunner? It allows us to have better test reporting in TeamCity for both the main Skyline configs and the Wine config.

brendanx67:
TeamCity test decoration seems cool. It would be nice to have the test report cover all the individual tests instead of 1 per test DLL. One thing I am not totally clear on is the difference between Console.WriteLine() and Log(). I guess it is arguable that you never want these lines written to the log file. So, maybe Console.WriteLine() is the right choice. I am sure it works, but it caught my eye.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
```diff
@@ -370,6 +379,23 @@ public bool Run(TestInfo test, int pass, int testNumber, string dmpDir)
...
+            if (TeamCityTestDecoration)
+            {
+                // ReSharper disable LocalizableElement
+                var tcMessage = new System.Text.StringBuilder(message);
+                tcMessage.AppendLine();
+                tcMessage.AppendLine(stackTrace);
+                tcMessage.Replace("'", "|'");
+                tcMessage.Replace("\n", "|n");
+                tcMessage.Replace("\r", "|r");
+                tcMessage.Replace("|", "||");
```
nickshulman: Usually, replacing "|" with "||" would be the absolutely first replacement. Otherwise, you end up replacing the "|n" with "||n", which is probably not what you want.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
```diff
@@ -172,6 +175,9 @@ private static string GetAssemblyPath(string assembly)
 
         public bool Run(TestInfo test, int pass, int testNumber, string dmpDir)
         {
+            if (TeamCityTestDecoration)
```
brendanx67: Please use a function for all 3 cases where you use if (TeamCityTestDecoration) Console.WriteLine(). Thanks.

chambm: You mean a function like Log() which is a no-op if TeamCityTestDecoration is false?

brendanx67: Exactly, and which just takes the status and TestInfo object as parameters.

brendanx67: And which does all the escaping.

brendanx67: Makes your impact on a 200-line function just 3 added lines, right?

chambm: Yes but I think it needs 2 functions. For start and end, where the end function takes an error message if the result is failure. Folding it into 1 function would require some kind of "message type" parameter or parsing the message itself.

brendanx67: Looks good to me.

PR #309, opened 2018-11-02
Skyline: Some initial fixes to Nick's new Lists feature

brendanx67:
- Make ListDesigner localizable, give it mnemonics, turn off minimize, maximize and taskbar, etc.
- Make View > Other Grids > Lists > Define New List actually work
- Rationalize mnemonics on the View menu to support the new options
- Make adding a new List to a document add audit log entries (needs more work)

### DIFF
--- a/pwiz_tools/Skyline/Model/Lists/ListDef.cs
+++ b/pwiz_tools/Skyline/Model/Lists/ListDef.cs
@@ -24,6 +24,7 @@ using System.Windows.Forms;
 using System.Xml;
 using System.Xml.Serialization;
 using pwiz.Common.Collections;
+using pwiz.Common.SystemUtil;
 using pwiz.Skyline.Controls.Lists;
 using pwiz.Skyline.Model.DocSettings;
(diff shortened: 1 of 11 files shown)


pwiz_tools/Skyline/Model/Lists/ListDef.cs
```diff
@@ -41,6 +42,10 @@ public ListDef(string name) : base(name)
             Properties = ImmutableList<AnnotationDef>.EMPTY;
         }
 
+        public override string AuditLogText { get { return null; } }
```
Tobias-Rohde: This works since ListDef has some properties with track attributes, therefore the reflection based ToString for this object would return something like {Properties=[...], IdProperty="...", DisplayProperty="..."}. But if it would not have any properties, the reflection ToStringwould try to use the AuditLogText. Generally there shouldn't be any IAuditLogObjects with IsName=false and no sub properties. ListDef shouldn't be an IAuditLogObject in the first place, but of course it is since XmlNamedElement implements it...

PR #312, opened 2018-11-02
Skyline/work/20180416 d and t

bspratt:
This adds the ability to understand IUPAC standard "elements" D and T (Deuterium and Tritium, or H' and H" in Skyline-speak) in chemical formulas and adduct descriptions.


pwiz_tools/Skyline/Util/Adduct.cs
```diff
@@ -315,7 +315,7 @@ private void ParseDescription(string input)
                                 success = int.TryParse(multMstr, out multiplierM);
                             }
                             var isotope = m.Groups["ion"].Value; // Not L10N
-                            var unlabel = isotope.Replace("'", "").Replace("\"", ""); // Not L10N
+                            var unlabel = BioMassCalc.DICT_HEAVYSYMBOL_TO_MONOSYMBOL.Aggregate(isotope, (current, kvp) => current.Replace(kvp.Key, kvp.Value));
```
nickshulman: Is "isotope" a chemical formula here? I'm worried that we're just doing string replace on a chemical formula, and so "T" will match any element which starts with a "T".

You might want to use SequenceUtil.TokenizeFormula.

bspratt: Excellent point, thanks!

bspratt: On closer inspection, this isn't anything to worry about until we start supporting heavy Ti labeling.

brendanx67:
Let's be sure to rebase and merge on this to avoid adding the month-long branch to the logs.

PR #313, opened 2018-11-05
Skyline/work/20181102 reorg tests

brendanx67:


brendanx67:
Yes, GUI tests stay in TestFunctional. I am considering a TestFunctionalData, but for now this only impacts Test and TestA. That is, TestA essentially becomes TestData and tests get redistributed based on whether they use pwiz_data_cli.dll

brendanx67:
What needs to happen to the Wine/Docker container build in TeamCity to support this? I noticed it failed. I think I have changed the jam files to replace TestA with TestData.

chambm:
The build config refers to Test.dll and TestA.dll by name so it's not really possible to have both the old code and the new code passing that config at the same time. When you're ready to merge to master you (or I) can edit that 3rd build step (it's really trivial) to change the name. After changing if you want to rerun the test to make sure this branch passes before merging, that's easy too.

An alternative would be a quick new flag like `nogui` to TestRunner in master which simply changes the default test list to Test and TestA (and in this branch, changes TestA to TestData).

brendanx67:
Okay. I will merge once I see code inspection passing. I will have a look at whether I can find the TestA use on TeamCity and change that. Although, I think it is unlikely this will change again soon, we should probably make this entirely controlled in the jam file, as you suggest. I would also like to see the new TeamCity decoration only happen when TeamCity is actually running since it makes it harder to look at the test output otherwise, as I found on my own dev machine last night running the target pwiz_tools/Skyline.

PR #315, opened 2018-11-06
Add all of the elements from IsotopeAbundances.cs to BioMassCalc

nickshulman:
This change makes it so that all of the elements in IsotopeAbundances.cs are available in the BioMassCalc class.
I had to change "SrmSettingsTest.cs" because it was assuming that "He" would not be a valid element.

What do you think?


pwiz_tools/Skyline/Util/BioMassCalc.cs
```diff
@@ -426,6 +426,15 @@ public BioMassCalc(MassType type)
             AddMass(Mn, 54.938045);
             AddMass(Mg, 24.305);
             AddMass(Si, 28.085); // Per Wikipedia
+            // Add all other elements
+            foreach (var entry in IsotopeAbundances.Default)
```
brendanx67: Can you add documentation of where the values in IsotopeAbundances come from?

pwiz_tools/Skyline/Test/SrmSettingsTest.cs
```diff
@@ -566,7 +566,7 @@ public void SerializeStaticModTest()
             // Bad terminus
             AssertEx.DeserializeError<StaticMod>("<static_modification name=\"Mod\" terminus=\"X\" formula=\"C23N\" />");
             // Bad formula
-            AssertEx.DeserializeError<StaticMod, ArgumentException>("<static_modification name=\"Mod\" aminoacid=\"K\" formula=\"C23NHe2\" />");
+            AssertEx.DeserializeError<StaticMod, ArgumentException>("<static_modification name=\"Mod\" aminoacid=\"K\" formula=\"C23NHx2\" />");
```
bspratt: Just ignorance on my part, but what is the significance of the change from C23NHe2 to C23NHx2?

nickshulman: This test is just trying to make sure that an error would be thrown if an invalid formula was provided. It used to be that we could count on "He" being an element that Skyline did not support, since we did not have it in our list of elements. However, now that we have added all of the elements from the periodic table, I had to pick some other string that was not a real element.

bspratt:
This seems fine from my end - all that remains is Brendan's question about documentation of where the values in IsotopeAbundances come from. I'd like to get this resolved so that I can move forward with my D T pull request, which has some potential code conflicts.

brendanx67:
I know where this comes from now (per Nick: old code of Mike Hoopman's). We should see what it would take to get  this normalized with pwiz/utility/chemistry/isotopes.txt, which clearly states its source as http://physics.nist.gov/PhysRefData/Compositions/index.html (though it was committed on 12/18/2009)

pwiz_tools/Skyline/Util/BioMassCalc.cs
bspratt: Can we make this a TODO and accept this merge? Another request for it today.

bspratt:
Add Matt to the reviewers list since I've touched something in Shared:

I wrote a little tool to generate an initialization code snippet for IsotopeAbundances.cs, it parses the output of https://physics.nist.gov/cgi-bin/Compositions/stand_alone.pl?ele=&all=all&ascii=ascii2&isotype=some

I used that tool to update and provide citations for values used in IsotopeAbundances.cs

I added a little code to carefully preserve the original CHONPS isotope values in IsotopeAbundances.cs, as changing these even a little creates havoc in Skyline testing (they aren't a lot different)

PR #319, opened 2018-11-10
Fix error "Conflict with transition Ion"

nickshulman:
This prevents the error "Conflict with transition Ion...Did you mean to use a different precursor charge state or label".

If the charge state and label type do not uniquely identify a transition group, then Skyline will ignore subsequent transition groups when calculating peptide-level ratios.

As far as I can tell, that error message would only show up when Skyline was importing results, and would not make sense to the user (such as when inserting a transition list).

I added a unit test which has some molecules with transition groups that have the same charge and label type, and it makes sure that the "Ratio To Standard" was calculated using the first transition group.


bspratt:
Did we ever get a data set that provokes this? While this looks reasonable I have an uncomfortable feeling that this is papering over a crack, and still worry about how a user got into this state in the first place.

nickshulman:
Yep, a user sent me their data here:
https://skyline.ms/_webdav/home/issues/files/%40files/bug608/

I'm not sure how they got their document to look like that, but the way I did it for my unit test in Skyline was by using the "Add Precursor" menu item in the Targets tree. I believe I was able to do it also from "Edit > Insert Transitions", but usually that did not work and the transition group that would have been conflicting never actually made it into the document. We should probably look into that.

bspratt:
OK, then I'm satisfied. 

:shipit:

PR #324, opened 2018-11-13
Fix side-by-side installation for Shimadzu

chambm:
* fixed side-by-side installation for Shimadzu's layout in x86/x64 subdirectories

@brendanx67 You want to test the clickonce installer for this to make sure things are getting into the right place before I merge to 4.2?

@nickshulman You want to sync it up with the admin installer? Gotta love those API stub DLLs.

Hopefully Shimadzu agrees to give us a version later which doesn't require this. I would be so happy to revert.


nickshulman:
If I understand these changes correctly, Skyline is now going to have two copies of files such as "api_ms_win_core_console_l1_1_0.dll". One copy will be installed in Skyline's root install folder, and the other copy will go in the x64 (or x86) subdirectory?

chambm:
Right. Because the Shimadzu native DLLs won't pick up their dependencies in the parent directory.

chambm:
@brendanx67 Can I merge this? How to test that DLLs are being placed properly with the ClickOnce installer?

chambm:
Note this also has the retention time fix. :)

chambm:
It's too bad symbolic links require admin permission or else x64 could just be a symbolic link to the parent directory. Which I tested and it does work. :) I tried setting PATH and that does not work.

PR #332, opened 2018-11-21
MSConvert: feature/profile-demultiplexing

austinkeller:
This adds the ability to demultiplex profile spectra.

Additional options for the demux filter:
* interpolateRT: Setting this option to false skips retention time interpolation (default true). Interpolation adds significant overhead when demultiplexing profile spectra.
* minWindowSize: Sets the window boundary in m/z (default 0.2). Using this flag to increase the window boundary size can serve as a workaround to salvage experiments with sloppily defined windows. Preventing poor windows boundary definitions needs to be done in Skyline where users generate their acquisition method.

Other changes:
* Add unit tests for PrecursorMaskCodec
* Cleaned up demux code, removing macros for profiling.
* Change demuxed spectrum id format. Previously, only demux= was added with a unique index for each spectrum derived from a single scan. Some software still relies on scan= being a unique number. To fix this, scan= is now preserved in originalScan= and scan= is replaced with a monotonically increasing index. demux= is still added.
* Remove selectedIon precursor intensity information since it doesn't make sense (or is no longer known) after demultiplexing.
* Update SHA1 library to fix excessive memory usage. The previous version would load the full file into memory before hashing. With this update the library no longer causes a memory usage spike.
* Fix msconvert issue interpreting boolean options (e.g. option=true or other_option=false would fail) by adding a LocaleBool struct.
* Fix issue with interpolation in FindNearbySpectra. One too many spectra were acquired from before the target spectrum rather than after. Updated DemuxHelpersTest to capture this change (the original test for this function was incorrect).
* Better demux error messages


austinkeller:
Note that the change to spectrum id format, while less wrong than before, still violates the schema. I'll be submitting my changes to fix mzML and mzXML schema validation of demuxed files in another pull request.

.gitignore
```diff
@@ -383,12 +383,27 @@ Version.cpp
 Reader*Test.data
 /pwiz_aux/msrc/utility/vendor_api/*/EULA.*
 /pwiz_aux/msrc/utility/vendor_api/ABI/LicenseKey.h
+/pwiz_aux/msrc/utility/vendor_api/ABI/*.dll
+/pwiz_aux/msrc/utility/vendor_api/Agilent/x64/
+/pwiz_aux/msrc/utility/vendor_api/Agilent/x86/
+/pwiz_aux/msrc/utility/vendor_api/Bruker/x64/
+/pwiz_aux/msrc/utility/vendor_api/Bruker/x86/
 /pwiz_aux/msrc/utility/vendor_api/Bruker/**/*.h
```
chambm: Would /pwiz_aux/msrc/utility/vendor_api/**/*.dll cover all these?

pwiz/utility/misc/SHA1.cpp
```diff
@@ -145,109 +145,68 @@ void CSHA1::Update(const UINT_8* pbData, UINT_32 uLen)
...
-		const INT_64 fileSize = (INT_64)boost::filesystem::file_size(tszFileName);
-
-		// don't use a 32-bit process's entire address space, but 64-bit processes can just use the file size
-		const INT_64 maxRegionSize = sizeof(void*) == 8 ? fileSize : (1 << 30);
-		const INT_64 lMaxBuf = SHA1_MAX_FILE_BUFFER;
+	FILE* fpIn = _tfopen(tszFileName, _T("rb"));
+	if(fpIn == NULL) return false;
 
-		using namespace boost::interprocess;
-		file_mapping mmFile(tszFileName, read_only);
```
chambm: There's a reason it's using memory mapped I/O: it's much faster! Or at least, it is much faster than the 2008 CSHA1 code. If you can show that the larger buffers in the newer CSHA1 code bring the performance up to par then we can switch back.

pwiz_tools/commandline/msconvert.cpp
```diff
@@ -841,8 +841,13 @@ void processFile(const string& filename, const Config& config, const ReaderList&
         MSData& msd = *msdList[i];
         try
         {
+            
             // calculate SHA1 checksums
+            
+            *os_ << "calculating source file checksums... ";
```
chambm: Let's just go with `*os_ << "calculating source file checksums" << endl;`
We don't do the "doing" ... "done" thing on any other progress message.

However, maybe you're adding the "done" message because the next step is wrapping the spectrum list, and if those ctors take a long time it can look like it's spending all that time hashing. I recently fixed this by adding IterationListRegistry to the wrapping. If there's a long running wrapper (like SpectrumList_Filter, or SpectrumList_Demux), it should send progress messages to the iteration listener. That will update the output and let the user know what's going on. In both msconvert.cpp and the GUI.

pwiz/analysis/demux/Jamfile.jam
```diff
@@ -52,9 +52,18 @@ lib pwiz_analysis_demux
         <library>$(PWIZ_ROOT_PATH)/pwiz/data/msdata//pwiz_data_msdata
     ;
 
+lib pwiz_analysis_demuxtestdata
```
chambm: That's an impressive solution for your simulated test data. What's the advantage of it vs. trimming down real data? I've had the same thought several times over the years for simulating peptide DDA data, but every time I decided it would be easier to use some real data and filter it down to a small size.

pwiz/analysis/demux/Jamfile.jam
```diff
@@ -52,9 +52,18 @@ lib pwiz_analysis_demux
...
+    :   <library>$(PWIZ_ROOT_PATH)/pwiz/data/msdata//pwiz_data_msdata
+    :
+    :   <library>$(PWIZ_ROOT_PATH)/pwiz/data/msdata//pwiz_data_msdata
+    ;
+
+
 unit-test-if-exists CubicHermiteSplineTest : CubicHermiteSplineTest.cpp pwiz_analysis_demux ;
 unit-test-if-exists DemuxDebugReadWriteTest : DemuxDebugReadWriteTest.cpp pwiz_analysis_demux ;
 unit-test-if-exists MatrixIOTest : MatrixIOTest.cpp pwiz_analysis_demux ;
+unit-test-if-exists PrecursorMaskCodecTest : PrecursorMaskCodecTest.cpp pwiz_analysis_demux ../../data/msdata//pwiz_data_msdata_examples pwiz_analysis_demuxtestdata ;
```
chambm: If this test is the only code using `DemuxTestData.cpp`, then you can forgo the lib target and just add it to the source file list (i.e. after `PrecursorMaskCodecTest.cpp`). If you do plan to use it elsewhere, then make it explicit:
`explicit pwiz_analysis_demuxtestdata ;`

.gitignore
austinkeller: Yes I can change that

pwiz/utility/misc/SHA1.cpp
austinkeller: The reason I had moved to the newer version was that the previous version loaded the entire file into memory, and some of my files for the fast scanning LIT in profile mode demultiplexed to 2x the number of scans were >50Gb. This provided a solution that could handle much larger input file sizes. I'm happy using whatever solution can scale to these larger file sizes.

I'm remembering now that I had wanted to merge the customized pwiz changes with the newer version, but didn't do that when I was just doing prototyping early on. I suppose I could try merging the pwiz customizations with this newer version and report back.

pwiz_tools/commandline/msconvert.cpp
austinkeller: Yep, the SpectrumList_Demux wrapper initialization can be time consuming so I added that done message for that reason. I'll remove that.

pwiz/analysis/demux/Jamfile.jam
austinkeller: In this case collecting just one of either MSX overlap or multi-overlap data would take longer than writing the code that I did :) and I figured that once I wrote the bulk of the code to simulate a single permutation then it'd be quick to try and validate many other permutations of demuxing. I want to verify that the demuxing code is as general as I think I'm making it, where n-many overlaps can be demuxed (instead of only up to 2) and that MSX and n-many overlaps can be mixed. At this point collecting any one of these would either be a beast or just not possible with current instrument control software, so it came down to a decision of how to simulate data and whether the code should be internal to pwiz or if I should generate some simulated mzMLs externally.

I should note that in this branch, neither overlap MSX or n-many overlap are working (probably). Testing those would be the next step.

pwiz/utility/misc/SHA1.cpp
chambm: Hmm. Memory mapping should load the whole file into virtual memory, but not physical memory. But after testing I've noticed it will use up all free physical memory as it churns through the file. It doesn't release the memory that it's already hashed. That is fixable. But I still benchmarked the several approaches. I think the upgrade to VS 2017 may have brought the two approaches closer together. Here's the performance of the various approaches on a 100GB file:
image

It originally looks like the 1GB region does best, but when I tried it with the larger buffer size from the new code, it was slower by a couple minutes. So maybe shrink_by with the larger buffer is best. But not by a large margin and it uses considerably more memory (but 200MB isn't much when dealing with a 100GB file). I'll have to think about this some more.

pwiz/analysis/demux/Jamfile.jam
chambm: I suppose when you want to test your code against data that is theoretically possible to acquire but nobody's actually done it, simulating it is definitely the next best thing. :100: :)

.gitignore
austinkeller: Actually, this would cut out some versioned files like the /pwiz_aux/msrc/utility/vendor_api/ABI/x64/SQLite_v1.0.98/System.Data.SQLite.DLL

Two alternatives:
Exclude the entire /pwiz_aux/msrc/utility/vendor_api/ directory with `/pwiz_aux/msrc/utility/vendor_api/**/*` and then either:
1) make future additions with `git add --force` (and maybe add a comment in the .gitignore indicating this is the intent), or
2) add each of our included files as explicit exceptions, e.g. 
```
!/pwiz_aux/msrc/utility/vendor_api/**/*.jam
!/pwiz_aux/msrc/utility/vendor_api/ABI/WiffFile.cpp
!/pwiz_aux/msrc/utility/vendor_api/ABI/WiffFile.hpp
...
```

pwiz/utility/misc/SHA1.cpp
austinkeller: Ahh okay, would it be fair then to branch fixing the physical memory consumption off into a separate issue? I can revert my edits to the SHA1 files.

.gitignore
chambm: Yeah let's just add the versioned DLLs as explicit exceptions. That damn SQLite DLL, I want to kill it!

pwiz/analysis/demux/Jamfile.jam
austinkeller: fixed in 2deaf41

pwiz_tools/commandline/msconvert.cpp
austinkeller: fixed in 2deaf41

.gitignore
austinkeller: My mistake, adding /pwiz_aux/msrc/utility/vendor_api/**/*.dll won't cover those four directories because they also contain some .dll.config, .xml, and .lib files as well. So I'm not really seeing a clean way to exclude them without just excluding those four directories.

chambm: These files are checked into git, right? So why does it matter whether they're included by an ignore rule? If we need to add a new DLL to one of these vendors that have non-private DLLs, it's true the ignore rule would make it harder to notice that it hasn't been added yet, but using exceptions wouldn't help that either. Only individually naming all the excluded DLLs would really solve that, right? It's not worth it. We'll just force add them on a case-by-case basis.

pwiz/utility/misc/SHA1.cpp
chambm: Yes let's move it to a separate issue.

.gitignore
austinkeller: Okay that makes sense. It didn't seem worth managing all of that. That's why I was leaning towards treating all of `/pwiz_aux/msrc/utility/vendor_api/` as something of a build output directory, the contents of which we don't really control, and then resort to `git add --force` for any changes to the vendor api.

austinkeller: Getting back to the original problem, here are the non-dll files that I was needing to exclude in order to reduce clutter on calls to `git status`:

```
pwiz_aux/msrc/utility/vendor_api/Agilent/x64/BaseDataAccess.dll.config
pwiz_aux/msrc/utility/vendor_api/Agilent/x86/BaseDataAccess.dll.config
pwiz_aux/msrc/utility/vendor_api/Bruker/x64/BDal.BCO.Constants.xml
pwiz_aux/msrc/utility/vendor_api/Bruker/x64/BDal.BCO.Objects.xml
pwiz_aux/msrc/utility/vendor_api/Bruker/x64/BDal.CCO.Constants.xml
pwiz_aux/msrc/utility/vendor_api/Bruker/x64/BDal.CCO.Objects.xml
pwiz_aux/msrc/utility/vendor_api/Bruker/x64/FlexVariableTable.xml
pwiz_aux/msrc/utility/vendor_api/Bruker/x64/ICRVariableTable.xml
pwiz_aux/msrc/utility/vendor_api/Bruker/x64/Interop.EDAL.SxS.manifest
pwiz_aux/msrc/utility/vendor_api/Bruker/x64/Interop.HSREADWRITELib.SxS.manifest
pwiz_aux/msrc/utility/vendor_api/Bruker/x86/BDal.BCO.Constants.xml
pwiz_aux/msrc/utility/vendor_api/Bruker/x86/BDal.BCO.Objects.xml
pwiz_aux/msrc/utility/vendor_api/Bruker/x86/BDal.CCO.Constants.xml
pwiz_aux/msrc/utility/vendor_api/Bruker/x86/BDal.CCO.Objects.xml
pwiz_aux/msrc/utility/vendor_api/Bruker/x86/FlexVariableTable.xml
pwiz_aux/msrc/utility/vendor_api/Bruker/x86/ICRVariableTable.xml
pwiz_aux/msrc/utility/vendor_api/Bruker/x86/Interop.EDAL.SxS.manifest
pwiz_aux/msrc/utility/vendor_api/Bruker/x86/Interop.HSREADWRITELib.SxS.manifest
```
The simplest pattern I saw in these was just
```
/pwiz_aux/msrc/utility/vendor_api/Agilent/x64/
/pwiz_aux/msrc/utility/vendor_api/Agilent/x86/
/pwiz_aux/msrc/utility/vendor_api/Bruker/x64/
/pwiz_aux/msrc/utility/vendor_api/Bruker/x86/
```

Maybe that's still the cleanest? It's my fault for adding confusion by at first saying that a *.dll exclusion would cover these all.

chambm: Yeah. Go for that (in addition to the *.dll one)

pwiz/utility/misc/SHA1.cpp
austinkeller: Moved to https://github.com/ProteoWizard/pwiz/issues/352

austinkeller: And reverted the changes in the latest commit

.gitignore
austinkeller: Will do 👍

austinkeller: Done in https://github.com/ProteoWizard/pwiz/pull/332/commits/a43b259829ab79b295536f1bac27d6964dfc42b3

austinkeller:
Hey @chambm, It looks like all tests are passing (except the skyline code inspection, and it seems like that's unrelated?). Am I good to automatically merge? I was thinking I'd do a squash and merge to concisely summarize all of the changes, unless you have another preference.

I'm going to branch off of these changes and start work on the mzXML schema issues with demux.

PR #336, opened 2018-11-26
Fix Bruker side-by-side VC execution

chambm:
A user reported a COM initialization failure when trying to convert a YEP file and I reproduced it on a Windows Server 2016 instance. 
* updated VC90 DLLs to 30729.6161 and updated versions in the embedded manifests of Bruker CompassXtract DLLs (it was the only way I could get it to run without VC9 redist, at least on Windows Server 2016)
* fixed name of vcomp90.dll; apparently it only lacks the suffix in its System32 hard-link


chambm:
@brendanx67 The scope of this PR expanded when I discovered that Thermo MSFileReader wouldn't run without the VC9 CRT installed, which was the case on one of the AWS boxes I tested. So due to some manifest updates, Skyline had to be updated too. When I looked at prerequisites in ClickOnce I didn't see any CRT/MFC ones added, but maybe I don't really understand what that old commented out section in the manifest is really saying.

brendanx67:
MSFileReader is now 32-bit only, right?

pwiz_tools/Shared/Lib/Microsoft.VC90.MFC/x86/Microsoft.VC90.MFC.manifest
```diff
@@ -1,13 +1,7 @@
...
-        name="Microsoft.VC90.MFC"
-        version="9.0.30729.1"
-        processorArchitecture="x86"
-        publicKeyToken="1fc8b3b9a1e18e3b"
-    />
-    <file name="mfc90u.dll" />
-</assembly>
+    <noInheritable></noInheritable>
+    <assemblyIdentity type="win32" name="Microsoft.VC90.MFC" version="9.0.30729.6161" processorArchitecture="x86" publicKeyToken="1fc8b3b9a1e18e3b"></assemblyIdentity>
+    <file name="mfc90.dll"/>
```
chambm: CompassXtract needs mfc90.dll, MSFR needs mfc90u.dll. So far I still have Skyline only including mfc90u.dll, but I'll have to see whether this manifest causes an issue when one of its files is missing.

chambm:
> MSFileReader is now 32-bit only, right?

Right.

PR #337, opened 2018-11-27
Enable vendor readers for BiblioSpec

chambm:
- enabled vendor readers for BiblioSpec
* removed msvc version requirements for Waters reader as they are no longer relevant


chambm:
@brendanx67 We'll want this as well as the ClickOnce fix in the next daily. Otherwise BiblioSpec will be looking for file extensions it can't read.

PR #347, opened 2018-12-03
- updated Shimadzu QTOF DLLs

chambm:
- updated Shimadzu QTOF DLLs which no longer need the platform-specific subdirectories


chambm:
@brendanx67 Any reason to delay merging this?

brendanx67:
Nope. High priority to solve this in both Skyline-daily and 4.2

PR #365, opened 2018-12-11
Skyline: working toward a UI that dynamically adjusts for small molec…

bspratt:
…ule usage (so, less "peptide" where we mean "molecule" when doing small molecule work)


bspratt:
The change history on this branch had gotten pretty twisted, so this is a fresh start for easy review.

PR #367, opened 2018-12-11
Skyline/work/20181207 code coverage

bspratt:
trying to work out a failure on TC code coverage that I can't repro anywhere else


bspratt:
OK, I believe I've figured it out. Started a new and cleaner branch for actual merge.

PR #368, opened 2018-12-11
Fix perf tests work on Wine

chambm:
- changed PathEx.GetDownloadsPath() to check SKYLINE_DOWNLOAD_PATH environment variable
* changed PerfImportResultsNativeVsMz5Test from functional tests to unit tests which don't require GUI (so now they work on Wine)


chambm:
Should `File.Copy(_baseSkyFile, _skyFile, true);` come after the TestFilesZipPaths extraction?

bspratt:
:shipit:

PR #377, opened 2018-12-13
kyline: deal with a race condition in some tutorial tests (TestMs1Tut…

bspratt:
…orial, TestTargetedMSMSTutorial, TestTargetedMSMSTutorialAsSmallMoleculeMasses, TestTargetedMSMSTutorialAsSmallMolecules) that could result in irtStandard being audit logged as either "" (empty) or "MISSING" (null) depending on whether UI events had time to fire during the test. It should never be "MISSING".


pwiz_tools/Skyline/Model/Irt/IrtStandard.cs
```diff
@@ -339,7 +339,7 @@ public IrtStandard(string name, string skyFile, IEnumerable<DbIrtPeptide> peptid
         public string Name { get; private set; }
         public ImmutableList<DbIrtPeptide> Peptides { get; private set; }
 
-        public string AuditLogText { get { return Equals(this, EMPTY) ? AuditLog.AuditLogStrings.None : Name; } }
+        public string AuditLogText { get { return Name; } }
```
Tobias-Rohde: Now the audit log will store the actual localized string, but we want to store the invariant form, namely {2:None}.
So you should return

`AuditLogParseHelper.GetParseString(ParseStringType.audit_log_string, nameof(AuditLog.AuditLogStrings.None))`

PR #378, opened 2018-12-13
Fix WIFF issues

chambm:
- fixed errors when opening multiple WIFF2s at the same time
- addressed issue with WiffFile's MassSpectrometerSample not being disposed (unverified, no repro)


chambm:
@brendanx67 Do you want this merged to 4.2?

PR #381, opened 2018-12-14
Centralize file type/extension handling in ReaderList

chambm:
- fixed reading Waters data with electromagnetic spectra (PDA/UV)
- fixed MSConvertGUI errors when typing/pasting source paths into the filepath text box
* added Reader::getFileExtensions(), ReaderList::getTypes(), ReaderList::getFileExtensions(), and ReaderList::getFileExtensionsByType() so that applications like MSConvertGUI can ask the ReaderList what formats are supported rather than hard-coding them
* split Reader_Bruker into one dummy reader per Bruker format (FID, BAF, TDF, YEP, U2); Reader_Bruker::identify() now returns a value matching one of those dummy readers, which allows for filtering an open file dialog according to Reader::getType()
* changed IDPicker's default source file extension list to read from ReaderList::getFileExtensions()
* added operator- for std::set in Container.hpp


pwiz/data/msdata/ReaderTest.cpp
```diff
@@ -249,6 +251,30 @@ void testIdentifyFileFormat()
...
+                                  "Shimadzu LCD", "Thermo RAW", "UIMF", "Waters RAW", "Waters UNIFI" };
+    auto expectedButNotFound = expectedTypeSet - readerTypeSet;
+    auto foundButNotExpected = readerTypeSet - expectedTypeSet;
+    unit_assert_operator_equal(set<string>(), expectedButNotFound);
+    unit_assert_operator_equal(set<string>(), foundButNotExpected);
+
+    auto extByType = readerList.getFileExtensionsByType();
+    unit_assert_operator_equal(expectedTypeSet.size(), extByType.size());
+    unit_assert_operator_equal(2, extByType["ABSciex WIFF"].size());
+    unit_assert_operator_equal(".wiff", extByType["ABSciex WIFF"][0]);
```
brendanx67: Can we drop the AB? They are no longer using AB

pwiz/data/vendor_readers/ABI/Reader_ABI.hpp
```diff
@@ -62,6 +62,7 @@ class PWIZ_API_DECL Reader_ABI : public Reader
                       const Config& config = Config()) const;
 
     virtual const char * getType() const {return "ABSciex WIFF";}
```
brendanx67: Let's drop the AB

pwiz/data/vendor_readers/ABI/T2D/Reader_ABI_T2D.hpp
```diff
@@ -62,6 +62,7 @@ class PWIZ_API_DECL Reader_ABI_T2D : public Reader
     }
 
     virtual const char * getType() const {return "ABSciex T2D";}
```
brendanx67: Drop AB

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
```diff
@@ -135,9 +135,12 @@ PWIZ_API_DECL SpectrumPtr SpectrumList_Waters::spectrum(size_t index, DetailLeve
     int msLevel;
     CVID spectrumType;
     translateFunctionType(WatersToPwizFunctionType(rawdata_->Info.GetFunctionType(ie.function)), msLevel, spectrumType);
+
```
brendanx67: What does this have to do with the file extensions?

pwiz/utility/bindings/CLI/msdata/MSDataTest.cpp
```diff
@@ -198,6 +199,23 @@ void testCatchAndForward()
 }
 
 
+void testReader()
+{
+    auto typeSet = gcnew System::Collections::Generic::HashSet<String^>(ReaderList::FullReaderList->getTypes());
+    unit_assert(typeSet->Contains("ABSciex WIFF"));
```
brendanx67: Drop the AB

pwiz/data/vendor_readers/ABI/T2D/Reader_ABI_T2D.hpp
chambm: Not sure about this one in particular. T2D is used for the old ABI TOF-TOF 4700/4800/5800. Maybe it should actually be "ABI T2D" instead.

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
chambm: It doesn't. It fixes a crash a user got converting their data with a PDA function. It makes no difference to MS data. Any particular reason I need to separate it out? It would have crashed Skyline too if it tried to import this data.

brendanx67: Okay. Fine. Looks like it is 2 separate commits. Just don't squash when you merge. Thanks for clarifying.

pwiz/data/vendor_readers/ABI/T2D/Reader_ABI_T2D.hpp
brendanx67: Okay. Anything that is ABSciex needs to become just Sciex or SCIEX, but if it is an old system pre-ABSciex, then maybe it is okay to make it ABI. Is the system supported by anyone? Is there a web page for it? If so, who supports it? What does the web page call it?

chambm: It looks like the 4700 and 4800 went to Thermo with the Applied Biosystems acquisition (https://assets.thermofisher.com/TFS-Assets/LSG/manuals/cms_041368.pdf). Meanwhile Sciex further developed the TOF-TOF into the 5800, which is no longer in production but has a page on their site (https://sciex.com/products/mass-spectrometers/tof/tof-systems/tof/tof-5800-system). They all use the same Oracle database backend and T2D file export. Murky waters indeed. "AB/Sciex" might be ok (either/or), but not "AB Sciex" (like the name of a merged company)

brendanx67: Okay. Agreed. "AB/Sciex" sounds like the way to go. Thanks for the detailed look.

PR #382, opened 2018-12-14
Skyline: add support for 14C labeling (as C" in Skyline-speak)

bspratt:


pwiz_tools/Skyline/Model/SequenceUtil.cs
```diff
@@ -1365,6 +1365,7 @@ public static string GetHeavyFormula(char aa, LabelAtoms labelAtoms)
             ALL_LABEL_SUBSTITUTIONS = ImmutableList.ValueOf(new[]
         {
             Tuple.Create(LabelAtoms.C13, BioMassCalc.C, BioMassCalc.C13),
+            Tuple.Create(LabelAtoms.C14, BioMassCalc.C, BioMassCalc.C14),
```
brendanx67: Let's just not do this. We don't have LabelAtoms.O17 here. This is just a short-cut to avoid needing to write out a more verbose function like:

5O" - 5O

i.e. Add 5 x 17O atoms to replace 5 x 16O atoms. This was the original implementation in Skyline before I added the checkboxes to denote simply labeling all atoms in the amino acid.

pwiz_tools/Skyline/Model/DocSettings/Modification.cs
```diff
@@ -175,6 +176,8 @@ public StaticMod(string name, string aas, ModTerminus? term, LabelAtoms labelAto
         [Track]
         public bool Label13C { get { return (LabelAtoms & LabelAtoms.C13) != 0; } }
         [Track]
+        public bool Label14C { get { return (LabelAtoms & LabelAtoms.C14) != 0; } }
```
brendanx67: I think, let's not have this shorthand, as we do not have a shorthand for labeling all O as 17O. We only have the shorthand for labeling all O as 18O.

pwiz_tools/Skyline/Model/SequenceUtil.cs
bspratt: Yes, I was just following a pattern without thinking too hard about it, I'll back out all the labelC14 stuff,

Issue #384, opened 2018-12-18
MSConvert: mzXML dataProcessing elements fail schema validation


austinkeller:
### Context
I'm testing schema validation of the output mzML and mzXML's from msconvert and finding that the mzXML writer produces validation errors when multiple processing steps occur. This seems like an issue with the mzXML 3.2 schema itself that disallows multiple occurrences of <software/> within a <dataProcessing/> group. Based on the code in pwiz/data/msdata/Serializer_mzXML.cpp, this seems to be an intentionally flexible field where processingOperation or comments can serve to document the processing done.
https://github.com/ProteoWizard/pwiz/blob/c71da0d84efd387920cd99ff7577ba25f4dcb2bc/pwiz/data/msdata/Serializer_mzXML.cpp#L358-L385
However, the schema itself seems to enforce unusually rigid

To reproduce, I converted a thermo raw file with the following msconvert config file:
```
mzXML=true
zlib=true
mz64=true
inten64=true
simAsSpectra=true
filter="peakPicking vendor msLevel=1-2"
filter="scanNumber 22289-22486"
```

This produced an mzXML containing the following lines:
```
<dataProcessing centroided="1">
    <software type="conversion" name="ProteoWizard software" version="3.0.18342"/>
    <processingOperation name="Conversion to mzML"/>
    <software type="processing" name="ProteoWizard software" version="3.0.18342"/>
    <comment>Thermo/Xcalibur peak picking</comment>
</dataProcessing>
```

Running a validator on the full mzXML with the appropriate schema gives the following validation error:
```
austin@austin-vm-ubuntu:~/gdrive/data/20181208_fix_demux_mzXML_schema$ xmllint --schema raw/mzXML_schema/schema_revision/mzXML_3.2/mzXML_idx_3.2.xsd data/01_trimmed/23aug2017_hela_serum_timecourse_4mz_narrow_1.mzXML --noout
data/01_trimmed/23aug2017_hela_serum_timecourse_4mz_narrow_1.mzXML:20: element software: Schemas validity error : Element '{http://sashimi.sourceforge.net/schema_revision/mzXML_3.2}software': This element is not expected. Expected is one of ( {http://sashimi.sourceforge.net/schema_revision/mzXML_3.2}processingOperation, {http://sashimi.sourceforge.net/schema_revision/mzXML_3.2}comment ).
data/01_trimmed/23aug2017_hela_serum_timecourse_4mz_narrow_1.mzXML fails to validate
```

[...]

Issue #359, opened 2018-12-07
MSConvert: Demultiplexing gives mzML and mzXML output that violate the schema/spec


austinkeller:
### Context
Brian Searle reported that OpenSWATH interprets the wrong isolation scheme of demuxed files. This appears to be because the selected ion m/z is not updated. Brian confirmed that manually updating the selected ion m/z values with centers corresponding to the synthetic demuxed windows fixes the import issue for OpenSWATH.

Here is an example of a line of the mzML file that should be updated:
```
<cvParam cvRef="MS" accession="MS:1000744" name="selected ion m/z" value="500.4774" unitCvRef="MS" unitAccession="MS:1000040" unitName="m/z"/>
```

Brian provided an awk script written to repair his mzML:
```
#!/usr/bin/awk -f
{
	if ($0 ~ /<processingMethod order=\"2\">/) {
		# fix the missing softwareRef attribute in prism
		print "        <processingMethod order=\"2\" softwareRef=\"pwiz\">";

	} else if ($0 ~ /<spectrum index=/) {
		# fix the scan numbers so they are consecutive, remember to save them in ids[]
		current=$0;
		split($0, a, "\"");
		num=a[2]+1;
		id=a[4];
		split(id, b, "scan=");
		idpre=b[1];
		newid=idpre"scan="num;
		ids[id]=newid;
		print a[1] "\"" a[2] "\"" a[3] "\"" newid "\"" a[5] "\"" a[6] "\"" a[7];

	} else if ($0 ~ /accession=\"MS:1000827\"/) {
		# grab the correct scan center, save it as target
		split($0, a, "\"");
		target=a[8];
		print $0;

	} else if ($0 ~ /accession=\"MS:1000744\"/) {
		# insert the previous target to correct the incorrect center
		split($0, a, "\"");
		print a[1] "\"" a[2] "\"" a[3] "\"" a[4] "\"" a[5] "\"" a[6] "\"" a[7] "\"" target "\"" a[9] "\"" a[10] "\"" a[11] "\"" a[12] "\"" a[13] "\"" a[14] "\"" a[15];

	} else if ($0 ~ /<precursor spectrumRef=/) {
		# fix the scan index to reflect the new scan numbers
		# for some reason, these DO NOT include the "demux=0" tag. Thanks Austin!
		split($0, a, "\"");
		key=a[2]" demux=0";
		print a[1] "\"" ids[key] "\"" a[3];

[...]

PR #390, opened 2018-12-19
MSConvert: feature/fix demux mzML and mzXML schema compliance

austinkeller:
### Issues Resolved

- https://github.com/ProteoWizard/pwiz/issues/359
- https://github.com/ProteoWizard/pwiz/issues/384
- MSConvert "checking source file checksums" status message was missing a newline

### Possible Next Steps

- Add a new CV term for demultiplexing to the .obo standard, rather than just using a generic "data processing" category
- Evaluate utility of the demux= and originalScan= attributes in the spectrum id of demultiplexed spectra. Pros: They are useful for testing and validating demultiplexing performance. Cons: Removes us from producing spectra that look as if they were simply collected at twice the scan rate, somewhat opening potential for incompatibility.


austinkeller:
Hey @chambm, while I was fixing issues with demultiplexed not conforming to the mzML or mzXML schema I found that applying any filter resulted in a non-conforming mzXML file. So I had to update the Serializer_mzXML to produce any valid mzXMLs. I tried to reason out the original intent of the schema design the best I could. Otherwise the changes are just specific to demultiplexing.

PR #402, opened 2019-01-04
iRT for small molecules

nickshulman:
Here's my work making it so that you can have iRT for small molecules.

I have tried to make as little change to the UI as possible, unlike my goals a few weeks ago.

The key to making this work is a class called "TargetResolver". A TargetResolver gets initialized with all of the molecules in a SrmDocument, as well as in an iRT database, and it decides how to represent molecules as strings. If the molecule's ToString() uniquely identifies it, then that's what's used, otherwise, it's that complicated thing with number signs in it.

Then, there's another class called "DataGridViewTargetColumn" which uses the TargetResolver to display the molecule in a grid.


bspratt:
I wanted to build this locally and was confused that I couldn't switch to the branch in question. I finally realized that the branch for this PR is in your private repo - why?

nickshulman:
These changes live in the "nickshulman" repository so that when I eventually do my squash and merge, the previous revisions of the files I've changed don't end up taking up space in the real repository.

You can add https://github.com/nickshulman/pwiz as another remote to your checkout:
TortoiseGit>Settings>Remote

chambm:
It's a legit concern. Although I think it would mainly be useful to prevent intermediate changes to large files from bloating the main repo. If you do a squash and merge when both branches are in main and delete the branch, the old commits get orphaned, don't they? So they wouldn't show up in a normal log/history view. The main disadvantage of doing it in your own fork is keeping it up to date with upstream. What's your process for that Nick?

nickshulman:
Right. Intermediate changes to large files.

My process for merging from master is the same as if I were pushing to the real repository.
The repository on my computer has two remotes: "origin" (https://github.com/ProteoWizard/pwiz) and "nickshulman" (https://github.com/nickshulman/pwiz).

So, I frequently fetch and merge from "origin/master", but I only ever push to "nickshulman".

PR #430, opened 2019-01-29
Be more helpful with setting obviously needed ion filter types when f…

bspratt:
…ull scan mode is enabled (we had some of this in the wizard already, but not in the Transition Settings dialog).


pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -895,6 +895,25 @@ public void CloseWizard(DialogResult result)
...
+            {
+                var list = TransitionSettingsControl.PeptideIonTypes.ToList();
+                if (e.PeptidePrecursorIonTypeEnabled.Value)
+                    list.Add(IonType.precursor);
+                else
+                    list.Remove(IonType.precursor);
+                if (list.Count > 0)
+                    TransitionSettingsControl.PeptideIonTypes = list.ToArray();
+            }
+            // TODO(bspratt): handle small mol ion types when this gets extended for UIModes
```
brendanx67: Can we make this a CONSIDER or FUTURE? Ideally, TODO should be for actionable work, and not future things we need to handle when something else changed.

pwiz_tools/Skyline/SettingsUI/FullScanSettingsControl.cs
```diff
@@ -303,6 +305,10 @@ private void comboPrecursorIsotopes_SelectedIndexChanged(object sender, EventArg
                 // Selection change should set filter m/z textbox correctly
                 comboPrecursorAnalyzerType.SelectedIndex = -1;
                 comboPrecursorAnalyzerType.Enabled = false;
+
+                // Any changes we may have made to isotope selection are moot, restore original state of "p" in isotope types
+                OnIonTypesChanged(new IonTypesChangedEventArgs() { PeptidePrecursorIonTypeEnabled = TransitionSettings.Filter.PeptideIonTypes.Contains(IonType.precursor) });
```
brendanx67: This is weird event handling code. I think the event should be more localized to what the FullScanSettingsControl knows about. It knows that MS1 filtering has been enabled or disabled. So the event should be MS1FilteringEnabledChanged with an event that tells the listener what the new state is. And there should only be one event fired for any change in this state. Then the listeners should check to see if they have any work to do based on this change. You are doing too much work on the event firing end to determine whether the listeners have work to do based on your foreknowledge of what they are going to be doing. This makes the code more tightly coupled than it needs to be.

PR #433, opened 2019-02-01
20190111 - AuditLog format changes

Tobias-Rohde:
- Made some changes to how log strings are passed
  to faster identify if strings contain expansion tokens
- Rewrote some of the hashing code for audit log entries
- Now writing english text for extra info & log messages
- Now also writing to auditlog:
     * document version to audit log root
     * undo_redo, summary, all_info elements instead of all message elements
     * root hash (hash of all message hashes)
- Added alert that shows the undo redo message of all audit log messages
   that changed (hashes differ)
- Now using base64 to store and compare hashes
- Added skyl xsd's to verify xml structure in tests
- Added an old log to audit log savings test to verify that
  old skyl format gets recognized
- Made Install.Version faster for dev builds as requested by Nick


pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -1224,17 +1224,18 @@ public byte[] GetAuditLogHash()
...
                 if (!string.IsNullOrEmpty(ExtraInfo))
+                {
                     blockHash.ProcessBytes(enc.GetBytes(LogMessage.ParseLogString(ExtraInfo, LogLevel.all_info,
                         CultureInfo.InvariantCulture)));
+                }
 
                 blockHash.ProcessBytes(enc.GetBytes(UndoRedo.ToString(CultureInfo.InvariantCulture)));
                 blockHash.ProcessBytes(enc.GetBytes(Summary.ToString(CultureInfo.InvariantCulture)));
                 _allInfoNoUndoRedo.ForEach(l => blockHash.ProcessBytes(enc.GetBytes(l.ToString(CultureInfo.InvariantCulture))));
-                // TODO: include other information in the hash?
```
rita-gwen: I believe there is reason element in the audit_log_entry schema. I think it should be included into the hash as well.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -1224,17 +1224,18 @@ public byte[] GetAuditLogHash()
             using (var sha1 = new SHA1CryptoServiceProvider())
             {
                 var enc = Encoding.UTF8;
-                BlockHash blockHash = new BlockHash(sha1);
+                var blockHash = new BlockHash(sha1);
                 blockHash.ProcessBytes(enc.GetBytes(User));
                 if (!string.IsNullOrEmpty(ExtraInfo))
+                {
                     blockHash.ProcessBytes(enc.GetBytes(LogMessage.ParseLogString(ExtraInfo, LogLevel.all_info,
```
rita-gwen: I have to parse this string already to include expanded extra info element. Wouldn't it be better to use that value instead of parsing it again?

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
Tobias-Rohde: Good catch, added that to the hash. The reason will be hashed right after its corresponding message. FYI, there is only one reason for the undo redo message, the summary message, and the first all info message (which is simply the undo redo message) and all remaining all info message can have their own reasons.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
Tobias-Rohde: I did this to avoid having to worry about hashing potentially modified english messages. Originally I was planning on not even reading the english strings from the XML since I don't use those in Skyline. I guess the resource name and its parameters could also be modified, but I think we should always prefer the data from the <message> and <name> elements.

When reading in the audit log and checking which log messages were modified I'll just construct a new audit log list and update the English strings for all entries that were modified. I also realized that some of my code that updates those English strings violates immutability, I'll fix that too.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -134,27 +158,13 @@ public void WriteXml(XmlWriter writer)
 
         private enum EL
         {
-            document_hash
+            document_hash,
+            en_root_hash
```
brendanx67: Let's drop the en_ part. It is not relevant that English is included in the hash. We have no ja_root_hash or any other language.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -1049,8 +1118,13 @@ private enum ATTR
...
         {
             message,
+            undo_redo,
+            summary,
+            all_info,
             reason,
-            extra_info
+            extra_info,
+            en_extra_info,
+            en_hash,
```
brendanx67: Again, let's drop en_ since there is only one hash and no plan to have other language-specific hashes (ja_hash, etc.)

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -1104,20 +1181,150 @@ public void ReadXml(XmlReader reader)
...
             var list = new List<LogMessage>();
-            while (reader.IsStartElement(EL.message))
-                list.Add(reader.DeserializeElement<LogMessage>().ChangeLevel(LogLevel.all_info));
+            while (reader.IsStartElement(allInfoEnum))
+                list.Add(reader.DeserializeElement<LogMessage>(allInfoEnum).ChangeLevel(LogLevel.all_info));
 
             AllInfo = list;
 
+            // Think about how and if we want to store these english strings
+            EnExtraInfo = reader.IsStartElement(EL.en_extra_info)
```
brendanx67: Normally with something like this, I would not use the stored value since it can be recalculated, but I guess since you are checking the hash to make sure it is exactly the same as what was written, it may not matter. What happens if the hash doesn't match? It seems like you could test whether the hash would match with the generated text to see if someone only tampered with the english text.

pwiz_tools/Skyline/Model/AuditLog/AuditLogStrings.Designer.cs
```diff
@@ -1,1643 +1,1667 @@
...
-        ///   Looks up a localized string similar to Upgraded background proteome &quot;{0}&quot;.
-        /// </summary>
-        public static string upgraded_background_proteome {
-            get {
-                return ResourceManager.GetString("upgraded_background_proteome", resourceCulture);
-            }
-        }
-    }
-}
+﻿//------------------------------------------------------------------------------
```
brendanx67: Why is this a complete diff of the entire file?

pwiz_tools/Skyline/Model/AuditLog/AuditLogStrings.resx
```diff
@@ -644,4 +644,16 @@
   <data name="start_log_existing_doc" xml:space="preserve">
     <value>Start of audit log for already existing document</value>
   </data>
+  <data name="AuditLogList_ReadFromFile_The_following_audit_log_entries_were_modified" xml:space="preserve">
```
brendanx67: Such a small change. Weird that the .designer.cs file changes completely.

pwiz_tools/Skyline/Util/CreateHandleDebugBase.cs
```diff
@@ -263,13 +263,5 @@ private void FormCreateHandle()
                 }
             }
         }
-
```
brendanx67: This seems like an unrelated change.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
Tobias-Rohde: Okay. For "en_extra_info", would "expanded_extra_info" be fine? We already have an element called "extra_info".

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
Tobias-Rohde: If the hashes are different I am already recalculating the hashes of the log messages and the extra info, I haven't push that code yet. (I actually just set them to null so that they only get calculated when we actually use them for the first time)

pwiz_tools/Skyline/Util/CreateHandleDebugBase.cs
Tobias-Rohde: I'll put this back in and remove it in a different pull request.

pwiz_tools/Skyline/Model/AuditLog/AuditLogStrings.Designer.cs
Tobias-Rohde: Fixed it.  I am still unsure as to how the line endings changed in the first place

PR #444, opened 2019-02-12
Skyline/work/20190124 improved cli

brendanx67:
Improves the Skyline command-line CommandArgs processing moving all help text from the Word doc to console output available through --help


brendanx67:
@nickshulman, @bspratt Can either of you reproduce the test failures? They look very clearly related to my changes. I just can't reproduce them either on the machine where I did the development or on another machine where I pulled the branch. They don't really look like the kinds of errors that would be system dependent, but it will be much harder to debug them if I can't find a way to reproduce them.

bspratt:
I'll have a look.

brendanx67:
@bspratt pointed out that it is only x86 release build that is failing. Taking a closer look.

bspratt:
It would be nice if the commandline help information went to stdout or stderr instead of console. I wanted to capture it to read elsewhere but redirection doesn't work. It's possible of course to select and copy in the cmd.exe window, but it's a bit nonstandard.

bspratt:
There are a handful of places where we could be more general as regards small molecules:

	Line 161: │--import-document-merge-peptides    │Matching peptides are merged if used, otherwise not.                            │
	Line 161: │--import-document-merge-peptides    │Matching peptides are merged if used, otherwise not.                            │
	Line 178: Importing peptide searches
	Line 180: │--import-search-file=path/to/file      │Import a peptide search results file into the open document, building a      │
	Line 182: │                                       │multiple files. Use –import-fasta argument to add matched peptides as        │
	Line 186: │                                       │library from peptide search results files, where 1 is for highest confidence │
	Line 192: │--import-search-include-ambiguous      │Prevent spectra with multiple ambiguous peptide matches from being discarded │
	Line 203: │--import-transition-list=path/to/file │Import a simple transition list with Q1, Q3 and peptide sequence.  Also       │
	Line 213: │--irt-standards-group-name=<name>     │The name of a protein or peptide list containing the iRT standards within an  │

brendanx67:
Can you propose changes you would like to see? These are all from the current documentation that is committed to the project.

PR #445, opened 2019-02-12
Skyline: performance improvements on import of lengthy small molecule…

bspratt:
… transition lists. A 12,000 line mz-only list now imports in 30 seconds where it used to take 2 minutes 52 seconds (nearly 6x faster)

### DIFF
--- a/pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
+++ b/pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
@@ -104,6 +104,10 @@ namespace pwiz.Skyline.Model
         public SrmDocument CreateTargets(SrmDocument document, IdentityPath to, out IdentityPath firstAdded)
         {
             firstAdded = null;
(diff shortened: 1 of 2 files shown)


bspratt:
Probably the most interesting part of this is the use of and slight change to SRMDocument.DeferSettingsChanges. We were running through that growing document again and again and it bogged things down.

nickshulman:
I think you should probably just build up a List of PeptideGroupDocNodes, and only call SrmDocument.ChangeChildren at the very end.
DeferSettingsChange gets used by the DocumentGrid because there is a bunch of completely unrelated bits of code that only know how to modify a document. But, in this case, where you have all of the code in one place it's probably straightforward to change it so that the document change only happens once at the very end.
That said, there's nothing wrong with using DeferSettingsChange, if the code happens to be difficult to rearrange like that.

bspratt:
I considered that, but as existing performance is fine for moderately sized imports I didn't want to get too in depth with the refactoring.

PR #448, opened 2019-02-21
Skyline/work/20190212 smallmol optimization libs

bspratt:


pwiz_tools/Skyline/Model/Peptide.cs
```diff
@@ -702,6 +702,7 @@ public override int GetHashCode()
     /// </summary>
     public class Target : IComparable<Target>, IEquatable<Target>, IAuditLogObject
     {
+        public static Target EMPTY = new Target(string.Empty);
```
nickshulman: Should be readonly

PR #463, opened 2019-03-05
20180301 zulu timestamp auditlog

bspratt:


pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -427,16 +426,22 @@ public bool IsRoot
             get { return ReferenceEquals(this, ROOT); }
         }
 
-        private AuditLogEntry(SrmDocument document, DateTime timeStamp, string reason,
+        private AuditLogEntry(DateTime timeStamp, string reason,
```
Tobias-Rohde: It's very nice how we don't have to pass around the document around anymore just to get the FormatVersion in the AuditLogEntry constructor. Originally I must've been using the document for other things in the constructor since otherwise I could've just passed around the format version.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -1344,7 +1346,7 @@ public byte[] GetAuditLogHash()
         // For hash creation
         private byte[] GetTimeStampBytesForHash()
         {
-            var bytes = BitConverter.GetBytes(TimeStamp.ToFileTime() / TimeSpan.TicksPerSecond); // We only serialize to hour:min:sec precision, lose the ticks so we can roundtrip
```
rita-gwen: If you are concerned about cross-platform stability then why to use a system-specific timestamp representation?

bspratt: You're right, that should really be in terms of unix epoch. Nice catch!

So now what? I guess cross platform code has to shift to windows file time before hashing. That's embarrassing.

rita-gwen: Yes, I'll have to reproduce file time calculation in Java. It's not that hard, but it introduces dependency on logic that we cannot control.

bspratt: Not sure I follow you there - what can't we control about our own hashing code? Or do you mean we can't control the windows file time epoch definition? That seems deeply unlikely to change.

rita-gwen: Yes, the later. I agree that it's stable enough so we don't need a fix here. It will be a bit awkward, but not a biggie.

PR #469, opened 2019-03-07
Prevent the user choosing negative charges for peptide precursors

nickshulman:
I am still working on writing a test for this change, but I wanted to get feedback about removing the TabControl overloads from MessageBoxHelper and just using my new method "MessageBoxHelper.EnsureControlTabPagesVisible" instead.

Is there any reason that we have all of these MessageBoxHelper overloads that take a TabControl and tab page index?
They don't really work right when there are multiple tab pages nested inside of each other, like on the Transition Settings Filter.

This change removes all of the overloads in MessageBoxHelper that take a tab control and page index.
Instead, MessageBoxHelper always calls "EnsureControlTabPagesVisible" which ensures that the correct tab is selected.
When the incorrect tab page is selected, "GetControlMessage" would return the wrong string, since GetControlMessage tries to find a Label that is not hidden.

Also, this change makes it so that you get an error message if you try to put a negative number in the list of Transition Settings Filter Peptide charges.

Also, SkylineWindow.ChangeSettingsMonitored was calling "ModifyDocument" from a background thread. ModifyDocument tries to bring up a message box if there is an error, which does not work. So, I made it so that ChangeSettingsMonitored calls a new method "ModifyDocumentOrThrow" which does not try to bring up a MessageDlg.


bspratt:
"They don't really work right when there are multiple tab pages nested inside of each other, like on the Transition Settings Filter." is a key realization, I'm sure we didn't consider this when we added that sub-tab.

bspratt:
Looks good from just reading the code. I'd probably want to see it in operation to be sure.

brendanx67:
"Is there any reason that we have all of these MessageBoxHelper overloads..?" Answer: It worked at the time it was written and I didn't think of the more general (and future-proof) implementation. But, it sounds like an improvement.

PR #479, opened 2019-03-13
Fix problem where modifications with negative mass cannot be found in…

nickshulman:
… iRT databases.

(Reported by Anatoly)


brendanx67:
Looks good to me.

PR #481, opened 2019-03-15
Skyline/work/20190315 mass only im predicitor

bspratt:
John F discovered that deriving ion mobility from loaded mass spec results didn't work for molecules defined by mass only. This fixes the problem and allows TargetResolver to work properly with molecules defined by mass only by tweaking the LibraryKey class.

### DIFF
--- a/pwiz_tools/Skyline/Model/Lib/LibKeyIndex.cs
+++ b/pwiz_tools/Skyline/Model/Lib/LibKeyIndex.cs
@@ -162,8 +162,10 @@ namespace pwiz.Skyline.Model.Lib
                 var thatPeptideKey = thatItem.LibraryKey as PeptideLibraryKey;
                 if (thatPeptideKey == null)
                 {
-                    result.Add(thatItem.LibraryKey);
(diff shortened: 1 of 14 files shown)


pwiz_tools/Skyline/Model/Lib/LibKeyIndex.cs
```diff
@@ -162,8 +162,10 @@ public IList<LibraryKey> MergeKeys(LibKeyIndex that)
                 var thatPeptideKey = thatItem.LibraryKey as PeptideLibraryKey;
                 if (thatPeptideKey == null)
                 {
-                    result.Add(thatItem.LibraryKey);
-                    nonPeptideKeySet.Add(thatItem.LibraryKey);
+                    if (nonPeptideKeySet.Add(thatItem.LibraryKey))
```
nickshulman: Is this necessary? Are there duplicates in your LibKeyIndex?

bspratt: Not really, no - it's just belt and suspenders since we're copying to a container that doesn't enforce uniqueness.

PR #482, opened 2019-03-18
Fix problem where pasting into the document grid can result in an Aud…

nickshulman:
…it Log Entry which has the wrong hash value the next time you try to open the Skyline document.


pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -660,7 +660,11 @@ public AuditLogEntry ChangeUndoAction(Action<AuditLogEntry> undoAction)
 
         public AuditLogEntry ChangeExtraInfo(string extraInfo)
         {
-            return ChangeProp(ImClone(this), im => im.ExtraInfo = extraInfo);
+            return ChangeProp(ImClone(this), im =>
+            {
```
brendanx67: Should the setter for ExtraInfo do this? Are their other cases where it makes sense to set ExtraInfo without doing this?

PR #486, opened 2019-03-20
Work in progress: UI modes in the grid

nickshulman:
Here is my work for having the set of available columns and the column names change depending on the uimode that the user is in.

This work is almost done. One thing that I have not figured out is whether we should do anything different with the set of Reports you get to start with ("Peptide Ration Results", "Small Molecule Transition List", etc.)-- whether they should always be visible, or whether the set of reports that you see in any particular mode gets filtered down based on the uimode that you are in.

The properties on the row objects (e.g. "Protein" etc. in the "pwiz.Skyline.Model.Databinding.Entities" namespace) can have attributes which only take effect in certain mode.
For instance, the "Name" property now looks like this:
        [ProteomicDisplayName("ProteinName")]
        [InvariantDisplayName("MoleculeListName")]
        public string Name
The "ProteomicDisplayName" attribute takes precedence over the "InvariantDisplayName" attribute. The logic for this is enforced in DataSchema.FilterAttributes which puts "ProteomicDisplayName" ahead of "InvariantDisplayName" because the latter is the base class of the former.

Anyway, I am going to keep working on polishing this up, but I wanted to give people a chance to look at it.


pwiz_tools/Skyline/Util/UiModes.cs
```diff
@@ -0,0 +1,88 @@
...
+        public const string SMALL_MOLECULES = "small_molecules";
+        public const string MIXED = "mixed";
+
+        public static readonly IEnumerable<string> ALL =
+            ImmutableList.ValueOf(new[] {PROTEOMIC, SMALL_MOLECULES, MIXED});
+
+        public static string FromDocumentType(SrmDocument.DOCUMENT_TYPE documentType)
+        {
+            switch (documentType)
+            {
```
bspratt: Does this take into account DOCUMENT_TYPE.none (ie empty document?)

bspratt:
Maybe not all that important, but potentially confusing for maintainers:
[ProteomicDisplayName("ProteinName")]
[InvariantDisplayName("MoleculeListName")]
is opposite to my convention of "invariant" meaning "never gets the peptide->molecule translation treatment, regardless of UI mode or document type". That's coming from the philospohical viewpoint that the Skyline UI is fundamentally proteomic.

bspratt:
Also, I'm unclear on how I can download and build this to play with given that it's not in the same repo the rest of us are using?

pwiz_tools/Skyline/Util/UiModes.cs
nickshulman: The mode that gets passed in is nearly always SkylineWindow.ModeUI, which I believe would never be "none". If it were "none", then the user would see the Mixed UI mode.

nickshulman:
> Maybe not all that important, but potentially confusing for maintainers:
> [ProteomicDisplayName("ProteinName")]
> [InvariantDisplayName("MoleculeListName")]
> is opposite to my convention of "invariant" meaning "never gets the peptide->molecule translation treatment, regardless of UI mode or document type". That's coming from the philospohical viewpoint that the Skyline UI is fundamentally proteomic.

It's been called the Invariant Name for a while because it's the column caption that you get if you choose "Invariant" as the language when you do "File > Export > Report".

It could be that we should come up with a better name for the language than "Invariant". It's really more like "computerese" since the column captions have no spaces in them, and all numbers are formatted to as many decimals as possible.

pwiz_tools/Skyline/Util/UiModes.cs
bspratt: Correct, there's never a "none" UI mode. But for that "nearly always" part it's probably better to default to Proteomic to avoid frightening the majority of our users who are traditionally proteomic.

PR #488, opened 2019-03-21
Skyline: Fix immutability loophole where ratios can get changed direc…

brendanx67:
…tly on a transition result

- Profiling gets mentioned in the comment, but it already looks like this is protected for only when the ratios are changing


brendanx67:
This causes an embarrassing inability to undo ratio changes due to changing Global Standards that I have run into several times during hands-on sessions at courses.

PR #493, opened 2019-03-28
Bug/fix plgs final fragment mod parser

chambm:
This PR moves the test final_fragment.csv file outside the inputs tarball.


chambm:
I think that loop is a questionable heuristic. But I just realized that it should always be running (the `mods.find()` call should always return `end()`) because the mod position string is part of the lookup but not in the map. And because of that, it is indeed finding the wrong mass for my added `SILAC 13C(6) 15N(4)` mod. Without exact matching, making it find the longest match instead of the shortest would probably make it break in some other cases (probably more contrived ones though).

nickshulman:
> Without exact matching, making it find the longest match instead of the shortest would probably make it break in some other cases (probably more contrived ones though).

It sounds like you're saying that the first thing we should do is find the location of the last open parenthesis, and then try to find an exact match for everything up to one space before that parenthesis?

chambm:
Yep I think that would be safest. A longest prefix matching heuristic just doesn't make sense to me in this case. An approximate string matching algorithm might be usable but would be more effort than it's worth IMO.

PR #494, opened 2019-03-29
WIP: Pull SampleID from Thermo RAW files as annotation in Skyline

chambm:
* added RawFile::getSampleID(), which is added to the MSData model as the MS_sample_name of a Sample element
* added MsDataFileImpl::GetSampleId(), which gets a nullable string from that MS_sample_name value

This is reference to [issue 628](https://skyline.ms/issues/home/issues/details.view?issueId=628):
> Jaffe lab has requested the ability to automatically set replicate attribute based on some properties in the Thermo RAW file header. ProteoWizard can write these properties to the MSData fileContent section as userParams if they are present and/or non-empty, where Skyline can look for them and use them appropriately.

The current branch has the requested property in the MSData model and an easy getter function in MSDataFileImpl.

@brendanx67 / @nickshulman Do you want me to go ahead and try to do the annotation part too?


chambm:
@nickshulman does approval mean you want to handle the Skyline side of this?

nickshulman:
> @nickshulman does approval mean you want to handle the Skyline side of this?

Yes, I should do the Skyline side of this. It will probably require changing our .skyd file format a little bit, since the easiest way to remember the value of this attribute that we read during chromatogram extraction is to store it in the .skyd file before eventually copying it into the .sky file.

PR #495, opened 2019-03-29
Skyline/work/20190320 transition level explicit values

bspratt:
Several explicit overrides that were formerly on the precursor level have been moved to the transition level where they more properly belong, so that individual fragments of a precursor can have individual overrides:

        CollisionEnergy  // For import formats with explicit values for CE
        DeclusteringPotential // For import formats with explicit values for DP
        SLens  // For Thermo
        ConeVoltage  // For Waters
        IonMobilityHighEnergyOffset// For import formats with explicit values for ion mobility


brendanx67:
Looks like it has a code inspection issue.

bspratt:
Skyline-Daily (left) vs revision

per_transition

PR #500, opened 2019-04-02
BiblioSpec i18n support

chambm:
- added support for non-ASCII filepaths to BiblioSpec
* added non-ASCII tests to Skyline's LibraryBuildTest
* removed `using namespace std` and `using namespace boost` from BiblioSpec code; the non-qualified cin/cout/cerr/fstream types now refer to boost::nowide's versions instead of std
* added `enable_utf8_path_operations()` function in Filesystem.hpp which adds UTF-8 capability to boost::filesystem::path; it must be called at the start of a program's `main()` function


chambm:
@bspratt @brendanx67 @nickshulman I added Skyline tests for BiblioSpec UTF-8 functionality. Needs review to determine if the scope of the testing goes too far (i.e. does it matter if we use all existing LibraryBuildTest tests to also validate UTF-8 functionality).

brendanx67:
Not sure I understand the question about "does it matter if we use all existing LibraryBuildTests" Where is that happening? I would feel better if I knew this had worked once. Makes me wonder if we should try building and running all tests in a project rooted below a path that contains far east characters.

chambm:
The command-line tests are indeed only for MaxQuant. Do you think I should move it into the test framework itself so that every test is run with CJK characters in the path? But the Skyline change I made gets used for almost all the tests in LibraryBuildTest.cs (all the ones that use libraryBaseName in their _libraryName).

The removal of `using namespace std` was a crucial change for using boost::nowide instead of std versions of various IO functions. I could have instead added full qualification for all those calls, but it would have taken more time and been more error prone for future maintenance (it's easy to forget using those fully qualified versions rather than using an unqualified cout, cin, cerr, etc).

brendanx67:
I think it would be great if you could at least come up with a way to run every test in a path with CJK characters. We don't have to run it that way every time, but before declaring this fixed, it would be good to run it that way once. That could be as simple as changing the inputs folder name from "inputs" to something CJK. But, yeah, just try it once at least.

brendanx67:
Fine with the "using namespace std" removal. Just commenting on why it gets a bit hard to see more important changes.

chambm:
> That could be as simple as changing the inputs folder name from "inputs" to something CJK

That was a good idea. It revealed some remaining issues with Mascot, MZ5, and the TSV reader. Mascot will now work with Unicode via a workaround (the Mascot Parser API doesn't support UTF-8), MZ5 will give a better error, and the TSV reader will work.

PR #501, opened 2019-04-03
Skyline/work/20190313 grid ui modes

nickshulman:
This is a new pull request for the same work in #486.

One thing I need to do is figure out what to do in terms of "Show Source Tab" (probably get rid of it) and "Show Hidden Fields" (probably make it into a button with a picture and get rid of the "Advanced" dropdown).
image

Also, need to address some points that Brian brought up in the old pull request.


bspratt:
This is looking good! I noticed this though:
image

PR #502, opened 2019-04-03
Skyline: A fairly common support question is "why won't this Skyline …

bspratt:
…file open?" when they are actually trying to open a .skyd file or somesuch.  Probably an artifact of Windows hiding file extensions. Try to work around it by finding a plausible matching .sky file when presented with a .sky? file for loading.


pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -285,6 +285,23 @@ public bool OpenFile(string path, FormEx parentWindow = null)
...
+            if (!path.EndsWith(SrmDocument.EXT))
+            {
+                var index = path.LastIndexOf(SrmDocument.EXT, StringComparison.Ordinal);
+                if (index > 0 && index == path.Length - (SrmDocument.EXT.Length + 1))
+                {
+                    // Looks like user picked a .skyd or .skyr or .skyl etc
+                    var likelyPath = path.Substring(0, index + SrmDocument.EXT.Length);
+                    if (File.Exists(likelyPath))
+                    {
+                        path = likelyPath;
```
brendanx67: Should we also throw a nicer error, if the .sky file does not exist but they have chosen a .skyd, .skyl, .skyr?

bspratt: Yeah, we should. Like this:
image

brendanx67: Not sure about the "and contain XML data". I would say "in XML format" but that is not true of a .sky.zip file. So maybe just leave it at the extensions expected.

bspratt: Can we go with "in XML format"? I wouldn't want to risk giving the impression that it was just the .ext that we were concerned with. And I was thinking that the user would understand that .zip isn't literally xml data, but that we expected to find xml data inside.

brendanx67: Okay.

bspratt: I mean, I can just imagine somebody renaming their .raw file to .sky in a frantic attempt to shoehorn it into Skyline without the bother of learning how to use it first.

bspratt: Approved, then?

pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -314,6 +383,11 @@ public bool OpenFile(string path, FormEx parentWindow = null)
             catch (Exception x)
             {
                 exception = x;
+                // Was that even a Skyline file?
+                if (!IsSkylineFile(path, out var explained))
+                {
+                    exception = new Exception(explained);
```
brendanx67: new IOException()

pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -204,7 +204,68 @@ private void openMenuItem_Click(object sender, EventArgs e)
             }
         }
 
-        public bool OpenSharedFile(string zipPath, FormEx parentWindow = null)
+        /// <summary>
```
brendanx67: Support functions go below the main function not above. Above is old-school for C compilers.

bspratt: On consideration, I realize that they're actually in the wrong file altogether - they're not specific to the UI and I've made them static methods of SrmDocument. This will be handy if we want to do something similar for the commandline (though I think scripters can probably be relied on to pass actual .sky files)

PR #503, opened 2019-04-03
Skyline: the Calibration Curve display needed a little work to says t…

bspratt:
…hings like "Use the Quantification tab on the Molecule Settings dialog to control the conversion of peak areas to concentrations." instead of "Use the Quantification tab on the Peptide Settings dialog to control the conversion of peak areas to concentrations." when needed.

image


bspratt:
We may be stumbling across these for a while, keep an eye out for thing like this...

PR #508, opened 2019-04-05
Improve OpenSWATH TSV error message

chambm:
- added 'Is this an OpenSWATH TSV file?' to the error message when expected columns aren't found in OpenSWATH .tsv file


chambm:
Agreed about the assumption. Is there really a difference between "Is this an OpenSWATH file?" and "Are you sure this is an OpenSWATH file?"? The important part in my mind is OpenSWATH TSV - implying to the user that this TSV file is being parsed an OpenSWATH TSV file, not any generic kind of TSV. We could make it more explicit by adding "Only .tsv files from OpenSWATH are supported"

bspratt:
To my eye there is, a difference, yes. "Is this an OpenSWATH file?" reads to me as "I suspect this is an OpenSWATH file" and  "Are you sure this is an OpenSWATH file?" reads to me as "I suspect this is NOT an OpenSWATH file".  Probably  "Only .tsv files from OpenSWATH are supported" is the better option.

PR #515, opened 2019-04-15
BPC and TIC for Bruker

chambm:
Also moved Bruker mzMLs out of tarball (but left raw data in tarball because it's 90 files for 5 raw data folders).

@bspratt : Brendan wants to restore the scan time / mobility filtering that you put in last year then reverted because it compromised the TIC and BPC trace extraction. By adding chromatograms to the Bruker readers, all readers have TIC chromatograms except Shimadzu and UNIFI. Only Sciex and Bruker have BPC. How hard would it be to make Skyline import these TIC chromatograms (rather than getting TIC from scans) so the performance filtering can be restored without losing the TIC trace? Brendan suggested an advanced scan extraction options dialog which would allow users to opt in to the slower, full extraction in order to guarantee they get BPC traces.


bspratt:
Changes seem fine. It shouldn't be too complicated to incorporate TIC and BPC that we don't derive ourselves, although I have to say it makes me vaguely nervous - is there any history of that being unreliable when available? Or did we just do it ourselves because it was convenient?

chambm:
I've never heard a complaint about it except perhaps that it doesn't distinguish between MS levels. So for MS1/MSn data you get very spiky traces. But that's what Skyline's doing now isn't it?

chambm:
We could add some basic sanity checking to alleviate your concerns, but I'm not sure where it belongs: make sure that the size of the TIC chromatogram is the same as the number of scans (but without filtering; the chromatogram doesn't care about preferMsLevelOnly for example).

bspratt:
I think we only look at MS1 for TIC, so yeah, that could be weird. OTOH I'm not sure anybody would really care. I have the impression that most Skyline users aren't terribly interested in TIC but I'll leave that call to Brendan.

bspratt:
>>  Brendan wants to restore the scan time / mobility filtering

@chambm Are you tasked with that as well?

chambm:
He didn't really specify. I think you'd be able to do that (much) more efficiently. But I'll make sure to review the code to understand how it's done.

bspratt:
Makes sense. Now I just gotta figure out what branch that was...

PR #517, opened 2019-04-15
BiblioSpec: tweaks to Shimadzu .mlb reader. Wasn't filtering for MS2 …

bspratt:
…only. Also, we now accept adducts "other(+)" and "other(-)" as meaning "[M+]" and "[M-]" respectively.


bspratt:
You're correct, the input file has other problems as well, which I'm discussing with them. But the change to filter out MS1 is still important, so that's what I'd like the blessing on. Thanks! Good point about referencing the support request, too.

PR #526, opened 2019-04-26
Skyline/work/20190415 per transition ims high energy offset fixes

bspratt:
1) per-transition ion mobility high energy offsets weren't being properly utilized in chromatogram extraction. 
2)  there was a display issue with not properly showing the per-transition ion mobility filter boundaries in  ion mobility heat maps.
3) high energy ion mobility offsets are now used for all-ions chromatogram extraction regardless of vendor.


pwiz_tools/Skyline/Model/Results/SpectrumFilterPair.cs
```diff
@@ -584,10 +597,62 @@ public override int GetHashCode()
...
                 hashCode = (hashCode*397) ^ FilterId;
+                hashCode = (hashCode * 397) ^ HighEnergyIonMobilityValueOffset.GetHashCode();
                 return hashCode;
             }
         }
 
         #endregion
     }
+
+    public class SpectrumFilterValues : IComparable<SpectrumFilterValues>, IComparable
```
nickshulman: This class only needs to exist if we are going to allow a SpectrumFilterPair to extract the same product m/z with two different high energy offsets.

However, if the m/z no longer uniquely identifies the chromatogram, then we probably need to add the high energy offset value to the class "ChromTransition" so that we will be able to use the correct chromatogram later on.

PR #527, opened 2019-04-26
Skyline: when trying to paste a seemingly well structure small molecu…

bspratt:
…le CSV into the Targets window, I was puzzled by a failure that turned out to be due to a trailing space in one of the column headers which prevented us from magically recognizing it. It's better, I think, to just trim() leading and trailing spaces rather than be pedantic about it.


pwiz_tools/Skyline/Util/Extensions/Text.cs
```diff
@@ -629,12 +629,13 @@ public void Initialize(TextReader reader, char separator, bool hasHeaders = true
             }
             for (int i = 0; i < fields.Length; ++i)
             {
-                FieldNames.Add(fields[i].Trim());
-                FieldDict[fields[i]] = i;
+                var trimmed = fields[i].Trim();
```
nickshulman: I would name this variable "fieldName" instead of "trimmed".

bspratt: sold!

PR #533, opened 2019-04-30
Skyline: adding a few Targets Window icons for non-peptide version of…

bspratt:
… IRT, IRT Library, Standard Library. Also fix a problem with not returning to same Molecule Settings tab when user reopens that dialog (seems to happen automagically normally, but as we hide tabs depending on UI mode it appears to need manual assistance).

molecule_irts


pwiz_tools/Skyline/Controls/SeqNode/PeptideTreeNode.cs
```diff
@@ -118,39 +118,30 @@ private static int GetTypeImageIndex(PeptideDocNode nodePep)
...
-                }
-                return (int)SequenceTree.ImageId.molecule;
+                                  : SequenceTree.ImageId.peptide_decoy); // Decoy has no small-molecule equivalent 
             }
             if (nodePep.GlobalStandardType == StandardType.IRT)
             {
                 return (int) (nodePep.HasLibInfo
-                                  ? SequenceTree.ImageId.peptide_irt_lib
-                                  : SequenceTree.ImageId.peptide_irt);
+                                  ? (nodePep.IsProteomic ? SequenceTree.ImageId.peptide_irt_lib : SequenceTree.ImageId.molecule_irt_lib)
```
brendanx67: Feels like this is getting a bit complicated. I wonder if there might be some way to have two parallel arrays for proteomic and non-proteomic, pick the right array at the top and then just index into it by another enum to get the right value.

pwiz_tools/Skyline/Skyline.cs
```diff
@@ -3303,7 +3303,11 @@ public void ShowPeptideSettingsUI(PeptideSettingsUI.TABS? tab)
 
         private void ShowPeptideSettingsUI(IWin32Window parent, PeptideSettingsUI.TABS? tab)
         {
-            using (PeptideSettingsUI ps = new PeptideSettingsUI(this, _libraryManager) { TabControlSel = tab })
+            using (PeptideSettingsUI ps = new PeptideSettingsUI(this, _libraryManager) { TabControlSel = tab ?? (PeptideSettingsUI.TABS)
```
brendanx67: Let's put this back inside PeptideSettingsUI in the constructor.

pwiz_tools/Skyline/Skyline.cs
```diff
@@ -3321,6 +3325,19 @@ private void ShowPeptideSettingsUI(IWin32Window parent, PeptideSettingsUI.TABS?
                     if (oldStandard != newStandard)
                         AddStandardsToDocument(newStandard, missingPeptides);
                 }
+
```
brendanx67: And this should go in something like OkDialog() or OnClosing() for the form. SkylineWindow is such a crazy long class already. Hate seeing logic like this leak out into it.

pwiz_tools/Skyline/Skyline.cs
bspratt: Done.

pwiz_tools/Skyline/Skyline.cs
bspratt: Done.

pwiz_tools/Skyline/Controls/SeqNode/PeptideTreeNode.cs
bspratt: I've created a small dictionary for images that have peptide/molecule versions.

pwiz_tools/Skyline/SettingsUI/PeptideSettingsUI.cs
```diff
@@ -85,10 +85,31 @@ public class QuantificationTab : IFormView { }
         private static readonly IList<int?> _quantMsLevels = ImmutableList.ValueOf(new int?[] {null, 1, 2});
         private readonly LabelTypeComboDriver _driverSmallMolInternalStandardTypes;
 
-        public PeptideSettingsUI(SkylineWindow parent, LibraryManager libraryManager)
+        public PeptideSettingsUI(SkylineWindow parent, LibraryManager libraryManager, TABS? selectTab)
         {
             InitializeComponent();
 
+            if (selectTab == null)
```
brendanx67: This is also already a pretty long function, and I would prefer to see the two switch statements closer together. Could we implement RestoreTabSel(TABS? selectTab) and SaveTabSel() which just get called by the constructor and the OnFormClosing(), and put the two functions right next to each other.

bspratt: Done.

PR #534, opened 2019-05-01
Add pressure traces for Sciex WIFF1 files

chambm:
Changes WIFF TIC chromatogram from summing all samples to only summing the experiments from the current sample
Add new test data for core and Skyline, and moves ABI test data mzMLs out of tarball
Fix bad collision energy values for Enolase_repeats data


bspratt:
Nothing jumps out at me - related to the work to be able to pull TIC without filtering in Skyline?

chambm:
No, this is unrelated to TIC. It's for AutoQC.

chambm:
@brendanx67 @nickshulman OK with `GetQcTraces()` and its semantics?

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -612,19 +612,59 @@ public double[] GetTotalIonCurrent()
...
+
+        public class QcTrace
+        {
+            public QcTrace(Chromatogram c)
+            {
+                Name = c.id;
+                if (Name.ToLowerInvariant().Contains(@"pressure"))
                 {
-                    return null;
+                    MeasuredQuality = "pressure";
```
brendanx67: I prefer to see text like this expressed as constants to avoid the same text used in multiple places, requiring you type it the same in both.

pwiz_tools/Skyline/TestData/PwizFileInfoTest.cs
```diff
@@ -102,6 +97,26 @@ public void TestTicChromatogram()
...
+        {
+            const string testZipPath = @"TestData\PressureTracesTest.zip";
+
+            var testFilesDir = new TestFilesDir(TestContext, testZipPath);
+
+            using (var msDataFile = new MsDataFileImpl(testFilesDir.GetTestPath("PressureTrace1" + ExtensionTestContext.ExtAbWiff)))
+            {
+                var pressureTraces = msDataFile.GetQcTraces();
+
+                VerifyQcTrace(pressureTraces[0], "Column Pressure (channel 1)", 1148, 0, 9.558333, 1470, 210, "pressure", "psi");
```
brendanx67: Better to refer to the string constants than have these values typed out in strings.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -612,19 +612,59 @@ public double[] GetTotalIonCurrent()
...
+                    IntensityUnits = "psi";
                 }
-                TimeIntensityPairList timeIntensityPairList = new TimeIntensityPairList();
-                chromatogram.getTimeIntensityPairs(ref timeIntensityPairList);
-                double[] intensities = new double[timeIntensityPairList.Count];
-                for (int i = 0; i < intensities.Length; i++)
+                else if (Name.ToLowerInvariant().Contains(@"flow"))
+                {
+                    MeasuredQuality = "volumetric flow rate";
```
brendanx67: Will these values ever get shown to the user? If so, they need to end up in .resx files for translation.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -612,19 +612,59 @@ public double[] GetTotalIonCurrent()
...
+                return chromatogram?.getIntensityArray()?.data.Storage();
+            }
+        }
+
+        public class QcTrace
+        {
+            public QcTrace(Chromatogram c)
+            {
+                Name = c.id;
+                if (Name.ToLowerInvariant().Contains(@"pressure"))
```
brendanx67: I guess okay, but any more of these and we'll want a dictionary at the top and look-up instead of adding more ifs.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
chambm: When the CV is updated to include PSI as a unit (https://github.com/bio-ontology-research-group/unit-ontology/issues/20) I plan to switch these to enumerated values instead of strings. It won't hurt to use string constants in the meantime though.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
chambm: If this will be used to plot the QcTraces in the GUI, then indeed there will need to be some translation from the precision-and-CV-oriented language I'm using here (which again I plan to convert to an enumerated value) to something more domain-specific (e.g. just 'flow rate' or localized equivalent).

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
chambm: When the MS CV is updated to add chromatogram types for pressure and flow rate, I will use that term as the if condition rather than the string matching. But I'd still plan to use if-blocks because I don't plan to re-use this logic anywhere else (indeed, the point of adding these properties was to avoid the need for any downstream code to do its own lookup other than translation). Is there another reason to use a dictionary/mapping in this case other than code-reuse? I assume you mean something like:
```
Dictionary<string, (string Quantity, string Units)> lookup = new Dictionary<string, (string, string)>()
{
   {"pressure", ("pressure", "psi") },
   {"flow", ("volumetric flow rate", "uL/min") }
}
```

As described above I would really imagining this to end up as:
```
Dictionary<CVID, (QcQuantity Quantity, QcUnits Units)> lookup = new Dictionary<CVID, (QcQuantity, QcUnits)>()
{
   {CVID.MS_pressure_trace, (QcQuantity.Pressure, QcUnits.PoundsPerSquareInch) },
   {CVID.MS_flow_rate_trace, (QcQuantity.VolumetricFlowRate, QcUnits.MicrolitersPerMinute") }
}
```
(QC quantity and units would also be in the CVID enum, but I don't want to force Skyline to import the CV since it doesn't currently do so)

chambm:
@nickshulman Brendan wanted me to hand this off to you to add the QC traces to the .skyd.

PR #540, opened 2019-05-09
Extract serial numbers for WIFF, MassHunter, BAF/TDF formats

chambm:


chambm:
Thermo's had serial numbers for about decade. However here I changed it so it won't put a CV term in if the serial number is empty in a Thermo file.

PR #549, opened 2019-05-22
UIMF updates

FarmGeek4Life:
* Use cvParams instead of userParams for the file format/native ID
* Add in support for SpectrumList3D and SpectrumListIonMobilityBase


FarmGeek4Life:
Hold on merging this for now; I finally tried popping a uimf file up in Skyline and it needs some more work to fix what I was trying to fix in the first place.

Skyline only shows a single retention time point (at zero) when looking at a peptide (and I tried zooming in all the way to 10^-15...). We can convert the same file to Agilent .D format (using internal tools developed in collaboration with Agilent) and have it appear "normal". I want to try to track this problem down before merging (because if I don't, it will end up as another pull request).

chambm:
I'm glad I procrastinated merging it. ;) It sounds like this PR needs to add a test to Skyline. Are you able to build Skyline? I've never made an IMS test myself so I'll ping @bspratt on this. But look at:
TestData\Results\WatersImsMseTest.zip and the corresponding .cs test class. Uimf probably needs something similar (but it can start simpler).

FarmGeek4Life:
Well, I did build Skyline, which is how I tested the functionality with this PR (I did forget that I could have downloaded the build artifacts from the teamcity builds). Figuring out where the retention time information is being lost/not set properly may be interesting...

chambm:
@bspratt Does a SpectrumList have to return canConvertIonMobilityAndCCS==true in order for Skyline to read it as ion mobility?

bspratt:
No. We can use direct ion mobility values, no problem.

bspratt:
Can you successfully convert it to mzML with retention times intact, using msconvert? That is generally enough to guarantee that it will work with Skyline. If you get that working I don't think you really need to create a specific Skyline test.

FarmGeek4Life:
It has worked for older files, I guess I didn't think about doing that with
one of the newer files (and that may very well be the problem...)

On Fri, May 24, 2019, 11:28 AM Brian Pratt <notifications@github.com> wrote

FarmGeek4Life:
So yeah, the files themselves are part of the problem; one of the changes that someone else made at some point was to stop writing the Frame start time to the file, and if that doesn't exist, '0' will be returned. So now I get to make other changes to calculate the frame start time based on other information in the file, since that will let me get it close to accurate.

bspratt:
A pwiz library test for correctly generated mzML would be appropriate, of
course.

FarmGeek4Life:
Now it's ready for merge.

This last set of commits resolves the issue I was seeing, another one I had missed, and adds support for outputting the TIC chromatogram. It is working as I expected it to, and allows loading some problematic UIMF files in Skyline. The retention time problem is most fixed within the UIMFLibrary DLL, because the same code is needed for other internal workflows. (Matt, I changed this to have just one copy of the UIMFLibrary DLL, since it is compiled as 'AnyCPU' now)

I haven't added additional mzML conversion tests to this, partly because (in this case) I am working with files that have several issues, including being 5+ GB in size, and missing important metadata (due to other issues with the acquisition software). I do want some newer tests files added, but I want to wait until those issues with the acquisition software are resolved.

bspratt:
If we already have UIMF tests in place that assert proper operation on
normal files, this seems fine.

FarmGeek4Life:
Okay, I'll have to ask for help on resolving the 'Core Windows x86_64 (no vendor DLLs)' build, because I did make changes to the jamfile that defines 'PWIZ_READER_UIMF', but don't know how I broke it (unless it's because I accidentally/unknowingly removed the check for '--i-agree-to-the-vendor-licenses' in the pwiz_aux/msrc/vendor_api/UIMF Jamfile?). The 'Core Windows x86' failure looks to be unrelated?

bspratt:
>>  The 'Core Windows x86' failure looks to be unrelated?

That's my impression, yes.

pwiz_aux/msrc/utility/vendor_api/UIMF/Jamfile.jam
```diff
@@ -28,13 +28,17 @@ path-constant API_PATH : $(PWIZ_ROOT_PATH)/pwiz_aux/msrc/utility/vendor_api/UIMF
 
 
 if [ modules.peek : NT ]
-{
+{
+
+if [ path.exists $(API_PATH)\\UIMFLibrary.dll ]
```
chambm: The --i-agree-to-the-vendor-licenses is used as a proxy for building with binary support. It doesn't need to be: there's a without-binary-msdata rule that could be used for it instead. But one or the other needs to be checked in the uimf-dll-location rule in Jamroot.jam, and the .hasDLLs value should be set based on the return value of that, not a simple path.exists check in the vendor_api/Jamfile.jam. Does that make sense?

FarmGeek4Life: okay, so what I put in was a mix between what was there and what is used in the Thermo jamfile; I didn't have a file search set up properly, so I couldn't find the source of 'uimf-file-location'. So, now knowing that, I basically should just revert the jamfile changes I did make, and modify the uimf-dll-location rule in Jamroot.jam...

chambm:
Windows x86 should be fixed now. See my comment about fixing the no vendor DLLs build.

pwiz_aux/msrc/utility/vendor_api/UIMF/Jamfile.jam
FarmGeek4Life: also, there are currently checks for msvc version 9.0 or 9.0 express; should this be changed to just a check for toolset msvc?

chambm:
Yeah our modular layout shows this as something of a drawback of Boost.Build. I haven't figured out how to include vendor_api variables/rules from vendor_reader Jamfiles. So that's why some stuff is in Jamroot.jam (as all child Jamfiles have access to it). But yeah, unqualified rule invocations are not obvious to find. :(

pwiz_aux/msrc/utility/vendor_api/UIMF/Jamfile.jam
chambm: No, it's checking for those versions as being explicitly NOT supported. Is UIMF .NET 4 now? 4.what?

FarmGeek4Life: Currently compiled as .NET 4.6.1 (The first UIMFLibrary.dll update I gave you back around June 2015 was already .NET 4.0, and I think the version already in use was as well).

So, as a remark: the Thermo jamfiles (I'm assuming due to the different APIs now in use for x86 vs. x64) do check for "--i-agree-to-the-vendor-licenses", rather than depending on jamroot.jam rules.

FarmGeek4Life:
I found one way to include vendor_api variables/rules in vendor_reader Jamfiles, but it isn't very pretty:
```
import modules ;
import project ;
uimf-api-module = [ project.load $(API_PATH) ] ;
if [ modules.call-in $(uimf-api-module) : uimf-dll-location $(API_PATH) : $(properties) ]
{
    .hasDLLs = true ;
}
```

Supposedly creating a separate module would be cleaner, but getting one to work appears to not be as straightforward as I hoped when I tried.

chambm:
Nice! I didn't expect you to dig into that: I thought you were just going to update the rule currently in `Jamroot.jam`. But I wouldn't mind refactoring it to go to vendor_api's Jamfile. The nastiness could be factored out in a function (which would be appropriate in Jamroot.jam):
```
rule get-project-expression ( project-path : expression )
{
    return [ modules.call-in [ project.load $(project-path) ] : $(expression) ] ;
}
```
Then `if [ get-project-expression $(API_PATH) : uimf-dll-location ] { .hasDLLs = true ; }` might work. The uimf-dll-location can be simplified so it doesn't need to take API_PATH: it should already know it.
(But again, the uimf-dll-location does need to respect --i-agree-to-the-vendor-licenses and --without-binary-msdata)

pwiz_aux/msrc/utility/vendor_api/UIMF/Jamfile.jam
FarmGeek4Life: Would you agree with me changing the lib rules for UIMF in vendor_api and vendor_reader to `include <conditional>@msvc-dotnet-requirement`, and removing the checks for msvc 9.0/9.0express?

chambm: Yes, good idea.

FarmGeek4Life:
I realized one addition that probably needs to be made to that project load: it should probably check if the projects exists first (otherwise it will probably just kill the build, rather than building without reader support, if a jamfile is missing).

FarmGeek4Life:
Anything holding this back, beyond the test failures that have occurred?

PR #553, opened 2019-05-23
Show QC traces and use global TIC/BPC chromatograms

chambm:
- Change TIC and BPC extractors to use global chromatograms from the source file (e.g. `<chromatogram id="TIC">`) instead of building TIC scan-by-scan
- Add option to view QC traces in Transition menu
- Add ShowQcTraceName setting which controls which trace is shown

Known issues:
1. `qcMenuItem` is not showing its DropDownItems submenu but `qcContextMenuItem` is. What's up with that?
2. ChromCacheMinimizerTest is failing but I don't understand why. Is the new way of getting TIC chromatograms somehow being filtered out by the minimizer? I couldn't tell the differences between the original and minimized versions when looking in the debugger.
3. I left this code intact in SpectrumFilter.cs:
```
if (EnabledMs)
{
    _isHighAccMsFilter = !Equals(_fullScan.PrecursorMassAnalyzer,
        FullScanMassAnalyzerType.qit);

    if (!firstPass && !_isFAIMS)
    {
        var key = new PrecursorTextId(SignedMz.ZERO, null, null, ChromExtractor.summed);  // TIC
        dictPrecursorMzToFilter.Add(key, new SpectrumFilterPair(key, PeptideDocNode.UNKNOWN_COLOR, dictPrecursorMzToFilter.Count,
            _instrument.MinTime, _instrument.MaxTime, _isHighAccMsFilter, _isHighAccProductFilter));
        key = new PrecursorTextId(SignedMz.ZERO, null, null, ChromExtractor.base_peak);   // BPC
        dictPrecursorMzToFilter.Add(key, new SpectrumFilterPair(key, PeptideDocNode.UNKNOWN_COLOR, dictPrecursorMzToFilter.Count,
            _instrument.MinTime, _instrument.MaxTime, _isHighAccMsFilter, _isHighAccProductFilter));
    }
}
```
Because when I comment it out, tests start failing. But as far as I can tell, the global chromatograms ARE getting used.


bspratt:
Leaving that code in kills any performance gains, so we'll need to sort
this out. What kind of failures are you seeing?

I'm not surprised if it's just that the TIC and BPC have changed somewhat -
my intuition was that this wouldn't agree with our own values. I notice,
for example, that the Bruker DIA PASEF files claim TIC 10 for scans with no
mz-intensity pairs, so we'd disagree on that for sure.

chambm:
> Leaving that code in kills any performance gains, so we'll need to sort this out. What kind of failures are you seeing? I'm not surprised if it's just that the TIC and BPC have changed somewhat - my intuition was that this wouldn't agree with our own values. I notice, for example, that the Bruker DIA PASEF files claim TIC 10 for scans with no mz-intensity pairs, so we'd disagree on that for sure.

In AgilentMseChromatogramTestAsNegativeSmallMolecules:
'pwiz.Skyline.Model.Results.NoFullScanDataException' in Skyline-daily.exe
Additional information: No scans in AgilentMse\BSA-AI-0-10-25-41_first_100_scans.mzML match the current filter settings.

It seems something I did fixed a lot of the other failures I saw earlier when commenting out that code. Now it's just this AgilentMse failure, the ChromCacheMinimizer failure, and a failure in TestRetentionTimeFilter because none of the old mzML/mz5 files and not all the vendor readers have base peak chromatograms in them. I think Brendan said he wants to allow the user to opt-in to the scan-by-scan BPC building to use when it's not available direct from the file. We would need that capability to make this test work without updating the mzML/mz5 files in RetentionTimeTest.zip (actually I need to add BPC support for Thermo too but that's quite simple).

bspratt:
Could we not selectively disable that based on whether the BPC is
available? Then we don't have to bug the user at all.

brendanx67:
My intent was to disable this completely and just have the UI show that BPC is unavailable when we cannot get it directly from the MSDataFile API and then wait to see if anyone complains, and if so, who and why, and only if we get important complaints consider adding a way to get back to the old way of doing things, but for now just go with the more limited, but higher performance method, rather than kill ourselves maintaining parity with the original implementation when we don't really know who cares, other than it was once requested by Michael Bereman when he was working on SProCoP as a post-doc in the MacCoss lab.

chambm:
OK, I fixed RetentionTimeTest by switching from BPC to TIC. Any tips about the ChromCacheMinimizer failure? What is that class doing with all ions chromatograms?

brendanx67:
ChromCacheMinimizer has an option to discard unused chromatograms. It may be testing that this does not discard TIC or BPC, since those are special cases, which your QC chromatograms now join, but if for some reason the test no longer includes, say, BPC, then it would fail testing that the minimizer did not discard BPC. Just a guess. One of us will have to have a closer look. Since you should think about modifiying this test to ensure QC chromatograms are not discarded by the minimizer, I vote for you to be the one to look more closely.

chambm:
When I add code to keep chromatograms if PrecursorMz==0, the first MinimizedRatio ChromCacheMinimizer test passes, but the next one fails:
`Assert.AreEqual(1.0, statsMissingFirstProtein.MinimizedRatio + statsWithOnlyFirstProtein.MinimizedRatio,
                                .00001);`
The logic here is to include or exclude the first protein and expect the minimized files to add up to the original file size. That only makes sense if the global chromatograms are not counted in either file total (i.e. they're excluded both from the original size and the minimized size). But in that case I don't see why the first MinimizedRatio failed (before I forced the minimizer to keep the global chromatograms). @nickshulman what am I missing?

bspratt:
If we find that folks do care about clickable TIC in Skyline it should be
generally possible and inexpensive to pull each scan's TIC and BPC from
relatively inexpensive header metadata.

bspratt:
... which I thought I'd commented on here, but that seems to have not stuck. I'll just fix them myself (mostly about one-liners, we generally put the if and the clause on seperate lines for ease of breakpoint setting).

PR #554, opened 2019-05-23
Skyline: Updates to command-line - Added two new command-line options…

vagisha:
- Added two new command-line options: --import-filename-pattern and --import-samplename-pattern (issue 626) 
- Import results from multi-injection wiff files that have the same sample names (issue 627)


vagisha:
Two of the tests in CommandLineTest no longer import data into a Skyline document.  They test CommandLine.ApplyFileAndSampleNameRegex() instead. Simplified CommandLineWiffTest.

PR #557, opened 2019-05-24
Fix the Reintegrate Dialog so that it treats "UserSet.FALSE" the same…

nickshulman:
… as "UserSet.REINTEGRATED", and does not required "Overwrite manual integration" in order to re-pick peaks that were changed by the model.


brendanx67:
Hello Nick? Can we get this passing tests and merged?

PR #566, opened 2019-06-07
Added CAD chromatogram for Thermo

chambm:
- added CAD chromatogram for Thermo (treating it as a kind of TIC)
* fixed enumeration of PDA chromatograms on 64-bit (really need a unit test file for this...)
* refactored repeated code out of ChromatogramList_Thermo::createIndex() into new private addChromatogram() function

### DIFF
--- a/pwiz/data/vendor_readers/Thermo/ChromatogramList_Thermo.cpp
+++ b/pwiz/data/vendor_readers/Thermo/ChromatogramList_Thermo.cpp
@@ -345,7 +345,7 @@ PWIZ_API_DECL void ChromatogramList_Thermo::createIndex() const
 
                                 for (size_t j=0, jc=scanInfo->scanRangeCount(); j < jc; ++j)
                                 {
-                                    double q1 = scanInfo->precursorMZ(0);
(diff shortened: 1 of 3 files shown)


jwinget:
Hi @chambm , sorry to bug you again but recent files that I've converted do not contain CAD data in the mzML, even though the data are in the Thermo RAW files. I've tried various options for `msconvert` to no avail. Did this commit get reverted in the interim?

Near the bottom of my mzMLs I see:
```
<chromatogramList count="2" defaultDataProcessingRef="pwiz_Reader_Thermo_conversion">
  <chromatogram index="0" id="TIC" defaultArrayLength="15860">
  [...]
  <chromatogram index="1" id="PDA 1" defaultArrayLength="6000">
```
The first of these is the total ion chromatogram. It does have 3 binaryDataArrays which is a bit odd (usually it's just two: time and intensity), but I think the third somehow encodes the MS level, as this is the value given in the preceding cvParam (accession MS:1000786). The PDA 1 chromatogram is the UV detector and has the expected two binaryDataArrays.

I'm on pwiz 3.0.20265.ed3a75814, downloaded today

PR #569, opened 2019-06-12
- fixed precursorList for Thermo MS3 data with "SPS Masses" 

chambm:
The first precursor is the MS1 precursor and the rest are MS2 precursors; the ms level is indicated by an "ms level" userParam in the <Precursor> element but will later switch to a standardized representation


pwiz_aux/msrc/utility/vendor_api/thermo/RawFile.cpp
```diff
@@ -1519,6 +1546,11 @@ vector<PrecursorInfo> ScanInfoImpl::precursorInfo() const
 
 long ScanInfoImpl::precursorCharge() const
 {
+    // "Charge State" header item for SPS spectra refers to MS2's precursor charge;
+    // CONSIDER: real charge states might be available in the instrument method
+    if (!spsMasses_.empty())
+        return 0;
```
chambm: Should use this charge state for the MS1 precursor, but not the MS2 precursors.

PR #583, opened 2019-07-17
- fixed BiblioSpec's off-by-one error when looking up scan-number ids

chambm:
- fixed BiblioSpec's off-by-one error when looking up scan-number ids (it was working fine for WIFFs, but now it uses a hacky check to see whether the mqpar.xml has a .wiff or .wiff2 in it before using the index lookup instead of scan number lookup)
* added Thermo RAW test for BiblioSpec::MaxQuantReader's external spectra scan number lookup


brendanx67:
So, it looks like this does not impact WIFF or WIFF2 files and it appears to fix Thermo RAW files pretty explicitly. Have you tested against Bruker files?

chambm:
No, but it should behave the same as it did before I made the switch to INDEX_ID (when I fixed WIFF support for MaxQuantReader). I've never run MaxQuant on Bruker data actually.

brendanx67:
Which could still be totally broken since the switch to using external rather than embedded spectra. So, we need to test this before we make the patch. Though, not necessarily before you merge this commit, which will fix Thermo without newly breaking Bruker, I agree.

brendanx67:
One thing I have been wondering about, also, is how using external spectra will work with Bruker timsTOF ddaPASEF data. Do we have any chance of getting back from the msms.txt to the same spectrum used in the search? Seems like that may be problematic.

chambm:
We will probably have to switch to using external spectra with Bruker TIMS because I don't see any definitive way to get from the MaxQuant txt files to an external scan number (single or merged). Shall I disable the default "prefer-external" behavior for TIMS?

Theoretically we could use precursor m/z, retention time, and ion mobility to determine a set of TIMS scans to read, but there are many variables that would go into which scans to use. MaxQuant's TIMS processing seems quite different than other data. Do we know for sure they even do the deconvolution that's been problematic for other MaxQuant spectra?

PR #589, opened 2019-07-23
(WIP) Add C++ stacktraces

chambm:
- added stacktraces to exception handlers for unit tests
* added boost::stacktrace lib target in ext-boost.jam
* added pwiz::util::throw_with_trace()
* changed most TEST_FAILED macros to TEST_FAILED_EX which just takes the exception instead of the message, and prints stacktrace if present
* changed a few global build settings for MSVC to make debug symbols leaner but always present; the PDB files are still really big though: probably too big to distribute to users, so I'm still not sure how to get stacktraces from users

@bspratt you might be interested in looking at this. I'm not really happy with the size of the PDBs so I don't see a way to redistribute this to users, but at least it works for developer builds (or rather, it would if I went through and replaced almost all the `throw XXX` calls with `throw_with_trace`).

Probably broken on Linux, let's see.

### DIFF
--- a/pwiz_tools/BiblioSpec/src/BuildParser.cpp
+++ b/pwiz_tools/BiblioSpec/src/BuildParser.cpp
@@ -43,7 +43,7 @@ BuildParser::BuildParser(BlibBuilder& maker,
     fileroot_ = getFileRoot(fullFilename_);
 
     preferEmbeddedSpectra_ = maker.preferEmbeddedSpectra().get_value_or(true);
-
+    pwiz::util::throw_with_trace(runtime_error("wtf"));
     this->parentProgress_ = parentProgress_;
     this->readAddProgress_ = NULL;
     this->fileProgress_ = NULL;
--- a/Jamroot.jam
+++ b/Jamroot.jam
@@ -357,9 +357,12 @@ project pwiz
         <threading>multi
 
         # change boost debug assertions to exceptions by force including Exception.hpp
-        <variant>debug,<toolset>msvc:<cxxflags>/FIpwiz/utility/misc/Exception.hpp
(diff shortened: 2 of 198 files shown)


pwiz_tools/BiblioSpec/src/BuildParser.cpp
```diff
@@ -43,7 +43,7 @@ BuildParser::BuildParser(BlibBuilder& maker,
     fileroot_ = getFileRoot(fullFilename_);
 
     preferEmbeddedSpectra_ = maker.preferEmbeddedSpectra().get_value_or(true);
-
+    pwiz::util::throw_with_trace(runtime_error("wtf"));
```
bspratt: (clutches pearls)

chambm: Oh yeah, bit of a test statement there. Probably won't help the blib tests pass. ;)

chambm:
> Any sense of potential download bloat, or possible performance costs, when carrying symbols around? (I like having them, just wondering what the hit might be).
Pure pain:
```
19M     chainsaw.pdb
36M     idcat.pdb
36M     idconvert.pdb
93M     msaccess.pdb
94M     msconvert.pdb
74M     msdiff.pdb
74M     msdir.pdb
5.6M    msistats.pdb
75M     mspicture.pdb
75M     peakaboo.pdb
7.1M    pepcat.pdb
7.0M    pepsum.pdb
76M     qtofpeakpicker.pdb
103M    pwiz_bindings_cli.pdb
98M     pwiz_data_cli.pdb
```
The overall size could be greatly reduced by putting everything into libpwiz.dll so there's just one big pdb (although there would still be a separate one for the bindings).

chambm:
May reopen in the future if I find/settle on a way to make the PDB sizes more manageable.

PR #598, opened 2019-07-30
Skyline: add a User Interface selection to the Settings menu, in case users don't notice it on the button bar

bspratt:
image


bspratt:
Two questions:
1) The capitalization matches that of the toolbar button but seems inconsistent with our menu conventions.
2) Do we want to indicate which choice is the current one with a check or highlight or similar?

brendanx67:
On the 2 questions:
1. I think you can drop the word "interface" because it seems redundant with the parent menu item "User Interface"
2. Yes, it needs to show the active "setting" with a checkbox

It needs to change to "Document Type" if we want to leave it here, and move above "Integrate All". This more accurately exposes what is happening to the SrmDocument, but seems a little inconsistent with what we are otherwise promoting on our other UI Mode settings. We still don't need the word "interface" in either case when it is on the menu.

bspratt:
OK, so to be clear,we want to:
put it on the View Menu, under Text Zoom
call it User Interface
drop "interface" from the submenu
indicate the current UI mode on the submenu

brendanx67:
Yes. I think that would be fine.

bspratt:
OK:
image

PR #599, opened 2019-07-31
- improved BlibFilter speed in best scoring mode

chambm:
- improved BlibFilter speed in best scoring mode (by not getting peaks when not used and transferring RefSpectra in bulk rather than one spectrum at a time); on a 30mb .blib, my time went from 3s to 0.5s; I'm putting together a bigger blib now for a more substantial test

* added try/catch for BlibBuild's main() code
* added tests for BlibFilter's best scoring mode


chambm:
OK, larger test, in seconds:


Before|After
------|-----
155|5

brendanx67:
30x improvement seems pretty compelling. Nice work.

chambm:
@brendanx67 I did some more testing to make sure the new code is getting the same results as the old. The only differences I found were due to picking different spectra with the same score (e.g. Q value 0). For both versions of the code the result is unpredictable because there's no additional ordering beyond score. I suggest we add that ordering now. What about sorting ties by fileID and scan index, then picking the first (or last)?

brendanx67:
I thought we sorted by probability score (e.g. q value 0) and then by raw score (e.g. xcorr) which is much less likely to be identical. Is that still possible to do? After that, I guess it is okay to make the sort deterministic as you suggest.

chambm:
No, blib only imports one score at this time.

brendanx67:
I guess that could be a lot of work to retrofit it to also store a raw
score for each type. I am pretty sure, though that not all q value zero
scores are created equal.

chambm:
It wouldn't be a lot of work to add a raw score column, but we'd still need a tie-breaker for older blibs which were created without that column. In the target-decoy world, a Q value of 0 means that there are no decoys with a better score. So yes of course the spectra have different raw scores, but with no decoys to compare them to, there's no way to talk about FDR the way we can with other target hits (with Q value > 0). It's for this reason that I switched IDPicker to use Andy Jones's FDR score, which just does linear interpolation between Q value spikes:

image

This way you always get a good ordering between spectra, even if it's not strictly tied to FDR. But IDPicker is calculating its own FDR scores, not reading them from the search engines files, so BiblioSpec couldn't really do this.

chambm:
Ah, nevermind I see what happened. The current code breaks ties by TIC, but it has to enumerate peaks to do so. The optimization to not load peaks meant that the TIC always returns 0. Shall I switch it to break ties by retention time, or somehow restore the TIC behavior (like adding a TIC column to RefSpectra)?

brendanx67:
Yeah. TIC was intentional and has positive impact. So, I guess add TIC to
RefSpectra.

I know I asked Kaipo to try the raw score approach to reduce our reliance
on TIC and make score differentiation more meaningful among high scoring
spectra. Maybe also add a raw score column... Though, I guess that would
require more score typing and scoring direction information, i.e. bigger or
smaller is better.

So, if we can maintain exactly how it works now but much faster, that would
be a big win.

I guess I understand why the old code is so slow if it needed to load and
sum spectra to break ties.

I do still find the idea of having BlibBuild write both the redundant and
non-redundant blib files appealing, when all the necessary information is
present. Because I expect that to be as fast as it is possible to achieve
and also to cover now something like 98% of the libraries that get built
with BiblioSpec. The existing split assumes the old model where redundant
libraries accumulate over time and the act of making them non-redundant
becomes impossible to achieve with what BlibBuild has in memory.

chambm:
The peak summing was a significant chunk of time but most of it was just suboptimal SQLite usage. It's always better to transfer rows in bulk rather than one at a time when possible.

I looked at making it write the non-redundant db at the same time as the redundant one but it would have made the code even messier than it already is. With a lot of refactoring it would be reasonable though.

chambm:
Ah, eat temp view, you awkward backward compatibility code!

chambm:
@bspratt What do you think about this TEMP VIEW approach?

bspratt:
I'm not enough of a database guy to have a worthwhile opinion. Nick would doubtless have valuable insights. Out of curiosity, it sounds like there may be drawbacks that you're thinking about?

chambm:
None that I can think of in this context. It's not hard to imagine cases where a temp view wouldn't be enough to convert from an old schema to a new one, but that was not the case here.

PR #601, opened 2019-08-01
Fix for the support issue 40871: error message for reversed audit log…

rita-gwen:
… timestamp order does not show the actual violating timestamp.

Link to the issue: https://skyline.ms/announcements/home/support/thread.view?rowId=40871
Brendan suggested that the error message should show the index and timestamp of the violating log entry. 
This fix does that. I also made sure that error message dialog shows the full error message and stack trace (including the inner exception chain).


bspratt:
There are a couple of inspection errors:
Format string placeholders mismatch (2)
Model/AuditLog
AuditLogStrings.ja.resx (1)
AuditLogStrings.zh-CHS.resx (1)

It's OK to just delete the entries causing the problem, our translators will fix it up on the next round. (In general when changing a formatting string its best to convert it back to a literal first, so that the naming convention holds and so that you don't have this mismatch issue in the L10N files.)

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -206,10 +206,31 @@ public static bool ReadFromFile(string fileName, out string loggedSkylineDocumen
...
                 {
-                    alert.Message = string.Format(
-                        AuditLogStrings.AuditLogList_ReadFromFile_An_exception_occured_while_reading_the_audit_log,
-                        fileName);
-                    alert.DetailMessage = ex.StackTrace;
+                    Exception exception = ex;
+                    StringBuilder msgBuilder = new StringBuilder("");
+                    List<string> stackList = new List<string>();
+
+
```
bspratt: Would our existing GetStackTraceText method work here?

rita-gwen: Well, I did not know about it, of course. It would have done, but it uses non-localized strings and it looks a bit inconsistent to me. Why would only part of the text visible to the user be translated.
I'll keep my code for now.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -206,10 +206,31 @@ public static bool ReadFromFile(string fileName, out string loggedSkylineDocumen
             {
                 using (var alert = new AlertDlg())
```
nickshulman: You should throw an exception instead of putting up an alert here.
This code gets called from both the commandline and from the user interface, and the commandline should never bring up a message box.
This exception can be caught in "SkylineWindow.OpenFile" and you can use the usual MessageDlg.ShowWithException to show the error in the usual way.

rita-gwen: Well, the dialog was there to start with, I just modified the way the message was displayed. But it's a good point. I considered just throwing the exception to catch is downstream, but the catch (no pun intended)  is that we still want to open the Skyline doc even if the audit log is invalid. And if this exception would propagate any further the loading of the document file is canceled.
Possible option would be to test if application has UI and just send the message into System.out if it has none.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -143,7 +143,7 @@ public void Validate()
             foreach (var entry in AuditLogEntries.Enumerate())
             {
                 Assume.IsTrue(entry.TimeStampUTC <= time && entry.LogIndex > logIndex,
-                    AuditLogStrings.AuditLogList_Validate_Audit_log_is_corrupted__Audit_log_entry_time_stamps_and_indices_should_be_decreasing);
+                    string.Format(AuditLogStrings.AuditLogList_Validate_Audit_log_is_corrupted__Audit_log_entry_time_stamps_and_indices_should_be_decreasing, entry.TimeStampUTC, entry.LogIndex.ToString()));
```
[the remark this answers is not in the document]
rita-gwen: Oh, come on! :-) It's always better to do explicit type conversions.

rita-gwen:
This commit should fix the failing inspection, I think.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -206,10 +207,31 @@ public static bool ReadFromFile(string fileName, out string loggedSkylineDocumen
...
+
+                    alert.Message = TextUtil.LineSeparate(
+                        string.Format(
+                            AuditLogStrings.AuditLogList_ReadFromFile_An_exception_occured_while_reading_the_audit_log,
+                            fileName),
+                        msgBuilder.ToString());
+
+                    alert.DetailMessage = TextUtil.LineSeparate(stackList);
                     alert.AddMessageBoxButtons(MessageBoxButtons.OK);
                     alert.ShowParentlessDialog();
```
brendanx67: This all looks pretty strange and contorted because of its placement inside Model. That is why the code ends up constructing its own AlertDlg and calling ShowParentlessDialog() (both really cheats to deal with the fact that we don't know what window the user is interacting with - and as Nick points out, there may be no window, if this is happening at the command-line). If the message code were properly placed, then the code to show a message with an OK button and an exception would be a 1-liner: MessageDlg.Show(parent, message, exception)

pwiz_tools/Skyline/Model/AuditLog/AuditLogStrings.resx
```diff
@@ -632,8 +632,8 @@
   <data name="modified_outside_of_skyline" xml:space="preserve">
     <value>The document was modified outside of Skyline: {0}</value>
   </data>
-  <data name="AuditLogList_Validate_Audit_log_is_corrupted__Audit_log_entry_time_stamps_and_indices_should_be_decreasing" xml:space="preserve">
-    <value>Audit log is corrupted. Audit log entry time stamps and indices should be decreasing</value>
+  <data name="AuditLogList_Validate_Audit_log_is_corrupted__Audit_log_entry_time_stamps_and_indices_should_be_decreasing_Entry_timestamp__0_G___entry_order_number" xml:space="preserve">
+    <value>Audit log is corrupted. Audit log entry time stamps and indices should be decreasing. Entry timestamp {0:G}, entry order number {1}.</value>
```
brendanx67: I would improve this message to give all the information in the message without requiring the user to look into the file to understand exactly what is going on.

Audit log is corrupted. Audit log entry timestamps and indices should be decreasing.

Entry {0} timestamp {1:G} occurs before entry {2} timestamp {3:G}.

or

Entry {0} appears after entry {1}.

Depending on whether the issue is out of order dates or indices.

You can use TextUtil.LineSeparate() to combine the two messages.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
rita-gwen: The recent commits should take care of this issue.

pwiz_tools/Skyline/Model/AuditLog/AuditLogStrings.resx
```diff
@@ -633,7 +633,7 @@
     <value>The document was modified outside of Skyline: {0}</value>
   </data>
   <data name="AuditLogList_Validate_Audit_log_is_corrupted__Audit_log_entry_time_stamps_and_indices_should_be_decreasing" xml:space="preserve">
-    <value>Audit log is corrupted. Audit log entry time stamps and indices should be decreasing</value>
+    <value>Audit log is corrupted. Audit log entry time stamps and indices should be decreasing. entry number {0} timestamp {1:G} occurs before entry number {2} timestamp {3:G}</value>
```
brendanx67: Use capital E at the beginning of the 3rd sentence.

pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -312,7 +313,16 @@ public bool OpenFile(string path, FormEx parentWindow = null)
...
 
-                        document = document.ReadAuditLog(path, skylineDocumentHash, AskForLogEntry);
+                        try
+                        {
+                            document = document.ReadAuditLog(path, skylineDocumentHash, AskForLogEntry);
+                        }
+                        catch (Exception e)
+                        {
+                            throw new AuditLogException(
+                                string.Format("Error when loading document audit log: {0}", path), e);
```
brendanx67: This text needs to be in a RESX file.

pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -321,11 +331,41 @@ public bool OpenFile(string path, FormEx parentWindow = null)
             }
             catch (Exception x)
             {
-                exception = x;
-                // Was that even a Skyline file?
-                if (!SrmDocument.IsSkylineFile(path, out var explained))
+                var ex = x;
```
brendanx67: This is getting too long and the logic needs to be moved elsewhere.

pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -321,11 +331,41 @@ public bool OpenFile(string path, FormEx parentWindow = null)
...
+                    }
+                    
+                    ex = ex.InnerException;
+
+                } while (ex != null);
+
+                if (isAuditLogException)
+                {
+                    MessageDlg.ShowWithException(parentWindow ?? this, 
+                        TextUtil.LineSeparate(msgStrings),
```
brendanx67: Probably move the logic to a static function on AudiLogException, since it seems like it only really has an impact when AuditLogException is involved. Maybe 2 functions AuditLogException.IsAuditLogInvolved(Exception e) and AuditLogException.GetMultiLevelMessage(Exception e).

rita-gwen:
I am going to work on adding a functional test to test the audit log validation logic.

rita-gwen:
Added functional tests for audit log validation. Had to make a minor change in the AbstractFunctionalTest class to implement the testing logic. It should not affect the rest of the code.

I am not sure I should address the failing code inspection above, the warnings it produces seem to be inconsequential.

bspratt:
>>  I am not sure I should address the failing code inspection above, the
warnings it produces seem to be inconsequential.

That's a per-issue judgement call - fix vs suppress - but they can't just
be left there as it raises a literal red flag on TeamCity which leaves
everyone wondering what went wrong.
https://teamcity.labkey.org/viewLog.html?buildId=816448&tab=Inspection&buildTypeId=ProteoWizard_WindowsX8664msvcProfessionalSkylineResharperChecks

In general it's usually best to accept ReSharper's suggestions for fixes.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -41,6 +40,63 @@
...
+        /// </summary>
+        /// <param name="ex">Exception to check</param>
+        /// <returns>True if this or any nested exception is of <see cref="AuditLogException"> AuditLogException </see> type.</returns>
+        public static bool IsAuditLogInvolved(Exception ex)
+        {
+            var isAuditLogException = false;
+
+            do      //traverse the nested exception chain to look for AuditLogException
+            {
+                if (ex.GetType() == typeof(AuditLogException))
```
nickshulman: It would probably be more correct to do "ex is AuditLogException" instead of matching the types exactly.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -41,6 +40,63 @@
...
+                {
+                    isAuditLogException = true;
```
nickshulman: You should just "return true" here. There is no reason to keep walking up the hierarchy since it won't change the result.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -140,11 +196,17 @@ public void Validate()
...
-                Assume.IsTrue(entry.TimeStampUTC <= time && entry.LogIndex > logIndex,
-                    AuditLogStrings.AuditLogList_Validate_Audit_log_is_corrupted__Audit_log_entry_time_stamps_and_indices_should_be_decreasing);
-
+                if (entry.TimeStampUTC > time || entry.LogIndex <= logIndex)
+                {
+                    string msg = string.Format(
+                        AuditLogStrings.AuditLogList_Validate_Audit_log_is_corrupted__Audit_log_entry_time_stamps_and_indices_should_be_decreasing,
+                        entry.LogIndex, entry.TimeStampUTC, logIndex, time);
+                    Exception e = new AuditLogException(msg);
+                    throw e;
```
nickshulman: Probably just do:
throw new AuditLogException(msg)
(There are a couple of other places where you split the construction of an exception and the throwing onto separate lines, but I cannot think of a reason that you would want to do that).

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -265,21 +320,20 @@ public static bool ReadFromXmlTextReader(XmlTextReader reader, out string logged
                 .ChangeActualHash(result.CalculateRootHash())
                 .ChangeSkylHash(Hash.FromBase64(rootHash));
 
+            // yell if audit log entry hash does not match.
             // If the docFormat is null, this is an old audit log and there won't be any entry hashes.
             // We can't just always ignore non-existent hashes, otherwise people could just delete the hash elements
             // and get Skyline to successfully load the audit log
-            if (docFormat != null && !result.RootHash.SkylAndActualHashesEqual())
+            var modifiedEntries =
+                result.AuditLogEntries.Enumerate().Where(entry => !entry.Hash.SkylAndActualHashesEqual());
```
nickshulman: You should call "ToArray()" or "ToList()" here so you don't get the Resharper warning about enumerating the collection more than once.

pwiz_tools/Skyline/Model/AuditLog/AuditLogStrings.resx
```diff
@@ -633,7 +633,7 @@
     <value>The document was modified outside of Skyline: {0}</value>
   </data>
   <data name="AuditLogList_Validate_Audit_log_is_corrupted__Audit_log_entry_time_stamps_and_indices_should_be_decreasing" xml:space="preserve">
-    <value>Audit log is corrupted. Audit log entry time stamps and indices should be decreasing</value>
+    <value>Audit log is corrupted. Audit log entry time stamps and indices should be decreasing. Entry number {0} timestamp {1:G} occurs before entry number {2} timestamp {3:G}.</value>
```
nickshulman: It's weird to say that the audit log entries should have their time stamps decreasing, since users would think of the audit log as being something in forward chronological order.

Maybe this error messsage could be changed to just say "in order" instead of either "decreasing" or "increasing".

pwiz_tools/Skyline/TestFunctional/AuditLogValidationTest.cs
```diff
@@ -0,0 +1,105 @@
...
+            RunFunctionalTest();
+
+        }
+
+        protected override void DoTest()
+        {
+            //testing scenarios:
+            // - valid audit log file - opens cleanly
+            var documentFile = TestFilesDir.GetTestPath(@"AuditLogValidationTest/MethodEditClean.sky");
+            WaitForCondition(() => File.Exists(documentFile));
```
nickshulman: I don't understand why you're doing "WaitForCondition" instead of just "Assert"

pwiz_tools/Skyline/TestFunctional/AuditLogValidationTest.cs
```diff
@@ -0,0 +1,105 @@
...
+            //making sure we get the expected error message
+            string dialogMessage = "(none)";
+            RunUI(() => dialogMessage = messageDialog.Message);
+            if (reTest.Match(dialogMessage).Value == String.Empty)
+            {
+                Console.Write("Error message: " + dialogMessage);
+                Assert.Fail("Unexpected exception type when opening document with entries out of order audit log.");
+            }
+            //close the dialog
+            RunUI(() => messageDialog.ClickOk());
```
nickshulman: It would probably be better to do:
OkDialog(messageDialog, messageDialog.ClickOk)
since that will make sure that the test waits for the dialog to disappear.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -265,21 +320,20 @@ public static bool ReadFromXmlTextReader(XmlTextReader reader, out string logged
...
-                var modifiedEntries =
-                    result.AuditLogEntries.Enumerate().Where(entry => !entry.Hash.SkylAndActualHashesEqual());
-                var dlg = new AlertDlg(AuditLogStrings.AuditLogList_ReadFromFile_The_following_audit_log_entries_were_modified +
-                                       string.Join(Environment.NewLine,
-                                           modifiedEntries.Select(entry => entry.UndoRedo.ToString())),
-                    MessageBoxButtons.OK);
-                dlg.ShowParentlessDialog();
+                throw new AuditLogException(
+                    AuditLogStrings.AuditLogList_ReadFromFile_The_following_audit_log_entries_were_modified +
+                    string.Join(Environment.NewLine, modifiedEntries.Select(entry => entry.UndoRedo.ToString())));
```
nickshulman: There's a method called "TextUtil.LineSeparate" which you might want to use instead of string.Join

pwiz_tools/Skyline/TestFunctional/AuditLogValidationTest.cs
rita-gwen: Well, this is the first Skyline functional test I wrote. This line was copied from AbstractFunctionalTestEx.OpenDocument() method. I would assume that ZIP extraction might be still in progress when the test starts, hence the wait. If it's not true then maybe the base class should be modified as well.

rita-gwen:
Ok, looks like everything is in order now. Merging into the trunk.

bspratt:
Looks like you still have a code inspection issues:
https://teamcity.labkey.org/viewLog.html?buildId=822986&tab=Inspection&buildTypeId=ProteoWizard_WindowsX8664msvcProfessionalSkylineResharperChecks
SkylineFiles.cs (3)
365
<https://teamcity.labkey.org/viewLog.html?buildId=822986&tab=Inspection&buildTypeId=ProteoWizard_WindowsX8664msvcProfessionalSkylineResharperChecks#>:
Possible 'System.NullReferenceException'
410
<https://teamcity.labkey.org/viewLog.html?buildId=822986&tab=Inspection&buildTypeId=ProteoWizard_WindowsX8664msvcProfessionalSkylineResharperChecks#>:
Possible 'System.NullReferenceException'
412
<https://teamcity.labkey.org/viewLog.html?buildId=822986&tab=Inspection&buildTypeId=ProteoWizard_WindowsX8664msvcProfessionalSkylineResharperChecks#>:
Possible 'System.NullReferenceException'

rita-gwen:
They are pre-existing warnings in the code I did not touch. I looked through that code and they do not look like real warnings - I did not see any way it could happen. If there is a way to suppress them then I think they should be suppressed.

bspratt:
Then yes, they should be suppressed - ReSharper can walk you through it.

But it's never OK to merge to trunk when the code inspection test is
failing - the reasoning being that as soon as you have any warnings that
are "OK to ignore" you'll likely never be alerted to the new ones that
actually matter.

There are no pre-existing warnings in master - so it's worth understanding
why these suddenly lit up for you. Although it may well just be ReSharper
weirdness (new version of ReSharper etc), it's weirdness worth looking at.

PR #612, opened 2019-08-14
Extracting chromatograms from MS3 scans

nickshulman:
This builds on Matt's work in the "feature/Thermo-SPS-Masses" branch.
This makes it so that Skyline will extract chromatograms from scans whose MS level is 2 or greater so long as the MS1 level isolation window matches the precursor in the Skyline document.


chambm:
Huh. Well the tests went from passing to failing, so the question is were the tests wrong or the changes? :) I suppose it needs an ms3 test added as well.

PR #630, opened 2019-09-05
Fixed issue with non-SPS spectra

chambm:
- fixed issue with non-SPS spectra being treated as having multiple precursors, which disabled utilization of the monoisotopic m/z


chambm:
@brendanx67 @nickshulman This should fix FAIMS. We really ought to run the perf tests nightly, except maybe TestPeakPerf.

brendanx67:
We do run the perf tests nightly. Don't you get the emails with these test failures every morning?

chambm:
Yes I get the emails. I mean having the perf tests run for PRs and reported to GitHub, so PRs get the benefit of perf testing before they are merged to master. This was a pretty serious bug (basically disabling monoisotopic m/z for non-SPS spectra from modern data that has that "SPS Masses" field left blank). The bug got through other testing because our Thermo unit test file has all monoisotopic m/z set to 0. And apparently all the other Skyline tests don't care about monoisotopic m/z being correct. :)

brendanx67:
It does take a lot of disk space (and time) to run the perf tests, which is why we dedicate 12 hours to those runs nightly and only run on certain machines. Unlike the other tests, we also do not wipe the system clean of the test files, because it takes so long to download them. Might be challenging in a TeamCity AWS environment.

PR #631, opened 2019-09-06
Do not try to find a peak if no explicit peak was found in EnyclopeDI…

nickshulman:
…A or other library.

If a library such as EncyclopeDIA has explicit peak information for a particular result file, and has explicit peak information for a peptide in other result files, but not the one being asked about, then conclude that the peptide was not found in that particular file.

It used to be that Skyline would then try to do its own peak finding if EncyclopeDIA said it couldn't find a peak for that peptide in that file.
Now what happens is Skyline does not choose a peak at all, and the chromatogram group gets an empty candidate peak list.


bspratt:
Sounds a bit similar to my work around less desperate peak picking when using "Explicit Retention Time" in PR 633, it would be great if you could review that while this is fresh in your mind.

PR #633, opened 2019-09-06
Skyline: make the "Explicit Retention Time" value behave as most user…

bspratt:
…s expect - only peaks that are within the explicit RT+window will be accepted, so if no peak falls within that range Skyline will simply not assign a peak (rather than picking the best nearby candidate outside the window, as before).


bspratt:
Here's the revised display when explicit RT has a window. This will appear instead of predicted RT if both are available.
image
And here it is when set to not display RT (formerly it showed explicit RT anyay)
image

brendanx67:
@bspratt please revive this and get it into master for our first post-20.1 skyline-daily

PR #637, opened 2019-09-09
Skyline: support KEGG compound IDs as molecular identifiers in transi…

bspratt:
…tion lists etc


pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
```diff
@@ -735,6 +735,15 @@ private MoleculeAccessionNumbers ReadMoleculeAccessionNumberColumns(Row row)
                 moleculeIdKeys.Add(MoleculeAccessionNumbers.TagSMILES, smiles);
             }
 
+            var keggCol = ColumnIndex(SmallMoleculeTransitionListColumnHeaders.idKEGG);
+            var kegg = NullForEmpty(row.GetCell(keggCol));
+            if (kegg != null)
+            {
+                // Should have form like C07481 but we'll accept anything for now, having no proper parser
+                kegg = kegg.Trim();
```
nickshulman: I think this is redundant since "NullForEmpty" already calls "Trim()".

pwiz_tools/Skyline/Model/Databinding/Entities/ColumnToolTips.resx
```diff
@@ -1269,6 +1269,9 @@ higher than the other extent.</value>
   <data name="InChiKey" xml:space="preserve">
     <value>Standardized molecular identifier</value>
   </data>
+  <data name="KEGG" xml:space="preserve">
+    <value>Standardized molecular identifier</value>
```
nickshulman: Maybe we should mention that this stands for "Kyoto Encyclopedia of Genes and Genomes"

pwiz_tools/Skyline/Model/Databinding/Entities/ColumnToolTips.resx
```diff
@@ -1269,6 +1269,9 @@ higher than the other extent.</value>
   <data name="InChiKey" xml:space="preserve">
     <value>Standardized molecular identifier</value>
   </data>
+  <data name="KEGG" xml:space="preserve">
+    <value>Standardized molecular identifier</value>
+  </data>
   <data name="SMILES" xml:space="preserve">
     <value>Standardized molecular identifier</value>
```
nickshulman: Maybe we should mention that this stands for "simplified molecular-input line-entry system" (which coincidentally has the same first letters as "Standardized molecular identifier").

pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
bspratt: I don't think it does, actually, but maybe it should. There are other implementations in our code that do.

nickshulman: If it were true that "NullForEmpty" does not do a "Trim()", then you would need to call "NullForEmpty" again after you do the Trim(), otherwise it's possible to end up with the empty string.

bspratt: Yeah, I've gone in and tidied that up -now NullForEmpty does the Trim().

PR #643, opened 2019-09-15
[mz5] fixed typo in data set ChromatogramTime

jeleclaire:
Fixed the typo in data set "ChomatogramTime" into "ChromatogramTime". Backward compatibility ensured.


chambm:
Thanks for the PR. I'd noticed that before, but left it alone because I didn't want to break anybody else's reader. Not that I'm actually aware of any other readers. Do you know of any?

jeleclaire:
Good point. I always thought OpenMS would support mz5 but I couldn't find a hint.

brendanx67:
Can we please decide whether we want this change and either merge it or close it?

jeleclaire:
I think, this change should be incorporated. First, even if it is just a typo, it is not the first time it was discovered and if it is not changed, it will be rediscovered again and again. Second, there were changes to the mz5 core anyway, so maybe now is a good time point to include it.

brendanx67:
I agree. As long as we are backward compatible, which it looks like we are, then I think we should just fix this before there are more readers that would be broken by the fix.

PR #644, opened 2019-09-17
Add option to sort by m/z on export method dialog

kaipot:


pwiz_tools/Skyline/CommandArgUsage.resx
```diff
@@ -724,4 +724,7 @@ greater peak area.</value>
   <data name="_refine_max_precursor_only" xml:space="preserve">
     <value>Only the precursor charge state with the maximum peak area will be retained.</value>
   </data>
+  <data name="_exp_order_by_mz" xml:space="preserve">
+    <value>Order exported method by m/z, ascending.</value>
```
brendanx67: Add "for optimized quadrupole switching" for a better explanation. This is the worst kind of documentation. It really just restates what you can read in the name other than "ascending". We could also add "(precursor m/z, then product m/z when applicable)" to be even more descriptive about what this is going to do.

pwiz_tools/Skyline/TestFunctional/ExportMethodDialogTest.cs
```diff
@@ -578,12 +579,48 @@ private void BrukerTOFMethodTest()
             string brukerTemplateMeth = TestFilesDir.GetTestPath("Bruker Template Scheduled Precursor List.m");
             WaitForDocumentLoaded();
 
+            // Test order by m/z
```
brendanx67: Need a corresponding command-line test to prove the new command argument works.

PR #650, opened 2019-09-24
Skyline: Uniprot now just drops the connection when presented with a …

bspratt:
…too-long search query, so treat that as a cue to back off on protein metadata search batch size

### DIFF
--- a/pwiz_tools/Shared/ProteomeDb/Fasta/WebEnabledFastaImporter.cs
+++ b/pwiz_tools/Shared/ProteomeDb/Fasta/WebEnabledFastaImporter.cs
@@ -1305,6 +1305,10 @@ namespace pwiz.ProteomeDatabase.Fasta
                                 return URL_TOO_LONG; // Probably asked for too many at once, caller will go into batch reduction mode
                         }
                     }
+                    else if (ex.Status == WebExceptionStatus.ConnectionClosed && searchType == UNIPROTKB_TAG) // Uniprot just drops the connection on too-large searches as of Sept 2019
+                    {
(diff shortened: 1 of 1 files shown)


pwiz_tools/Shared/ProteomeDb/Fasta/WebEnabledFastaImporter.cs
```diff
@@ -1305,6 +1305,10 @@ private int DoWebserviceLookup(IList<ProteinSearchInfo> proteins, char searchTyp
                                 return URL_TOO_LONG; // Probably asked for too many at once, caller will go into batch reduction mode
                         }
                     }
+                    else if (ex.Status == WebExceptionStatus.ConnectionClosed && searchType == UNIPROTKB_TAG) // Uniprot just drops the connection on too-large searches as of Sept 2019
+                    {
+                        return URL_TOO_LONG; // Probably asked for too many at once, caller will go into batch reduction mode
```
nickshulman: I think you should only return URL_TOO_LONG if proteins.Count > 1.
(Either that, or you should check whether _maxBatchSize[searchType] is greater than 1 when handling URL_TOO_LONG on line 829)

bspratt: Agreed - the code already makes sure that _maxBatchSize[searchType] never drops below 1. 

_maxBatchSize[searchType] = Math.Max(1, (int)(_maxBatchSize[searchType] * 0.75));

Or it's meant to anyway - am I not seeing something?

PR #651, opened 2019-09-25
Skyline: Fix ExportMethodDlg threading to cancel when form closes and…

brendanx67:
… keep from Invoking on a destroyed form

### DIFF
--- a/pwiz_tools/Skyline/FileUI/ExportMethodDlg.cs
+++ b/pwiz_tools/Skyline/FileUI/ExportMethodDlg.cs
@@ -62,7 +62,7 @@ namespace pwiz.Skyline.FileUI
         {
             InitializeComponent();
 
-            _exportProperties = new ExportDlgProperties(this, _cancellationTokenSource);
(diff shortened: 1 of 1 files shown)


pwiz_tools/Skyline/FileUI/ExportMethodDlg.cs
```diff
@@ -54,11 +56,13 @@ public sealed partial class ExportMethodDlg : CreateHandleDebugBase, IMultipleVi
...
         private readonly ExportDlgProperties _exportProperties;
 
+        private CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();
+
         public ExportMethodDlg(SrmDocument document, ExportFileType fileType)
         {
             InitializeComponent();
 
-            _exportProperties = new ExportDlgProperties(this);
+            _exportProperties = new ExportDlgProperties(this, _cancellationTokenSource);
```
nickshulman: You should probably pass _cancellationTokenSource.Token in here.

Objects that just need to know whether the job is cancelled should be given a CancellationToken.
The only objects that should see the CancellationTokenSource are the ones that need to be able to cancel the job.

PR #655, opened 2019-09-25
Fixes for ReSharper PossibleNullReference code inspection failure.

rita-gwen:


pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -407,10 +407,15 @@ public bool OpenFile(string path, FormEx parentWindow = null)
             // Once user has opened an existing document, stop reminding them to set a default UI mode
             if (string.IsNullOrEmpty(Settings.Default.UIMode))
             {
-                var mode = document.DocumentType == SrmDocument.DOCUMENT_TYPE.none
-                    ? SrmDocument.DOCUMENT_TYPE.proteomic
-                    : document.DocumentType;
-                Settings.Default.UIMode = mode.ToString();
+                if (document != null)
+                {
+                    var mode = document?.DocumentType == SrmDocument.DOCUMENT_TYPE.none
```
brendanx67: The ? in document?. now seems redundant since you are checking document != null. Surprised this doesn't cause some kind of warning.

rita-gwen: Well, just adding the document != null check did not disable the NullReferenceException inspection warning. 
In any case this PR is redundant because Brian already disabled them in master.

PR #668, opened 2019-10-03
iRT calibration form improvements and iRT standards as Settings list

kaipot:


pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
```diff
@@ -66,246 +120,537 @@ private void OnLoad(object sender, EventArgs e)
...
             {
-                if (GridView.IsCurrentCellDirty)
+                yield return new RegressionOption(Resources.RegressionOption_All_Fixed_points, null, null);
+
+                var docPeptides = document.Peptides.Where(nodePep => nodePep.PercentileMeasuredRetentionTime.HasValue && !nodePep.IsDecoy)
+                    .ToDictionary(nodePep => nodePep.ModifiedTarget, nodePep => nodePep); // TODO duplicate peptides?
+
+                var removedValues = new List<Tuple<double, double>>();
+
+                foreach (var standard in IrtStandard.ALL.Where(standard => !ReferenceEquals(standard, IrtStandard.EMPTY)))
```
nickshulman: Should this be iterating over Settings.Default.IrtStandardList instead of IrtStandard.ALL so that it includes user-generated standards?
(I don't understand what's going on here, so I might be wrong).

kaipot: I don't know that there would be a use case for this and it might just be confusing for the user, but I could be wrong. Maybe if they have two sets of their own standards and want to calibrate one against the other?

pwiz_tools/Skyline/Properties/Settings.cs
```diff
@@ -2143,7 +2165,65 @@ public bool CanEditItem(RetentionScoreCalculatorSpec item)
...
+        {
+            return (IrtStandard) item.ChangeName(string.Empty);
+        }
+
+        public override IEnumerable<IrtStandard> GetDefaults(int revisionIndex)
+        {
+            return IrtStandard.ALL;
+        }
+
+        public void EnsureDefault()
```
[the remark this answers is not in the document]
kaipot: I am not sure either, I'll remove it.

pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
brendanx67: I agree with Nick, and I think you have the right use case. Imagine none of the list of standards we know of actually existed. Then the only way to calibrate multiple standards to the same scale would be using Settings.Default.IrtStandardList, because IrtStandard.All would be empty. Since we can't foresee what kind of standards users will create, it is totally valid that they may eventually want to calibrate one custom set with another custom set.

kaipot: Makes sense, I committed a change to iterate over the settings list instead of just the built in ones.

PR #676, opened 2019-10-09
Store document XML in irtdb files

kaipot:
When saving an irtdb file, we check the document for standard peptides with ranked transitions and store document XML containing this information into the database. When switching iRT calculator, we check if the irtdb has any document XML and if so, prompt the user to import it into their document (if no document XML, fall back to old behavior, check if any of our built in standards match)


pwiz_tools/Skyline/Model/Irt/IrtStandard.cs
```diff
@@ -454,11 +454,9 @@ private static DbIrtPeptide MakePeptide(string sequence, double time)
             return new DbIrtPeptide(new Target(sequence), time, true, TimeSource.peak);
         }
 
-        public static IrtStandard WhichStandard(ICollection<Target> peptides, out HashSet<Target> missingPeptides)
+        public static IrtStandard WhichStandard(ICollection<Target> peptides)
         {
-            var standard = ALL.FirstOrDefault(s => s.ContainsAll(peptides.Select(p => new DbIrtPeptide(p, 0, true, TimeSource.peak)).ToList(), null)) ?? EMPTY;
-            missingPeptides = new HashSet<Target>(standard.Peptides.Where(s => !peptides.Any(p => p.Equals(s.Target))).Select(s => s.Target));
-            return standard;
+            return ALL.FirstOrDefault(s => s.ContainsAll(peptides.Select(p => new DbIrtPeptide(p, 0, true, TimeSource.peak)).ToList(), null)) ?? EMPTY;
```
nickshulman: Probably should get:
peptides.Select(p => new DbIrtPeptide(p, 0, true, TimeSource.peak)).ToList()
out of the loop so it only gets called once

PR #678, opened 2019-10-10
Skyline/work/20191004 mru chromatograms

bspratt:
When loading many (i.e. thousands) of replicates it's possible for Skyline to run out of Windows handles. This change avoids the problem by limiting the number of chromatogram displays to no more than 100 replicates at a time.

We may wish to make that limit configurable, but for now it's hard coded. Nick points out that "1" is actually a pretty useful value when doing manual inspection.


pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -2998,6 +3037,14 @@ private void closeAllChromatogramsMenuItem_Click(object sender, EventArgs e)
             }
         }
 
+        /// <summary>
```
brendanx67: Normally, we do this by extracting a public function and having the event handler call that, rather than adding a public function that calls the event handler. This is safer and avoids the confusion of someone trying to use the passed in objects when they are null (for your test method).

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -2584,10 +2618,15 @@ private GraphChromatogram CreateGraphChrom(string name)
             graphChrom.ChangedPeakBounds += graphChromatogram_ChangedPeakBounds;
             graphChrom.PickedSpectrum += graphChromatogram_PickedSpectrum;
             graphChrom.ZoomAll += graphChromatogram_ZoomAll;
-            _listGraphChrom.Add(graphChrom);
+            AddGraphChromToList(graphChrom);
             return graphChrom;
         }
 
+        private void AddGraphChromToList(GraphChromatogram graphChrom)
```
brendanx67: Why this change? It doesn't seem to get used throughout the code, and it doesn't seem to provide any encapsulation, being private.

pwiz_tools/Skyline/Controls/Graphs/AllChromatogramsGraph.cs
```diff
@@ -436,6 +436,10 @@ private void AddProgressControls(MultiProgressStatus status)
         {
             // Match each file status with a progress control.
             bool first = true;
+            var width = flowFileStatus.Width - 2 -
```
brendanx67: Not sure what this is. Bad merge?

pwiz_tools/Skyline/Model/Results/MsDataFileScanHelper.cs
```diff
@@ -67,9 +67,9 @@ public ChromSource Source
                 var newTimeIntensities = GetTimeIntensities(Source);
                 if (newTimeIntensities != null)
                 {
-                    if (oldTimeIntensities != null)
+                    if (oldTimeIntensities != null && oldTimeIntensities.NumPoints > 0)
```
brendanx67: Not sure what this is. Bad merge?

PR #682, opened 2019-10-21
Fix settings persistence with AssemblyInfo versioning

chambm:
* added SettingsUpgradeRequired setting and Settings.Upgrade() call to Skyline, SkylineTester, and SkylineNightly to persist settings between versions (settings between different locations will still be different)
* removed Skyline.Properties.Settings section from app.config after testing that settings are still carried over from a release ClickOnce installation


chambm:
It probably looks for a lower version number one directory up from user.config.

PR #683, opened 2019-10-21
Try to improve iRT peptide choices based on qvalue/score

kaipot:


pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
```diff
@@ -692,46 +693,67 @@ private static List<MeasuredPeptide> FindBestPeptides(SrmDocument doc, int pepti
...
             /// The returned list is guaranteed to be sorted by retention time.
             /// </summary>
+            /// <param name="doc">Document containing the peptides to choose from</param>
             /// <param name="docPeptides">Peptides to choose from</param>
             /// <param name="peptideCount">The number of peptides desired</param>
-            private static List<MeasuredPeptide> FindEvenlySpacedPeptides(List<MeasuredPeptide> docPeptides, int peptideCount)
+            private static List<MeasuredPeptide> FindEvenlySpacedPeptides(SrmDocument doc, List<MeasuredPeptide> docPeptides, int peptideCount)
             {
-                docPeptides.Sort((x, y) => x.RetentionTime.CompareTo(y.RetentionTime));
                 if (docPeptides.Count == peptideCount)
```
nickshulman: I think this should be "<=".

pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
```diff
@@ -692,46 +693,67 @@ private static List<MeasuredPeptide> FindBestPeptides(SrmDocument doc, int pepti
...
-                List<MeasuredPeptide> standardPeptides = new List<MeasuredPeptide>();
-                double gradientLength = maxRT - minRT;
-                for (int i = 0; i < peptideCount; i++)
-                {
-                    double targetRT = minRT + i * (gradientLength / (peptideCount - 1));
-                    for (int j = 0; j < docPeptides.Count; j++)
+                var resultsHandler = new MProphetResultsHandler(doc, model);
+                resultsHandler.ScoreFeatures(null, true);
+
+                var docTargets = docPeptides.ToDictionary(pep => pep.Target, pep => pep);
```
nickshulman: I think you need to guard against an exception caused by the duplicate peptides here.

pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
```diff
@@ -692,46 +693,67 @@ private static List<MeasuredPeptide> FindBestPeptides(SrmDocument doc, int pepti
...
+                    minRt + rtRange * 4 / 8,
+                    minRt + rtRange * 6 / 8,
+                    minRt + rtRange * 7 / 8,
+                    double.MaxValue
+                });
+                var endBuckets = new[] {buckets.First(), buckets.Last()};
+                var midBuckets = buckets.Skip(1).Take(buckets.Length - 2).ToArray();
+
+                var bestPeptides = new List<MeasuredPeptide>();
+                while (bestPeptides.Count < peptideCount)
```
nickshulman: Maybe you should put something in this loop to make it so that you break out of the loop if all of the buckets are empty. I know that can't happen right now, but it's always good to be safe.

pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
```diff
@@ -692,46 +693,67 @@ private static List<MeasuredPeptide> FindBestPeptides(SrmDocument doc, int pepti
...
+                    scoredPeptides.Add(new ScoredPeptide(docTargets[nodePep.ModifiedTarget], value));
                 }
 
-                return standardPeptides;
+                var retentionTimes = scoredPeptides.Select(pep => pep.Peptide.RetentionTime).OrderBy(rt => rt).ToArray();
+                var minRt = retentionTimes.First();
+                var maxRt = retentionTimes.Last();
+                var rtRange = maxRt - minRt;
+
+                var buckets = PeptideBucket.BucketPeptides(scoredPeptides, new[]
```
nickshulman: I would have probably done something with "IEnumerable.ToLookup" instead of making a separate PeptideBucket class, but this works too.

pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
kaipot: Good

pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
```diff
@@ -692,46 +693,63 @@ private static List<MeasuredPeptide> FindBestPeptides(SrmDocument doc, int pepti
...
+            {
+                if (docPeptides.Count <= peptideCount)
+                    return docPeptides.Select(pep => pep.Item1).OrderBy(pep => pep.RetentionTime).ToList();
+
+                var model = doc.Settings.PeptideSettings.Integration.PeakScoringModel;
+                if (model == null || !model.IsTrained)
+                    model = LegacyScoringModel.DEFAULT_MODEL;
+
+                var resultsHandler = new MProphetResultsHandler(doc, model);
+                resultsHandler.ScoreFeatures(null, true);
```
brendanx67: Probably should introduce a progress monitor here. This rescoring could take serious time and it looks like you will just lock up the UI right now.

pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
```diff
@@ -924,4 +942,76 @@ public class StandardPeptide : MeasuredPeptide
...
+            _peptides.RemoveAt(0);
+            return pep;
+        }
+
+        public static PeptideBucket[] BucketPeptides(IEnumerable<ScoredPeptide> peptides, double[] rtBoundaries)
+        {
+            var buckets = rtBoundaries.OrderBy(x => x).Select(boundary => new PeptideBucket(boundary)).ToArray();
+            var curBucketIdx = 0;
+            var curBucket = buckets[0];
+            foreach (var pep in peptides.OrderBy(pep => pep.Peptide.RetentionTime))
```
brendanx67: The sorts seem potentially slow. Probably not a huge issue, but because you essentially use multiples of 1/8 your range, you could also just us an array of 8 indexes and RT math to decide which bucket a specific RT goes into.

bucketIndex = bucketIndexes[(rt - minRT)/8]

bucketIndexes = new[] {0, 1, 1, 2, 2, 3, 3, 4};

kaipot: I realized that the sort isn't necessary, since we already had to do the sort before to find the min/max RT.

pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
```diff
@@ -662,87 +680,157 @@ public void SetPeptides(IEnumerable<MeasuredPeptide> peps)
...
+                var removedValues = new List<Tuple<double, double>>();
+
+                RegressionLine cirtRegression = null;
+                var cirtUsePredefined = false;
+
+                if (RCalcIrt.TryGetRegressionLine(rts, irts, peptideCount, out var regression, removedValues))
+                {
+                    for (var i = scoredCirtPeptides.Count - 1; i >= 0; i--)
+                    {
+                        if (removedValues.FirstOrDefault(tuple => tuple.Item1.Equals(rts[i]) && tuple.Item2.Equals(irts[i])) != null)
```
nickshulman: It is simpler to do:
if (removedValues.Contains(Tuple.Create(rts[i], irts[i])))

pwiz_tools/Skyline/TestFunctional/IrtTest.cs
```diff
@@ -143,8 +143,8 @@ private void RunIrtTest()
                       {
                           Assert.AreEqual(numStandardPeps, irtDlg1.StandardPeptideCount);
                           //And that there are 3 below 0 and 3 above 100
-                          Assert.AreEqual(3, irtDlg1.StandardPeptides.Count(pep => pep.Irt < 0));
-                          Assert.AreEqual(3, irtDlg1.StandardPeptides.Count(pep => pep.Irt > 100));
+                          Assert.AreEqual(3, irtDlg1.StandardPeptides.Count(pep => Math.Round(pep.Irt, 2) < 0));
```
nickshulman: Would it make more sense if this were:
Assert.AreEqual(3, irtDlg1.StandardPeptides.Count(pep => pep.Irt < -0.005));

pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
```diff
@@ -523,22 +545,34 @@ private void ShowGraph(string title, double[] xValues, double[] yValues, Diction
         public class RegressionOption
         {
             public string Name { get; }
-            public List<Tuple<DbIrtPeptide, PeptideDocNode>> MatchedPeptides { get; }
+            public RegressionLine RegressionLine { get; }
+            public List<Tuple<DbIrtPeptide, PeptideDocNode>> MatchedRegressionPeptides { get; }
+            public List<StandardPeptide> StandardPeptides { get; }
+            public bool AllowEditEquation { get; }
```
nickshulman: It might be simpler to just have this property return "RegressionLine == null".
That is, you are allowed to edit the equation if the RegressionLine is null.

pwiz_tools/Skyline/TestFunctional/IrtTest.cs
kaipot: Maybe, I like the rounding though because it's comparing the numbers that are visible to the user in the dialog.

pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
kaipot: Originally I had it like that, then made it a separate property. But you're right, I don't think there will be any case where we have a regression line and want it to be editable.

chambm:
Definitely squash this one merge. :)

PR #691, opened 2019-10-23
Match heavy iRT peptides

kaipot:


pwiz_tools/Skyline/Model/Irt/RCalcIrt.cs
```diff
@@ -243,6 +244,38 @@ public IEnumerable<DbIrtPeptide> GetDbIrtPeptides()
...
+                            MeasuredResults.MergeAction.remove,
+                            false,
+                            null,
+                            Settings.Default.StaticModList,
+                            Settings.Default.HeavyModList,
+                            null,
+                            out _,
+                            out _,
+                            false);
+                        standardPeptideList = standardPeptideList.Select(pep => new DbIrtPeptide(pep)).ToArray();
```
nickshulman: You end up replacing every element in this list, so you would probably just start with an empty list.

pwiz_tools/Skyline/Model/Irt/RCalcIrt.cs
```diff
@@ -243,6 +244,38 @@ public IEnumerable<DbIrtPeptide> GetDbIrtPeptides()
             IEnumerable<IRetentionTimeProvider> providers, int countProviders,
             DbIrtPeptide[] standardPeptideList, DbIrtPeptide[] items)
         {
+            var matchedStandard = IrtStandard.WhichStandard(standardPeptideList.Select(pep => pep.ModifiedTarget));
+            if (matchedStandard != null)
+            {
+                var dummyDoc = new SrmDocument(SrmSettingsList.GetDefault());
+                using (var reader = matchedStandard.DocumentReader)
```
nickshulman: You should change the DocumentReader property into a function "GetDocumentReader()".

It's not a good idea to have a property where the caller is expected to call "Dispose" on the value.

PR #699, opened 2019-10-30
Absolute quantification tutorial and test revision

rita-gwen:
Test code is updated. I left some comments in the doc file. We can work on downloading the binaries tomorrow.


pwiz_tools/Skyline/TestTutorial/AbsoluteQuantTutorialTest.cs
```diff
@@ -49,16 +50,25 @@ public void TestAbsoluteQuantificationTutorial()
             // Set true to look at tutorial screenshots.
             // IsPauseForScreenShots = true;
 
-            ForceMzml = true;   // Mzml is ~8x faster for this test.
-
+            ForceMzml = false;   // Mzml is ~8x faster for this test.
```
brendanx67: We don't actually want to do this for the main test implementation. It is quite a bit slower to use the .raw files. If we need this for the tutorial screenshots, then this should be based on whether PauseForScreenShots is on.

pwiz_tools/Skyline/TestTutorial/TutorialAuditLogs/en/TestAbsoluteQuantificationTutorial.log
```diff
@@ -31,40 +31,40 @@ Undo Redo : Imported results from 8 files
 Summary   : Imported results from 8 files
 All Info  :
 Imported results from 8 files
-Import results settings > File names : contains "Standard_1.mzML"
```
brendanx67: This file shouldn't have to change if you get the mzML use correct based on screenshot mode.

pwiz_tools/Skyline/TestTutorial/AbsoluteQuantTutorialTest.cs
```diff
@@ -237,14 +258,15 @@ protected override void DoTest()
                 int transitionGroupCount = SkylineWindow.DocumentUI.Peptides.First().TransitionGroupCount;
                 CheckGstGraphs(transitionGroupCount, transitionGroupCount - 1);
             });
+            RestoreViewOnScreen(11);
             PauseForScreenShot("Main window with totals graphs for light and heavy and FOXN1-GST", 11);
             
             // Peptide Quantitification Settings p. 11
-            var peptideSettingsUi = ShowDialog<PeptideSettingsUI>(SkylineWindow.ShowPeptideSettingsUI);
+            peptideSettingsUi = ShowDialog<PeptideSettingsUI>(SkylineWindow.ShowPeptideSettingsUI);
+            RunUI(() => peptideSettingsUi.SelectedTab = (PeptideSettingsUI.TABS)5);
```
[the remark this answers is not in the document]
rita-gwen: Yes, it should. But unfortunately it does not work. Tab control indexes only visible tabs and Labels tab is before Quantification ans is hidden so it messes up PeptideSettingsUI.Tabs.Labels enum. `(PeptideSettingsUI.TABS)5` is a workaround for test purposes but I think it should be fixed in the code at some point even though it does not manifest itself in the existing tests. I mentioned it to Brian P already.

rita-gwen:
This was a bit premature. There is still an issue that Nat reported above that I am still working on. I guess I'll create another pull request for it them.

PR #702, opened 2019-10-31
Skyline/work/20191030 prosit secure connection

nickshulman:
This change makes it so that you can put a file:
PrositConfig_production.xml
in your source tree. That XML file is expected to contain a server name and some certificates. This enables Skyline to be built with those credentials.
If that file is not present, then "PrositConfig_development.xml" gets used instead.

Other changes: You are no longer allowed to edit the server name in Tools > Options. The server name comes from PrositConfig.xml and cannot be changed after Skyline is built. (The combo box is now a read-only textbox).

Also, the .gitignore file "UpdatePrositConfiguration.bat" exist so that you can have PrositConfig_production.xml in your source tree and not be worried about accidentally committing it.


brendanx67:
@nickshulman Can you get the Prosit integration branch up to date with master (currently it has conflicts) and then get it and this PR passing tests and then merge to one PR? We can make the resulting branch the integration branch and start running nightly tests on it.

PR #707, opened 2019-11-01
Protein combobox on ChangeIrtPeptidesDlg; peptide modification matching

kaipot:


pwiz_tools/Skyline/SettingsUI/Irt/ChangeIrtPeptidesDlg.cs
```diff
@@ -119,5 +186,45 @@ private void btnOk_Click(object sender, EventArgs e)
...
+            public ComboBoxProtein(PeptideGroupDocNode protein)
+            {
+                Protein = protein;
+            }
+
+            public string[] PeptideStrings => Protein?.Molecules.Select(pep => pep.ModifiedSequenceDisplay).ToArray() ?? new string[0];
+
+            public IEnumerable<PeptideDocNode> RemovedPeptides(IEnumerable<string> pepLines)
+            {
+                var textTargets = new TargetMap<bool>(pepLines.Select(line => line.Trim())
```
nickshulman: You need to use the "TargetResolver" object to convert from strings back to peptides.
That object knows wheter the ModifiedTarget was able to uniquely identify the molecule, or whether it had to fall back to using a more complicated string with some "#" characters in it.

PR #708, opened 2019-11-01
Agilent /Waters 3 array IMS

chambm:
I made the PR against the DIA branch for convenience (you can easily integrate it into changes there but also back them out to do a Bruker-only release).


bspratt:
Running perf tests now.

brendanx67:
I think things are looking pretty good on this branch and I would like to merge it to master. @chambm 
Do I need to worry about failures on Core builds?

chambm:
Yes. That's not a spurious failure. I thought Brian's "can do" response
implied he was going to handle it but now I realize since the phrase was
missing the subject it could also mean "you can do" or "we can do". :) I'm
traveling today but I will try to make the change (it should be as simple
as moving the logic from the core to msdatafileimpl).

chambm:
OK it's not QUITE that simple because Brian's logic had it limited to Bruker TDF format. Currently preferOnlyMsLevel only works on Bruker TDF format, but even if that was not the case, the indexing stability issue Brian is concerned about would affect all ion mobility formats that support both combineIonMobilitySpectra and preferOnlyMsLevel. So it doesn't make sense to do this filter inside each reader. It makes more sense on the client side, where the client can decide whether index stability is needed or not.

chambm:
I'm testing the fix now.

brendanx67:
Thanks, Matt. Based on your earlier email, I had started to work on that, but had to get an apple pie in the oven. So, nice to find this solved in the intervening period. Your solution is, of course, more informed. I was going to re-introduce my less than ideal test for Bruker TDF with a TODO to come up with something better. Thanks again for the quick fix on TG.

PR #712, opened 2019-11-04
Encountered some damaged Agilent files where it appears that centroid…

bspratt:
… data wasn't written and spectrum representation was never declared, which eventually led to a crash. More robust handling now.


bspratt:
I think maybe this duplicates a fix in the diaPASEF branch - if so, we can delete this branch

chambm:
Sorry I should have said fixing Agilent reader, not data. My fix was for not crashing with the Bruker TDF combined data which previously didn't have any spectrum representation written for it (but I fixed that as well in the same PR).

PR #725, opened 2019-11-14
Support for opening .skyp files in Skyline

vagisha:
Support for opening .skyp files in Skyline.  These files can be downloaded from a Panorama server, and contain the URL of a shared Skyline .zip file on the server. When a user opens a .skyp file the Skyline document it points to should be downloaded and opened in Skyline.


pwiz_tools/Skyline/Model/SkypFile.cs
```diff
@@ -0,0 +1,278 @@
+﻿using System;
```
kaipot: License text should be added here

PR #734, opened 2019-11-21
Handle exception on loading midas library

kaipot:


nickshulman:
Do you think this exception might have been caused by the "|" character that Skyline uses when a .wiff file contains more than one sample? I did send an email to the person who reported this exception but I have not heard back.

kaipot:
> Do you think this exception might have been caused by the "|" character that Skyline uses when a .wiff file contains more than one sample? I did send an email to the person who reported this exception but I have not heard back.

That's a good thought, they mentioned that it used to work for them though so I'm not sure.

pwiz_tools/Skyline/Model/Lib/Midas/MidasLibrary.cs
```diff
@@ -294,7 +294,15 @@ private bool Load(IProgressMonitor monitor)
             _spectra = null;
             if (FilePath == null)
                 return false;
-            var info = new FileInfo(FilePath);
+            FileInfo info;
```
brendanx67: How exactly are you expecting the user to see issues with their Midas library? It seems like all error information is being swallowed and not clearly reported to the user. Even if the eventual result is to report that loading the file failed, it seems like the exception, in this case, might have more information about why.

PR #740, opened 2019-11-26
Verify decoys on Import Results

kaipot:


pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -2601,6 +2638,39 @@ private static IEnumerable<Target> CheckMissingIrtPeptides(SrmDocument document)
                 .Where(target => !existingPeptides.ItemsMatching(new LibKey(target, Adduct.EMPTY).LibraryKey, false).Any());
         }
 
+        public static bool CheckDecoys(SrmDocument document, out int numDecoys, out int numNoSource, out int numWrongTransitionCount)
+        {
+            var targets = new Dictionary<Target, List<PeptideDocNode>>();
```
nickshulman: I think you should do:
targets = document.Peptides.Where(pep=>!pep.IsDecoy).ToLookup(pep=>pep.ModifiedTarget)

pwiz_tools/Skyline/Properties/Resources.resx
```diff
@@ -10799,4 +10799,25 @@ Click 'Retry' to try building again with original spectrum files placed next to
...
+    <value>Continue</value>
+  </data>
+  <data name="SkylineWindow_ImportResults_Generate" xml:space="preserve">
+    <value>Generate</value>
+  </data>
+  <data name="SkylineWindow_ImportResults__0__decoy_s__do_not_have_a_matching_target" xml:space="preserve">
+    <value>{0} decoy(s) do not have a matching target</value>
+  </data>
+  <data name="SkylineWindow_ImportResults__0__decoy_s__do_not_have_the_same_number_of_transitions_as_the_matching_target" xml:space="preserve">
+    <value>{0} decoy(s) do not have the same number of transitions as the matching target</value>
```
nickshulman: I think it should be "their matching target" instead of "the matching target"

pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -2498,6 +2498,43 @@ public void ImportResults()
...
+                    Resources.SkylineWindow_ImportResults_Generate, Resources.SkylineWindow_ImportResults_Continue, true))
+                {
+                    switch (dlg.ShowDialog(this))
+                    {
+                        case DialogResult.Yes:
+                            if (!ShowGenerateDecoysDlg(dlg))
+                                return;
+                            break;
+                        case DialogResult.No:
+                            using (var dlg2 = new MultiButtonMsgDlg(
```
nickshulman: This second confirmation dialog seems excessive.

pwiz_tools/Skyline/SkylineFiles.cs
kaipot: You are right, this is a great spot for a lookup.

PR #748, opened 2019-11-28
Fix exception if a Peptide Modification formula contains "D".

nickshulman:
(Reported by Sarah via exception web)

This change makes it so that the IsotopeAbundances object which we use for calculating masses of the M, M+1, etc in full scan data knows about "D" and "T".
This problem was reported into by someone who used "D" in their peptide isotope modification, but it could also happen with ordinary small molecules.
IsotopeEnrichments calls BioMassCalc.AddHeavyNicknames in order to create an IsotopeAbundances object that can handle "D" and "T".


brendanx67:
Would be nice if it passed the tests.

PR #750, opened 2019-11-30
Skyline/bugfix/20191130 open from zip

brendanx67:


brendanx67:
This comes up in workshops sometimes. Someone drills into a .zip file (or .sky.zip) and just double-clicks on the .sky file inside it and then can't understand why they are having problems. This fix should make it so that they get an error message rather than an open Skyline document that lives in a temporary folder disconnected from its supporting files, e.g. libraries.

PR #752, opened 2019-12-02
Fix intermittent failure in TestManageResults

nickshulman:
I believe that this change will fix the intermittent failure in TestManageResults.
If two status updates come in at the same time for the same file, then we would create two FileProgress controls. Also, we would throw a "item with the same key already added" exception, and that would get displayed as an import failure error.

(This error was made possible between the work I did to make the AllChromatogramsGraph a little faster when importing thousands of files at once: the controls get added to flowFileStatus with one "AddRange" call instead of one at a time. The fact that I moved the adding to "_fileProgressControls" made it possible to create controls for the same URI more than once, but it could only happen if we were processing two statuses for the same URI in the same batch, which I think only became possible to happen because of this 3array work)


bspratt:
Seems proper - a small enough change, certainly.

PR #755, opened 2019-12-03
Skyline/bugfix/20191202 corrected bruker ccs calc

bspratt:
This changes the Bruker CCS calculation to use mass N2 instead of mass N.

### DIFF
--- a/pwiz/data/vendor_readers/Bruker/SpectrumList_Bruker.cpp
+++ b/pwiz/data/vendor_readers/Bruker/SpectrumList_Bruker.cpp
@@ -731,8 +731,9 @@ PWIZ_API_DECL bool SpectrumList_Bruker::hasCombinedIonMobility() const
 
 // Per email thread Aug 22 2017 bpratt, mattc, Bruker's SvenB:
 // The gas is nitrogen(14.0067 AMU) and the temperature is(according to Sven) assumed to be 305K.
+// Turns out it's N2, actually, which seems obvious in retrospect (bpratt Dec 2 2019)
 static const double ccs_conversion_factor = 18509.863216340458;
-static const double MolWeightGas = 14.0067;
+static const double MolWeightGas = 28.0134; // 2* 14.0067
 static const double Temperature = 305;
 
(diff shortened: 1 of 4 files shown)


pwiz_tools/Skyline/Controls/Graphs/AllChromatogramsGraph.cs
```diff
@@ -470,14 +470,11 @@ private void AddProgressControls(MultiProgressStatus status)
                 progressControl.ShowGraph += (sender, args) => ShowGraph();
                 progressControl.ShowLog += (sender, args) => ShowLog();
                 controlsToAdd.Add(progressControl);
+                _fileProgressControls.Add(filePath.GetLocation(), progressControl);
```
brendanx67: This seems to be a repeat or removal of Nick's recent change and not related to this PR.

pwiz/data/vendor_readers/Bruker/SpectrumList_Bruker.cpp
```diff
@@ -731,8 +731,9 @@ PWIZ_API_DECL bool SpectrumList_Bruker::hasCombinedIonMobility() const
 
 // Per email thread Aug 22 2017 bpratt, mattc, Bruker's SvenB:
 // The gas is nitrogen(14.0067 AMU) and the temperature is(according to Sven) assumed to be 305K.
+// Turns out it's N2, actually, which seems obvious in retrospect (bpratt Dec 2 2019)
 static const double ccs_conversion_factor = 18509.863216340458;
-static const double MolWeightGas = 14.0067;
+static const double MolWeightGas = 28.0134; // 2* 14.0067
```
brendanx67: Does C++ not accept the equation in the comment? Normally, I would just write the equation in the code and not have a comment. I might even have a constant called something like NitrogenAtomicWeight and then have NitrogenAtomicWeight*2 for the MolWeightGas value... But, as this is likely to be replaced soon by a function from Bruker, not a big deal. More of a stylistic thing anyway.

PR #769, opened 2019-12-05
Fix BiblioSpec ProteinPilot group.xml progress indication

chambm:
- fixed progress indication when reading large ProteinPilot group.xml files
* fixed saxhandler::getCurrentByteIndex() for large file sizes
* fixed some compile warnings in BiblioSpec


pwiz_tools/BiblioSpec/src/ProteinPilotReader.h
```diff
@@ -76,6 +76,9 @@ namespace BiblioSpec {
         vector<PEAK_T> curPeaks_;
         string peaksStr_;
         size_t expectedNumPeaks_;
+        int totalSpectrumCount_;
```
kaipot: This variable doesn't look like it's being used anywhere

PR #770, opened 2019-12-05
adding a new perftest for Bruker prmPASEF data, which also covers an …

bspratt:
…issue with recovering IM data from libraries for peptides with heavy modifications.


chambm:
Hmm... how can we test prmPASEF when we don't have proper data reading for it yet (at least for "high target density" sections)?

bspratt:
It seems likely that we will have to adjust this test later, then. But it does check another problem found along the way so I'm letting it ride for now.

PR #772, opened 2019-12-06
Fix diaPASEF scanEnd

chambm:
- fixed diaPASEF's scanEnd treated as inclusive; it's actually exclusive and there's a skipped scan between diaPASEF windows

Actually didn't break Skyline tests as far as I can tell.


bspratt:
Not entirely surprising, since it's at the fringe of the frame. Did you run the perf tests? That's where it would tank, if anywhere.

chambm:
@brendanx67 Do you want this in 19.2?

brendanx67:
Let's add this to both master and 19_1. Thanks, Matt.

PR #774, opened 2019-12-07
Import Assay Library improvements

kaipot:


pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -2126,24 +2126,61 @@ public string AssayLibraryName
                 db = IrtDb.CreateIrtDb(AssayLibraryFileName);
 
                 // Try to guess iRT standards
-                var matchingStandards = IrtStandard.ALL.Where(standard => standard.IsSubset(dbIrtPeptides, null)).ToArray();
-                if (matchingStandards.Length == 2 && matchingStandards.Contains(IrtStandard.BIOGNOSYS_10) && matchingStandards.Contains(IrtStandard.BIOGNOSYS_11))
-                    matchingStandards = new[] {IrtStandard.BIOGNOSYS_11};
-                if (matchingStandards.Length == 1)
+                var matchingStandards = IrtStandard.ALL.Where(standard => standard.IsSubset(dbIrtPeptides, null)).ToList();
```
brendanx67: There is also assay library import code in CommandLine. I don't see it changed here, but I would expect it to need changes similar to SkylineFiles. Maybe we should add comments in both places explaining that they need to be kept in synch.

PR #780, opened 2019-12-10
Skyp: add Panorama server and open

vagisha:
If the skyline document that a skyp file points to requires login on the Panorama server, and the server is not configured in Skyline, let the user add the Panorama server and try to download and open the file again.

### DIFF
--- a/pwiz_tools/Skyline/Model/SkypFile.cs
+++ b/pwiz_tools/Skyline/Model/SkypFile.cs
@@ -125,14 +125,16 @@ namespace pwiz.Skyline.Model
     {
         private SkylineWindow _skyline;
 
-        public IDownloadClient DownloadClient { get; set; }
+        public DownloadClientCreator DownloadClientCreator { private get; set; }
 
         public const string ERROR401 = "(401) Unauthorized";
         public const string ERROR403 = "(403) Forbidden";
+        private int _retryRemaining = 3;
 
         public SkypSupport(SkylineWindow skyline)
         {
             _skyline = skyline;
+            DownloadClientCreator = new DownloadClientCreator();
         }
 
         public bool Open(string skypPath, IEnumerable<Server> servers, FormEx parentWindow = null)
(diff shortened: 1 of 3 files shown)


pwiz_tools/Skyline/Model/SkypFile.cs
```diff
@@ -152,20 +155,50 @@ public bool Open(string skypPath, IEnumerable<Server> servers, FormEx parentWind
...
+        }
 
         private void Download(SkypFile skyp, IProgressMonitor progressMonitor, FormEx parentWindow = null)
         {
             var progressStatus =
                 new ProgressStatus(string.Format(Resources.SkypSupport_Download_Downloading__0_, skyp.SkylineDocUri));
             progressMonitor.UpdateProgress(progressStatus);
 
-            if (DownloadClient == null)
+            if (DownloadClient == null || DownloadClient is WebDownloadClient)
```
brendanx67: I don't follow this. Why do you create a new DownloadClient when you have one that is a WebDownloadClient. Isn't that the same as always creating one? Seems like a comment here explaining would be helpful.

pwiz_tools/Skyline/Model/SkypFile.cs
```diff
@@ -152,20 +155,50 @@ public bool Open(string skypPath, IEnumerable<Server> servers, FormEx parentWind
...
+                    {
+                        // Let the user add a Panorama server.
+                        dlg.NavigateToTab(ToolOptionsUI.TABS.Panorama);
+                        dlg.ShowDialog(alertDlg);
+                        if (dlg.DialogResult != DialogResult.OK)
+                        {
+                            return false;
+                        }
+                    }
+                    return Open(skypPath, Settings.Default.ServerList /*get the updated server list*/, parentWindow);
```
brendanx67: Does this recurse every time the request fails? Probably not a huge concern, but maybe impose some limit where we stop showing the ToolsOptionsUI and leave the user to actually use Tools > Options as described in the message or try again from scratch.

vagisha: Thanks.  I added a commit limiting retries to 3.

vagisha: I committed a change limiting retries to 3

pwiz_tools/Skyline/Model/SkypFile.cs
vagisha: This was for SkypTest to be able to use its custom implementation of IDownloadClient.  I replaced this with a DownloadClientCreator to avoid confusion in this method.

PR #781, opened 2019-12-10
Fix building a prosit spectral library when the precursors in the doc…

nickshulman:
…ument are heavy.

Prosit always requests the light spectra from the server. PrositLibraryBuilder was incorrectly writing to the database that the modified sequence was heavy.


nickshulman:
This pull request is going to need a lot more work. For IRT peptides, we really do want to write the heavy version of the spectrum to the .blib file.

PR #782, opened 2019-12-10
- fixed score type for OpenSWATH assay input to BiblioSpec

chambm:


kaipot:
In the file I'm testing with, some lines have the same precursor m/z as the previous line but different sequence. They were getting grouped and only the 1st sequence was appearing in the library.

chambm:
Same precursor m/z but different sequence? Isobaric residues?

kaipot:
I think it's just a coincidence, here is an example of two consecutive lines with the same precursor m/z:
```
628.845884379887  399.187425536375  2 1 32.39792  47.6336 ETVSILLNPDR ETVSILLNPDR ETVSILLNPDR2  Q8TEL6;Q8TEL6-3|TP4AP_HUMAN Q8TEL6;Q8TEL6-3|TP4AP_HUMAN b 4 -1  -1  ETVSILLNPDR2  0
628.845884379887  381.249631990492  2 1 9.18923 21.92965  VPAINVNDSVTK  VPAINVNDSVTK  VPAINVNDSVTK2 P23526|SAHH_HUMAN P23526|SAHH_HUMAN b 4 -1  -1  VPAINVNDSVTK2 0
```

chambm:
That is pretty wild. I'd expect that if they were sorting on precursor m/z but they are clearly not doing that. Maybe they were sorted at one point in the past and some other sorting step shuffled them in blocks.

Easy to fix, anyway. Glad you caught it!

brendanx67:
As a test case, we are building the library with both BlibBuild and File > Import > Assay Library and fixing the issues which cause the two to come up different. Doing this it was pretty easy to see that the number of matched peptides differed between the two. On the Skyline side, the modifications use single-decimal precision and fail to set the RT value. Kaipo is working on the latter, having solved the former.

PR #783, opened 2019-12-10
Remaining fixes in the AbsoluteQuant tutorial

rita-gwen:


rita-gwen:
Since there are some changes in the framework code (AbstractFunctionalTest) I would like a brief code review. Thanks.

PR #789, opened 2019-12-11
MSn format nativeID and scan number fix

chambm:
- added nativeID field (as an 'I' entry) for ms1 and ms2 formats
- fixed ms1/ms2 format scan numbers defaulting to 0 instead of index for nativeID formats that can't translate to scan numbers


chambm:
No, only to converters. It could conceivably break other software's MSn format readers if they refuse to process an 'I' line with an unknown field, but nothing to do with Skyline.

PR #790, opened 2019-12-11
Skyline/SkylineNightly - added retry logic to re-download data sets o…

bspratt:
…n test failure, and retry the test if the file is shown to be different from what was already on disk. This means that the nightly tests no longer have to delete+download existing tutorial data sets, which should improve throughput as well as making us more tolerant of transitory test data server failures.


bspratt:
I tested this manually by:
deleting the local zip file for MethodEditTutorial
damaging the local zip file for MethodEditTutorial
dmaging the local zip file for MethodEditTutorial and simulating a download failure for its replacement

bspratt:
OK, download retry behavior now happens only when requested by commandline
arg to TestRunner. SkylineNightly uses this new commandline argument, which
is off by default.

PR #791, opened 2019-12-11
ProcessRetentionTimes bug fix

kaipot:


pwiz_tools/Skyline/Model/Irt/RCalcIrt.cs
```diff
@@ -245,7 +245,7 @@ public IEnumerable<DbIrtPeptide> GetDbIrtPeptides()
             var matchedStandard = IrtStandard.WhichStandard(standardPeptideList.Select(pep => pep.ModifiedTarget));
             if (matchedStandard != null && matchedStandard.HasDocument)
             {
-                var standardDoc = matchedStandard.GetDocument();
+                var standardDoc = matchedStandard.ImportTo(new SrmDocument(SrmSettingsList.GetDefault()), null, out _);
```
nickshulman: Should probably have a comment here saying why this call is being made.
Someone might be confused as to why we are importing into a document that we immediately discard.

PR #793, opened 2019-12-12
Transfer spectra in bulk when not in best score mode

kaipot:


chambm:
Is this intended to fix BlibFilter's speed with the unknown score type with the OSW assay?

kaipot:
Yes though it looks like the change broke something, I'll need to fix that first.

kaipot:
The output looks close but I think the test is failing because the spectra are being inserted in a different order. If the changes look correct to you, I'll just update the test to reflect the new output.

chambm:
I haven't built it myself yet. I just added .blib.observed artifacts to the TC artifacts list so failures like this should be easier to check from TC builds. (rebuild is queued on TC)

chambm:
I'm going to try to get the bulk transfer to sort on insert so it's easier to compare. If it turns out to be tricky then we'll just go with this.

chambm:
Yes that's more trouble than it's worth. I thought I could use auto-increment and `order by` to create ids the same way as before, but syncing those up with the ids used by other tables would take a merge table with both the old and new id.

kaipot:
Okay, I don't think the order in the database ever matters anyway. I'll merge once the tests have completed, thanks for updating the test.

PR #794, opened 2019-12-12
Skyline: Make Skyline generated .blib files use the same number of di…

brendanx67:
…gits (5) in modified sequences as BlibBuild


bspratt:
Seems pretty straightforward. Might want to drop a comment into BlibBuild indicating that there's an external expectation of the rounding value which should be tracked.

brendanx67:
I am fine with that if you want to add it to this PR.

bspratt:
OK, I added that comment in BlibBuild.

PR #798, opened 2019-12-13
Add Reader::Config::globalChromatogramsAreMs1Only

chambm:
* added Reader::Config::globalChromatogramsAreMs1Only to satisfy Skyline's requirement of getting TIC chromatograms from only MS1 spectra
* added TestFilesDir.GetSkylineDir() and TestFilesDir.GetVendorTestData() (any objection to these @nickshulman  @brendanx67 ?)
* changed TestTicChromatogram() to use vendor reader test data (too many of the existing test files had no MS1s, e.g. MRM data)
* tweaked VendorReaderTestHarness for some more convenient functionality


chambm:
Thanks for checking the TIC values. I wasn't sure how to do that.

To avoid potential problems with non-typical layouts, shall I - for now - just make the test a no-op if the data isn't found? Right now the test is basically just checking that Skyline can read the vendor data and that the globalChromatogramsAreMs1Only flag is working: the actual data is being checked in the pwiz tests against reference mzML files.

I don't think we should duplicate test files between pwiz and Skyline unless necessary, but to make it work with SkylineTester as a standalone zip (without the source tree), I can see about adding a vendor reader test data zip to SkylineTester on the fly. That way it won't bloat the repo with binaries more than necessary and it won't need to be synchronized when the pwiz test data gets updated.

nickshulman:
I think you might as well leave the tests the way that they are.
If it turns out that they fail in a scenario that we care about then we can add code to conditionally disable them later. (I believe they will pass in Nightly).

(It's too bad we check in our test files as .zip's. If we checked them in as regular files, then I would not see it as wasteful to have identical files checked in multiple times in different spots in the tree since it won't affect the size of the repository).

nickshulman:
BTW, the way to see the total ion current area in Skyline is to bring up the Document Grid.
None of the built-in reports show the Total Ion Current Area column by default, but you can customize (Reports dropdown -> "Customize Report") the Replicates report to show that column. The column is located in the View Editor under:
Replicates > Files > Total Ion Current Area

brendanx67:
Let's put the files Skyline needs in the Skyline project. Otherwise, the SkylineTester artifact will be incomplete.

chambm:
OK, done. The pwiz vendor reader test data is now added in `SkylineTester Files` while generating SkylineTester.zip and I tested that TestTicChromatogram() will work from both locations.

PR #801, opened 2019-12-13
Fix support for MSFragger via PeptideProphet

chambm:
- fixed support for MSFragger _[un]calibrated.mgf files when pepXML is run through PeptideProphet


chambm:
@brendanx67 Do you want this in 19.2?

brendanx67:
Yes. I want this in 19.2. Thanks.

brendanx67:
I also added the necessary support on the Skyline side, which should follow this to 19.2. Probably best to squash and merge. Did my own manual testing and both now work as expected. Though, having an mzML file around with the right basename was preferred to the MGF file. But, when I switched to a .d file with analysis.tdf in it, BlibBuild did not recognize that as a possible spectrum source.

PR #817, opened 2019-12-17
CustomReports tutorial test and screenshot generation.

rita-gwen:
A would like to have a code review since this change involves some framework classes.


pwiz_tools/Skyline/TestUtil/ScreenshotManager.cs
```diff
@@ -0,0 +1,309 @@
...
+                // ReSharper enable PossibleNullReferenceException
+            }
+
+            public SkylineScreenshot(ShotType pShotType, SkylineWindow pSkylineWindow)
+            {
+                _type = pShotType;
+                _skylineWindow = pSkylineWindow;
+            }
+
+            protected float getScalingFactor()
```
nickshulman: This function name should be capitalized.
Also, it seems like maybe this method should return a "PointFactor" instead of a float

pwiz_tools/Skyline/TestUtil/ScreenshotManager.cs
```diff
@@ -0,0 +1,309 @@
...
+            {
+                //Have to do it this way because of the limitation on OLE access from background threads.
+                Thread clipThread = new Thread(() => Clipboard.SetImage(shotPic));
+                clipThread.SetApartmentState(ApartmentState.STA);
+                clipThread.Start();
+                clipThread.Join();
+            }
+        }
+
+        public void SaveToFile()
```
nickshulman: I think this method should take the file path as a parameter, and the caller should be the one deciding where the screenshot should be saved (and it would have something to do with TestContext.TestDir).

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -456,6 +470,18 @@ private static IEnumerable<Form> OpenForms
             return null;
         }
 
+        public static Form FindOpenForm(Type formType) 
+        {
+            foreach (var form in OpenForms)
+            {
+                if (((formType.IsInstanceOfType(form) || formType.DeclaringType != null && formType.DeclaringType.IsInstanceOfType(form))) && form.Created)
```
nickshulman: Why do we need the "formType.DeclaringType" here?
Which form class are we looking for?

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -972,17 +1049,44 @@ private void PauseForScreenShot(string description, int? pageNum, Type formType)
...
             }
             else
             {
                 PauseForForm(formType);
             }
         }
 
+        public Form FindFloatingWindow<T>() where T:FloatingWindow
+        {
+            foreach (var panel in SkylineWindow.DockPanel.FloatingWindows)
```
nickshulman: Usually you would write something like this in one line:
return SkylineWindow.DockPanel.FloatingWindows.OfType<T>().FirstOrDefault();

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -1074,6 +1202,15 @@ protected void RunFunctionalTest(string defaultUiMode = UiModes.PROTEOMIC)
                 }
             }
 
+            try
+            {
+                _shotManager.SaveToFile();
+            }
+            catch (Exception e)     //TODO: it probably shouldn't interfere with the test like this, but this exception should be logged somehow.
```
nickshulman: It might be better to not catch this exception, and let it get handled by the test framework as a failure.
The only reason to catch it would be if you were afraid that you might be hiding some exception that happened earlier in the test.

pwiz_tools/Skyline/TestUtil/ScreenshotManager.cs
```diff
@@ -0,0 +1,309 @@
...
+        private string filePath
+        {
+            get
+            {
+                var exeDir = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+                return Path.Combine(exeDir ?? "", _tutorialName + "_shots");
+            }
+        }
+
+        public ScreenshotManager([NotNull] string pTutorialName, [NotNull] SkylineWindow pSkylineWindow)
```
nickshulman: We never put "p" at the front of variable names.

pwiz_tools/Skyline/TestUtil/ScreenshotManager.cs
rita-gwen: Agree.

pwiz_tools/Skyline/TestUtil/ScreenshotManager.cs
rita-gwen: Well, TestContext.TestDir points to the "SkylineTester Results" folder that is volatile and dropped/recreated every time the test is run. 
The idea here is for the user to create a customized set of screenshot configurations that can be re-used on subsequent runs, so the location should be persistent. Also, since the file is saved after each change there is no need to call it externally. I actually removed the call to SaveToFile() from the AbstractFunctionalTest as redundant and made the method private.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
rita-gwen: Because tests often use a nested class when calling PauseForScreenShot like here:
`PauseForScreenShot<ViewEditor.ChooseColumnsView>("Edit Report form", 4);`
The nested class is usually one of the tabs in the parent form, so when we look for an open window we need the declaring class which is an actual form containing this tab.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
rita-gwen: This is nice, thanks. I didn't know about OfType<>() method.

pwiz_tools/Skyline/TestUtil/ScreenshotManager.cs
rita-gwen: Replaced this parameter with TestContext anyways.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
rita-gwen: This code is removed, see above.

rita-gwen:
Nat, are you done with the tutorial doc? Should I merge this?

PR #827, opened 2019-12-21
Skyline: adding a "Product Neutral Loss" column to small molecule tra…

bspratt:
…nsition list reader, which makes it convenient to describe fragments in terms of a chemical formula to be subtracted from the precursor's chemical formula.


pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
```diff
@@ -1452,6 +1398,12 @@ public static string GetAcceptedIonMobilityUnitsString()
                     errMessage =
                         string.Format(Resources.SmallMoleculeTransitionListReader_ReadPrecursorOrProductColumns_Cannot_derive_charge_from_adduct_description___0____Use_the_corresponding_Charge_column_to_set_this_explicitly__or_change_the_adduct_description_as_needed_, adduct.AdductFormula);
                 }
+                else if (countValues < 2)
+                {
+                    errMessage = string.Format(getPrecursorColumns
+                            ? Resources.PasteDlg_ValidateEntry_Error_on_line__0___Precursor_needs_values_for_any_two_of__Formula__m_z_or_Charge_
+                            : Resources.PasteDlg_ValidateEntry_Error_on_line__0___Product_needs_values_for_any_two_of__Formula__m_z_or_Charge_, row.Index + 1);
```
nickshulman: You might want to show a different error message here if the "Product Neutral Loss" column is displayed: in that case, they do not need "Formula".

bspratt: I think you can make the case the product neutral loss is a variant of formula. The message doesn't use any formal column names, so the ambiguity works in our favor.

pwiz_tools/Skyline/Util/Adduct.cs
```diff
@@ -702,6 +702,38 @@ public static Adduct ProtonatedFromFormulaDiff(string left, string right, int ch
             return new Adduct(string.Format(@"[M{0}{1}{2}H]", sign, adductFormula, charge>0?@"+":@"-"), ADDUCT_TYPE.non_proteomic) { AdductCharge = charge };
         }
 
+        /// <summary>
+        /// Changes text like "[M+H]" to just "M+H"
+        /// </summary>
+        public static string StripBrackets(string adductText)
```
brendanx67: These functions just still look very loose to me, which is why it is good to pull them out like this. They don't look like they belong in an Adduct class, because all they do is tell if a formula contains brackets and strip all brackets out of any kind of text. I would expect them to expect more out of well-formed adduct text. Why not HasAdduct(formula) instead, and actually look for some kind of proper placement? Couldn't you just make everything based on SplitFormulaAndTrailingAdduct? Do that up front and then test string.IsNullOrEmpty(formulaAdduct), and then if you still feel you need to strip brackets to make an equals comparison... Write an Adduct.AreEqual() function instead?

pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
```diff
@@ -1467,6 +1419,156 @@ public static string GetAcceptedIonMobilityUnitsString()
...
+                        Line = row.Index,
+                        Message = Resources.SmallMoleculeTransitionListReader_ProcessNeutralLoss_Cannot_use_product_neutral_loss_chemical_formula_without_a_precursor_chemical_formula
+                    });
+                    return false;
+                }
+
+                // Get neutral loss formula
+                try
+                {
+                    BioMassCalc.MONOISOTOPIC.CalculateMassFromFormula(neutralLoss); // For syntax check - throws ArgumentException if there's a problem
```
brendanx67: I still don't get what is the extra checking beyond Molecule.Parse() below. It still feels like this should not be necessary and the fact that you need the comment indicates the use and result are unexpected.

bspratt:
I suppose the confusion is from "MONOISOTOPIC". As the comment indicates,
I'm not interested in the mass at this point, just the ArgumentException in
case of parsing trouble. Molecule.Parse doesn't offer the same exception
behavior, it assumes that the input has already been checked.

I suppose the right thing to do is try to split out the syntax check in
CalculateMassFromFormula as a static method of BioMassCalc, so there's none
of the MONO vs AVG weirdness.

PR #828, opened 2019-12-23
Skyline: Move combined mode IMS spectra attribute to ChromCachedFile …

brendanx67:
…and ChromFileInfo from MsDataFileUri
- Passing existing tests but need a new backwards compatibility test


brendanx67:
I have added a new backward compatibility test, and this now passes all tests that I have tried so far, including all of the IMS perf tests and AgilentSpectrumMillRampedIMSImportTest and WatersLockmassPerfTest which failed last night on TeamCity.

I would like to complete this transition back to where we do not store any attributes on the URI but use it only for the location of the data. Putting attributes on the URI has mostly been problematic and was never complete, except from the perspective of how MsDataFile is implemented and what is necessary to construct one with an index capable of retrieving the IDs returned. I found it a little surprising that preferOnlyMsLevel was not included in this, since it seemed equally plausible that this attribute would impact spectrum IDs in a way similar to combined ion mobility.

I left Waters lock-mass attributes on the URI, because it seemed more involved to implement, and should impact a relatively small number of documents created with Skyline, but I would like to remove it also in the next release cycle.

I can't really think of any issue that this introduces, since we had already had to implement special-case code to change URI parameters to match the current document during Edit > Manage Results - Re-import. I did switch to using attributes taken directly from the MsDataFileImpl instead of asking the ChromatogramProvider, which could, in theory, be less flexible in the future, but we never really needed these attributes on the URI when interacting with Chorus. We sent Chorus a full XML document including the complete Full-Scan extraction parameters and all of the targets, and they sent back numeric IDs that allowed us to query the spectra they had used without needing us to specify any extra parameters.

brendanx67:
This PR also now contains fixes for issues that could happen on our existing branches regardless of how these attributes are stored.
- Not dealing correctly with a document ID change in MemoryDocumentContainer
- Not handling correctly reimport of a single replicate in a single-replicate document

bspratt:
I've gone through the changes. There's generally less code than before which is always a good sign. 

As discussed, in moving load parameters to the exclusive province of the chromatogram cache we strengthen the assumption that the .skyd file is present, and weaken the ability to re-create the .skyd file on demand - though as also discussed our ability to do that was always sketchy at best since we were very incomplete in preserving every full-scan parameter which might affect the extraction process. So this simplifying assumption is also a disambiguation, which is excellent.

I'm eager to see how this fares in Nightly - all tests are passing on my local machine.

brendanx67:
Lots more testing performed manually: 

- trimmed down a DIA document and imported first 1 file in profile mode (ms2 only),
- then 1 in centroid mode (ms2 only), then added precursors and imported 1 file in profile mode again (both), 
- 1 file in centroid mode (both),
- 1 file mixed with ms1 centroiding and ms2 profile, and 
- 1 file mixed with m1 profile and ms2 centroiding. 
- Tested that all chromatograms showed the right type of spectra. 
- Tested that attempting to open this file with 19.1 gave the expected error. 
- Used File > Share to 19.1 format and tested that 19.1 could open the file and
- Full-Scan view showed the right spectra
- Did the same with the current build and saved
- Reviewed all .sky files to confirm the expected presence or lack of URI parameters
- Changed Full-Scan settings to centroided for both and used Edit > Manage Results to reimport all files: confirmed that all chromatograms were populated and when clicked on showed centroided spectra for both ms1 and ms2
- File > Share to 19.1 and check compatibility with installed Skyline 19.1
- Changed Full-Scan settings to TOF for both and used Edit > Manage Results to reimport all files: confirmed that all chromatograms were populated and when clicked on showed profile mode spectra for both ms1 and ms2
- File > Share to 19.1 and check compatibility with installed Skyline 19.1

brendanx67:
Files were big and being SCIEX WIFF files made them very slow to centroid. Not easily integrated into our automated tests. It did uncover a typo where ms2 was used instead of ms1.

PR #829, opened 2019-12-27
Skyline: fix for Issue 689 "File > Export > Spectral Library losing i…

bspratt:
…nformation from small molecule documents" - Ion annotations do not preserve the ion names from the document but instead always use "ion[mass]" for the annotations when the document gave them names


pwiz_tools/Skyline/Model/RefinementSettings.cs
```diff
@@ -889,12 +889,12 @@ public enum ConvertToSmallMoleculesChargesMode
             string pathForLibraryFiles, // In case we translate libraries etc
```
brendanx67: Is there a benefit to keeping this in RefinementSettings? Or could we move it into its own class in TestUtil?

pwiz_tools/Skyline/Model/Lib/BlibData/DbRefSpectraPeakAnnotations.cs
```diff
@@ -78,6 +78,10 @@ public static ICollection<DbRefSpectraPeakAnnotations> Create(DbRefSpectra refSp
             return resultList;
         }
 
+        public override string ToString()
+        {
+            return (Name??string.Empty)+@" i"+PeakIndex; // For debugging convenience
```
brendanx67: string.Empty seems like a strange choice here. Is that really what will be shown for this annotation? I would think something like "<unknown>" or "<unnamed>". The spacing is also nonstandard here. It is one thing to bang this in for your own debugging and another to commit it to the project for the benefit of others.

brendanx67:
Just noticed this is a PR to master. This seems minor enough that it could go into skyline_19_2.

pwiz_tools/Skyline/Model/Lib/BlibData/DbRefSpectraPeakAnnotations.cs
bspratt: Fair enough, but what *is* our standard for this kind of "user never sees it, but just showing a class name in the debugger isn't at all helpful to developers" formatting? If there's a right way to do it I'm happy to go that route. I do think it's worth leaving in. 

But I can certainly leave it out if there's a philosophical objection to it, particularly if I'm remaking this as a commit to 19.2 with cherry-pick back to master.

pwiz_tools/Skyline/Model/RefinementSettings.cs
bspratt: It's test only code, for sure. We put it where it is on purpose, back in the day, but the landscape has grown more complicated. I'll see about moving it, as long as it already has a change.

bspratt: Oops, it's not purely test code - we use portions of it, for example, to display the molecule qualities of a peptide (e.g. chemical formula) in the document grid. So it's not a cheap refactor after all.

PR #830, opened 2019-12-27
Changes to ignore case when looking for library files based on .ext

bspratt:
Now uses centralized code for library filename extension check - potentially unwanted side effect is that Midas is now recognized as a spectral library by the commandline processor. If that's not desirable we can easily add an exclusion list to the call.


bspratt:
Thanks - I should have looked at PathEx. 

As discussed I think I'll leave PathEx.HasExtension() implementation unchanged as it's not causing any problems.

I'll apply to these changes to 19_2 first then cherry pick back to master.

PR #832, opened 2019-12-31
Skyline: fix for Issue 708 "Renaming a molecule in Targets tree cause…

bspratt:
…s Skyline to miscalculate mass of transition with heavy label adduct". In effect it was adding the mass of the heavy isotopes twice due to needlessly complex code to derive mass from mz (modeled on the peptide code, where it's perfectly reasonable). But we can just get the mass of the molecule directly, no need for conversion from mz.


bspratt:
Note this is a PR against 19.2, I'll cherry-pick to master when approved.

PR #838, opened 2020-01-04
Adds support for Golm Metabolome Database GCMS spectral library in .MSP format

bspratt:
Currently treats this as just another NIST file, but as Golm Metabolome Database (GMD) entries lack any correctness value like NIST's TF_ratio, perhaps there should be a stronger distinction? Also note that it currently stashes the Retention Index value in Skyline's Retention Time slot.


pwiz_tools/Skyline/Model/Lib/NistLibSpec.cs
```diff
@@ -1199,11 +1256,28 @@ private static int GetMod(string mod, out string massDiff)
...
+            var valString = rtString.Split(MINOR_SEP).First();
+            if (!TryParseDoubleUncertainCulture(valString, out rt))
                 return null;
             return isMinutes ? rt : rt / 60;
         }
 
+        private static bool TryParseDoubleUncertainCulture(string valString, out double dval)
+        {
+            if (!double.TryParse(valString, NumberStyles.Float, CultureInfo.InvariantCulture, out dval))
+                if (!double.TryParse(valString.Replace(',', '.'), NumberStyles.Float, CultureInfo.InvariantCulture, out dval)) // .MSP from Golm GMD may have European decimals
```
brendanx67: I prefer && here.

pwiz_tools/Skyline/Model/Lib/NistLibSpec.cs
```diff
@@ -1199,11 +1256,28 @@ private static int GetMod(string mod, out string massDiff)
...
+                    return false;
+            return true;
+        }
+
+        private static bool TryParseFloatUncertainCulture(string valString, out float fval)
+        {
+            if (!float.TryParse(valString, NumberStyles.Float, CultureInfo.InvariantCulture, out fval))
```
brendanx67: I prefer && here.

pwiz_tools/Skyline/Model/Lib/NistLibSpec.cs
```diff
@@ -871,6 +874,14 @@ private bool CreateCache(ILoadMonitor loader, IProgressStatus status, int percen
                             continue;
                         }
 
+                        match = REGEX_RI_LINE.Match(line);
+                        if (match.Success)
+                        {
+                            rt = GetRetentionTime(match.Groups[1].Value, true); // Note using RT as stand-in for RI (retention Time vs retention Index, LC vs GC)
```
brendanx67: There is precedent for this. We store iRT values as RT values with Prosit and Assay Libraries. This only gets a bit weird with the Show IDs feature "From Other Runs" which plots the RT values from the library regardless of run matching. Then we end up plotting iRTs on an RT axis. I suppose we could consider some kind of flag in the library indicating it is an iRT and never to plot it on a Measured RT axis. But, that can be "FUTURE" work. This is not the first case.

brendanx67:
And the code inspection warnings need to be fixed.

pwiz_tools/Skyline/Model/Lib/NistLibSpec.cs
bspratt: OK, I've added a CONSIDER comment that suggests we might someday want to track RI and RT simultaneously.

PR #842, opened 2020-01-08
Fix tutorial tests including TestExistingExperimentsTutorial.

nickshulman:
It looks like some of the tutorial tests can't handle "ForceMzml".


rita-gwen:
Am I right to assume that tutorial tests are part of the nightly build, but are not in the on-demand pull request build? Because otherwise it's not clear how it passed the checks.

chambm:
Right, tutorial and perf tests run overnight (and even then not all the perftests: the ones with 20GB+ zip files are not run). I have not yet made that test a required check, which is why it passes checks until the nightly test. If the nightly test fails and you still haven't merged, then you'll see that it doesn't pass all checks. @brendanx67 do you want me to make the nightly test required? (it won't change the ability to override the required checks as administrator)

PR #849, opened 2020-01-11
Increase the threshold scan range at which we consider an MS1 scan to…

nickshulman:
… be a "SIM Scan".

The cutoff used to be 200 and now it has been raised to 500.
(requested by Alex)

### DIFF
--- a/pwiz_tools/Skyline/Model/Results/SpectrumFilter.cs
+++ b/pwiz_tools/Skyline/Model/Results/SpectrumFilter.cs
@@ -677,9 +677,10 @@ namespace pwiz.Skyline.Model.Results
 
         private static bool IsSimIsolation(IsolationWindowFilter isoWin)
         {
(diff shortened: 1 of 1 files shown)


chambm:
@nickshulman I asked on the call yesterday but didn't hear the answer due to background noise. Why doesn't Skyline always use scan range, when available (which is it _almost_ always), to avoid adding an intensity sample of 0 when the targeted m/z is outside the range?

brendanx67:
Probably we should. We still need a range at which we consider something SIM, because we want to be able to distinguish between SIM and survey MS1 scans to allow those to have different chromatograms. i.e. an MS1 scan from 50 to 2000 and one spanning 500 to 504 should not contribute points to the same chromatogram. But, I think it makes sense to limit even the chromatograms we consider "survey" to only the range they actually cover.

PR #852, opened 2020-01-13
Fix checking of decoys against targets

kaipot:


brendanx67:
Please run all 4 DiaSwath tutorial tests and get them passing. The two shorter ones that TeamCity runs failed last night. An explanation of why they changed would also be helpful.

PR #864, opened 2020-01-16
Skyline: Set PeptideLibraries.HasDocument library to false if the document library is removed

vagisha:
If multiple libraries are selected for a document, un-checking or removing the document library does not set PeptideLibraries.HasDocumentLibrary to false. If the document is then saved and opened again Skyline looks for the document library and adds it back.  If the document is saved as a shared .zip file opening the .zip results in an error about the missing document library and chromatograms are not loaded.


vagisha:
Closing this PR.  Created a PR against the skyline_20_1 branch: https://github.com/ProteoWizard/pwiz/pull/868

PR #866, opened 2020-01-18
Allow iRT peptides to match heavy or light

kaipot:


brendanx67:
We need this in the 20.1 branch

PR #870, opened 2020-01-21
Match heavy or light iRT standards

kaipot:


pwiz_tools/Skyline/Model/Irt/RCalcIrt.cs
```diff
@@ -696,12 +698,23 @@ public void AddIrt(double irt)
 
     public sealed class RetentionTimeProviderData
     {
-        public RetentionTimeProviderData(IRetentionTimeProvider retentionTimes, IEnumerable<DbIrtPeptide> standardPeptides)
+        public RetentionTimeProviderData(IRetentionTimeProvider retentionTimes, DbIrtPeptide[] standardPeptides, DbIrtPeptide[] heavyStandardPeptides)
```
nickshulman: Personally, I would make it so that "heavyStandards" was an array of "Target" instead of array of "DbIrtPeptide".
That way, it would be clear that the only difference between standardPeptides and heavyPeptides is the ModifiedTarget, and the iRT values etc. are the same.

(Your way of doing this is fine too, it's just that the code later on in this function gets a little confusing where you first assign the variable "standard" to "heavyPeptides[i]" and then "standardPeptides[i]" and I thought maybe it would be less confusing if they were completely different datatypes).

PR #871, opened 2020-01-21
* added CCS <-> 1/k0 conversion functions to CompassData interface an…

chambm:
…d implemented them in TimsData; SpectrumList_Bruker::ionMobilityToCCS/ccsToIonMobility return the value from the vendor API but throws an exception if the value doesn't match the value from the builtin calculation

@bspratt DLLs not yet included (they're prerelease AFAIK), so I'll need to send updated 7z file separately.


brendanx67:
Can we revive this PR and get it to the point that it can be merged?

chambm:
@bspratt You think we should strip out the built-in calculation and validation logic?

bspratt:
Maybe retain for test purposes to detect any shifts in the DLL behavior,
but otherwise yeah, let's be clear on where that calculation is formally
implemented.

PR #875, opened 2020-01-22
Skyline: fix a problem reading Agilent GCMS SIM SIC chromatograms

bspratt:


pwiz_tools/Skyline/Model/Results/ChromHeaderInfo.cs
```diff
@@ -1920,7 +1920,18 @@ public static ChromKey FromId(string idIn, bool parseCE)
...
                 else if (id.StartsWith(MsDataFileImpl.PREFIX_PRECURSOR))
                 {
-                    precursor = double.Parse(id.Substring(MsDataFileImpl.PREFIX_TOTAL.Length), CultureInfo.InvariantCulture);
+                    var str = id.Substring(MsDataFileImpl.PREFIX_PRECURSOR.Length);
+                    if (!double.TryParse(str, NumberStyles.Float, CultureInfo.InvariantCulture, out precursor))
+                    {
+                        if (str.StartsWith(@"Q1="))
+                        {
+                            str = str.Substring(3);
+                            var tail = str.IndexOf(' ');
```
brendanx67: str.Substring(3).Split(' ')[0] would also work and is more concise.

pwiz_tools/Skyline/Model/Results/ChromHeaderInfo.cs
```diff
@@ -1920,7 +1920,18 @@ public static ChromKey FromId(string idIn, bool parseCE)
                 }
                 else if (id.StartsWith(MsDataFileImpl.PREFIX_PRECURSOR))
                 {
-                    precursor = double.Parse(id.Substring(MsDataFileImpl.PREFIX_TOTAL.Length), CultureInfo.InvariantCulture);
+                    var str = id.Substring(MsDataFileImpl.PREFIX_PRECURSOR.Length);
+                    if (!double.TryParse(str, NumberStyles.Float, CultureInfo.InvariantCulture, out precursor))
+                    {
+                        if (str.StartsWith(@"Q1="))
```
brendanx67: Maybe just move this to immediately after str is initialized and drop the TryParse(). Seems strange to TryParse and then double.Parse() again regardless of whether the string starts with "Q1=".

pwiz_tools/Skyline/Model/Results/ChromHeaderInfo.cs
bspratt: Sounds good.

pwiz_tools/Skyline/Model/Results/ChromHeaderInfo.cs
bspratt: The intent is to retain the original behavior, which is to throw when it can't parse out a double. I'll add a comment to that effect.

PR #879, opened 2020-01-23
* added Thermo scan description as a userParam when present

chambm:
* added MsDataFileImpl.GetScanDescription() and added optional maxDetailLevel to MsDataFileImpl.GetMetaDataValue()

I added maxDetailLevel because when the level of a metadata item is known, it makes sense to directly start at that level and not allow it to go higher if it's missing (which means it's not going to be available at any level).


chambm:
Whose idea was it to make it a warning when "The parameter sampleIndex has the same default value"? :unamused:

chambm:
@nickshulman Shall I merge this now or wait till you make use of it in Skyline?

nickshulman:
Let's hold off on merging this.
I am going to send another email to the SureQuant team.
Now that I see what the scan descriptions look like (label type, charge,
and labeled amino acid encoded in a weird way) I'm not sure how we would
want to implement this as a feature in Skyline. Are we thinking Skyline
should encode that specific logic, or are users expected to fill in an
extra attribute on the things in their Skyline document?
-- Nick

brendanx67:
We do already use modified peptide charge state (not yet label type) in the spectrum extraction code, at this point for matching extracted chromatograms to their targets in Skyline to be more specific than just precursor m/z. If we can translate the "labeled amino acid encoded in a weird way" into our normal modified peptide sequence, I could see using this to match spectra to the targets that are allowed to extract from them.

chambm:
He did say that, at least for SureQuant methods that are created from the new template, there would be a limited set of controlled possible values.

nickshulman:
These changes were committed as part of Pull Request #1040

PR #887, opened 2020-01-28
Add support for lowess and logarithmic iRT regressions

kaipot:


pwiz_tools/Shared/Common/Database/SqliteOperations.cs
```diff
@@ -35,5 +35,20 @@ public static bool TableExists(IDbConnection connection, string tableName)
...
+
+        public static bool ColumnExists(IDbConnection connection, string tableName, string columnName)
+        {
+            using (var cmd = connection.CreateCommand())
+            {
+                cmd.CommandText = @"PRAGMA table_info(" + tableName + ")";
+                using (var reader = cmd.ExecuteReader())
+                {
+                    while (reader.Read())
+                        if (columnName.Equals(reader.GetString(1)))
```
nickshulman: I think you should put some curly braces for the while statement here.

pwiz_tools/Skyline/Model/Irt/Regression.cs
```diff
@@ -0,0 +1,197 @@
...
+using System.Linq;
+using pwiz.Common.DataAnalysis;
+using pwiz.Skyline.Model.DocSettings;
+using pwiz.Skyline.Model.RetentionTimes;
+using pwiz.Skyline.Properties;
+using pwiz.Skyline.Util;
+
+namespace pwiz.Skyline.Model.Irt
+{
+    public static class IrtRegressionType
```
nickshulman: You should probably extend "LabeledValues" here so that you get a "Nick enum" where you keep track of the non-localized name and the localized label.
You definitely do not want to be writing localized regression names to the database.

pwiz_tools/Skyline/Model/Irt/IrtDb.cs
```diff
@@ -527,31 +524,63 @@ public IrtDb SetDocumentXml(SrmDocument doc, string oldXml)
...
+            {
+                if (!SqliteOperations.TableExists(session.Connection, @"DocumentXml"))
                 {
                     using (var cmd = session.Connection.CreateCommand())
                     {
-                        cmd.CommandText = !rowExists
-                            ? @"INSERT INTO DocumentXml (Xml) VALUES (?)"
-                            : @"UPDATE DocumentXml SET Xml = ?";
-                        cmd.Parameters.Add(new SQLiteParameter { Value = documentXml });
+                        cmd.CommandText = @"CREATE TABLE IF NOT EXISTS DocumentXml (Xml TEXT, RegressionType TEXT)";
```
nickshulman: Can you consolidate all of the places where we ensure the "DocumentXml" table exists into one central function?
Maybe that function would have to take parameters depending on whether we needed to also ensure the "RegressionType" column exists, and whether to create the table if it does not exist, or just return false.

PR #895, opened 2020-02-01
Small Molecule Method Development tutorial review.

rita-gwen:


pwiz_tools/Skyline/TestTutorial/SmallMoleculesMethodDevAndCEOptimizationTutorial.cs
```diff
@@ -257,23 +266,32 @@ protected override void DoTest()
...
-                PauseForScreenShot<SkylineWindow>("Skyline window showing relative peak seperation", 12);
+                PauseForScreenShot<SkylineWindow>("Skyline window showing relative peak separation", 12);
 
                 // Set time window
                 var peptideSettingsDlg = ShowDialog<PeptideSettingsUI>(SkylineWindow.ShowPeptideSettingsUI);
                 RunUI(() =>
                 {
-                    peptideSettingsDlg.SelectedTab = PeptideSettingsUI.TABS.Prediction;
+                    // ReSharper disable once RedundantCast
+                    peptideSettingsDlg.SelectedTab = (PeptideSettingsUI.TABS)0; //regular enum does not work because of the hidden tabs in the Small Molecule mode.
```
nickshulman: We should fix the implementation of PeptideSettingsUI.SelectedTab so that it works correctly in terms of returning valid values from the enum PeptideSettingsUI.TABS.

Probably, we would make it so that "SelectedTab" has a switch statement in it, and maps between TabPage objects such as "tabQuantification" and enum values TABS.Quantification

PR #896, opened 2020-02-01
Small Molecule Quantification tutorial review.

rita-gwen:


rita-gwen:
This tutorial has a few comments about manual construction. Unless we decide to spend some time on improving automated screen capture, menus, dropdowns and such will remain manual since it's impossible to freeze them on the screen.

PR #900, opened 2020-02-05
* refactored LeakExceptionTests

chambm:
* refactored LeakExceptionTests to LeakCheckIterationsOverrideByTestName dictionary so each test can have a customized expanded iteration count, and removed the IsFixedLeakIterations requirement for GetLeakCheckIterations()
* changed GetLeakMessage to always be reported at LeakCheckIterations, but leak tracking can continue after the message to see if it stabilizes


brendanx67:
Or was this entirely disabled before? If so, then we probably could just remove the other tests in there now. It may be that I was just playing around with this as a way of tightening our leak limits but eventually decided against that and just expanded the limits.

Either way, I would like the behavior to not really change for all tests except for TestGroupStudiesTutorial which is what we are trying to diagnose. I worry that the current changes may be impacting other 5 files in the list.

chambm:
Yes, it was disabled before because of how IsFixedLeakCheckIterations was handled. I changed that handling, so I went ahead and removed those tests from the list. So the overall behavior should only change for TestGroupStudiesTutorial now.

chambm:
@brendanx67 Shall I merge this now to get it into tomorrow's nightly?

PR #904, opened 2020-02-05
Fix regression graph exception

kaipot:


pwiz_tools/Skyline/Controls/Graphs/RTLinearRegressionGraphPane.cs
```diff
@@ -1396,8 +1396,11 @@ private string XAxisName
                 {
                     if (IsRunToRun)
                     {
-                        return string.Format(Resources.GraphData_CorrelationLabel_Measured_Time___0__,
-                            _document.MeasuredResults.Chromatograms[_originalIndex].Name);
+                        return _document.MeasuredResults != null && !_document.MeasuredResults.IsEmpty &&
```
nickshulman: I think you should make this an "if/else" statement instead of "?:" so that it's easier to understand.

Also, we should probably get rid of the property "MeasuredResults.IsEmpty". We never create empty MeasuredResults objects (it's always null instead).

PR #905, opened 2020-02-06
Calibrate iRT w/log regression

kaipot:


pwiz_tools/Skyline/Model/Irt/Regression.cs
```diff
@@ -132,9 +133,10 @@ public static void GetCurve(IIrtRegression regression, RetentionTimeStatistics s
                 minHydro = Math.Min(minHydro, hydroScore);
                 maxHydro = Math.Max(maxHydro, hydroScore);
             }
-            var addMin = minHydro < regression.XValues[0];
-            var addMax = maxHydro > regression.XValues.Last();
-            var points = regression.XValues.Length + (addMax ? 1 : 0) + (addMin ? 1 : 0);
+            var sortedX = regression.XValues.OrderBy(x => x).ToArray();
+            var addMin = minHydro < sortedX[0];
```
nickshulman: I think you should make "sortedX" into a List, and just add "minHydro" and "maxHydro" to the start or end of that list.
That way you don't have to be so careful about "addMin", "addMax" and "offset".

PR #919, opened 2020-02-11
SkylineNightly sends email notifications by posting to TestResults module

kaipot:


kaipot:
> 
> 
> Seems good. What kind of testing were you able to do?
> 
> The code would benefit from a short comment at the top of the file indicating the code's purpose, and general strategy. That's not to say the code is unclear, but it's nice for the casual reader to have a synopsis.

I tested that the SendEmailNotification function posts to the TestResults module and sends out an email; I didn't test the hang detection but hopefully overnight we will see if it works.

At Brendan's suggestion I wrapped the new code in a try-catch so hopefully it won't blow up the nightlies tonight.

I'll add some comments to the new file.

chambm:
I didn't get an email related to that testing as far as I can see.

FYI it would not be hard to update the smart build triggering logic so that changes to SkylineNightly (but not the rest of Skyline) trigger a special TeamCity build configuration that is allowed to interact with skyline.ms. That way devs could get immediate automated feedback that the changes are working, or at least didn't break anything horribly. :) The hard part though is actually adding tests to SkylineNightly since that would require running on demand rather than nightly. But of course it wouldn't be a full nightly run, just a short subset. So I'm not sure it's worth it.

kaipot:
> 
> 
> I didn't get an email related to that testing as far as I can see.
> 

I had only myself as recipient during testing 🙂

PR #940, opened 2020-02-20
Fix Bruker BAF returning empty spectra

chambm:
- fixed Bruker BAF returning empty spectra when combineIonMobilitySpectra is on (the default mode in Skyline and SeeMS)
* added ExtraZeroSamplesFilter::count_non_zeros(); tried using it (along with remove_zeros()) for Bruker profile spectra when config_.ignoreZeroIntensityPoints is off, but it didn't help as much as I expected due to lack of baseline removal in Bruker profile spectra
* refactored Bruker reader to use BinaryData instead of automation_vector which will avoid an extra copy of the unmanaged arrays

Sorry about this dropping off the radar @nickshulman .

@brendanx67 This certainly needs to be cherry picked over to 20.1, I'll do that as soon as tests pass.

### DIFF
--- a/pwiz/analysis/common/ExtraZeroSamplesFilter.cpp
+++ b/pwiz/analysis/common/ExtraZeroSamplesFilter.cpp
@@ -91,5 +91,41 @@ void ExtraZeroSamplesFilter::remove_zeros(const vector<double>& x, const vector<
 }
 
 
+int ExtraZeroSamplesFilter::count_non_zeros(const vector<double>& x, const vector<double>& y, bool preserveFlankingZeros)
+{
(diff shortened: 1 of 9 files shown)


chambm:
That Skyline tests didn't catch this seems to indicate there's no Bruker BAF perftest. It slipped by the pwiz vendor tests because I only test combineIonMobilitySpectra on the TDF data. :( Which I chose because the BAF-derived output file should be exactly the same with and without that option, but it would duplicate the mzML file twice in the repo. I'll ponder a way to have the pwiz vendor tests test all configurations, but some of them will redirect to reference mzML files from a different config.

PR #941, opened 2020-02-20
Fix lockmass being applied/removed on metadata-only DetailLevels

chambm:
- fixed lockmass being applied/removed when only metadata DetailLevels are requested (since it only impacts the binary data)
- fixed centroiding being applied for fast and instant metadata levels (it's still needed for accurate defaultArrayLength on full metadata)


chambm:
@bspratt I didn't test that this fixes the Skyline import time. Did you test that (for your patch variant)?

bspratt:
I did - it made it half as awful, which is to say still awful.

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
```diff
@@ -449,12 +455,11 @@ PWIZ_API_DECL pwiz::analysis::Spectrum3DPtr SpectrumList_Waters::spectrum3d(doub
 
 PWIZ_API_DECL void SpectrumList_Waters::getCombinedSpectrumData(int function, int block, BinaryData<double>& mz, BinaryData<double>& intensity, BinaryData<double>& driftTime, bool doCentroid) const
 {
-    MassLynxRawScanReader& scanReader = doCentroid ? rawdata_->CentroidRawDataFile()->Reader : rawdata_->GetCompressedDataClusterForBlock(function, block);
+    MassLynxRawScanReader& scanReader = doCentroid ? rawdata_->CentroidRawDataFile()->Reader : (MassLynxRawScanReader&) rawdata_->GetCompressedDataClusterForBlock(function, block);
```
bspratt: Just cleaning up warnings here?

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
```diff
@@ -449,12 +455,11 @@ PWIZ_API_DECL pwiz::analysis::Spectrum3DPtr SpectrumList_Waters::spectrum3d(doub
 
 PWIZ_API_DECL void SpectrumList_Waters::getCombinedSpectrumData(int function, int block, BinaryData<double>& mz, BinaryData<double>& intensity, BinaryData<double>& driftTime, bool doCentroid) const
 {
-    MassLynxRawScanReader& scanReader = doCentroid ? rawdata_->CentroidRawDataFile()->Reader : rawdata_->GetCompressedDataClusterForBlock(function, block);
+    MassLynxRawScanReader& scanReader = doCentroid ? rawdata_->CentroidRawDataFile()->Reader : (MassLynxRawScanReader&) rawdata_->GetCompressedDataClusterForBlock(function, block);
     vector<float>& imsMasses = imsMasses_;
     vector<float>& imsIntensities = imsIntensities_;
 
     int numScansInBlock = rawdata_->Info.GetDriftScanCount(function);
-    const auto& mzMobilityFilter = config_.isolationMzAndMobilityFilter;
```
bspratt: and here?

chambm:
Hmm. I did some local testing with msbenchmark and I started to suspect the reason for the long import is that Skyline is asking for metadata without lockmass refinement, and binary data with lockmass refinement. That would cause it to cycle through apply/remove/apply/remove/apply/remove and each application takes about 15 seconds on this file. (Once it's applied it has no cost to request the next spectrum with refinement).

Since there's no reason to apply OR remove lockmass if only metadata is requested, adding an if condition seemed like an easy fix. However if it still has awful import time (msbenchmark takes 22 seconds to enumerate all spectra with binary data for `190731_BSAd_dil45_MSE_W_03_10_50B_2uL.raw`) even with this fix, then I guess what I suspected above is wrong. Oh I see. Your patch would be removing lockmass if metadata only and lockmass==0. I'll test this version in Skyline myself then. Thanks.

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
chambm: Unused variable.

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
chambm: Yep.

bspratt:
I don't think so - see line 537 in MsDataFileImpl.cs

bspratt:
I also played with msconvert with and without lockmass correction, dramatic difference in processing speed. It really does seem to be something on the vendor side.

chambm:
I do not see the difference in msconvert times with and without lockmassRefiner. What command-lines were you using?

chambm:
Well it seems to be fixed. I can import `190731_BSAd_dil45_MSE_W_03_10_50B_2uL.raw` in about 60 seconds.

brendanx67:
Fixed by these changes? Or fixed by something else? If it is a simple and low risk enough fix, then we would obviously like to see it in the upcoming 20.1 patch. i.e. merged to the 20.1 branch.

chambm:
By these changes.

chambm:
I'll let perftests run overnight.

bspratt:
I just tested it here - seems to be working well. No idea why I was seeing
what I did yesterday.

This is probably worth backporting to the release branch.

PR #962, opened 2020-02-28
Fix deadlock when error is found by SpectrumList::checkNativeIdFindResult()

chambm:


chambm:
@brendanx67 Nevermind, this doesn't need cherry picking because the original bug was added after 20.1 (2/11/2020).

PR #965, opened 2020-03-02
Skyline: Make closing the AllChromatogramsGraph asynchronous…

brendanx67:
… and dependent on CreateHandle() having completed successfully

- Trying to fix the Dispose() in CreateHandle() errors happening in nightly test runs

### DIFF
--- a/pwiz_tools/Skyline/Controls/Graphs/AllChromatogramsGraph.cs
+++ b/pwiz_tools/Skyline/Controls/Graphs/AllChromatogramsGraph.cs
@@ -22,6 +22,7 @@ using System.Diagnostics;
 using System.Drawing;
 using System.Linq;
 using System.Text;
+using System.Threading;
 using System.Windows.Forms;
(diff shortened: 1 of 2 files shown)


pwiz_tools/Skyline/Controls/Graphs/AllChromatogramsGraph.cs
```diff
@@ -58,11 +60,31 @@ public partial class AllChromatogramsGraph : FormEx
...
 
+        private void Notification_HandleCreated(object sender, EventArgs e)
+        {
+            _windowCreatedEvent.Set();
+        }
+
+        public void RemoveAsync()
+        {
+            // Avoid closing the ACG during CreateHandle()
+            var closeThread = new Thread(() =>
```
nickshulman: I wonder whether you should use "ActionUtil.RunAsync" here so that it gets an exception handler (and also because it's good to centralize all of our thread creating code).

PR #973, opened 2020-03-05
Fix null exception

kaipot:


pwiz_tools/Skyline/SettingsUI/BuildLibraryNotification.cs
```diff
@@ -391,8 +391,11 @@ public void LibraryBuildCompleteCallback(LibraryManager.BuildState buildState, b
                                     {
                                         lib = NotificationContainer.LibraryManager.TryGetLibrary(buildState.LibrarySpec) ??
                                               NotificationContainer.LibraryManager.LoadLibrary(buildState.LibrarySpec, () => new DefaultFileLoadMonitor(monitor));
-                                        foreach (var stream in lib.ReadStreams)
-                                            stream.CloseStream();
+                                        if (lib != null)
+                                        {
+                                            foreach (var stream in lib.ReadStreams.Where(stream => stream != null))
```
nickshulman: The stream != null check is not necessary.

PR #978, opened 2020-03-07
Fix calculation of TIC Area 

nickshulman:
Make it so that Skyline falls back to the old way of extracting the Total Ion Current chromatogram and calculating the TIC Area if the Tic Chromatogram that is present in the data file is not usable.

It might not be usable if:
1. The chromatogram contains MS2 scans. We don't know for sure whether it contains MS2 scans, but the heuristic is that if the point count in the TIC chromatogram is more than 1/2 (number chosen because of typical MSe data) as large as the total spectrum count, we don't use it.
2. The chromatogram has zero points in it. This can happen if all of the spectra in the .raw file were marked as SIM spectra.

The logic for this is in GlobalChromatogramExtractor.IsTicChromatogramUsable

### DIFF
--- a/pwiz_tools/Skyline/Model/Results/ChromDataProvider.cs
+++ b/pwiz_tools/Skyline/Model/Results/ChromDataProvider.cs
@@ -135,7 +135,7 @@ namespace pwiz.Skyline.Model.Results
         /// </summary>
         public int IndexOffset { get; set; }
 
-        public int? TicChromatogramIndex { get; }
+        public int? TicChromatogramIndex { get; set; }
         public int? BpcChromatogramIndex { get; }
 
         public IList<int> GlobalChromatogramIndexes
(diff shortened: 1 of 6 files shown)


pwiz_tools/Skyline/Model/Results/ChromDataProvider.cs
```diff
@@ -179,6 +179,39 @@ public bool GetChromatogram(int index, out float[] times, out float[] intensitie
...
+
+            if (times.Length == 0)
+            {
+                return false;
+            }
+
+            // If the number of points in the chromatogram is more than a quarter of the total
+            // spectra, then this chromatogram probably includes MS2 scans, and should not 
+            // be used for calculating TIC Area.
+            if (times.Length >= _dataFile.SpectrumCount / 4)
```
brendanx67: I just realized that this would cause us to extract our own TIC for all MSe-type data where every other spectrum is MS1. This may need to change to > count/2.

chambm:
Due to Waters lockmass scans, which get counted as MS1s, you might want to make the divisor like 1.9 instead of 2.

PR #982, opened 2020-03-09
Nightly: Fix for DST and skip run if projected end time is after the next scheduled start time

kaipot:


pwiz_tools/Skyline/SkylineNightly/Program.cs
```diff
@@ -18,6 +18,7 @@
  */
 
 using System;
+using System.Globalization;
```
bspratt: This is causing a code inspection failure.

PR #984, opened 2020-03-12
Add MSData integer data arrays and ms level / function arrays to TIC/BP chromatograms

chambm:
And use the ms level / function arrays, when present, to filter for only MS1/function=1 points in Skyline.


chambm:
It doesn't make sense, but it's possible in the code. It's not a path used by any current client code though (msconvert would numpress the double arrays but not the integer ones).

PR #989, opened 2020-03-17
Skyline (20.1): Fixes to Prosit integration

brendanx67:
- Failure to annotate fragment ions in heavy labeled Prosit predicted spectra
- Failure to match m/z values between heavy labeled library spectra and Prosit mirror plot spectra
- Mistake in parenting ToolOptionsUI when shown as a result of CheckPrositSettings
- Fix canceling of PrositPingRequest during closing of ToolOptionsUI

### DIFF
--- a/pwiz_tools/Skyline/Model/Prosit/Models/PrositModel.cs
+++ b/pwiz_tools/Skyline/Model/Prosit/Models/PrositModel.cs
@@ -336,21 +336,23 @@ namespace pwiz.Skyline.Model.Prosit.Models
 
             public PrositRequest(PrositPredictionClient client, PrositIntensityModel intensityModel,
                 PrositRetentionTimeModel rtModel, SrmSettings settings,
-                PeptideDocNode peptide, TransitionGroupDocNode precursor, int nce, Action updateCallback)
+                PeptideDocNode peptide, TransitionGroupDocNode precursor, IsotopeLabelType labelType,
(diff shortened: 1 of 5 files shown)


pwiz_tools/Skyline/Model/Prosit/Models/PrositModel.cs
```diff
@@ -432,8 +435,9 @@ public override int GetHashCode()
                     hashCode = (hashCode * 397) ^ (IntensityModel != null ? IntensityModel.GetHashCode() : 0);
                     hashCode = (hashCode * 397) ^ (RTModel != null ? RTModel.GetHashCode() : 0);
                     hashCode = (hashCode * 397) ^ (Settings != null ? Settings.GetHashCode() : 0);
-                    hashCode = (hashCode * 397) ^ (Precursor != null ? Precursor.GetHashCode() : 0);
                     hashCode = (hashCode * 397) ^ (Peptide != null ? Peptide.GetHashCode() : 0);
+                    hashCode = (hashCode * 397) ^ (Precursor != null ? Precursor.GetHashCode() : 0);
```
nickshulman: In GetHashCode, we should use "RuntimeHelpers.GetHashCode" for fields that the Equals method uses ReferenceEquals to compare.
(Not something we should worry about in this pull request, and also has no actual impact other than being slightly less efficient)

PR #991, opened 2020-03-18
Add support for specifying tests with regex

chambm:
Like:
`TestRunner test=~.*Waters.*`
to run all tests with Waters in the name.


pwiz_tools/Skyline/TestRunner/Program.cs
```diff
@@ -958,6 +964,19 @@ private static List<string> LoadList(string testList)
                 }
             }
 
+            foreach (var testDll in TEST_DLLS)
+            {
+                foreach (var testInfo in RunTests.GetTestInfos(testDll))
```
brendanx67: What about using the GetTestList() function? I would prefer to see this inverted a bit. Put GetTestList() at the top if its performance is why you put this new blob of code at the bottom. Then write a single function that takes a RegEx and the test list and returns an IEnumerable that you can do outputList.AddRange() with any time you have name.StartsWith("~")

chambm: OK I think I did what you're asking here (except making the regex test a separate function because it's brief and I don't have any ideas of it being used elsewhere in the code). But let me know if I misunderstood.

PR #994, opened 2020-03-22
Skyline: Audit logging fixes based on disallowing ModifyDocument skipping…

brendanx67:
…undo/logging when given no AuditLogEntry

Because of Nick's mistake which caused the lack of an undo/redo record for changing Quantitative through the Targets view, I realized that Tobi's decision to skip undo/redo and audit logging when an audit logging function returned null was too permissive and easy to cause accidentally.

Instead, I added AuditLogEntry.SKIP for the few cases where we really want this behavior, but requiring more explicit intent to cause the log/undo/redo skipping behavior.

I ended up leaving this test-only for now and even needed to turn off the behavior (again explicitly) in TestIrtTutorial, because we currently cannot detect a settings change consisting solely of a modification to a .irtdb file.

But there are a fair number of improvements in this PR and it will make it more difficult for us to end up with document modifications which cannot be undone and lack an audit log record.


pwiz_tools/Skyline/SettingsUI/PeptideSettingsUI.cs
```diff
@@ -343,7 +343,10 @@ private PeptideSettings ValidateNewSettings(bool showMessages)
             BackgroundProteome backgroundProteome = BackgroundProteome.NONE;
             if (!backgroundProteomeSpec.IsNone)
             {
-                backgroundProteome = new BackgroundProteome(backgroundProteomeSpec);
+                if (_peptideSettings.BackgroundProteome.EqualsSpec(backgroundProteomeSpec))
```
nickshulman: I do not understand why you aren't just using regular .Equals here.

brendanx67: I guess I need to add some comments. It has to do with the _cache being not truly immutable, but being used in the BackgroundProteome.Equals().

PR #999, opened 2020-03-25
Remove maximum number of peptides limit on "Import Fasta", automatically resize LongWaitDlg to accomodate multiline messages

nickshulman:
The support board user "tobi" wants to be able to import large FASTA files without running into the 200,000 peptide limit.

This change removes that limit. I believe that if the user makes a mistake with their Enzyme settings so that they end up with far too many peptides, they will have plenty of time to hit cancel on the LongWaitDlg.

I have also made the LongWaitDlg automatically size itself (just like AlertDlg) so that I could display a 2-line status message like this:
image

The LongWaitDlg only grows itself vertically. I would expect that wide messages will get wrapped appropriately.
What was the really long message which we originally made LongWaitDlg resizable for? I want to make sure that message appears readably.


nickshulman:
There's something wrong with the way that the LongWaitDlg is resizing. I don't think it's displaying long messages correctly. I need to work on this one a little more.

PR #1000, opened 2020-03-26
TestRunner heap diagnostics

kaipot:


pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
```diff
@@ -501,11 +505,50 @@ public static HeapAllocationSizes[] GetProcessHeapSizes()
                 {
                     var h = buffer[i];
                     HeapLock(h);
+                    sizes[i].CommittedSizes = new Dictionary<long, int>();
```
brendanx67: I think you are going to need a flag to keep from doing this every time we calculate heap sizes. I think it will be too slow to always collect all of the strings. We had to turn off just to total committed by counts for the perf tests, because that was too slow. Please add a flag that can be turned on in the debugger to have this analysis, but otherwise skip it. i.e. it only happens during an active diagnosis.

kaipot: I wrapped the code into a conditional block with a flag but the debugger is giving me issues when I do this (hangs on "evaluating local variables", then gives an error saying it timed out). Trying to figure out how to work around it.

kaipot:
Didn't figure out why Visual Studio was giving me that error, but it seemed like it just didn't like breakpoints in certain places. I added a boolean that I was able to switch on/off during debugging to enable the counting.

kaipot:
> This doesn't look like it would ever trigger the hang outside IsTestMode. I think we need to turn leak hanging back on for this to be at all useful for diagnosing the problem in SkylineNightly test runs.

You're right, I had that condition backwards, fixed now.

PR #1002, opened 2020-03-26
Add Q-Values to Precursor Results Summary.

rita-gwen:


rita-gwen:
Nick, do you know if there are unit tests for the existing Precursor Results Summary values that I can use as an example or possibly just add q-value there?

pwiz_tools/Skyline/Model/Databinding/Entities/ColumnCaptions.resx
```diff
@@ -117,1080 +117,1117 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <data name="AbsFoldChange">
+  <data name="AbsFoldChange" xml:space="preserve">
```
brendanx67: This is a bit heavy on changes. If we need all these xml:space="preserve" added, let's do it in a separate commit.

pwiz_tools/Skyline/Model/Databinding/Entities/ColumnToolTips.resx
```diff
@@ -1,479 +1,497 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-	<!-- 
-		Microsoft ResX Schema
+  <!--
```
brendanx67: Again, too much changing here. Please limit the diff to the things you actually want to change.

pwiz_tools/Skyline/Model/Databinding/Entities/ColumnToolTips.resx
```diff
@@ -548,6 +548,10 @@ The options for determining this value can be specified on the Quantification ta
   <data name="MaxBestRetentionTime">
     <value>Maximum of the precursor BestRetentionTime values</value>
   </data>
+  <data name="MaxDetectionQValue" xml:space="preserve">
+    <value>A  maximum of the false discovery rate (FDR) score assigned to each chosen target peak after applying a mPro
```
nickshulman, line 552: How about:
"The median false discovery rate (FDR) assigned to the chosen peak for this peptide across all of the replicates."

FDR values usually come from mProphet models, but they can also come from SWATH or EncyclopeDIA search results. (Although I think EnyclopeDIA always assigns the same q-value to each replicate so these summary values are not as useful).

pwiz_tools/Skyline/Model/Databinding/Entities/SummaryValue.cs
```diff
@@ -80,6 +82,34 @@ public override string ToString()
...
+    {
+        public DetectionQValueSummary(Statistics statistics)
+        {
+            Min = statistics.Min();
+            Max = statistics.Max();
+            Median = statistics.Median();
+        }
+
+        [Format(Formats.PValue, NullValue = TextUtil.EXCEL_NA)]
+        public double? Min { get; private set; }
```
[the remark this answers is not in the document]
rita-gwen: Well, q value itself can be null if the mProphet model is not run, that's why I marked them as nullable.

pwiz_tools/Skyline/TestPerf/DiaSwathTutorialTest.cs
```diff
@@ -108,6 +109,11 @@ private string InstrumentTypeName
         [Timeout(int.MaxValue)] // These can take a long time
         public void TestDiaTtofTutorial()
         {
+            //DEBUG
+            RunPerfTests = true;
```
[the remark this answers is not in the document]
rita-gwen: Yes, thank you!

pwiz_tools/Skyline/Model/Databinding/Entities/ColumnToolTips.resx
rita-gwen: I copied the description from the Detection Q Value and I think it's better to be consistent. Also the current description points to the origin of the q values, which is helpful if a user would be wondering why they are null.

pwiz_tools/Skyline/Model/Databinding/Entities/ColumnToolTips.resx
rita-gwen: The last commit reverted all the unnecessary changes, it's much more compact now.

PR #1010, opened 2020-04-01
Skyline: make it possible for selected tests on selected machines to …

bspratt:
…invoke a debugger instead of quitting.

### DIFF
--- a/pwiz_tools/Skyline/TestUtil/AssertEx.cs
+++ b/pwiz_tools/Skyline/TestUtil/AssertEx.cs
@@ -37,17 +37,128 @@ using pwiz.Skyline.Util.Extensions;
 
 namespace pwiz.SkylineTestUtil
 {
+    /// <summary>
+    /// AssertEx provides an extended set of Assert functions, such as collection equality, as well as
+    /// new implementations of common Assert functions such as Assert.IsTrue.
+    /// 
+    /// These implementations give two advantages over standard Assert.* functions:
+    ///    Easier to set breakpoints while debugging
+    ///    Tests can optionally invoke a debugger instead of quitting, using the <see cref="Assume.DebugOnFail"/> mechanism.
+    /// </summary>
     public static class AssertEx
     {
         public static void AreEqualDeep<TItem>(IList<TItem> l1, IList<TItem> l2)
         {
-            Assert.AreEqual(l1.Count, l2.Count);
+            AreEqual(l1.Count, l2.Count);
             for (int i = 0; i < l1.Count; i++)
             {
                 if (!Equals(l1[i], l2[i]))
-                    Assert.AreEqual(l1[i], l2[i]);  // For setting breakpoint
+                {
+                    AreEqual(l1[i], l2[i]);  // For setting breakpoint
+                }
             }
         }
+
+        public static void AreEqual<T>(T expected, T actual, string message = null)
+        {
+            if (!Equals(expected, actual))
+            {
+                if (Assume.InvokeDebuggerOnFail)
+                {
+                    Assume.Fail(message); // Handles the debugger launch
+                }
+                Assert.AreEqual(expected, actual, message);
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


bspratt:
This is what you'd find when this kicks in:

image

pwiz_tools/Skyline/Util/Util.cs
```diff
@@ -1957,6 +1966,36 @@ public static string NullableDoubleToString(double? d)
     /// </summary>
     public static class Assume
     {
+
+        private static string _machineNamePatternForInvokeDebuggerOnFail; // When set, if machine name contains this string we will invoke the debugger rather than fail.
+        public class DebugOnFail : IDisposable
+        {
+            private string _initialValue;
+
+            public DebugOnFail(string machineNamePattern)
```
brendanx67, line 1975: Let's make this a Func<bool> instead of a machine name. That way the function could be anything and it could be testing for a set of 5 different machines or whatever. You could put utility functions in our test utility code without needing to implement them inside Assume.

pwiz_tools/Skyline/TestUtil/AssertEx.cs
```diff
@@ -37,18 +37,129 @@
...
+            {
+                if (Assume.InvokeDebuggerOnFail)
+                {
+                    Assume.Fail(message); // Handles the debugger launch
+                }
+                Assert.AreEqual(expected, actual, message);
+            }
+        }
+
+        public static void AreNotEqual<T>(T expected, T actual, string message)
```
brendanx67, line 74: Why not message = null?

pwiz_tools/Skyline/TestUtil/AssertEx.cs
```diff
@@ -37,18 +37,129 @@
...
+            {
+                if (Assume.InvokeDebuggerOnFail)
+                {
+                    Assume.Fail(message); // Handles the debugger launch
+                }
+                Assert.AreNotEqual(expected, actual, message);
+            }
+        }
+
+        public static void AreNotSame(object expected, object actual)
```
brendanx67, line 86: Why not message = null?

pwiz_tools/Skyline/Util/Util.cs
```diff
@@ -2001,6 +2040,14 @@ public static void AreEqual(double expected, double actual, double delta, string
 
         public static void Fail(string error = "")
         {
+            if (InvokeDebuggerOnFail)
```
brendanx67, line 2043: Let's just put this all in AssertEx for now, which makes it more limited to testing. It will likely produce cleaner code. In general, it was a mistake that you seemed to have made to use Assume in tests instead of Assert[Ex]. Maybe we should clean that up. Assume is really supposed to be for implementation code and not testing. This just further muddies that picture.

bspratt: OK, per our video chat we convinced ourselves that it does need to be in Assume as well as AssertEx if it's going to catch the bug we're chasing in TestLibraryBuild. where the Assert.Fail passes through Assume rather than AssertEx.

The ctor now just takes a simple bool, for best generality and simplicity, with a default of True,

pwiz_tools/Skyline/Util/Util.cs
```diff
@@ -2001,6 +2028,14 @@ public static void AreEqual(double expected, double actual, double delta, string
 
         public static void Fail(string error = "")
         {
+            if (InvokeDebuggerOnFail)
+            {
+                Console.WriteLine();
+                if (!string.IsNullOrEmpty(error))
+                    Console.WriteLine(error);
+                Console.WriteLine(@"error encountered, launching debugger as indicated by Assume.MachineNamePatternForInvokeDebuggerOnFail");
```
brendanx67, line 2036: This text is no longer valid.

bspratt: Right, thanks. I'll clean that up then merge to master.

PR #1014, opened 2020-04-07
Heap output if test leaks

kaipot:


pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
```diff
@@ -363,6 +363,27 @@ public bool Run(TestInfo test, int pass, int testNumber, string dmpDir)
...
 
+                if (heapOutput && ReportSystemHeaps)
+                {
+                    const int sizeOutputs = 50;
+                    const int stringOutputs = 50;
+                    var allSizes = new List<KeyValuePair<long, int>>();
+                    var allStrings = new List<KeyValuePair<string, int>>();
+                    foreach (var heapCount in heapCounts)
+                    {
+                        allSizes.AddRange(heapCount.CommittedSizes);
```
brendanx67, line 374: Merge sort? Since they are already in sorted order. Probably not worth it. But you can at least use .Take(size) here as well to reduce the number of things you will need to sort, since sorts are O(n*log(n)). There is no possible way you will use any value greater than you .Take() below. I guess most other clever things you could do are not worth the effort, but certainly that seems simple enough.

kaipot: Thanks for catching that. I agree that merge sort probably would not add much benefit, since most of the heaps are relatively small compared to the 0th one (?)

PR #1026, opened 2020-04-23
Skyline: improved handling of small molecule .MSP library files witho…

bspratt:
…ut Precursor_type values. Formerly we failed to load them, now we derive adduct type from declared mz and chemical formula by searching against mz vallues created by the most common adduct types, including the possibility of H2O loss. Reported by Christina Lucas.

Also improved handling of imported MSP files with multiple instances of the same molecule adduct - these are meant to load with a warning, but we would show the warning then not finish loading.


pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
```diff
@@ -577,47 +577,102 @@ private int INDEX_DECLUSTERING_POTENTIAL
...
-                TransitionCalc.MassShiftType.none, out massShift, out nearestCharge);
-            if (calculatedCharge.IsEmpty)
+
+            // See if this can be explained by (de)protonation within acceptable charge range
+            var minCharge = (isPositive ?? false)
+                ? TransitionGroup.MIN_PRECURSOR_CHARGE
+                : -TransitionGroup.MAX_PRECURSOR_CHARGE;
+            var maxCharge = (isPositive ?? true)
+                ? TransitionGroup.MAX_PRECURSOR_CHARGE
+                : -TransitionGroup.MIN_PRECURSOR_CHARGE;
```
kaipot, line 607: Subjective, but I think this would be more clear to me if written w/ if blocks.

pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
```diff
@@ -577,47 +577,102 @@ private int INDEX_DECLUSTERING_POTENTIAL
...
+                        var err = Math.Abs(adductInferred.MzFromNeutralMass(mass) - mz);
+                        if (err <= tolerance)
+                        {
+                            matches.Add(err, adductInferred);
+                        }
+                        else
+                        {
+                            // Try water loss
+                            var parts = text.Split('+', '-'); // Only for simple adducts like M+H, M+Na etc
+                            if (parts.Length == 2)
```
kaipot, line 631: Do we anticipate ever needing to support more than water loss and double water loss? If so it could be helpful to rewrite this in a more generalized form.

bspratt: Possibly? It seems like most other possibilities are already covered by that list of "most common" adducts put out by the Fiehn lab.

pwiz_tools/Shared/Common/SystemUtil/ProgressStatus.cs
```diff
@@ -32,6 +32,7 @@ public interface IProgressStatus
         int PercentComplete { get; }
         bool ProgressEqual(IProgressStatus status);
         Exception ErrorException { get; }
+        string WarningMessage { get; }
```
brendanx67, line 35: Nit, but can you move this up to below Message instead of under ErrorException. I might not have flagged this in the first place, if it had been there as it is in the ProgressStatus class.

PR #1038, opened 2020-05-03
Fix bogus validation errors in EditIrtCalcDlg when switching standards.

nickshulman:
This fixes the problem that Brendan was running into where if you put the focus in the last row of the EditIrtCalcDlg standard grid, and then choose a standard with fewer peptides in it, you get into a state of endless row validating errors.

### DIFF
--- a/pwiz_tools/Skyline/SettingsUI/PeptideGridViewDriver.cs
+++ b/pwiz_tools/Skyline/SettingsUI/PeptideGridViewDriver.cs
@@ -17,11 +17,9 @@
  * limitations under the License.
  */
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Windows.Forms;
-using NHibernate.Transform;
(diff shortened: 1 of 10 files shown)


pwiz_tools/Skyline/SettingsUI/PeptideGridViewDriver.cs
```diff
@@ -17,9 +17,11 @@
  * limitations under the License.
  */
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Windows.Forms;
+using NHibernate.Transform;
```
brendanx67, line 24: Seems like this might be the cause of the code inspection failure? Can't really see how NHibernate would be required for your changes.

brendanx67:
Nick, can you cherry-pick this to Skyline/skyline_20_1? It seems pretty simple to run into and in my experience it requires killing Skyline with the Task Manager, a pretty high penalty for stumbling into it.

PR #1044, opened 2020-05-05
implemented MS Amanda search within skyline

viikiid:
here is a first version of MS Amanda (and percolator) running in Skyline


pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -1271,6 +1275,16 @@ private static int GetMsLevel(Precursor precursor)
 
         }
 
+        private static int? GetChargeStateValue(Precursor precursor)
+        {
+            if (precursor.selectedIons == null || precursor.selectedIons.Count == 0)
+                return null;
+            var param = precursor.selectedIons[0].cvParam(CVID.MS_charge_state);
```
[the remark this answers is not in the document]
chambm: Oops, this was part of my ongoing review. I didn't want to finalize the review yet but it seems it finalized it when I made the one-off comment about percolator.exe.

chambm:
For mzIdentML, different charge states for the same spectrum should be put in the same SIR: each spectrum should have only one SIR, at least for single-run mzIdentMLs.

chambm:
I got a failed search without getting an apparent error:

> Starting search...
> Input files contain 2436 ms2 spectra
> Searching file 'Rpal_Std_2d_FullMS_Orbi30k_MSMS_Orbi7k_Centroid_Run1_102006_01.RAW'...
>   Searching spectra 1 - 2000 of 2436 spectra
>      Starting search against fwd database...
>      Search against fwd database finished.
>      Starting search against decoy database...
>      Search against decoy database finished.
>   Search finished for 2000/2436 spectra
>   Searching spectra 2001 - 2030 of 2436 spectra
>      Starting search against fwd database...
>      Search against fwd database finished.
>      Starting search against decoy database...
>      Search against decoy database finished.
>   Search finished for 2030/2436 spectra
> File 'Rpal_Std_2d_FullMS_Orbi30k_MSMS_Orbi7k_Centroid_Run1_102006_01.RAW' finished.
> 
> Running percolator ...
> Generating .mzid file...
> 
> Done.
> Search failed.
>

chambm:
OK, that is because success is a global and not initialized each time Run() is called.

chambm:
When I search 2 files, the 2nd file gets skipped. `ParseNextSpectra` isn't being called on it.

chambm:
If search fails, it seems to continue importing the FASTA anyway, without filtering on the DDA results.

chambm:
@viikiid Is MSAmanda calling percolator within its own DLLs? So the CMD window that pops up isn't something Skyline can control (and route that console output into the Skyline search log)?

pwiz_tools/Skyline/FileUI/PeptideSearch/BuildPeptideSearchLibraryControl.cs
```diff
@@ -250,16 +250,41 @@ private void listSearchFiles_SelectedIndexChanged(object sender, EventArgs e)
...
+            {
+                
+                using (OpenFileDialog dlg = new OpenFileDialog
+                {
+                    //todo add resources here
+                    Title = "Add spectra files to analyze",
+                    InitialDirectory = Path.GetDirectoryName(DocumentContainer.DocumentFilePath),
+                    CheckPathExists = true,
+                    Multiselect = true,
+                    Filter = "spectrum files(*.mgf;*.MGF;*.mzML;*.mzml;*.MZML; *.raw)|*.mgf;*.MGF;*.mzML;*.mzml;*.MZML;*.raw"
```
chambm, line 263: This should be search-engine dependent, or we can add a user-transparent msconvert step if the engine does not support the given file type.

pwiz_tools/Skyline/FileUI/PeptideSearch/BuildPeptideSearchLibraryControl.cs
```diff
@@ -280,12 +305,32 @@ public bool BuildOrUsePeptideSearchLibrary(CancelEventArgs e)
...
             }
         }
 
+        private bool PresetNecessarySettingsWithoutLibrary(CancelEventArgs cancelEventArgs)
+        {
+            //ImportPeptideSearch.DocLib = new EmptyLibraryForDDASearch();
+            foreach (string rawFile in ImportPeptideSearch.SearchFilenames)
+            {
+                if (!Path.GetExtension(rawFile).ToUpper().Equals(".MGF"))
+                    ImportPeptideSearch.SpectrumSourceFiles.Add(rawFile, new ImportPeptideSearch.FoundResultsFilePossibilities(rawFile){ExactMatch = rawFile});
```
chambm, line 323: Am I understanding correctly that with the Perform DDA Search option, FoundResultsFiles indicates the search engine source files (mzML, RAW, MGF, etc)?

pwiz_tools/Skyline/FileUI/PeptideSearch/MatchModificationsControl.cs
```diff
@@ -116,38 +130,73 @@ public bool Initialize(SrmDocument document)
...
+            newHeavyMods = (from mod in ImportPeptideSearch.MatcherHeavyMods
                 from ListBoxModification checkedMod in modificationsListBox.CheckedItems
                 where mod.Equivalent(checkedMod.Mod)
                 select mod).ToList();
+          }
 
             // Update document modifications
             return ImportPeptideSearch.AddModifications(document,
-                new PeptideModifications(newStructuralMods, new[] {new TypedModifications(IsotopeLabelType.heavy, newHeavyMods)}));
+                new PeptideModifications(structuralMods, new[] {new TypedModifications(IsotopeLabelType.heavy, newHeavyMods)}));
```
chambm, line 175: The way mods are presented in the wizard seems backward to me for DDA. I expect to first be asked about what mods to include in the search, and later asked whether to import them into the document from the library. At least the caption needs to change because it's talking about search results when the search hasn't been conducted yet. :)
image

pwiz_tools/Skyline/FileUI/PeptideSearch/BuildPeptideSearchLibraryControl.cs
brendanx67: It could mean that, but I think it is trying for the files which Skyline will extract chromatograms from, which usually are not MGF files and may not even be mzXML files, if the searched mzXML files had only MS/MS spectra, as was once the case in the Aebersold lab.

pwiz_tools/Skyline/Model/AbstractDdaSearchEngine.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public abstract class AbstractDdaSearchEngine
+    {
+        public abstract string[] FragmentIons { get; }
+        public abstract string EngineName { get; }
+        public abstract Bitmap SearchEngineLogo { get; }
+        protected string[] SpectrumFileNames { get; set; }
+        protected string[] FastaFileNames { get; set; }
+
+        public abstract void SetMS1Tolerance(double mass, string unit);
```
chambm, line 20: It's unfortunate Skyline doesn't have an MzTolerance class right now. There's MassTolerance in FastOverlapDemultiplexer but it's ppm-only. If we had the class, I'd say this should be a get/set property instead of SetXX() function. Same with the others below. MaxMissedCleavages could be its own property, and there will be search-engine-specific settings to add, somehow. Perhaps via a dictionary here? I guess we need to decide whether each search engine will get its own UI control or whether they will all share a single generic control.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -526,20 +559,68 @@ private void NextPage()
...
+
+            }
 
             var newPage = CurrentPage + 1;
             while (_pagesToSkip.Contains(newPage))
                 ++newPage;
 
             // Skip import FASTA if user filters for document peptides
-            if (newPage > Pages.import_fasta_page)
+            if (newPage > Pages.import_fasta_page && !ImportPeptideSearch.IsDDASearch)
```
chambm, line 623: Unless we are going to be writing out a FASTA file if the user pastes in FASTA text, we should make the FASTA tab only accept a FASTA file. I.e. replace the big textbox with a filepath box.

pwiz_tools/Skyline/SettingsUI/SearchSettingsControl.Designer.cs
```diff
@@ -0,0 +1,248 @@
...
+            this.clbFixedModifs.TabIndex = 6;
+            // 
+            // lblFixedModifications
+            // 
+            this.lblFixedModifications.AutoSize = true;
+            this.lblFixedModifications.Location = new System.Drawing.Point(16, 182);
+            this.lblFixedModifications.Name = "lblFixedModifications";
+            this.lblFixedModifications.Size = new System.Drawing.Size(99, 13);
+            this.lblFixedModifications.TabIndex = 24;
+            this.lblFixedModifications.Text = "Fixed modifications:";
```
chambm, line 171: Where is the variable modifications box? And why is the mods checklist repeated here? I think we can just fix the caption text on the Add Modifications tab and only use that tab for modifications. All the selected mods on that page will be included in the search. If desired, the previously selected mods could be presented again on this page in read-only text.

pwiz_tools/Skyline/Model/AbstractDdaSearchEngine.cs
```diff
@@ -0,0 +1,51 @@
...
+        public abstract void SetMS2Tolerance(double mass, string unit);
```
chambm, line 21: Should we plan now to support MS3 searches? I think they are typically handled similarly but with different mods selected, right? So this should be MSn tolerance or product/fragment ion tolerance.

pwiz_tools/Skyline/Model/MSAmanda/MSAmandaSearchWrapper.cs
```diff
@@ -0,0 +1,315 @@
...
+                    if (tokenSource.Token.IsCancellationRequested)
+                    {
+                        
+                        tokenSource.Token.ThrowIfCancellationRequested();
+                    }
+
+                    amandaInputParser = new MSAmandaSpectrumParser(rawFileName, Settings.ConsideredCharges, true);
+                    SearchEngine.SetInputParser(amandaInputParser);
+
+                    SearchEngine.PerformSearch(_outputParameters.DBFile);
```
chambm, line 251: I don't really understand what's going on when I give it multiple files to search. After calling this function on the first file in SpectrumFileNames, MSAmanda apparently calls GetTotalNumberOfSpectra() on all the files in SpectrumFileNames. This seems to be a bug in MSAmanda.

viikiid:
> For mzIdentML, different charge states for the same spectrum should be put in the same SIR: each spectrum should have only one SIR, at least for single-run mzIdentMLs.

thanks, didn't know that, we will change our mzIdentML generation accordingly

viikiid:
> @viikiid Is MSAmanda calling percolator within its own DLLs? So the CMD window that pops up isn't something Skyline can control (and route that console output into the Skyline search log)?

yes, I fixed the popup, it is not going to show a window any more and there is a log file generated for percolator in the percolator directory

pwiz_tools/Skyline/Model/AbstractDdaSearchEngine.cs
viikiid: good point, renamed the properties to precursor and fragmentIon tolerance such that each search engine can handle the tolerances in its own way!

pwiz_tools/Skyline/SettingsUI/SearchSettingsControl.Designer.cs
viikiid: the idea behind that was reuse the MatchModificationsControl to select the modifications for the search engine. If you added Oxidation(M) here, the SearchSettingsControl would look like this:
modiSettings

to have the user re-check the modifications and which ones are fixed and which ones are optional. But I can definitely make it readonly!

pwiz_tools/Skyline/FileUI/PeptideSearch/MatchModificationsControl.cs
viikiid: true! changed the caption!
asking the user which modifications shall be imported is still missing (added a todo). do you think we should show again the MatchModificationsControl?

pwiz_tools/Skyline/Model/AbstractDdaSearchEngine.cs
viikiid: true, this is an important decision. MS Amanda has some additional settings, that are left at default right now, but I added a button for additional settings that could open kind of a table view with the name of the parameter and its value to be changed and those could be read from the dictionary?!

pwiz_tools/Skyline/FileUI/PeptideSearch/BuildPeptideSearchLibraryControl.cs
viikiid: yes, the search engine searches all files in the SearchFileNames list but already uses these files to indicate Skyline where to extract the chromatograms from. After the search is finished, the SpectrumSourceFiles are replaced by the correct search result and the original source file. We might have to extend this if clause by other extensions, though... I don't know which file formats are fine to extract chromatograms... shall I add the mzXML here?

pwiz_tools/Skyline/SettingsUI/SearchSettingsControl.Designer.cs
chambm: OK, now I understand what it's saying. I previously thought the checklist items were only the fixed mods, and whether they were checked or not was whether they would actually be included in the search. I'm sure we could make that UI more clear if necessary, but is there a reason to let the user change them from fixed to variable (or vice versa) at this point rather than in the earlier mods selection dialog?

chambm:
@brendanx67 @nickshulman  I'm not sure yet why the DDA functional test is failing here; it's working locally. I'll fix that this week, but I think the code is ready for a preliminary review from you or Nick to see where you think it's not up to par. I haven't made sweeping changes to Viktoria's code. I'd like to know if there's something major here that needs to change before working on polishing it.

brendanx67:
I would like to start with a screen sharing review of the user interface before diving into the code. That usually yields enough required change that is harder to see in a straight code review, and harder to come up with on my own, when I am less prepared with test data sets and test cases. I usually reserve looking at the code for once I am completely happy with how it looks to the user.

chambm:
OK well I suppose it's ready for that then. Will need another one later when multiple search engine support is added, but we'll cross that bridge when we get there.

nickshulman:
The file system on the TeamCity computer wouldn't happen to be
case-sensitive, would it?
I noticed that it's called "Enzymes.xml" in TestSettings_x64.testsettings,
but the real file name is "enzymes.xml".
(Although, this is a long shot since I don't think the testsettings file
gets used on Team City-- it just uses TestRunner.exe).

By the way, on my machine TestDdaSearch passes in English, but fails in
French with the wrong number of peptides being found.
-- Nick

chambm:
Actually I guess what must be happening is the current directory is getting changed. Is there a util function to get Skyline's exe's path? (otherwise I'd resort to Assembly.GetCallingAssembly)

chambm:
French language failure looks like a problem in percolator. I wonder if we can set the language settings just for running percolator.exe? If so, it'll be a change in Amanda DLLs because it handles calling percolator. If not, it'll be a change in percolator.exe.

chambm:
Nevermind. Fixed just by setting `Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;` in the Task's worker function. Does that have a potential effect on anything other than that Task? Or is a threadpool thing?

chambm:
Huh. What should I do to solve 
```
Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException: Assert.Fail failed. Timeout 10 seconds exceeded in WaitForConditionUI. Open forms: SkylineWindow (Skyline), AllChromatogramsGraph (-> Z:\pwiz\pwiz_tools\Skyline\TestResults\TestRunner results\DdaSearchTest\Rpal_Std_2d_FullMS_Orbi30k_MSMS_Orbi7k_Centroid_Run1_102006_03.mz5: 82%

Total complete: 82%)
   at pwiz.SkylineTestUtil.AbstractFunctionalTest.WaitForConditionUI(Int32 millis, Func`1 func, Func`1 timeoutMessage, Boolean failOnTimeout, Boolean throwOnProgramException) in Z:\pwiz\pwiz_tools\Skyline\TestUtil\TestFunctional.cs:line 941
   at pwiz.SkylineTestUtil.AbstractFunctionalTest.EndTest() in Z:\pwiz\pwiz_tools\Skyline\TestUtil\TestFunctional.cs:line 1648
```

Seems like it's importing results or something but it's already in `EndTest()`?

nickshulman:
You should probably put a "WaitForDocumentLoaded()" near the end of
"TestAmandaSearch()". That will make sure that Skyline is finished
extracting chromatograms before it tries to go into the test cleanup code.

Functional tests are required to have closed all of their dialogs and be
finished with all their background processing by the time that they exit
their DoTest() method.
-- Nick

brendanx67:
Just noticed watching TestDdaSearch that the Search progress text box does not scroll down as it fills with text. It needs to keep the selection at the end of the text to force scrolling unless the user manually changes the selection or scrollbar, and it should be possible to restore the scrolling by pressing Ctrl-End on the keyboard. At least that is how we implemented build progress in Visual Studio in the 90s.

pwiz_tools/Skyline/Model/MSAmanda/MSAmandaSearchWrapper.cs
```diff
@@ -0,0 +1,306 @@
...
+            Settings.ConsideredCharges.Add(2);
+            Settings.ConsideredCharges.Add(3);
+            Settings.ChemicalData.UseMonoisotopicMass = true;
+            Settings.ReportBothBestHitsForTD = false;
+            mzID.Settings = Settings;
+            SearchEngine = new MSAmandaSearch(helper, _baseDir, _outputParameters, Settings, token);
+            //OutputMzid mzid = new OutputMzid();
+            SearchEngine.InitializeOutputMZ(mzID);
+
+            //List<FastaDBFile> fastaFiles = GetFastaFileList();
```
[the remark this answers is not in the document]
chambm: @viikiid Can you answer why there's so much commented out code in the wrapper? Is it dead code or something you planned to use eventually?

pwiz_tools/Skyline/Model/MsgfPlusSearchEngine.cs
```diff
@@ -0,0 +1,189 @@
...
+        {
+            @"Low-res LCQ/LTQ",
+            @"Orbitrap/FTICR/Lumos",
+            @"TOF",
+            @"Q-Exactive"
+        };
+
+        private static readonly string[] ENZYMES = new[]
+        {
+            @"Unspecific cleavage",
```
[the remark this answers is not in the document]
chambm: Interesting. These terms are technical enough that I'd think users wouldn't expect them to be translated. But what about translating enzyme names as well? I could imagine both, or either, or neither. I tried downloading pFind to see what they do in their Chinese language setting but they haven't sent me a license key yet. License keys with academic software, ugh.

pwiz_tools/Skyline/Model/ImportPeptideSearch.cs
```diff
@@ -377,8 +383,20 @@ public List<string> GetUnmatchedSequences()
...
+            {
+                _matcher.MatcherPepMods = modifications;
+                return document.Settings.ChangePeptideModifications(mods => _matcher.SafeMergeImplicitMods(document));
+            }
+            else
+            {
+                return document.Settings.ChangePeptideSettings(
+                    document.Settings.PeptideSettings.ChangeModifications(modifications));
+
+                //todo add heavy modifications
```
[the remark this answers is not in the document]
chambm: @viikiid Are heavy mods supported by MSAmanda? I.e. is this a todo in Skyline only, or also in MSAmanda?

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -401,10 +430,13 @@ private void NextPage()
                         // did not, so hide the "early finish" button for
                         // the rest of the wizard pages.
                         ShowEarlyFinish(false);
-
-                        if (FastaOptional)
+                        
+                        if (BuildPepSearchLibControl.PerformDDASearch) {
```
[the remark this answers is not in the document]
chambm: What is the style guideline for using brackets for one-line blocks?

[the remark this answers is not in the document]
brendanx67: Leaving off brackets is allowed for a single line if or when both if and else are single-line, but never on the else when the if has brackets. Or you can flip the conditional and not user brackets on the if, but use them on the else. For getters and setters single line brackets can be used like

private string value { get { return _value; } }

But in most other cases they are avoided, and we never use KR bracket style for if...else

if (test) {
    ...
} else {
    ...
}

pwiz_tools/Skyline/SettingsUI/SearchSettingsControl.Designer.cs
```diff
@@ -0,0 +1,227 @@
...
+            this.cbMS1TolUnit.TabIndex = 2;
+            // 
+            // lblMs1Tolerance
+            // 
+            this.lblMs1Tolerance.AutoSize = true;
+            this.lblMs1Tolerance.Location = new System.Drawing.Point(13, 53);
+            this.lblMs1Tolerance.Name = "lblMs1Tolerance";
+            this.lblMs1Tolerance.Size = new System.Drawing.Size(79, 13);
+            this.lblMs1Tolerance.TabIndex = 9;
+            this.lblMs1Tolerance.Text = "MS1 tolerance:";
```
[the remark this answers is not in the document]
chambm: I'd never heard about using mnemonics for controls other than buttons. Apparently it's possible (https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.label.usemnemonic?view=netcore-3.1) but I don't see it used for other textboxes in Skyline controls (find in files didn't find UseMnemonic at all in the Skyline project). Is that something you've just started to enforce or is there another way to do non-button mnemonics I'm not aware?

pwiz_tools/Skyline/Model/MSAmanda/MSAmandaSearchWrapper.cs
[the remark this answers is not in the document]
viikiid: I deleted the commented code, it was something I tried but don't need it any longer.

pwiz_tools/Skyline/Model/MSAmanda/MSAmandaSearchWrapper.cs
```diff
@@ -0,0 +1,306 @@
...
+                        var elem = elemsFromUnimod.Find(m => m.AA == aa);
+                        if (elem != null)
+                        {
+                            Modification modClone = new Modification(elem);
+                            modClone.Fixed = item.IsVariable;
+                            Settings.SelectedModifications.Add(modClone);
+                        }
+                        else
+                        {
+                            //Settings.SelectedModifications.Add(GenerateNewModification(item.Key, item.Value));
```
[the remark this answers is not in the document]
viikiid: that depends: in Skyline is there a way to define your own modification? something like "MyNewFancyModification - K(+17.3Da)" if yes, I need to add code here

pwiz_tools/Skyline/Model/MSAmanda/MSAmandaSearchWrapper.cs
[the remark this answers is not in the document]
chambm: Thanks! FYI I've started work on all these comments Brendan left in cases where I didn't ask you about them, so no need to double the work.

pwiz_tools/Skyline/Model/ImportPeptideSearch.cs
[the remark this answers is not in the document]
viikiid: yes and no. the heavy mods are not in unimod but I will add a "user defined modification" to represent the heavy mods. so it is a todo in skyline

chambm:
> Just noticed watching TestDdaSearch that the Search progress text box does not scroll down as it fills with text. It needs to keep the selection at the end of the text to force scrolling unless the user manually changes the selection or scrollbar, and it should be possible to restore the scrolling by pressing Ctrl-End on the keyboard. At least that is how we implemented build progress in Visual Studio in the 90s.

Turns out that using `AppendText()` instead of `Text +=`  does this. Or at least it scrolls it to the end, I don't know if it pauses the autoscrolling if you manually scroll up (is that important?).

Another issue though: you mentioned before that you wanted a successful search to segue seamlessly into the importing results step. Right now the user has to click "Finish" after the search completes, then set up the library import (e.g. OK the 'ambiguous peptides' dialog if not configured to be transparently allowed, set up minimum peptides per protein), and then it goes into importing results. Should we embed the library import settings control into the search settings tab (and then copy the settings forward automatically instead of waiting for user to interact with the dialog)?

brendanx67:
Yes, it is important that scrolling or moving the caret from the end of the existing text stops this behavior. Otherwise, you end up with the highly undesirable behavior of jumping to the end every time new content is added. If I am trying to look at something or select and copy something at the beginning or middle frequent changes to the log can make this impossible.

But, if you have two ways to achieve adding log text and one scrolls and the other doesn't, then you could just choose which ti used based on the positions of the scroll thumb and the insertion caret.

brendanx67:
On the second question, let's go through it again together and review where things stand today.

pwiz_tools/Skyline/FileUI/PeptideSearch/BuildPeptideSearchLibraryControl.cs
chambm: NB: Brendan suggests the transparent msconvert step. Fortunately not needed for MSAmanda.

pwiz_tools/Skyline/Model/MSAmanda/MSAmandaSpectrumParser.cs
```diff
@@ -0,0 +1,133 @@
...
+        private int amandaId = 0;
+        private List<int> consideredCharges;
+        private bool useMonoIsotopicMass;
+        private string rawFileName; 
+        public Dictionary<int, string> SpectTitleMap { get; }
+
+        public MSAmandaSpectrumParser(string file, List<int> charges, bool mono)
+        {
+            consideredCharges = charges;
+            if (file.EndsWith(".raw"))
```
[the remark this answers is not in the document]
chambm: Replaced with `MsDataFileImpl.SupportsVendorPeakPicking()`

pwiz_tools/Skyline/FileUI/PeptideSearch/BuildPeptideSearchLibraryControl.cs
chambm: I don't see any way to do this reliably with only file extension. Sure, MGF will never have MS1s, but most other open formats might or might not. I don't see any way to tell whether those contain the MS1s other than opening them to check. Simpler would be to use the searched files for XIC extraction only if a vendor raw file with the same basename can't be found nearby (not sure how this would work for WIFF though). Then if the searched file ends up being used, if it happens to only contain MS2s, a suitable error message can be given.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -526,20 +563,82 @@ private void NextPage()
...
+                    ImportPeptideSearch.SearchEngine.SetFastaFiles(ImportFastaControl.FastaFile);
+                    SearchControl.OnSearchFinished += SearchControl_OnSearchFinished;
+                    btnNext.Enabled = false;
+                    btnCancel.Enabled = false;
+                    btnBack.Enabled = false;
+                    SearchControl.RunSearch();
+                    break;
+
+                case Pages.dda_search_page: // this is really the last page
+                                            //todo check if search was correct
```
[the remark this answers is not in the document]
chambm: @viikiid What did this "check if the search was correct" comment mean?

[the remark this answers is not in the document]
viikiid: Sorry, I forgot to delete this todo, this is resolved, as it is already checked in DDASearchControl.RunSearch()

PR #1050, opened 2020-05-07
Refactor LibraryRankedSpectrumInfo 

nickshulman:
I am working on refactoring "LibraryRankedSpectrumInfo" because I could not understand the old code well enough to eventually add support for crosslinked peptides.
(The things that I could not understand was there was a class called "RankParams" that was being passed around, and it was not clear when its fields were assigned their values, and also the class "RankedMI" was not immutable).

I have a new class called "SpectrumRanker", which has a few objects in it to hold the parameters used to do the ranking ("FragmentFilter", "TargetInfo"), and a class to hold the calculated masses of the fragments ("MoleculeMasses"), and then "RankingState" which holds the numbers that it keeps track of while it's assigning ranks to the matched ions.


pwiz_tools/Skyline/Model/Lib/SpectrumRanker.cs
```diff
@@ -0,0 +1,835 @@
+﻿using System;
```
bspratt, line 1: Don't forget copyright stuff

pwiz_tools/Skyline/Model/Lib/SpectrumRanker.cs
```diff
@@ -0,0 +1,835 @@
...
+using System.Linq;
+using pwiz.Common.Chemistry;
+using pwiz.Common.Collections;
+using pwiz.Common.SystemUtil;
+using pwiz.Skyline.Model.DocSettings;
+using RankedMI = pwiz.Skyline.Model.Lib.LibraryRankedSpectrumInfo.RankedMI;
+using pwiz.Skyline.Util;
+
+namespace pwiz.Skyline.Model.Lib
+{
```
bspratt, line 13: As you further hone your sense of just what it is this code does, it would be great to have a comment here describing that/

PR #1058, opened 2020-05-11
C++ port of DIA Umpire for creating DDA pseudo-spectra from DIA

chambm:
- added C++ port of a subset of DIA Umpire code to create DDA pseudo-MS/MS spectra in pwiz
- fixed user_errors (like unknown filter not being fatal in msconvert
- added SpectrumList_DiaUmpire and "diaUmpire" filter to SpectrumListFactory
* added SpectrumList::min_level_accepted() for succinctly determining what DetailLevel is needed to get a spectrum to pass some predicate
* changed IterationListener::UpdateMessage::ctor to cap iterationIndex at iterationCount-1
* changed IterationListenerRegistry::broadcastUpdateMessage() to always update on last iteration for time-based update intervals
* added commonly used static sets to IntegerSet
* added deallocate (via swap) function to Container.hpp
* fixed DiffConfig.ignoreDataProcessing not affecting all dataProcessing comparisons


chambm:
@brendanx67 My initial focus was on porting the existing Java functionality. I didn't implement several obvious ideas about how to make the filter more user-friendly. For example:
- read instrument type from input file and pick appropriate settings based on that (and allow overriding this from filter parameters)
- allow overriding individual config parameters (you can see there are a lot in InstrumentParameter.hpp) from filter parameters
- make params file optional if instrument is specified in either filter parameters or the input file

chambm:
@brendanx67 Any reason I can't merge this now? It contains some Linux compile fixes since I also had to upgrade TeamCity's linux builds to GCC 5 to get C++14.

chambm:
Only the command-line for now. The params file is required and there are a ton of parameters (way more than any other filter in pwiz) so without more work on deriving the parameters from a single instrument setting, it doesn't make much sense in MsconvertGUI.

PR #1059, opened 2020-05-12
Bruker method export

kaipot:


chambm:
@kaipot Can you handle updating the Skyline-side to make use of the progress delegate?

kaipot:
> @kaipot Can you handle updating the Skyline-side to make use of the progress delegate?

Yes, the test is working for me now, I'll make the changes in Skyline.

kaipot:
> > @kaipot Can you handle updating the Skyline-side to make use of the progress delegate?
> 
> Yes, the test is working for me now, I'll make the changes in Skyline.

I've added this to Skyline. By the way Matt, the test changes the prmsqlite file. Maybe it should copy it (or not write)?

brendanx67:
That should get fixed. We don't want a test that causes a tree diff.

chambm:
Should be ready to merge now. Kaipo I'm assuming you tested that the progress/cancel function is working ok in Skyline. :)

kaipot:
> Should be ready to merge now. Kaipo I'm assuming you tested that the progress/cancel function is working ok in Skyline. :)

Yep, I think this is good to go. I will send out a link to the artifacts once the build is done.

PR #1066, opened 2020-05-13
Skyline: the progress display for Import Results should show only chr…

bspratt:
…omatograms for targets (in color) - it was also showing chromatograms for TIC, BPC, etc (in grey) which tended to eclipse the target chromatograms,


pwiz_tools/Skyline/Model/Results/ChromDataProvider.cs
```diff
@@ -430,7 +430,7 @@ public override bool GetChromatogram(int id, Target modifiedSequence, Color colo
 
             // Display in AllChromatogramsGraph
             var loadingStatus = Status as ChromatogramLoadingStatus;
-            if (loadingStatus != null)
+            if (loadingStatus != null && modifiedSequence != null) // Don't show TIC, BPC etc
```
brendanx67, line 433: I think this is likely too broad and it will get rid of anything that is currently shown as gray. We currently import SRM chromatograms which have no target specified. That is the point of showing them in gray. I think you are going to need to come up with a way of determining whether it is a global chromatogram or not.

bspratt: Based on what I observed running the tutorial test in an older Skyline, I thought that we only wanted to see matched chromatograms (I wasn't seeing any gray). But if we do, then yes, this is too broard and I will rework it.

brendanx67: Try the AbsoluteQuant tutorial. That is SRM where the targeted peptide is not the only thing that was measured in the SRM. I would expect you to see gray even before this issue. Also the iRT tutorial, where we import complex SRM targeting only the iRT peptides in it.

bspratt: You're correct - and I don't think there's actually any new issue here after all.

When I go back and run SmallMoleculesQuantificationTest in 19.1, which is where it first appeared, it does show gray (I was running the wrong test on my previous comparison with older Skyline versions).

In fact in 19.1 SmallMoleculesQuantificationTest shows nothing but gray, as you noticed in the current version. As it happens, the only chromatograms with document transitions have much lower intensities that some others, so they get blown out visually. So I was mistaken about the BPC and TIC being the culprits.

So, not a *new* issue but still arguably an issue since it raised an alarm in your mind - not what you expected to see as a user.

A few ideas:
1) use the "fix" in this branch?
2) log scale?
3) do nothing

brendanx67: So, you are saying that this SRM data actually measured much more intense things that just are not included in the tutorial, similar to what was done in the peptide AbsoluteQuant tutorial? Okay, then maybe we do nothing. Not great for this tutorial, but also not something users should see a lot with their own data. I thought there might be some other issue and didn't want to just continue past it any longer after going through the tutorial a dozen times.

bspratt: That's correct, the document happened to target only relatively low intensity transitions, so those got visually blown out by the much more intense unmatched ones. 

This episode suggests to me that the user expectation is that the progress window only shows target progress, as on the full-scan side of things.  It was unusual enough to see gray that we thought we had a bug. Which means, I think, that we do have a bug.

I'd still show gray if it happens that no transition in the file has a target match, so that there's at least some indicator of activity.

brendanx67: And yet, I feel it warrants deeper consideration than we have to give it now that I understand why it is behaving that way, which is expected behavior given the original implementation. I see that it is not ideal in this case, but no one else seemed to care at all and it took me 10 runs of the tutorial to decide I should understand more fully. If you like, post an issue to yourself on our issues list. I am not ready to accept this PR and I don't even want to consider it further right now. We both have higher priorities right now than rehashing this design for SRM data. Please close the PR.

brendanx67: For instance, a better fix might be just to set the y-axis max based on only the curves that match the targets. That way the gray would still show up, but would not be able to reduce the targets to being invisible or close to it. I am sure there are other possibilities even. I just don't want to give it the time required at this time, given that it has always been this way and I was the first person to complain.

bspratt: OK, I'll delete this branch and this PR.

PR #1075, opened 2020-05-16
Filter out invalid peptide sequences from library files when creating cache.

nickshulman:
Filter out invalid peptide sequences in libraries and show warning message.

The warning message is shown after creating the library cache file.
When Skyline processes complete IProgressStatus objects, SkylineWindow will now display a MessageDlg if IProgressStatus.WarningMessage is not empty.

The warning message looks like this:
image


pwiz_tools/Skyline/Model/Lib/EncylopeDiaLibrary.cs
```diff
@@ -205,7 +205,7 @@ private bool Load(ILoadMonitor loader)
         // ReSharper disable LocalizableElement
         private bool LoadLibraryFromDatabase(ILoadMonitor loader)
         {
-            var status = new ProgressStatus(
+            IProgressStatus status = new ProgressStatus(
```
bspratt, line 208: Just curious - why not var?

brendanx67: Because you need IProgressStatus and not ProgressStatus. Pretty common in our code.

bspratt: Huh, I would have thought that amounted to the same thing.

pwiz_tools/Skyline/Model/Lib/Library.cs
```diff
@@ -997,6 +997,62 @@ protected virtual void SetLibraryEntries(IEnumerable<TInfo> entries)
...
+            status = WarnInvalidEntries(status, validEntries.Count, invalidKeys);
+            return validEntries;
+        }
+
+        protected bool IsValidLibKey(LibKey libKey)
+        {
+            try
+            {
+                var libraryKey = libKey.LibraryKey;
+                if (libraryKey is PeptideLibraryKey peptideLibraryKey)
```
brendanx67, line 1025: This implies the need for a base class.

pwiz_tools/Skyline/TestFunctional/InvalidPeptidesInLibraryTest.cs
```diff
@@ -0,0 +1,86 @@
...
+        [TestMethod]
+        public void TestInvalidPeptidesInLibrary()
+        {
+            TestFilesZip = @"TestFunctional\InvalidPeptidesInLibraryTest.zip";
+            RunFunctionalTest();
+        }
+
+        protected override void DoTest()
+        {
+            VerifyInvalidPeptideMessage("BiblioSpecInvalidPeptides.blib", 2, 30);
```
brendanx67, line 45: Seems like we should include a small molecule library in this test.

pwiz_tools/Skyline/Properties/Resources.resx
```diff
@@ -10803,4 +10803,8 @@ There might be something wrong with default web browser on this computer.</value
   <data name="CommandLine_AddDecoys_Decoys_discarded" xml:space="preserve">
     <value>Decoys discarded</value>
   </data>
+  <data name="CachedLibrary_WarnInvalidEntries_" xml:space="preserve">
+    <value>{0} of the {1} entries in library '{2}' were invalid, including:
```
brendanx67, line 10807: Not a big fan of messages starting with a number. How about this?

Invalid entries found in the library '{0}' ({1} of {2}), including:

This gets to the problem statement quicker with the numbers being more supplementary information, which they are.

pwiz_tools/Skyline/Model/Lib/EncylopeDiaLibrary.cs
brendanx67: ReSharper knows better. Give it a try. It has to do with the use of "ref" in the function call below:

SetLibraryEntries(FilterInvalidLibraryEntries(ref status, spectrumInfos));

nickshulman:
Brian,

Can you give me an example of an invalid small molecule string that we
might potentially find in a .blib file that I could use in a unit test?

I know you're fixing something to do with "[M+]" as an adduct.
-- Nick

nickshulman:
The issue you're thinking of was a garbage chemical formula. Here's the
data in my test for that.  Change  "Formula: [C11H22NO4]+\n" to
"Formula:GARBAGE\n" and you should have what you need.

        public const string TEXT_LIB_FORMULA_PLUS =
            "\n" +
            "Name: ACar 4:0\n" +
            "Synon: [M]+\n" +
            "Synon: $:00in-source\n" +
            "DB#: LipidBlast000001\n" +
            "InChIKey: QWYFHHGCZUCMBN-UHFFFAOYSA-O\n" +
            "Precursor_type: [M]+\n" +
            "Spectrum_type: MS2\n" +
            "PrecursorMZ: 232.15433\n" +
            "Instrument_type: in-silico QTOF\n" +
            "Instrument: SCIEX 5600\n" +
            "Ion_mode: P\n" +
            "Collision_energy: 45 V\n" +
            "Formula: [C11H22NO4]+\n" +
            "MW: 232\n" +
            "ExactMass: 232.1543346040907\n" +
            // TODO(bspratt) - MoNA embeds a lot of information in Comments
that we should pick out, like SMILES, InChIKey, retention time etc
            "Comments: \"SMILES=CCCC(=O)OC(CC(O)=O)C[N+](C)(C)C\"
\"compound class=ACar\" \"computed SMILES=O=C(O)CC(OC(=O)CCC)C[N+](C)(C)C\"
\"computed
InChI=InChI=1S/C11H21NO4/c1-5-6-11(15)16-9(7-10(13)14)8-12(2,3)4/h9H,5-8H2,1-4H3/p+1\"
\"retention time=0.51\" \"collision energy spread=15 V\" \"author=Tobias
Kind, Hiroshi Tsugawa\" \"computed mass accuracy=2.3431646471704717\"
\"computed mass error=5.439758187435473E-4\"
\"SPLASH=splash10-001r-7090000000-aa12589a2481560ea0d5\" \"submitter=Tobias
Kind (University of California, Davis)\" \"MoNA
Rating=3.6363636363636367\"\n" +
            "Num Peaks: 2\n" +
            "85.02841 80.080080\n" +
            "232.1543 100.000000\n";

PR #1080, opened 2020-05-19
Skyline: Fix EditPeakScoringModelDlg to keep mProphet histograms from…

brendanx67:
… extending too far left with invisible buckets


brendanx67:
Turns this:
image
Into this:
image

PR #1085, opened 2020-05-20
Skyline: when in Molecules UI mode with no spectral libraries loaded,…

bspratt:
… the "View | Spectral Libraries" menu item would try to take you to the Molecule Settings "Library" tab but would actually drop you on the "Quantification" tab instead, due to some tabs being hidden in this UI mode. Now using a map instead of a simple index to pick the right tab.


pwiz_tools/Skyline/SettingsUI/PeptideSettingsUI.cs
```diff
@@ -1258,8 +1296,8 @@ public IFormView ShowingFormView
 
         public TABS SelectedTab
         {
-            get { return (TABS)tabControl1.SelectedIndex; }
-            set { tabControl1.SelectedIndex = (int)value; }
+            get { return ControlIndexToTab(tabControl1.SelectedIndex); }
```
nickshulman, line 1299: I think you could just make this:
return tabMap[tabControl1.SelectedTab];

bspratt: The enum is used in serialization, so we still need a mapping, but I agree that I could simplify this code by just using a dictionray (and matching reverse dictionary). Thanks as always for a thoughtful review.

bspratt:
Tidy, I like it and will implement.

PR #1087, opened 2020-05-21
Skyline: automatically add small molecule detail columns for datagrid…

bspratt:
…s that use TargetColumn


pwiz_tools/Skyline/Controls/TargetColumn.cs
```diff
@@ -71,5 +127,284 @@ public override object Clone()
...
+    public interface ISmallMoleculeColumnsManagementProvider
+    {
+        TargetResolver TargetResolver { get; }
+        SmallMoleculeColumnsManager SmallMoleculeColumnsManager { get; }
     }
+
+    /// <summary>
+    /// Handles updating small molecule detail columns when target is set
+    /// </summary>
+    public class SmallMoleculeColumnsManager
```
nickshulman, line 146: This class should go in a separate file.

bspratt: Done. Anything else?

pwiz_tools/Skyline/Controls/SmallMoleculeColumnsManager.cs
```diff
@@ -0,0 +1,307 @@
...
+using pwiz.Skyline.Properties;
+
+namespace pwiz.Skyline.Controls
+{
+
+    /// <summary>
+    /// Interface which guarantees automatic management of small molecule details in
+    /// data grids originally designed for peptides
+    /// </summary>
+    public interface ISmallMoleculeColumnsManagementProvider
```
nickshulman, line 34: Maybe the SmallMoleculeColumnsManager should have a TargetResolver on it, and then you would not need this interface.

bspratt: Here's a screenshot from the iRT tutorial running with the UI set to Mixed mode:
image

bspratt: and I'll see about eliminating that interface

bspratt:
Here's an example of how a TargetResolver equipped grid looks currently:
image

and with the new work - more detail, which is nice:
image

but the big difference is that if you wanted to add a new entry to that list that wasn't already in the document, you'd have to understand our crazy string encoding. Now you just give it a name, formula etc and it all works.

brendanx67:
I guess I see this as more informational than a useful feature that anyone would use. Certainly, I doubt anyone will ever enter a single row into these grids manually. I have used pasting in lists of peptides, but here I think it is not as clear what columns you need, and there does not seem to be a way to turn off unused columns and rearrange them, as in Edit > Insert > Transitions. So, to me it looks more like you get the information rather than this makes it possible to enter the information. The Inche and SMILES columns seem very long, possibly less useful in this context. If the KEGG column is another ID and much shorter, maybe it should go next to CAS and IncheKey instead of at the end.

bspratt:
It presents columns for any information that it has, leaving out any columns that it hasn't got values for. If there isn't any initial information available, it presents all possible fields so you can paste in, as in populating an ion mobility database (which is what drove this work in the first place). It's true that there's no way to hide or rearrange columns at this point, that would be a nice enhancement but it's not a big deal to do it in Excel first. The current order of all those ID columns reflects our priority of IDs in the event that more than one are available, but that's not terribly important from a UI point of view and they could certainly be ordered for decreasing brevity if that's what we want.

bspratt:
A picture of friendlier UI for small molecule users
image

brendanx67:
Masses seem a bit high precision for UI. Let's limit them to the same 4 decimal places used in the Targets tree.

bspratt:
It's a round trip thing, but I'll see what I can do about it.

brendanx67:
You need the full value for the round trip, but you can display whatever precision you want.

bspratt:
Certainly - you just have to make sure you don't pull it back in again and
decide that the molecule description has changed.

nickshulman (commented):
> Certainly - you just have to make sure you don't pull it back in again and decide that the molecule description has changed.

Microsoft Access has logic in it to decide whether a the format string loss data for a particular value. When you go into edit mode on a cell in Microsoft Access, it checks whether parsing currently formatted value is equal to the cell value. If the parsed formatted value and the real value are not equal, then the value that gets put in the edit box is the unformatted real value.

You would think that this logic would have been duplicated somewhere in the current DataGridView, but I do not see anything like that. There might be some way to make it work where you change the formatting if necessary on a DataGridViewCell right before it goes into edit mode, but I do not know whether that is feasible.

bspratt:
Excellent, thanks!

bspratt:
Closing this until I can return to it

PR #1106, opened 2020-06-05
Skyline: restore per-precursor Explicit Collision Energy

bspratt:
In version 4.2 we changed Explicit Collision Energy from being an optional per-precursor value to being an optional per-transition value. We have now restored the per-precursor capability, which is overridden by per-transition values when available.

For transition list inputs with an "Explicit Collision Energy" column, behavior has changed slightly:
   - For any given precursor, the first-seen explicit collision energy value is taken as the precursor explicit collision energy.
   - If all transitions for a precursor claim the same explicit CE, then only the precursor explicit CE is set, and all its transitions use that value at output time.
   - If any transitions declare different explicit CE values than the per-precursor value, their per-transition explicit CE value is set and that overrides the per-precursor value at output time.

The per-precursor explicit CE value can also be set using the "Precursor Explicit Collision Energy" column in the document grid.


bspratt:
image
image

pwiz_tools/Skyline/Model/Databinding/Entities/ColumnToolTips.resx
```diff
@@ -1,479 +1,497 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-	<!-- 
-		Microsoft ResX Schema
+  <!--
```
brendanx67, line 3: Please make this change less extensive.

pwiz_tools/Skyline/Model/SrmDocument.cs
```diff
@@ -2115,10 +2115,15 @@ public double GetCollisionEnergy(PeptideDocNode nodePep, TransitionGroupDocNode
         public static double GetCollisionEnergy(SrmSettings settings, PeptideDocNode nodePep,
             TransitionGroupDocNode nodeGroup, TransitionDocNode nodeTran, CollisionEnergyRegression regression, int step)
         {
-            var ce = nodeTran==null // If we're only given a precursor, use the explicit CE of its children if they all agree
-                ? (nodeGroup.Children.Any() && nodeGroup.Children.All( node => ((TransitionDocNode)node).ExplicitValues.CollisionEnergy == ((TransitionDocNode)nodeGroup.Children.First()).ExplicitValues.CollisionEnergy) 
-                    ? ((TransitionDocNode)nodeGroup.Children.First()).ExplicitValues.CollisionEnergy : null)
-                : nodeTran.ExplicitValues.CollisionEnergy;
+            // Collision Energy explicitly declared at the transition level is taken to be the correct value.
```
brendanx67, line 2118: This is a mess. Can it be broken out as a separate function using ifs and return statements, which would also make it more debuggable.

pwiz_tools/Skyline/SettingsUI/EditCustomMoleculeDlg.resx
```diff
@@ -977,6 +1045,6 @@
     <value>EditCustomMoleculeDlg</value>
   </data>
   <data name="&gt;&gt;$this.Type" xml:space="preserve">
-    <value>pwiz.Skyline.Util.FormEx, Skyline, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null</value>
+    <value>pwiz.Skyline.Util.FormEx, Skyline-daily, Version=20.1.1.155, Culture=neutral, PublicKeyToken=null</value>
```
brendanx67, line 1048: Please revert this.

pwiz_tools/Skyline/Skyline.csproj
```diff
@@ -3001,6 +3000,7 @@
     <EmbeddedResource Include="Model\Prosit\Config\PrositConfig.xml">
       <SubType>Designer</SubType>
     </EmbeddedResource>
+    <None Include="ClassDiagram1.cd" />
```
brendanx67, line 3003: This file contains changes we don't want. Please limit the changes to just those required for this fix.

pwiz_tools/Skyline/Model/Databinding/Entities/ColumnToolTips.resx
chambm: Visual Studio's treatment of resx and designer-managed files is infuriating. They just get rearranged for no damn (discernable) reason. You practically have to text-edit the things for simple changes, or give up on being able to easily track the real changes (rather than reorderings).

pwiz_tools/Skyline/Model/SrmDocument.cs
bspratt: Its in a pretty brief function already, so I've just changed it in place use if statements.

pwiz_tools/Skyline/Model/Databinding/Entities/ColumnToolTips.resx
bspratt: Kinda makes you wonder about their interior change control processes, doesn't it?

chambm: It does indeed. Maybe they use some special sort-before-diffing engine for resx and designer files, like this guy did back in 2009: https://www.codeproject.com/Articles/37022/Solving-the-resx-Merge-Problem - but Googling reveals surprisingly little about dealing with the problem.

bspratt:
OK, this should all be tidied up now.

pwiz_tools/Skyline/Model/SrmDocument.cs
```diff
@@ -2115,12 +2115,26 @@ public double GetCollisionEnergy(PeptideDocNode nodePep, TransitionGroupDocNode
...
+            if (!ce.HasValue)
+            {
+                if (nodeGroup.Children.Select(node => ((TransitionDocNode)node).ExplicitValues.CollisionEnergy).Distinct().Count() == 1)
+                    ce = ((TransitionDocNode)nodeGroup.Children.First()).ExplicitValues.CollisionEnergy;
+            }
+            // Failing that, explicitly declared value at the precursor level is used.
+            if (!ce.HasValue)
+            {
+                ce = nodeGroup.ExplicitValues.CollisionEnergy;
+            }
```
brendanx67, line 2134: I would still prefer to see it as a function which would make this a single line double? ce = GetCE()

bspratt: Done.

pwiz_tools/Skyline/Model/SrmDocument.cs
```diff
@@ -2132,6 +2130,30 @@ public double GetCollisionEnergy(PeptideDocNode nodePep, TransitionGroupDocNode
             return ce ?? 0.0;
         }
 
+        private static double? GetExplicitCollisionEnergy(TransitionGroupDocNode nodeGroup, TransitionDocNode nodeTran)
```
brendanx67, line 2133: Can you just use returns instead of the ce = null at the top. Then just return null at the bottom?

brendanx67: Just try to make it as simple as possible now. This is a mess function, when you allow for simply returning as soon as you have a value.

bspratt: perhaps you'll prefer this next formulation.

brendanx67: I guess it wasn't as easy to simplify as I had been thinking. I took a closer look in VS and did my best to get rid of the cruft, which did end up using a double? ce = null at the start of the function and a single return. Good to go now.

brendanx67: By the way, I got rid of all the (TransitionDocNode) casting by using TransitionGroup.Transitions instead of TransitionGroup.Children, but even with the latter, you could have used Children.Cast<TransitionDocNode>() for cleaner code.

PR #1109, opened 2020-06-08
Skyline would not load chromatogram files with fewer than 3 chromatog…

bspratt:
…rams, apparently on the assumption that they were pure metadata. Now it actually looks at them to see if they are SIC chromatograms.

Reported by user egauth in https://skyline.ms/announcements/home/support/thread.view?rowId=47502

### DIFF
--- a/pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
+++ b/pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
@@ -995,11 +995,7 @@ namespace pwiz.ProteowizardWrapper
             get
             {
                 var len = ChromatogramCount;
-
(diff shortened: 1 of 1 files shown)


bspratt:
I simply removed what seems like a pretty trivial optimization - does anyone see any performance implications here?

brendanx67:
Okay. Surprised this was a problem. The title is just plain wrong about what the code did. It only skipped files with just 1 chromatogram. "len < 2" would be 1 or 0. So, this was a case where the file had 1 chromatogram that was not a TIC chromatogram. i.e. it had no TIC.

Seems super strange to me. Why didn't it have a TIC?

chambm:
Shimadzu MRM is one of the very few readers that don't provide TIC. Maybe the only one.

PR #1116, opened 2020-06-15
Workaround for 32 transition limit on Waters methods

kaipot:


brendanx67:
Have you asked Waters to have a look and verify the fix?

PR #1118, opened 2020-06-16
Skyline: fix zero vs missing values logic in ion mobility filtered chromatogram extraction

bspratt:
Fixes a problem with performing chromatogram extraction with ion mobility filtering where there was not a proper distinction between failure to detect signal because a scan simply has none in a given IM filter range (zero value) versus an IM filter range being outside the measurement range of the instrument (no value). The pwiz CLI API has been modified to provide this instrument IM range metadata when available. Currently only Bruker TDF has this, the others we derive by inspecting the first few scans.

This clears up a mystery around subtle differences in Skyline peakpicking results between 2-array and 3-array representations of IMS data, as seen in BrukerPrmPasefImportTest. A zero intensity chromatogram time point leads to slightly different RT peak placement than a missing time point does, and in the extreme can mean a red dot vs no dot chromatogram quality indicator in the Targets tree.


chambm:
I suggest changing this to a UserParam in a file-level MSData element (like `<instrumentConfiguration>`). Not only is it simpler to implement (not too relevant since you've already done the work), but the information would still be accessible in converted output files. So if someone's giving Skyline an mzML/mz5 file instead of the raw Bruker data they don't lose the ion mobility range information.

bspratt:
Fair point. I'll have a look at that.

brendanx67:
All the failures make me a little nervous. Is something broken?

pwiz_aux/msrc/utility/vendor_api/Bruker/CompassData.hpp
```diff
@@ -331,6 +331,9 @@ struct PWIZ_API_DECL CompassData
     virtual InstrumentSource getInstrumentSource() const = 0;
     virtual std::string getAcquisitionSoftware() const = 0;
     virtual std::string getAcquisitionSoftwareVersion() const = 0;
+    virtual double getOneOverK0AcqRangeLower() const = 0;
```
chambm, line 334: Add stubs in CompassData.cpp's CompassDataImpl similar to Baf2Sql.cpp.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -1644,8 +1667,7 @@ public override int GetHashCode()
         {
             unchecked
             {
-                int result = 0;
-                result = (result * 397) ^ (Model != null ? Model.GetHashCode() : 0);
+                int result = (Model != null ? Model.GetHashCode() : 0);
```
brendanx67, line 1670: This doesn't seem necessary to me. Cleaning up generated code.

chambm: You wouldn't want to use the minimum and maximum observed IMS values, but it is better (for missing vs. zero) to use the min/max possible IMS values on a spectrum by spectrum basis. Those values would be based on the scan range for the combined spectrum. We should probably put the IM range in the <scan> element for combined spectra. Brian and I probably discussed that before but (I) forgot about it.

brendanx67: Sure. That sounds good. I was just going off how I would have expected it to work when we had 2D spectra for IMS. Then, our only option would have been whether we saw any spectra in the given range. But, I agree, if it were possible to get a range from a header on the 3D spectra, that would be great.

pwiz_aux/msrc/utility/vendor_api/Bruker/CompassData.hpp
bspratt: Right, missed that one

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
bspratt: >> Cleaning up generated code.
It eliminates an inspection warning.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -1,1709 +1,1738 @@
...
+                }
+                return null;
+            }
+        }
+    }
+
+    public sealed class MsDataSpectrum
+    {
+        private IonMobilityValue _ionMobility;
+        private double? _ionMobilityMeasurementRangeLow, _ionMobilityMeasurementRangeHigh; // The range actually measured (for zero vs missing value determination)
```
brendanx67, line 1548: Define as get-set property like Id, Level, Index

pwiz_tools/Skyline/Model/Results/SpectrumFilterPair.cs
```diff
@@ -510,6 +532,8 @@ internal class IonMobilityRangeHelper
             }
         }
 
+        public bool OutsideInstrumentRange { get; private set; }
```
brendanx67, line 535: What does this do?

pwiz_tools/Skyline/TestPerf/PerfImportDiaPasefTest.cs
```diff
@@ -91,7 +91,6 @@ protected override void DoTest()
             AssertEx.IsDocumentState(doc1, null, 1, 34, 34, 204);
             loadStopwatch.Stop();
             DebugLog.Info("load time = {0}", loadStopwatch.ElapsedMilliseconds);
-
```
brendanx67, line 94: Revert

pwiz_tools/Skyline/TestPerf/PerfImportPrmPasefTest.cs
```diff
@@ -140,27 +140,23 @@ private void TestReports(SrmDocument doc1, string msg = null)
...
-                10.07, 24.29, 23.28, 15.83, 14.87, 28.53, 14.85, 28.5, 22.49, 19.35, 24.81, 24.75, 24.63, 24.36, 15.46, 22.7, 19.85, 22.56,
-                21.91, 15.97, 20.82, 13.62, 19.52, 15.55, 23.71, 30.35, 19.57, 13.57, 29.08, 15.56, 21.73, 16.05, 13.95, 25.39, 14.3, 25.3,
-                15.72, 10.48, 9.88, 21.31, 9.92, 26.56, 27.89, 29.85, 22.12, 25.06, 21.53, 15.25, 21.13, 22.81, 20.44, 18.23, 11.49, 25.59,
-                20.13, 16.32, 13.19, 9.63, 12.11, 21.84, 10.83, 12.49, 16.99, 18.83, 11.71, 12.84, 17.31, 11.13, 12.88, 24.35, 16.83, 15.26,
-                22.72, 8.45, 11.17, 18.41, 18.85, 22.18, 16.56, 10.02, 10.13, 25.91, 29.07
+                12.45, 21.48, 16.93, 22.93, 13.63, 19.12, 28.97, 14.88, 14.24, 27.25, 14.97, 14.26, 25.7, 15.06, 11.93, 26.37, 12.89, 15.88,
+                18.34, 11.16, 10.46, 10.98, 28, 24.01, 11.15, 18.97, 23.33, 26.56, 11.94, 19, 24.2, 23.42, 26.1, 27.86, 27.76, 20.99,
+                26.15, 21.16, 25.99, 14.29, 26.04, 15.04, 24.07, 28.15, 20.1, 26.03, 15.07, 19.78, 27.2, 21.26, 25.39, 25.04, 11.71, 21.1,
```
brendanx67, line 145: Fix the wrapping to be the same as old values.

bspratt:
moved this work to a new branch to work around unexplained line ending changes from a merge from trunk

PR #1121, opened 2020-06-17
Make it so that the current .sky file format is always described in a file called "Skyline_Current.xsd".

nickshulman:
This is the change that I proposed in email a few weeks ago.

This makes it so that the current document schema file is always "Skyline_Current.xsd", and that is the only file that you ever edit. That way, it will always be easy to see what changes were made with a particular checkin. Also, if two people were planning to make changes to the Skyline_Current.xsd during the same time period, everything merges easily.

Files called "Skyline_###.xsd" only get created when you need to bump the DocumentFormat version number and make a change to "Skyline_Current.xsd".
When that happens, the "Skyline_###.xsd" file that you will be adding will represent the old DocumentFormat version number.

The class "SchemaDocuments" takes care of the logic of deciding what the schema filename is for a particular version number.


bspratt:
I like the idea in principle, but how would it aid in writing code like this?
`        /// <summary>
        /// Serializes any optionally explicitly specified CE, RT and DT information to attributes only
        /// </summary>
        private void WriteExplicitTransitionGroupValuesAttributes(XmlWriter writer, ExplicitTransitionGroupValues importedAttributes)
        {
            if (DocumentFormat < DocumentFormat.VERSION_4_22 || DocumentFormat >= DocumentFormat.VERSION_20_12) // Format supports per-precursor explicit CE?
                writer.WriteAttributeNullable(ATTR.explicit_collision_energy, importedAttributes.CollisionEnergy);
            writer.WriteAttributeNullable(ATTR.explicit_ion_mobility, importedAttributes.IonMobility);
            if (importedAttributes.IonMobility.HasValue)
                writer.WriteAttribute(ATTR.explicit_ion_mobility_units, importedAttributes.IonMobilityUnits.ToString());
            writer.WriteAttributeNullable(ATTR.explicit_ccs_sqa, importedAttributes.CollisionalCrossSectionSqA);
        }`

or 
`
        public static readonly DocumentFormat BINARY_RESULTS = VERSION_3_62; // First version that had binary results`

?

nickshulman:
I'm not sure if I understand your question.

That code would stay the same.
The DocumentFormat really is still a double.
The only difference is that the name of the .xsd file that corresponds to
the current version is called "Skyline_Current.xsd" instead of
"Skyline_20.13.xsd".
-- Nick

bspratt:
When writing backward compatibility code for a new feature I need to be able to define a watershed document version for determining whether or not we need to do anything special for (de)serialization.

I guess if I always use an alias
`public static readonly DocumentFormat MY_NEW_THINGEE = DocumentFormat.CURRENT; // First version that had binary result`
and we're super careful to track down *all* uses of DocumentFormat.CURRENT before release this works fine. If we miss one, though, it would make for some very confusing behavior.

nickshulman:
You still get to add constants to the class DocumentFormat the same as you
currently do.

The situation that I'm trying to fix is where you and I have both been
working on a new feature for a month, so we both have a new file in our
branches called "Skyline_20.14.xsd". Git does not know we both created that
file by copying "Skyline_20.13.xsd". When I check in my version of
"Skyline_20.14.xsd" before you, you suddenly have to change the name of
your "Skyline_20.14.xsd" to "Skyline_20.15.xsd", and also make sure that
your Skyline_20.15.xsd includes the changes that I added to my
Skyline_20.14.xsd.
-- Nick

bspratt:
I totally get why we want this. Just wanted to make sure we weren't going to accidentally create new defect opportunities that wouldn't pop up until after a release or two. 

So part of the release process becomes
1 Compare skyline_current.xsd to whatever the latest numbered xsd might be
2 If they differ: 
  2a Create a new DocumentFormat version number YY.n
  2b Copy skyline_current.xsd to skyline_<YY.n>.xsd
  2c Add skyline_<YY.n>.xsd to the project and the repo
  2d Change any constants in class DocumentFormat using DocumentFormat.VERSION_CURRENT to use DocumentFormat.VERSION_<YY.n> instead

brendanx67:
I think Nick is saying that it doesn't change our use of constants at all. Just that there is a new XSD file, which always describes the DocumentFormat which equals VERSION_CURRENT. So, when we do our XSD check in testing, the code would look at the document version and if it is equal to VERSION_CURRENT, then it gets checked against Skyline_Current.xsd.

Then, I think you are right that we need to understand when we copy Skyline_Current.xsd to Skyline_###.xsd before making a modification. That should be any time we create a new version constant and set it to VERSION_CURRENT. As developers, we still need to keep an eye on when that is necessary, which would be any time we make a change that would be incompatible with what has already been released.

I am not sure this is much different. To me it seems you will end up working as you have in the past, but when you create VERSION_20_13, instead of creating Skyline_20.13.xsd, you would:

1. Copy Skyline_Current.xsd to Skyline_20.12.xsd and add it to the project.
2. Apply any necessary modifications to Skyline_Current.xsd and commit them.
 
In this way Skyline_Current.xsd will accumulate diffs of what is changing. Otherwise, I don't think much is changing. Do I have this right, Nick?

bspratt:
I LOVE the idea of a schema with a proper history. In fact I'd argue that
when we create that xsd file we should start with a copy of  our oldest
schema version and commit changes from version to version until we're
current, and not do a squash when it gets added to master..

As to my concerns, I'm specifically thinking about a scenario such as
moving ion mobility from Peptide Settings to Transition Settings. I need to
create a constant such that serialization code knows whether to output the
old or new style.

    if (outputDocumentFormat <
DocumentFormat.ION_MOBILITY_IN_TRANSITIONSETTINGS)
    {
        // Write IM settings as part of PeptideSettings


If I set that constant as

   DocumentFormat.ION_MOBILITY_IN_TRANSITIONSETTINGS =
DocumentFormat.VERSION_20_13;

then every time I merge from master it's likely to be creating a bug since
there's a chance that we've moved on to 20_14 since the last update. This
misses the point of NIck's work.

So I should set it as

   DocumentFormat.ION_MOBILITY_IN_TRANSITIONSETTING =
DocumentFormat.VERSION_CURRENT;

That creates a different problem if I don't remember to change it to
the proper DocumentFormat.VERSION_xx_y  right before I merge to master.
This would be a pernicious bug because it wouldn't reveal itself until we
bump the version number again and backward compatibility suddenly goes
funny.

So it would be ideal if there was some automated check on master that
insists that no
         xxx  =  DocumentFormat.CURRENT;
 assignments are allowed in master.

chambm:
Can the enum have an overridden operator= that checks whether the rhs is CURRENT and throws an exception if so? Probably only used if lhs if also the same type rather than casting to int or something.

bspratt:
Master has been updated to have a Skyline_Current.xsd file with a complete version history, and has been merged into this branch.

PR #1122, opened 2020-06-18
Skyline: support for even more .MSP formatting variants in small mole…

bspratt:
…cule spectral libraries from various sources including MoNA and MINE

Make keyword matches case insensitive
Support nonstandard chemical formula formatting
Ignore peaks with zero intensity (some examples had tens of thousands per precursor)

Resolves Issue 731: "Varying small molecule implementations of .MSP format aren't handled well by Skyline" as reported in https://skyline.ms/announcements/home/support/thread.view?rowId=46784  

reported by user Thomas Stoll


pwiz_tools/Skyline/Model/Lib/NistLibSpec.cs
```diff
@@ -829,7 +832,7 @@ private bool CreateCache(ILoadMonitor loader, IProgressStatus status, int percen
 
 
                     // Read until name line
-                    if (!line.StartsWith(NAME))
+                    if (!line.StartsWith(NAME, true, CultureInfo.InvariantCulture)) // Accept Name:, NAME:, nAmE: etc
```
nickshulman, line 835: You could also do "line.StartsWith(NAME, StringComparison.InvariantCultureIgnoreCase)" which does exactly the same thing but might be more self-documenting

pwiz_tools/Skyline/Test/LibraryLoadTest.cs
```diff
@@ -295,7 +306,7 @@ public UpdateProgressResponse UpdateProgress(IProgressStatus status)
             TEXT_LIB_YEAST_NIST4;
 
         public const string TEXT_LIB_YEAST_NIST1 =
-            "Name: QQGPLEPTVGNSTAITEER/2\n" +
+            "NaMe: QQGPLEPTVGNSTAITEER/2\n" + // This oarticular strange capitalization not actually seen in the wild, but be ready
```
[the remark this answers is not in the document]
bspratt: hah nice catch

pwiz_tools/Skyline/Model/Lib/NistLibSpec.cs
bspratt: Spiffy! Will do.

pwiz_tools/Skyline/Model/Lib/NistLibSpec.cs
```diff
@@ -937,7 +940,14 @@ private bool CreateCache(ILoadMonitor loader, IProgressStatus status, int percen
                             match = REGEX_PRECURSORMZ.Match(line);
                             if (match.Success)
                             {
-                                precursorMz = double.Parse(match.Groups[1].Value, CultureInfo.InvariantCulture);
+                                try
+                                {
+                                    precursorMz = double.Parse(match.Groups[1].Value, CultureInfo.InvariantCulture);
```
brendanx67, line 945: Maybe if (!double.TryParse()) instead

pwiz_tools/Skyline/Model/Lib/NistLibSpec.cs
```diff
@@ -975,16 +985,29 @@ private bool CreateCache(ILoadMonitor loader, IProgressStatus status, int percen
...
                     {
-                        formula = formula.Substring(1).Split(']')[0];
+                        // Watch for strange formula specification as seen in MoNA
+                        if (formula.StartsWith(@"["))
+                        {
+                            // In MoNA LipidBlast we've seen "[C11H22NO4]+" associated with adduct [M]+
+                            formula = formula.Substring(1).Split(']')[0];
+                        }
+                        // In MoNA Fiehn-HILIC we see "C7H9N2O+" associated with adduct [M]+
+                        else if (formula.Contains(@"+"))
```
brendanx67, line 1003: Don't really need the Contains here. string.Split()[0] works no matter what in this case.

pwiz_tools/Skyline/Properties/Resources.resx
```diff
@@ -10860,4 +10860,7 @@ Invalid peptide sequence at position 18
   <data name="BrukerTimsTofMethodExporter_ExportMethod_Getting_scheduling___" xml:space="preserve">
     <value>Getting scheduling...</value>
   </data>
+  <data name="NistLibraryBase_CreateCache_Could_not_read_precursor_mz_value___0__" xml:space="preserve">
+    <value>Could not read precursor mz value "{0}"</value>
```
brendanx67, line 10864: Typo: m/z, and "read the precursor m/z" (programmers too often drop articles)

pwiz_tools/Skyline/Model/Lib/NistLibSpec.cs
bspratt: It's a nullable, so that would look something like
`
                                if (double.TryParse(match.Groups[1].Value, NumberStyles.Float, CultureInfo.InvariantCulture, out var parsedPrecursorMz))
                                {
                                    precursorMz = parsedPrecursorMz;
                                }
                                else
                                {
                                    ThrowIOException(lineCount, string.Format(Resources.NistLibraryBase_CreateCache_Could_not_read_the_precursor_m_z_value___0__, line) );
                                }
                                continue;
`

pwiz_tools/Skyline/Model/Lib/NistLibSpec.cs
bspratt: Easy enough to change to 

                        // In MoNA Fiehn-HILIC we see "C7H9N2O+" associated with adduct [M]+ or "C7H9N2O-" with [M]-
                        else
                        {
                            formula = formula.Split('+', '-')[0];
                        }

PR #1124, opened 2020-06-19
Skyline/work/20200619 xsd history

bspratt:
Created Skyline_Current.xsd with a full history of previous versions. When this gets merged to trunk we want to keep all that history.


bspratt:
Yes, especially the timestamps.

I'll rebase this into master, then you can update your branch.

bspratt:
Ugh, I messed up the order. Going to try again then rebase into master. But I'll consider this the OK for that.

bspratt:
Wait, no I didn't. Not in my script anyway. Why does the log seem to show that they're out of order?

PR #1128, opened 2020-06-25
Update WiffFile.cpp to handle wiff files with missing data in last cycle of acquisition

david-cox-sciex:
On some Analyst wiff files the very last cycle is corrupted. Ideally, BasePeakChromatogramSettings would ignore this but it throws an exception in the current reader. This workaround would ignore the last cycle for BPC data and enable successful conversion of wiff to mzML.


chambm:
@david-cox-sciex This change is supposed to ALWAYS remove the last BPC cycle, whether it's corrupt or not? Can we instead catch the exception when generating the BPC and retry it without the last cycle?

PR #1138, opened 2020-07-01
Detection plots implementation

rita-gwen:


pwiz_tools/Skyline/Controls/Graphs/DetectionsToolbar.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        private Dictionary<ToolStripDropDown, ToolStripItem> _selectedItems =
+            new Dictionary<ToolStripDropDown, ToolStripItem>(4);
+
+        public override bool Visible => true;
+
+        public DetectionsToolbar(GraphSummary graphSummary) : base(graphSummary)
+        {
+            InitializeComponent();
+            _timer = new Timer { Interval = 100 };
```
nickshulman, line 25: This Timer might need to get Dispose'd.

pwiz_tools/Skyline/Controls/Graphs/DetectionPlotData.cs
```diff
@@ -0,0 +1,234 @@
...
+                        else
+                        {
+                            if (runningNaN)
+                            {
+                                mins[i] = maxes[i] = qValues[i];
+                                runningNaN = false;
+                            }
+                            else
+                            {
+                                mins[i] = (mins[i - 1] > qValues[i]) ? qValues[i] : mins[i - 1];
```
nickshulman, line 189: It is also possible to do:
mins[i] = Math.Min(mins[i-1], qValues[i]);

pwiz_tools/Skyline/Controls/Graphs/DetectionsHistogramPane.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        public override ImmutableList<int> GetDataSeries()
+        {
+            return TargetData.Histogram;
+        }
+
+
+        public override void UpdateGraph(bool selectionChanged)
+        {
+            if (!_detectionData.IsValid)
```
nickshulman, line 36: BUG: I believe this check needs to go after the line where you assign _detectionData to the value from the cache.

Otherwise, if you get yourself in a state where the detectionData is invalid, it will never fetch new data, and the graph will never update.

rita-gwen: Thank you, Nick, appreciate your comments. Good catch on _detectionData.IsValid logic, this definitely should be fixed. I didn't see the other exception you mentioned, but I agree that Dispose in not really necessary there.

pwiz_tools/Skyline/Controls/Graphs/DetectionPlotData.cs
```diff
@@ -0,0 +1,417 @@
...
+                return false;
+            }
+
+            public void Cancel()
+            {
+                new Task(CancelWorker, new object[]{null, true}).Start();
+
+            }
+
+            private void CancelWorker(object param)
```
nickshulman, line 327: It would be easier to understand this function if it took two arguments: DataRequest, bool

pwiz_tools/Skyline/Controls/Graphs/DetectionToolbarProperties.cs
```diff
@@ -0,0 +1,124 @@
...
+            Settings.TargetType = IntLabeledValue.GetValue(cmbTargetType, Settings.TargetType);
+            IntLabeledValue.PopulateCombo(cmbCountMultiple, Settings.YScaleFactor);
+        }
+
+        #region Functional test support
+
+        public void SetQValueTo(float qValue)
+        {
+            if (qValue == .01f)
+                rbQValue01.Select();
```
nickshulman, line 112: The usual way to select a radio button is to set Checked to true:
rbQValue01.Checked = true;
but, I guess there's nothing wrong with using "Select()" if it works.

pwiz_tools/Skyline/TestPerf/DetectionsPlotTest.cs
```diff
@@ -0,0 +1,215 @@
...
+        {
+            "Replicate	2_SW-B",
+            "Precursor Count	137",
+            "Cumulative Count	143",
+            "All Count	133",
+            "-log10 of Q-Value Median	6.0"
+        };
+        private static readonly string[] HISTOGRAM_TIP_TEXT =
+        {
+            "Replicate Count	5",
```
nickshulman: These strings are going to be different in different languages, so your test should make use of things like:
Skyline.Properties.Resources.DetectionHistogramPane_Tooltip_ReplicateCount

(Also, the code would probably be easier to understand if you declared the string right were you are using it in the test code)

pwiz_tools/Skyline/Controls/Graphs/DetectionPlotData.cs
```diff
@@ -0,0 +1,417 @@
...
+
+        }
+
+        public List<string> ReplicateNames { get; private set; }
+
+        public static DetectionPlotData INVALID = new DetectionPlotData(null, 0.001f);
+        public const int REPORTING_STEP = 3;
+
+        public DetectionPlotData(SrmDocument document, float qValueCutoff, 
+            CancellationToken? cancellationToken = null, [CanBeNull] Action<int> progressReport = null)
```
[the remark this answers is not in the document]
rita-gwen: Agree.

pwiz_tools/Skyline/TestPerf/DetectionsPlotTest.cs
```diff
@@ -0,0 +1,215 @@
...
+                                           && DetectionPlotData.DataCache.Status == DetectionPlotData.DetectionDataCache.CacheStatus.idle,
+                () => $"Retrieving data for qValue {qValue}, refIndex {refIndex} took too long.");
+            WaitForGraphs();
+            Assert.IsTrue(data.IsValid);
+
+            if (record)
+            {
+                Debug.WriteLine("Peptides");
+                pane.CurrentData.GetTargetData(DetectionsGraphController.TargetType.PEPTIDE).TargetsCount
+                    .ForEach((cnt) => { Debug.Write($"{cnt}, "); });
```
nickshulman, line 182: We usually use "Console.WriteLine" for things (such as record mode output) that we actually want the developer to see.

Debug.Write gets completely stripped out in the Release build (which is how we usually run tests).

pwiz_tools/Skyline/Controls/Graphs/DetectionsPlotPane.cs
```diff
@@ -0,0 +1,429 @@
...
+            }
+
+            //used for testing only
+            public List<string> TipLines
+            {
+                get
+                {
+                    return _table.Select((rowDesc) =>
+                    {
+                        return Enumerable.Aggregate(
```
nickshulman, line 103: It's probably simpler to use "string.Join":
                    return _table.Select(rowDesc=>
                        string.Join(TextUtil.SEPARATOR_TSV_STR, rowDesc.Select(cell=>cell.Text))
                    ).ToList();

pwiz_tools/Skyline/TestPerf/DetectionsPlotTest.cs
```diff
@@ -0,0 +1,215 @@
...
+            //display and hide tooltip
+            evt = FindBarPoint(paneHistogram, graphHistogram, 5);
+            Trace.WriteLine(this.GetType().Name + ": showing histogram tooltip.");
+            RunUI(() =>
+            {
+                paneHistogram.HandleMouseMoveEvent(graphHistogram.GraphControl, evt);
+                Assert.IsNotNull(paneHistogram.ToolTip, "No tooltip found.");
+                WaitForConditionUI(() => paneHistogram.ToolTip.IsVisible, "Tooltip is not visible.");
+                //verify the tooltip text
+                Assert.IsTrue(HISTOGRAM_TIP_TEXT.SequenceEqual(paneHistogram.ToolTip.TipLines));
```
nickshulman, line 154: You might want to use:
CollectionAssert.AreEqual

pwiz_tools/Skyline/Controls/Graphs/DetectionPlotData.cs
```diff
@@ -0,0 +1,234 @@
...
+
+            _data[DetectionsGraphController.TargetType.PRECURSOR] = new DataSet(_precursorData, ReplicateCount, QValueCutoff);
+            _data[DetectionsGraphController.TargetType.PEPTIDE] = new DataSet(_peptideData, ReplicateCount, QValueCutoff);
+
+            _isValid = true;
+        }
+
+        public void Dispose()
+        {
+            _precursorData.Clear();
```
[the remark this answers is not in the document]
rita-gwen: Yes, I removed the IDisposable interface.

pwiz_tools/Skyline/Controls/Graphs/DetectionsToolbar.cs
rita-gwen: I do not see any memory leaks, but I guess it's better to be on the safe side.

pwiz_tools/Skyline/Controls/Graphs/DetectionPlotData.cs
rita-gwen: Agree. I didn't realize I can use lambda in the Task constructor.

pwiz_tools/Skyline/TestPerf/DetectionsPlotTest.cs
rita-gwen: Right. I replaced the strings with resource reference. Not sure about the latter, I think it's better to keep bulky static definitions outside of the main code, unless it carries a performance penalty

rita-gwen:
So, the detection plots are ready to merge. Should I go ahead with it? Code review is done, tests are stable, all checks are green.

pwiz_tools/Skyline/TestPerf/DetectionsPlotTest.cs
```diff
@@ -0,0 +1,228 @@
...
+            Resources.DetectionHistogramPane_Tooltip_ReplicateCount + TextUtil.SEPARATOR_TSV_STR + 
+            5.ToString( CultureInfo.CurrentCulture),
+            String.Format(Resources.DetectionHistogramPane_Tooltip_Count, DetectionsGraphController.TargetType.PRECURSOR) + 
+            TextUtil.SEPARATOR_TSV_STR + 119.ToString( CultureInfo.CurrentCulture),
+        };
+
+        [TestMethod]
+        public void TestDetectionsPlot()
+        {
+            RunPerfTests = true;
```
nickshulman, line 53: Make sure you do not check this line in.

pwiz_tools/Skyline/TestPerf/DetectionsPlotTest.cs
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        private static MouseEventArgs FindBarPoint(GraphPane pane, Control graph, int index)
+        {
+            using (var g = graph.CreateGraphics())
+            {
+                var iterationCount = 0;
+                var randomInt = new Random();
+                var evt = new MouseEventArgs(MouseButtons.None, 0, 0, 0, 0);
+                while (!(pane.FindNearestObject(evt.Location, g, out var nearestObject, out var objectIndex) &&
```
nickshulman, line 216: When I run this test under TestRunner (which sets Program.SkylineOffscreen to true), I always get stuck on this line.

brendanx67:
Please be sure to run the test in all languages and off-screen (as Nick suggests) using SkylineTester.

pwiz_tools/Skyline/TestPerf/DetectionsPlotTest.cs
rita-gwen: Thanks, I commented it out.

pwiz_tools/Skyline/TestPerf/DetectionsPlotTest.cs
rita-gwen: This is very strange. I can the test both on and off screen without any problems. May be I should explicitly seed the random number generator to make it reproducible.

rita-gwen: Does it happen only in offscreen mode?

rita-gwen:
I did run it in SkylineTester in both on and off-screen modes and it runs fine. Good point about the languages, though.

nickshulman:
The hang in DetectionsPlotTest.FindBarPoint does not involve the random number generator.
The problem is that the pane._chartRect is an empty rectangle (0,0,0,0), so there are not coordinates that would correspond to the bar graph that you are trying to find.
I am not sure why the rectangle is empty (since it's offscreen, I can't see what the graph looks like).

I would recommend that you not try simulating a mouse move event in this way. It looks like you're trying to have your test make sure that tooltips work. Some things are too difficult to write an automated test for. You could add a helper method that returns to you what text the tooltip would be, and you could verify that text, and not worry about whether the tooltip displaying code might break.

rita-gwen:
OK. I wonder why it works on my computer, though. I can also check for the off-screen flag and skip the show tooltip part if it's on.

rita-gwen:
Does it work on your machine in on-screen mode?

nickshulman:
One thing that I am seeing is that the StackWorker in DetectionDataCache prevents the Skyline process from terminating.
You need to call Dispose on the StackWorker so that it destroys its threads.
One way to get it disposed would be to change it so that DetectionPlotData.DataCache is not a static variable, but instead gets created and owned by the DetectionsPlotPane that needs it.
Or, if you want there to be only one cache, then you should make it so that a DetectionPlotPane increments and decrements the DataCache's reference count as appropriate and it disposes itself when its reference count gets to zero.

brendanx67:
We have done this numerous times in our code. We need to stop reinventing and try to follow patterns already in use.

pwiz_tools/Skyline/Controls/Graphs/DetectionPlotData.cs
```diff
@@ -0,0 +1,416 @@
+﻿using pwiz.Skyline.Model;
```
nickshulman, line 1: You should add the standard file header comment to this file. Make sure to change the Original author to your own name.

rita-gwen:
I released the cache worker in SkylineWindow.OnClosed() method. It looks like an appropriate place to do so. It is already used to stop half a dozen of assorted background workers.

pwiz_tools/Skyline/Controls/Graphs/DetectionPlotData.cs
```diff
@@ -0,0 +1,435 @@
...
+        public static DetectionDataCache GetDataCache()
+        {
+            if (_dataCache == null)
+                _dataCache = new DetectionDataCache();
+            return _dataCache;
+        }
+
+        public static void ReleaseDataCache()
+        {
+            _dataCache.Dispose();
```
nickshulman, line 60: Make sure you check if _dataCache is null

PR #1156, opened 2020-07-24
SkylineRunner: Increase wait timeout for connecting to Skyline

vagisha:
Problem reported by Sonnet Davis.  Skyline was installed in the expected location but SkylineRunner (via AutoQC Loader) would exit with the following error:
Error: Could not connect to Skyline.
Make sure you have a valid Skyline installation.
SkylineRunner sometimes takes longer than 5s to connect to the named pipe in WaitForConnection().  Increasing the timeout fixed the problem.

### DIFF
--- a/pwiz_tools/Skyline/Executables/SkylineRunner/Program.cs
+++ b/pwiz_tools/Skyline/Executables/SkylineRunner/Program.cs
@@ -127,14 +127,13 @@ namespace pwiz.SkylineRunner
             }
         }
 
-
         private bool WaitForConnection(NamedPipeServerStream serverStream, string inPipeName)
         {
             Thread connector = new Thread(() =>
             {
                 serverStream.WaitForConnection();
                 lock (SERVER_CONNECTION_LOCK)
-                {         
+                {
                     _connected = true;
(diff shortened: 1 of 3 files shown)


pwiz_tools/Skyline/Executables/SkylineRunner/Program.cs
```diff
@@ -127,13 +127,14 @@ private int Run(IEnumerable<string> args)
             }
         }
 
+
```
brendanx67, line 130: Remove

pwiz_tools/Skyline/Executables/SkylineRunner/Program.cs
```diff
@@ -127,13 +127,14 @@ private int Run(IEnumerable<string> args)
...
 
+
         private bool WaitForConnection(NamedPipeServerStream serverStream, string inPipeName)
         {
             Thread connector = new Thread(() =>
             {
                 serverStream.WaitForConnection();
                 lock (SERVER_CONNECTION_LOCK)
-                {
+                {
```
brendanx67, line 137: Revert

PR #1165, opened 2020-07-31
Qualitative Ion Ratio

nickshulman:
This implements the Qualitative Ion Ratio as requested by some scientists at NCSU.

The Qualitative Ion Ratio is the ratio of the sum of the non-quantitative transition peak areas to the sum of the quantitative transition peak areas.
The Target Qualitative Ion Ratio is the average Ion Ratio across all of the external standards that have not been excluded from calibration.
The Qualitative Ion Ratio Threshold is a number that the user can define on the Peptide Settings Quantification tab which is used to determine whether the Qualitative Ion Ratio Status is a Pass or Fail.
image
image
image
image
image

There were other things that were requested that I do not think we should do yet:
1. Ability to use something other than "Quantitative" to indicate which transitions go in the numerator vs. denominator for Ion Ratio
2. Ability to use something other than "Exclude from Calibration" to indicate that a replicate should not be used in the averaging for Target Ion Ratio
3. Ability to specify asymmetric range around the Target Ion Ratio so that things are allowed to be a larger percent higher than the target vs. lower, etc.

If there is enough demand we could add these features in the future.


brendanx67:
I feel like the naming is too vague. Given that both precursors and products are ions, it just is not clear which ion ratios you are talking about. Especially when you are normalizing light to heave, which should be relatively common, that would seem to be the most obvious ion ration (light:heavy). Maybe it should be "Product Ion Ratio"? Though even looking at this I am not sure if it is Qualitative to Quantitative or vice versa. Maybe it should be "Qualitative Ion Ratio" (if Qual:Quan) or "Ratio To Qualitative Ions" (if Quan:Qual).

nickshulman:
I think we should call it "Qualitative Ion Ratio".

Should the other columns be called "Target Qualitative Ion Ratio" and "Qualitative Ion Ratio Status"?
Or, should they remain "Target Ion Ratio" and "Ion Ratio Status"?

brendanx67:
I think the complete name change makes most sense to tie them all together in an intuitive way. Thanks, Nick.

brendanx67:
Seems strange that they show up at the top of the Quantification element in the report editor. Clearly the other 3 elements are more important and they are not sorted alphabetically. You should put them under Normalized Area, Calculated Concentration, and Accuracy.

brendanx67:
Otherwise, it is looking really good.

brendanx67:
Though your choices of mnemonics in the Quantification tab are lousy. You generally want to use very narrow characters (lowercase i and l) and characters with descenders (lowercase g, y, p, q) as a last resort. It looks like you have now used lowercase L twice. "MS &level" and "Qua&litative ion ratio threshold:" Can you change "threshold" to either min or max to be more specific about what kind of threshold it is? I think "Min qualitative ion ratio:". This makes it more consistent with the other thresholds in this form and shortens the text a bit.

Mnemonics should be: &R, N, w, &M, U, b, V, C, &a (because in "Max qu&alitative ion ratio:" the q has a descender)

I put ampersands by the ones that should change.

nickshulman:
The Qualitative Ion Ratio Threshold is the amount (as a percentage) that the observed Qualitative Ion Ratio is allowed to be different than the Target Qualitative Ion Ratio.

So, if the Target Qualitative Ion Ratio is 2, and the Qualitative Ion Ratio Threshold is 20, then the Qualitative Ion Ratio for a particular replicate needs to be somewhere between 1.6 and 2.4 in order to be considered a "pass".
"Threshold" was the word that Jeffrey Enders used. We could maybe call it "Tolerance" if we wanted, but "Min" or "Max" would not be right.
(Chris Shuford also requested that we allow separate upper and lower tolerances, but I was planning on doing that in the future).

The reason that the columns are appearing at the top in the Report Editor has to do with the fact that they're defined on the class "PeptideQuantificationResult", which is a subclass of "QuantificationResult" which is where the rest of the properties live. The Live Reports code always puts subclass properties above base class properties (a side effect of how it traverses the class hierarchy). It will be straightforward to fix the order in this case.

brendanx67:
Well, just checking, but isn't "Max LOQ bias" also a +/- value? i.e. accuracy is allowed to vary within +/- the value specified? Though, I guess it is challenging to come up with words that are concise enough. "Max qualitative ion ratio deviation"

Okay, let's just stay with what you have. Maybe something more general using attributes for fixing the ordering. I would like the fix to only change this one case rather than having to worry about it changing other cases. Maybe just add an attribute that causes what you described for a subclass?

PR #1167, opened 2020-08-02
Fix tooltips on Peptide Settings dialog.

nickshulman:
The ToolTip class gets confused if any of the controls with tooltips are no longer part of the form. The ToolTip class figures out what the top level form is by iterating through a Hashtable and taking the first Control.TopLevelControl property. If the first control in the HashTable is no longer attached to the Form, then TopLevelControl ends up being a bogus value (usually the TabPage that got removed from its parent), and the tooltips do not work in the dialog.

This is particularly likely to happen when the UIMode is Small Molecules, since several pages get removed from the TabControl.

This fix adds a method "FormUtil.RemoveTabPage" which removes the tab page and nulls out all the tooltips for the descenents.
I fixed all of the places in the code where we remove tab pages.
I found that RefineDlg was already working around this bug, so I the workaround now that it is using FormUtil.RemoveTabPage.


bspratt:
>> FormEx does not have access to the list of components
IIRC the code that does the "peptide" -> "molecule" text replacement crawls the form looking for tooltips

nickshulman:
Thanks for the tip, Brian.
I see that PeptideToMoleculeTextMapper.TranslateForm uses reflection to enumerate the fields in the form class and finds the ToolTip objects that way.
I will look into whether I can leverage that.

pwiz_tools/Skyline/EditUI/RefineDlg.cs
```diff
@@ -85,25 +85,12 @@ public RefineDlg(IDocumentUIContainer documentContainer)
...
-
-                tabControl1.TabPages.Remove(tabResults);
-
-                helpTip.RemoveAll();
-                foreach (Control control in tabControl1.TabPages[0].Controls)
-                {
-                    helpTip.SetToolTip(control, listTips[0]);
-                    listTips.RemoveAt(0);
-                }
+                FormUtil.RemoveTabPage(tabResults, helpTip);
```
bspratt, line 88: Always a good sign when a comment with "not sure why" in it goes away

pwiz_tools/Skyline/TestFunctional/PeptideSettingsTooltipTest.cs
```diff
@@ -0,0 +1,76 @@
...
+        public void TestPeptideSettingsTooltip()
+        {
+            RunFunctionalTest();
+        }
+
+        protected override void DoTest()
+        {
+            RunUI(()=>SkylineWindow.SetUIMode(SrmDocument.DOCUMENT_TYPE.small_molecules));
+            var fieldComponents = typeof(PeptideSettingsUI).GetField("components", BindingFlags.NonPublic | BindingFlags.Instance);
+            Assert.IsNotNull(fieldComponents);
```
bspratt, line 48: Moving forward it would be nice to use AssertEx in tests instead of Assert, as it allows one to set a breakpoint rather than just bailing out.

bspratt:
Seems good. Only change I'd request is that AssertEx thing, which is obviously minor, but helps set precedent for other work.

PR #1169, opened 2020-08-04
Skyline: completing yesterday's fix to ion-mobility-filtered 2D Full …

bspratt:
…Scan graphs - the previous fix restored the ability to view the scan but did not actually apply ion mobility filtering to the view.
image
image

### DIFF
--- a/pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
+++ b/pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
@@ -47,7 +47,7 @@ namespace pwiz.Skyline.Controls.Graphs
         private HeatMapData _heatMapData;
         private double _maxMz;
         private double _maxIntensity;
-        private double _maxIonMobility; // Max ion mobility across the entire file, for scaling purposes
+        private double _maxIonMobility;
         private bool _zoomXAxis;
         private bool _zoomYAxis;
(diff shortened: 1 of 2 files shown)


pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -537,16 +544,17 @@ private static double SumIntensities(MsDataSpectrum[] spectra, double mz, int[]
 
         private void GetMaxMzIntensity(out double maxMz, out double maxIntensity)
         {
-            var fullScans = _msDataFileScanHelper.GetFilteredScans();
+            var fullScans = _msDataFileScanHelper.GetFilteredScans(out _, out _);
```
nickshulman, line 547: Does this code want the scans filtered by ion mobility, or not?
If it wants the unfiltered scans, then shouldn't it be using "MsDataSpectra" directly, instead of calling "GetFilteredScans"?
If it wants the filtered scans, then doesn't it need to do something with the two out parameters that it's ignoring here?

The way the code is now, I think the behavior is going to end up being different depending on whether the data is stored as 3 arrays or not.

bspratt: At this point it just wants to know the mz and intensity range of scans have data that passes within the IM filter, but doesn't need to know the IM range.

PR #1171, opened 2020-08-05
Bruker timsTOF use IM window width from settings

kaipot:
Fix for: https://skyline.ms/announcements/home/support/thread.view?rowId=47966


pwiz_tools/Skyline/Model/Export.cs
```diff
@@ -3185,6 +3185,8 @@ protected override string InstrumentType
                 var result = Document.Settings.GetIonMobilityFilter(nodePep, nodeTranGroup, nodeTran, null, null, 1.2);
                 if (result.HasIonMobilityValue)
                     ionMobility = result.IonMobility.Mobility.Value;
+                if (result.IonMobilityExtractionWindowWidth.HasValue)
```
bspratt, line 3188: Assuming that we want the same window size used in chromatogram extraction ion mobility filtering, this seems correct. I don't know enough about this code are to say if that's a valid assumption, though,

kaipot: After looking at this some more, I guess we should actually be using:
windowIM = Document.Settings.TransitionSettings.IonMobilityFiltering.FilterWindowWidthCalculator.WidthAt
Since the user mentioned Ion Mobility under Transition Settings.
I'm not sure what value we should use for the ion mobility max though - does it make sense to use the 1/K0 upper limit defined in the method template?

bspratt: That ion mobility max value is only used when the window type is "Linear", which would be unusual for Bruker data. So  the 1/K0 upper limit defined in the method template is a reasonable value that likely won't get used anyway.

PR #1174, opened 2020-08-06
Automatic iRT standard

kaipot:


kaipot:
> Okay to commit once it is passing tests, but then you also need to add new test code to cover the new functionality. If we weren't past feature complete, I would request tests before committing.

I didn't have any search results containing CiRT and multiple standards, but now that I think about it I can build blibs from multiple sources and then use those blibs as test inputs.

kaipot:
> Looks good to me.
> One thing that would be nice would be to hide "Automatic" from the list of options available on the EditIrtCalcDialog, since that option only does anything when you are building a library. (But I understand that might be tricky).

I played around with this a bit, since it isn't a relevant option there or building a Prosit library. Since it uses the SettingsListComboDriver, I had to remove the item from the list when the form is constructed and re-add it when the form closes. This worked but I suspect it might lead to bugs where the option doesn't get re-added.

brendanx67:
Maybe, those classes should be limited to an interface so that you can easily introduce a wrapper class that exposes the interface, rather than using the base class.

PR #1178, opened 2020-08-09
Stop trying to match small molecule transitions between light and heavy unless they have been assigned names

nickshulman:
Back in January, Will Thompson had a small molecule document where the light precursor had 4 transitions (2 of which were non-quantitative) and the heavy precursor had 2 transitions.
https://skyline.ms/announcements/home/support/thread.view?rowId=43696
image

There were a couple of things that were going weird in this scenario:
1. If you delete the first or second transition from the first precursor, the corresponding transition would get deleted from the second precursor.
2. When doing the light to heavy ratio, only the second transition in each precursor participates. This is because the first transition in light precursor is non-quantitative, and so it's partner is the first transition in the heavy precursor.

The way that I propose fixing this is that if any of the small molecule transitions under a precursor has not be given a name, then we no longer try to do heavy/light matching. Instead, all of the quantitative transitions participate in the ratio, and the ratio is just sum of light transition areas over sum of heavy transition areas. Also, you do not get a "Ratio dot product", since Skyline never tries to match up light and heavy transitions unless they all have names.

I added a method "TransitionDocNode.CanBeMatchedAcrossPrecursors" which indicates whether it is a nameless small molecule transition. If it is nameless, then the TransitionLossKey and TransitionKey end up being something which will only be equal to itself. In my implementation, I made it so that for nameless small molecule transitions, the CustomIonEquivalenceTestValue ends up being an IdentityPath with the Transition in it. I did this is because the TranstionKey is also used to figure out which of the items in a "Pick Children" list are already in the document, so I needed something in the TransitionKey that could be matched back to something in the document, and I wanted to stop using the ("_mzSortIndex_" + parent.Children.IndexOf(transition)) that we had been using.

I made it so that unnamed transitions never show up if you do a search for mismatched transitions.

Maybe I should have solved this differently. Maybe I should have made it so that Skyline does try to match up heavy and light transitions by position so long as the light and heavy precursors have exactly the same number of transitions, and exactly the same set of transitions have been marked as non-quantitative. In this way. What do you think? Should we require that users give all of their transitions a name if they want to get a "ratio dot product", or should it be sufficient for the light and heavy precursors have the same number of transitions?


bspratt:
It's often the case in small molecules that there isn't enough information to assign names to fragments, so relying on their availability too heavily isn't a good thing. Maybe there's no good answer but I guess I'd want to loop Will in on this.

nickshulman:
This looks like it's going to be hard to implement correctly. I will try again next version.

PR #1180, opened 2020-08-10
Skyline/work/20200807 use blib ref spectra proteins

bspratt:
Preserve protein names (or Molecule List names for small molecules) when exporting .blib spectral libraries, and use them in Library Explorer's "Add to Document" logic when they are found in imported .blib files.

When .blib files contain this grouping information for small molecules, it is used during Library Explorer's "Add to Document" to create named Molecule Lists instead of a single "Library Molecules" list.

N.B. currently the following is currently explicitly disabled for peptides: [For peptide spectra, if the library provides a protein name for a peptide spectrum it will be used by Add to Document only if "Associate Proteins" is not in use, or if no associated document protein can be found. When it is used, an peptide group with the protein's name will be created and used instead of the "Library Peptides" group.]


pwiz_tools/Skyline/Model/Lib/BlibData/BlibDb.cs
```diff
@@ -136,6 +136,90 @@ public int GetSpectraCount()
...
+                {
+                    cmd.CommandText = @"DROP TABLE IF EXISTS RefSpectraProteins";
+                    cmd.ExecuteNonQuery();
+                }
+
+                if (_namedProteinIds.Any())
+                {
+                    using (var cmd = _session.Connection.CreateCommand())
+                    {
+                        cmd.CommandText = @"CREATE TABLE Proteins (id INTEGER not null, accession TEXT)";
```
nickshulman, line 183: Make sure you say that "id" is the primary key.
I recommend:
CREATE TABLE Proteins (id INTEGER primary key autoincrement, accession TEXT)

pwiz_tools/Skyline/Model/Lib/BiblioSpec.cs
```diff
@@ -154,12 +159,18 @@ public override void WriteXml(XmlWriter writer)
                 writer.WriteAttribute(ATTR.score, Score.Value);
                 writer.WriteAttribute(ATTR.score_type, ScoreType);
             }
+            if (!string.IsNullOrEmpty(Protein))
```
nickshulman, line 162: You can use "WriteAttributeIfString" here.

pwiz_tools/Skyline/Model/Lib/BlibData/BlibDb.cs
```diff
@@ -136,6 +136,90 @@ public int GetSpectraCount()
...
+                    _peptideIdProteinId.Add((int)(refSpectra.Id ?? 0), proteinTableId);
+                }
+            }
+
+            public void Write()
+            {
+                // Output the protein - peptide relationships
+                using (var cmd = _session.Connection.CreateCommand())
+                {
+                    cmd.CommandText = @"DROP TABLE IF EXISTS Proteins";
```
nickshulman, line 170: It does not seem right to drop a table like this. I imagine it does not matter because this only ever gets called on blank databases. But, it seems like the right thing to do would be if the table already exists, just assume it has the right schema, and append your new rows to it.

pwiz_tools/Skyline/Model/Lib/Library.cs
```diff
@@ -1698,6 +1699,8 @@ public virtual float GetRankValue(PeptideRankId rankId)
         }
 
         public abstract IEnumerable<KeyValuePair<PeptideRankId, string>> RankValues { get; }
+        public abstract string Protein { get; } // Some .blib files provide a protein accession (or Molecule List Name for small molecules)
```
nickshulman, line 1702: I would probably have made this virtual and have the base class implementation return null, so that BiblioSpecLiteLibrary would be the only class that needed to override it.

pwiz_tools/Skyline/Model/Lib/BiblioSpecLite.cs
```diff
@@ -823,6 +826,17 @@ private bool CreateCache(ILoadMonitor loader, IProgressStatus status, int percen
                         outStream.Write(BitConverter.GetBytes(info.Copies), 0, sizeof (int));
                         outStream.Write(BitConverter.GetBytes(info.NumPeaks), 0, sizeof (int));
                         outStream.Write(BitConverter.GetBytes(info.Id), 0, sizeof (int));
+
+                        // Optional protein name or molecule list name
+                        var proteinOrMoleculeList = info.Protein ?? string.Empty;
+                        int len = proteinOrMoleculeList.Length;
```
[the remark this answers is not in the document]
bspratt: Right, thanks.

pwiz_tools/Skyline/Model/Lib/BiblioSpecLite.cs
```diff
@@ -897,6 +911,33 @@ private bool CreateCache(ILoadMonitor loader, IProgressStatus status, int percen
...
+                        @"SELECT RefSpectraId, accession FROM [Proteins] as t INNER JOIN [RefSpectraProteins] as s ON t.[id] = s.[ProteinId]";
+                    using (var dataReader = cmd.ExecuteReader())
+                    {
+                        while (dataReader.Read())
+                        {
+                            var refSpectraId = dataReader.GetInt32(0);
+                            var accession = dataReader.GetString(1);
+                            if (!string.IsNullOrEmpty(accession) && proteinsBySpectraID.ContainsKey(refSpectraId))
+                            {
+                                proteinsBySpectraID.Add(refSpectraId, accession);
```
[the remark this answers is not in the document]
bspratt: Right, again, thanks.

pwiz_tools/Skyline/Model/Lib/BlibData/BlibDb.cs
bspratt: Yup.

pwiz_tools/Skyline/Model/Lib/BlibData/BlibDb.cs
bspratt: I suppose you're right. I'll have to inspect for available IDs.

pwiz_tools/Skyline/Model/Lib/BiblioSpec.cs
bspratt: Spiffy.

pwiz_tools/Skyline/Model/Lib/Library.cs
bspratt: That has merit - less code, fewer changes - but I like the idea that if a new derived class is made the author has to at least think about this.

pwiz_tools/Skyline/SettingsUI/ViewLibraryPepInfo.cs
```diff
@@ -34,13 +34,15 @@ namespace pwiz.Skyline.SettingsUI
...
-        public ViewLibraryPepInfo(LibKey key)
+        public ViewLibraryPepInfo(LibKey key, SpectrumHeaderInfo libInfo)
         {
             Key = key;
+            LibInfo = libInfo;
             UnmodifiedTargetText = GetUnmodifiedTargetText(key.LibraryKey);
         }
 
         public LibKey Key { get; private set; }
+        public SpectrumHeaderInfo LibInfo { get; private set; } //
```
brendanx67, line 45: Add a comment or remove the slashes

pwiz_tools/Skyline/SettingsUI/ViewLibraryPepMatching.cs
```diff
@@ -732,63 +737,76 @@ public SrmDocument AddPeptides(SrmDocument document, ILongWaitBroker broker, Ide
...
             int processedCount = 0;
             int totalMatches = listMatches.Count;
+            selectedPath = null;
 
-            var listPeptides = new List<PeptideDocNode>();
-            var hasSmallMolecules = false;
-            foreach (var match in listMatches)
+            // Some .blib files provide protein accessions (understood as Molecule List Names fro small molecules).
+            // If those are provided we will use them as node names.
+            // TODO(bspratt) for now we will do this only for small molecules since it might be a surprise to proteomics users. We should revisit that decision.
```
brendanx67, line 744: Maybe CONSIDER. You shouldn't use TODO for something you don't feel confident you will address for the next release. We should talk about this as a policy decision and run a pass where people either address all TODOs or convert the ones they are not going to address into CONSIDER.

pwiz_tools/Skyline/TestFunctional/LibraryBuildTest.cs
```diff
@@ -110,7 +110,7 @@ private void MainTest()
                 Assert.AreEqual(z, key.Charge);
                 Assert.AreEqual(adduct, key.Adduct);
                 Assert.AreEqual(caffeineInChiKey, key.Target.ToString());
-                var viewLibPepInfo = new ViewLibraryPepInfo(key);
+                var viewLibPepInfo = new ViewLibraryPepInfo(key, null);
```
brendanx67, line 113: Feels like this could have a default null value to remove the need for this explicit "null" use.

brendanx67 (approved):
Looks good. Some minor style suggestions.

pwiz_tools/Skyline/Model/Lib/BiblioSpec.cs
```diff
@@ -84,17 +84,20 @@ public static BiblioSpecLibSpec Deserialize(XmlReader reader)
...
             SpectrumCount = spectrumCount;
             Score = score;
             ScoreType = scoreType;
+            _protein = protein;
         }
 
         public int SpectrumCount { get; private set; }
         public double? Score { get; private set; }
         public string ScoreType { get; private set; }
+        private string _protein;
```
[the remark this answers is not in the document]
bspratt: It's this way because the parent class declares Protein as read only: 
    public abstract string Protein { get; }
so private set isn't implied, and in fact isn't allowed.

So, we have to use a backing variable to be able to set the value during XML deserialization.

pwiz_tools/Skyline/Model/Lib/ChromLib/ChromLibSpectrumHeaderInfo.cs
```diff
@@ -54,6 +54,9 @@ public override float GetRankValue(PeptideRankId rankId)
                     };
             }
         }
+
+        public override string Protein => null; // Only .blib provides protein accessions / molecule list names
```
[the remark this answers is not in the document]
bspratt: That's not information that Skyline currently takes any notice of. And the only test example .clib that I've found doesn't populate any of that information.

I'll change this comment to
// CONSIDER(bspratt) .clib files may (or may not) contain protein/peptide associations but Skyline currently takes no notice of this.

pwiz_tools/Skyline/Model/Lib/BiblioSpec.cs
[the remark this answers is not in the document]
brendanx67: Maybe a comment, then. Strange, since interfaces do allow { get; private set; } on a property with only { get; } in the interface. You would think an abstract would act like an interface.

pwiz_tools/Skyline/TestFunctional/LibraryBuildTest.cs
bspratt: Agreed, thanks.

pwiz_tools/Skyline/SettingsUI/ViewLibraryPepMatching.cs
bspratt: Agreed, thanks.

pwiz_tools/Skyline/SettingsUI/ViewLibraryPepInfo.cs
bspratt: Agreed, thanks.

pwiz_tools/Skyline/Model/Lib/BiblioSpec.cs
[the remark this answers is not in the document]
bspratt: Will do, thanks.

PR #1183, opened 2020-08-13
Performance improvements with large numbers of peptides

nickshulman:
Here are some mundane performance tweaks which speed up working with "phl004_canonical_s64.sky".

1. Use TreeNodeCollection.AddRange when expanding nodes in the Sequence Tree
This reduces by 50% the time it takes to expand a protein that has 20,000 peptides.
(Edit > Collapse All is still pretty slow-- haven't figured out a way to speed up TreeNodeCollection.Clear)

2. Avoid calling IEnumerable.ToArray() in SrmSettings.GetExplicitPeakBounds
We call that in order to get the peak's q-value if present, but most libraries have no explicit peak boundaries, and never look at the peptide modified sequences. Speeds up PeptideDocNode.ChangeSettings by 10%.

3. Reduce number of calls to IList.Count when iterating over elements in a list in things like "ArrayUtil.ReferencesEqual", etc.
4. Reduce places where ChromInfoList has to call ImmutableList.Singleton and immediately throw it away (e.g. in "Count")
These last 2 things reduce the time to open "phl004_canonical_s64.sky" by one or two percent


pwiz_tools/Skyline/Controls/SeqNode/SrmTreeNode.cs
```diff
@@ -752,6 +752,8 @@ public void Pick(IEnumerable<DocNode> chosen, bool autoManageChildren, bool sync
             while (i < treeNodes.Count && treeNodes[i] is TNode);
 
 
+            int firstInsertPosition = i;
+            List<TreeNode> nodesToInsert = new List<TreeNode>(docNodes.Count - firstInsertPosition);
```
brendanx67, line 756: Why not var?

PR #1186, opened 2020-08-15
Update Agilent export format

kaipot:
from: https://skyline.ms/announcements/home/support/thread.view?rowId=47798

Removed "Ion Name" and "Library Rank".
Added "Polarity" for all export types; "Threshold", "Trigger Entrance Delay (cycles)", "Trigger Delay (cycles)", "Trigger Window", "IsLogicEnabled", "Trigger Logic Flag", "Trigger Ratio", "Trigger Ratio Window", "Ignore MRM" for non-standard export.
Replaced "TRUE" and "FALSE" values with 1 and 0 respectively.


kaipot:
Not sure if there is an Agilent template somewhere I can use for testing

nickshulman (approved):

kaipot:
Waiting to hear back from Agilent before merging.

PR #1189, opened 2020-08-18
- fixed BiblioSpec files created from MaxQuant to include file extension

chambm:
- fixed BiblioSpec files created from MaxQuant without the -E option to include the file extension

@vagisha I assume file extension is ok, you don't need absolute path?


vagisha:
Thanks for doing this, Matt! For Panorama Public just the filename with extension is enough.  However, it may be easier for a user to find the files if they know the full path.  What do you think?

chambm:
I think we're safe with just `filename.ext`. Thanks!

PR #1190, opened 2020-08-18
* fixed publish category for enzymes.xml and Instruments.xml

chambm:
* fixed publish category for enzymes.xml and Instruments.xml (defaulted to Data File, overrode them to Include)
* fixed missing MSAmanda dependencies giving NRE instead of sensible error message
* removed Method\Bruker\BDal.Submethod.Wrapper.xml from Skyline project since it's an XMLDOC file for the corresponding assembly

### DIFF
--- a/pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs
@@ -39,6 +39,11 @@ namespace pwiz.SkylineTestFunctional
         public void TestDdaSearch()
         {
             TestFilesZip = @"TestFunctional\DdaSearchTest.zip";
+
+            // test error parsing an MSAmanda installed file (enzymes.xml)
+            //var skylineDir = System.IO.Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location);
+            //File.Move(System.IO.Path.Combine(skylineDir, "enzymes.xml"), System.IO.Path.Combine(skylineDir, "not-the-enzymes-you-are-looking-for.xml"));
+
             RunFunctionalTest();
         }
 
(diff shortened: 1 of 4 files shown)


pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs
```diff
@@ -39,6 +39,11 @@ public class DdaSearchTest : AbstractFunctionalTest
         public void TestDdaSearch()
         {
             TestFilesZip = @"TestFunctional\DdaSearchTest.zip";
+
```
brendanx67, line 42: Do we need this commented out code?

chambm: I left it in intentionally. I was trying to add an automatic test to check that it throws the correct exception but that turned out to be more difficult than I thought and not worth the effort. But being able to cause that kind of exception manually seems useful and relevant to that test.

nickshulman:
We also need to add 
FHOOE_IMP.MSAmanda.Core.dll
FHOOE_IMP.MSAmanda.InOutput.dll
FHOOE_IMP.MSAmanda.Utils.dll
enzymes.xml 
Instruments.xml
to:
pwiz_tools\Skyline\Executables\Installer\SkylineInstaller\Product.wxs

in order for the admin installer to work.

chambm:
I'm going to add a check (maybe a GitHub action) that validates Skyline.csproj such that any Content/None that isn't dll or exe should have a corresponding PublishFile element to override the publish type. While I'm messing with that I might come up with a way to generate that WXS automatically. I'll add those manually in the meantime. Although it looks like those entries are in alphabetical order. Do you already have a script managing that file?

nickshulman:
That file is maintained manually. We carefully add the files in alphabetical order.

The file "FileList64.txt" has the list of all of the files that are copied directly from the Skyline Properties > Publish > Application Files dialog. The idea is that when you want to see what the new files are, you will paste in the new list of files, and see if FileList64.txt has changed. Then, if that file is different, then you go and add all the new files to Product.wxs

Brendan changed FileList64.txt yesterday, but should have also updated Product.wxs at the same time.

chambm:
Definitely need a test file from Shimadzu that prompted the DLL upgrade. There are new DLLs (both native and .NET) but none of them seem necessary for the new API to read the existing test files. I didn't remember to add them to the Shimadzu Jamfile either.

PR #1194, opened 2020-08-20
Fix to allow multiple Skyline instances to run MSAmanda searches at the same time

chambm:
* fixed Skyline and MSAmanda to allow multiple Skyline instances to run MSAmanda searches at the same time

Tested with a batch script `ParallelTestRunner.bat` (NB: absolute path for results argument seems necessary):
```
set PWIZ_ROOT=c:\pwiz.git\pwiz
FOR /L %%I IN (1,1,2) DO start cmd /c "%PWIZ_ROOT%\pwiz_tools\Skyline\bin\x64\Release\TestRunner.exe results="%PWIZ_ROOT%\pwiz_tools\Skyline\TestResults\TestRunner%%I results" log=TestRunner%%I.log %* & pause"
```


pwiz_tools/Skyline/Model/DdaSearch/MSAmandaSearchWrapper.cs
```diff
@@ -211,6 +212,7 @@ public override bool Run(CancellationTokenSource tokenSource)
             try
             {
                 using (var c = new CurrentCultureSetter(CultureInfo.InvariantCulture))
+                using (var d = new CurrentDirectorySetter(_baseDir.DirPath))
```
nickshulman, line 215: I am a little worried about this line of code because the current directory is a process-wide value, and not thread-local like "CurrentCulture". So, setting it to one value and then setting it back when you are done does not actually protect other code in Skyline which might be relying on the current directory being something specific.

However, I can't think of anything better for you to be doing here.

Also, I don't think Skyline actually uses the current directory for anything useful.

brendanx67: Well, the current directory probably impacts where some of our FileOpenDialogs start when they are brought up, but since this processing takes over the UI and doesn't allow work to continue, it probably doesn't impact those.

chambm: Yes, I only figured it was safe because the ImportPeptideSearchDlg is modal. If that ever changes, this will need to be revisited. It could probably be avoided with some changes to MSAmanda.

PR #1203, opened 2020-08-25
MSAmanda fixes

chambm:
- fixed non-Unimod mods causing errors in MSAmanda
- fixed error caused by going back from DDA settings page to spectra page, changing input files, and forward again
* fixed search engine UIUpdate not connecting after the first time the DDASearchControl is shown
* added test for canceling search, adding another spectrum file, and starting search again

@brendanx67 I assume you'll want these in 20.2 as well?


chambm:
Exception web.

PR #1207, opened 2020-09-01
Feature/sciex wiff2 data sdk update

chambm:
- changed output WIFF sample filename to be only WIFF filename if filename contains sample name (e.g. if sample name is "Sample" and WIFF name is "WiffSample", output is "WiffSample" instead of "WiffSample-Sample")
- added collision energy for WIFF2 and enabled it for 32-bit

* updated Sciex API DLLs
* removed restrictions on WIFF2 testing in Skyline
* removed isolation window functions from WiffFile::Experiment
* added INSTALLER_VENDOR_FILES.txt target for installer scripts to generate their vendor dependency file list automatically (pwiz and IDPicker installers updated, Skyline admin installer could come next)
* added retries in read_file_header() to avoid file lock failures with the alternate WIFF1 constructor
* added --test option to SeeMS to check that it can open a file

TODO: add SQLite.Interop.dll to Skyline.csproj and update FileList64.txt/Product.wxs or make admin installer automatically take vendor files from INSTALLER_VENDOR_FILES.txt target


chambm:
`TODO: add SQLite.Interop.dll to Skyline.csproj and update FileList64.txt/Product.wxs or make admin installer automatically take vendor files from INSTALLER_VENDOR_FILES.txt target` (the INSTALLER_VENDOR_FILES.txt is a list of files required by the vendors that is automatically generated by Boost.Build, so it's always generated up-to-date)

@nickshulman This is the last thing to do for this PR to go to master. The way I do it with pwiz and Bumbershoot installers is to have a WiX template which gets updated by a Python or C# script (pwiz uses Python and Bumbershoot uses C# because they used to be separate projects). That's the `pwiz-setup.py` file. I'm confident that I could adapt either script to do the same thing for the Skyline administrative installer. This would allow us to build and test it from the command-line (and thus from TeamCity). More importantly, TeamCity can test that the installed application actually works (for example we could run a test like PwizFileInfoTest, which uses all vendor readers).

Do you know if the current administrative installer can actually be installed as per-user, i.e. reg keys to HKCU and files not to Program Files, and thus not require admin permissions, which makes installing it on TeamCity trickier? If not, any objection to me (trying to) add the per-user capability?

My proposal:
1. Change the Product.wxs, FileList64.txt, and .testsetting files to templates that have everything like they are now except the required vendor files are not there (replaced by a placeholder).
2. Adapt the pwiz python script (or C# if you prefer) to instantiate those templates at Boost-Build-time to add the vendor files from INSTALLER_VENDOR_FILES.txt.
3. Leave Skyline.csproj alone? New native dependencies would need to be added manually to maintain ClickOnce (the .NET ones seem to be inherited automatically from ProteowizardWrapper).
4. Add Product.wxs, FileList64.txt, and the .testsetting files to .gitignore and clean.bat so they won't get committed.
5. The current approach of copying the ClickOnce list to FileList64.txt would stay mostly the same. I'll have to annotate the lines added by the setup script to make it clear that you don't need to add those lines to the wxs template (and of course you'd only modify the template, not the output wxs).

pwiz_tools/Skyline/Skyline.csproj
```diff
@@ -52,6 +52,10 @@
     <UseApplicationTrust>false</UseApplicationTrust>
     <PublishWizardCompleted>true</PublishWizardCompleted>
     <BootstrapperEnabled>true</BootstrapperEnabled>
+    
+    <!-- Make the assembly name configurable -->
+    <AssemblyName Condition=" '$(MSBuildAssemblyName)' == '' ">Skyline-daily</AssemblyName>
+    <AssemblyName Condition=" '$(MSBuildAssemblyName)' != '' ">$(MSBuildAssemblyName)</AssemblyName>
   </PropertyGroup>
```
chambm, line 59: What do you think about this change @brendanx67 ? There should be no change in the IDE, but it lets the command-line build override the default "Skyline-daily" assembly name. I used that ability to make --official override the name to "Skyline", which was necessary to fully test the admin installer's logic (because it changes the expected files based on the --official option).

chambm:
@nickshulman How often do you think we should run the installer test? Currently it runs TestData.dll and TestDdaSearch but the list of tests is arbitrary. It's currently a separate TeamCity build config that runs ~17m but ~12m of that is building the C++ core, so it'd be better to combine it with the existing build config. One option is to replace the current TeamCity run of TestData.dll with the installer test.

.gitignore
```diff
@@ -444,3 +444,8 @@ SkylineTester Results
 /pwiz_tools/Skyline/SkylineNightlyShim/Properties/AssemblyInfo.cs
 /pwiz_tools/Skyline/SkylineTester/Properties/AssemblyInfo.cs
 /pwiz_tools/Skyline/TestRunner/Properties/AssemblyInfo.cs
+/pwiz_tools/Skyline/Executables/Installer/FileList64.txt
+/pwiz_tools/Skyline/Executables/Installer/Test.passed
+/pwiz_tools/Skyline/Executables/Installer/setup.exe
+/pwiz_tools/Skyline/Executables/Installer/Skyline*.*
```
nickshulman, line 450: Can you make it so that these files get created in some subdirectory such as "bin" or "bin/x64"? It seems weird to have these output files created in the source directory like this.

chambm: I was following the existing README which mentioned finding the MSI in the same directory as the README. I'm generating the dynamic files (FileList64.txt and the WXS file) in the same place the static files used to be. `Test.passed` and `setup.exe` (which are basically flags to say that the MSI has been built and tested already) could certainly be moved to the build directory though. But if you want to move them all to a build directory I can do that.

chambm: I also updated the clean script so it would clean these artifacts.

pwiz_tools/Skyline/CleanSkyline.bat
```diff
@@ -41,5 +41,9 @@ IF EXIST ..\Shared\MSGraph\obj rmdir /s /q ..\Shared\MSGraph\obj
 IF EXIST ..\Shared\Crawdad\obj rmdir /s /q ..\Shared\Crawdad\obj
 IF EXIST ProtocolBuffers\tmp rmdir /s /q ProtocolBuffers\tmp
 IF EXIST ProtocolBuffers\GeneratedCode\*.cs del /q ProtocolBuffers\GeneratedCode\*.cs
+IF EXIST TestSettings_x64.testsettings del /q TestSettings_x64.testsettings
+IF EXIST TestSettings_x86.testsettings del /q TestSettings_x86.testsettings
+
+git clean -f -d -X Executables/Installer > nul
```
nickshulman, line 47: I don't think this batch file should use "git clean".
If a developer has files in their source tree that have not yet added to git, we would not expect that those files will be deleted by doing a clean build.

chambm: I agree it's really bad to delete something that isn't generated by the build. But note that git clean is used by the main pwiz clean scripts to clean up the vendor API and data directories, and the command above that deletes GeneratedCode/*.cs, and the ones that delete entire directories, are equally non-discriminating. But I'd accept that unadded files are more likely to happen in Executables/Installer than GeneratedCode. ;) If you're ok with the MSI being built in a bin directory, I'll switch to that and avoid the cleaning issue entirely.

nickshulman: Yes, bin directory would be a great place for the .msi file. That's where it currently gets put in the master branch, right?

chambm: The README implies it gets build in the Installer directory but I didn't actually test it. The vdproj doesn't load in my Visual Studio. Didn't they disable it in 2017? Are you using 2015 to build it?

chambm:
@nickshulman Still need a decision on whether to replace some/all of the TestRunner tests on TeamCity's current Skyline x64 build config with tests run from the installed location.

nickshulman:
> @nickshulman Still need a decision on whether to replace some/all of the TestRunner tests on TeamCity's current Skyline x64 build config with tests run from the installed location.

I think this is 100% your call.

chambm:
OK, I'll make the existing TestData target an alias to the installer test if WIX is installed, else fall back to the old, non-installed TestData test.

PR #1209, opened 2020-09-02
Change "psi-ms.obo" and "unimod.obo" to be "links" in Skyline.csproj …

nickshulman:
…so they do not need to be copied by the jamfile.

I noticed that the files "unimod.obo" and "psi-ms.obo" always show up in pwiz_tools\Skyline source tree as new files.

This change makes it so that Skyline.csproj references those files in "..\..\pwiz\data\common" as Visual Studio project linked files, so they do not need to be copied.

I think this fix works perfectly, but if it turns out that there really is some reason they need to get copied then we should update .gitignore.

### DIFF
--- a/pwiz_tools/Skyline/Jamfile.jam
+++ b/pwiz_tools/Skyline/Jamfile.jam
@@ -222,9 +222,6 @@ if [ modules.peek : NT ] && --i-agree-to-the-vendor-licenses in [ modules.peek :
         ) > "$(<)"
     }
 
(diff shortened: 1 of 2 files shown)


nickshulman:
> Definitely seems like a better option than copying the file. Do you have to edit the csproj manually for that?

In Visual Studio, when you do "Add Exisiting Item", the "Add" button on the File Dialog has a dropdown option for "Add As Link".
image

PR #1210, opened 2020-09-02
New menu item: Refine > Permute Isotope Modifications

nickshulman:
This change adds a menu item to the Refine menu: Permute Isotope Modifications
image

Users can choose which isotope modification to apply, and whether to do "Simple" (only the number of modified residues matters) or "Complete" (they want all combinations of labeled and unlabeled residues so they get 2^N precursors)
image

Here is what it looks like when they did "Simple" on a peptide with 8 leucines in it.
image

We might want to look into making the Edit Modifications dialog faster when there are a lot of label types.


chambm:
Does that last window scroll horizontally? What does the complex permutation look like for that peptide? :stuck_out_tongue:

nickshulman:
> Does that last window scroll horizontally? What does the complex permutation look like for that peptide? 😛

Yes, that window gets a scrollbar if it has too many columns. The complex permutation results in that form having 128 columns of combo boxes. It actually takes a few minutes for that form to come up once you have that many Isotope Label Types in your document. This is because every one of those combo boxes is a Window that needs to be created.
In the future, I was going to propose a fix to make that window come up faster. That fix would either involve somehow making it so that the combo boxes do not get constructed until they are needed, or making that entire thing into a DataGridView.

PR #1222, opened 2020-09-16
Fix ExportMethodDlg layout

kaipot:


kaipot:
> What about the Chinese files? Maybe this should wait until I merge the new localizations. Sorry to get you going on this, but doing this now will likely complicate my merge.

Sure, I will check again after the new localizations. I left the Chinese as-is, since it didn't look like there were any layout issues with it like on the English/Japanese ones.

PR #1224, opened 2020-09-16
Total Ion Current normalization

nickshulman:
This change adds Total Ion Current normalization:
image
image
image
image

"Total Ion Current" shows up in the list of normalization choices if either:
a) At least one replicate has a Total Ion Current Area
b) The document has no replicates at all

Until the user has imported their first replicate, we do not know whether Skyline will be able to get a TIC area out of it, so we make that normalization method available if there are no results in the document yet. This logic is in SrmSettings.HasTicArea

When normalizing to TIC, we do it in a similar way to how we do median normalization: the raw area is divided by the TIC Area, and then multiplied by the median TIC area across all of the replicates. In this way, the normalized area and the raw area will be close to each other in value.


pwiz_tools/Skyline/Controls/Graphs/AreaCVToolbar.resx
```diff
@@ -235,7 +235,7 @@
     <value>toolStripNumericDetections</value>
   </data>
   <data name="&gt;&gt;toolStripNumericDetections.Type" xml:space="preserve">
-    <value>pwiz.Skyline.Controls.ToolStripNumericUpDown, Skyline, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null</value>
+    <value>pwiz.Skyline.Controls.ToolStripNumericUpDown, Skyline-daily, Version=20.1.1.259, Culture=neutral, PublicKeyToken=null</value>
```
brendanx67, line 238: Revert this, or I just have to undo them at release time.

pwiz_tools/Skyline/Skyline.csproj
```diff
@@ -2483,8 +2483,8 @@
     </EmbeddedResource>
     <EmbeddedResource Include="Model\AuditLog\EnumNames.resx">
       <Generator>PublicResXFileCodeGenerator</Generator>
-      <LastGenOutput>EnumNames.Designer.cs</LastGenOutput>
       <SubType>Designer</SubType>
+      <LastGenOutput>EnumNames.Designer.cs</LastGenOutput>
```
brendanx67, line 2487: Revert file. Seems like just an ordering change.

pwiz_tools/Skyline/Skyline.resx
```diff
@@ -121,14 +121,23 @@
     <value>17, 17</value>
   </metadata>
   <metadata name="contextMenuTreeNode.TrayLocation" type="System.Drawing.Point, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
-    <value>17, 173</value>
+    <value>763, 17</value>
   </metadata>
   <assembly alias="System.Drawing" name="System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
+  <data name="contextMenuTreeNode.Size" type="System.Drawing.Size, System.Drawing">
```
brendanx67, line 127: Changes seem overly extensive for such a small addition.

brendanx67 (approved):
Looks good. A few places where changes seem unnecessary.

brendanx67:
Seems like we have some inconsistency in the way we refer to normalization. Either "Normalize to" or "Normalization method". With the former the values are "Medians" and "Total ion current" (sentence case) and the latter uses "Equalize Medians" and "Total Ion Current" (title case).

PR #1225, opened 2020-09-16
code inspection improvements

bspratt:
Can specify that an inspection only applies if the file contains a certain string
Inspections can check for missing text as well as unwanted text
An inspection can exclude certain directories (e.g. code areas where we don't bother with L10N when the inspection is about enforcing L10N)


pwiz_tools/Skyline/Test/CodeInspectionTest.cs
```diff
@@ -65,46 +71,141 @@ public void CodeInspection()
...
+                return string.IsNullOrEmpty(path) ||
+                       IgnoredDirectories != null && IgnoredDirectories.Any(path.Contains);
+            }
+        }
+        private readonly Dictionary<string, Dictionary<string, PatternDetails>> forbiddenPatternsByFileMask = new Dictionary<string, Dictionary<string, PatternDetails>>();
+        private readonly Dictionary<string, Dictionary<string, PatternDetails>> requiredPatternsByFileMask = new Dictionary<string, Dictionary<string, PatternDetails>>();
+
+        private enum Inspection { Forbidden, Required }
+
+        private string[] NonLocalizedProjectDirectories = { "Skyline\\Executables", "Skyline\\TestUtil", "Skyline\\SkylineNightly" };
```
nickshulman, line 194: The list of folders that we localize is in:
pwiz_tools\Skyline\Executables\KeepResx\Program.cs

It would be nice to have some way of keeping that folder list, and your folder list here in sync with each other.

bspratt: That's a handy list of things we do not localize, which I was unaware of. I'll see about using that.

PR #1234, opened 2020-09-21
Skyline/work/09032020 tutorial small mol libraries

bspratt:
Skyline: adding new tutorial "Skyline Small Molecule Multidimensional Spectral Libraries" by Kaylie Kirkwood, and associated test


pwiz_tools/Skyline/Controls/Startup/ActionTutorial.cs
```diff
@@ -187,7 +187,13 @@ public void ExtractTutorial(ILongWaitBroker waitBroker)
                 var skyFileToOpen = Path.Combine(extractDir ?? string.Empty, SkyFileLocationInZip);
                 foreach (var entry in zip.Entries.ToList())
                 {
-                    entry.FileName = entry.FileName.Substring(entry.FileName.IndexOf('/')); // Gets rid of everything up to and including first '/'.
+                    // Remove top level of directory structure if any - unless it's a data directory
```
nickshulman, line 190: Should we just change the .zip file that is on the website instead of having this logic here?

pwiz_tools/Skyline/Controls/Startup/ActionTutorial.cs
```diff
@@ -234,7 +241,20 @@ public void ExtractTutorial(ILongWaitBroker waitBroker)
                         MessageDlg.ShowWithException(Program.MainWindow, message, e);
                     }
                 }));
-               
+
+                // Make it convenient for user to locate tutorial files if we haven't already opened anything
+                if (!hasSkylineFile && !string.IsNullOrEmpty(ExtractPath))
+                {
+                    Directory.SetCurrentDirectory(ExtractPath);
+                    Settings.Default.LibraryDirectory =
```
nickshulman, line 249: Setting all of these directories at the same time looks like neat functionality. I wonder if we should make it a method on "DirectoryEx".

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -1789,11 +1800,25 @@ public void RestoreCoverViewOnScreen(bool hasSavedView = true)
...
+            SetSkylineWindowSize(1200, 800);
+        }
+
+        // Make the Skyline window as large as possible, without actually putting it into
+        // Maximized state which prevents further resizing
+        const int marginW = 14;
+        const int marginH = 7;
+        public void MaximizeSkylineWindow()
+        {
+            SetSkylineWindowSize(int.MaxValue - marginW, int.MaxValue - marginH);
```
nickshulman, line 1812: int.MaxValue is more than 2 billion.
does it really make a difference if you subtract marginW?

nickshulman (commented):
Some optional comments.

nickshulman (approved):

pwiz_tools/Skyline/Controls/Startup/ActionTutorial.cs
brendanx67: Yes. Tutorial ZIP files have a defined structure where they always include a toplevel folder that includes the name of the tutorial (e.g. MethodEdit, MethodRefine, iRT, etc.) That way things don't become a mess if you just right-click > Extract All to the current folder. This tutorial ZIP file SmallMoleculeLibraries.zip should contain a single subfolder "SmallMoleculeLibraries" which contains all the tutorial files and where the user will be instructed to save the .sky file.

pwiz_tools/Skyline/Controls/Startup/StartPage.cs
```diff
@@ -434,6 +434,18 @@ private void PopulateTutorialPanel()
                         string.Empty
                     ),
                     Description = TutorialTextResources.HiResMetabolomics_Description
+                },
+                new TutorialActionBoxControl
+                {
+                Caption = TutorialTextResources.SmallMolLibraries_Caption,
+                Icon = TutorialImageResources.SmallMolLibraries_start,
+                EventAction = () => Tutorial(
+                    ActionTutorial.TutorialType.hi_res_metabolomics,
```
brendanx67, line 443: All other tutorials seem to have their own type. So, I am guessing this is a copy-paste error where you forgot to give this new tutorial its own value in the enum. Though, I am not seeing how this value gets used. Seems vestigial and should maybe be removed entirely.

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
```diff
@@ -0,0 +1,431 @@
...
+            //   �	In the Precursor Adducts field, enter �[M-H], [M+HCOO], [M+CH3COO]�. 
+            //   �	In the Fragment Adducts field, enter �[M-]�.
+            RunUI(() =>
+            {
+                transitionSettingsUI.SmallMoleculePrecursorAdducts = "[M-H], [M+HCOO], [M+CH3COO]";
+                transitionSettingsUI.SmallMoleculeFragmentAdducts = "[M-]";
+                transitionSettingsUI.SmallMoleculeFragmentTypes = "f, p";
+            });
+            //   �	The Transition Settings form should now look like this:
+            PauseForScreenShot("Transition Settings: Filter", 4);
```
brendanx67, line 114: Use the IForView type in this pause for the Transition Settings - Filter tab

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
```diff
@@ -0,0 +1,431 @@
...
+                //   �	Set the Isolation scheme field to All Ions.
+                transitionSettingsUI.IsolationSchemeName = IsolationScheme.SpecialHandlingType.ALL_IONS;
+                //   �	Enter �20,000� in the Resolving power field.
+                transitionSettingsUI.ProductRes = 20000;
+                //   �	Check Use high-selectivity extraction.
+                transitionSettingsUI.UseSelectiveExtraction = true;
+            });
+
+            //   The Transition Settings form should look like this:
+            PauseForScreenShot("Transition Settings - Full-Scan", 5);
```
brendanx67, line 137: Use the IForView type in this pause for the Transition Settings - Full-Scan tab

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
```diff
@@ -0,0 +1,431 @@
...
+            OkDialog(addLibDlg, addLibDlg.OkDialog);
+            //   �	Click the OK button in the Edit Libraries form.
+            OkDialog(libListDlg, libListDlg.OkDialog);
+
+            //   The Libraries list in the Molecule Settings form should now contain the Drosophila Lipids library you just created.
+            //   �	Check the Drosophila Lipids checkbox to tell Skyline to use this library in the current document.
+            //   �	If you have any other libraries in this list checked, uncheck them now.
+            RunUI(() => { peptideSettingsUI.PickedLibraries = new[] {drosophilaLipids}; });
+            //   The Molecule Settings form should now look like:
+            PauseForScreenShot("Molecule Settings - Library", 7);
```
brendanx67, line 175: Use the IForView type in this pause for the Molecule Settings - Library tab

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
```diff
@@ -0,0 +1,431 @@
...
+            //   �	Click the OK button in the Molecule Settings form.
+            OkDialog(peptideSettingsUI, peptideSettingsUI.OkDialog);
+            doc = WaitForDocumentChangeLoaded(doc);
+
+            //   To open the library explorer and view the contents of the library you just added, do the following: 
+            //   �	From the View menu, click Spectral Libraries.
+            var viewLibUI = ShowDialog<ViewLibraryDlg>(SkylineWindow.ViewSpectralLibraries);
+
+            //   The library explorer should now resemble the image below:
+            PauseForScreenShot("Library Explorer (probably need to resize wider)", 8);
```
brendanx67, line 185: Do exact resizing and remove the "probably"

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
```diff
@@ -0,0 +1,431 @@
...
+            {
+                RunUI(confirmAddDlg.OkDialog);
+                doc = WaitForDocumentChangeLoaded(doc);
+            }
+
+            //   �	Close the Spectral Library Explorer window.
+            OkDialog(viewLibUI, viewLibUI.CancelDialog);
+
+            //   Your Skyline window should now resemble:
+            PauseForScreenShot("Populated Skyline window", 9);
```
brendanx67, line 201: Usually want to resize SkylineWindow to an exact size before the first full window screenshot.

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
```diff
@@ -0,0 +1,431 @@
...
+            //   �	From the View menu, choose Peak Areas and click Replicate Comparison (F7).
+            RestoreViewOnScreen(15);
+
+            //   �	Attach the Peak Areas view above the Retention Times view by clicking in the title bar and dragging until the mouse cursor is inside the up-side docking icon.
+            //   �	Right-click in the Peak Areas view and click Legend to hide the legend in its graph.
+            RunUI( () => SkylineWindow.ShowPeakAreaLegend(false));
+            //    You should end up with a similar layout to that below:
+            RunUI(() => SkylineWindow.AutoZoomBestPeak());
+            WaitForGraphs();
+            PauseForScreenShot("Layout", 15);
```
brendanx67, line 283: This isn't really a description of what I am taking the screenshot of. Is it the main window again?

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
```diff
@@ -0,0 +1,431 @@
...
+
+            //Understanding and Utilizing the IMS Separation
+            //To this point, we have ignored the IMS dimension in this data.To better understand the IMS separation, you need to look at the underlying spectra from which these chromatograms were extracted by doing the following:
+            //   �	On the View menu, choose Auto-Zoom and click Best Peak (F11).
+            //   �	Select the molecule PE(16:1_18:3).
+            FindNode("PE(16:1_18:3)");
+            WaitForGraphs();
+            RunUI(() => SkylineWindow.AutoZoomBestPeak());
+            WaitForGraphs();
+            PauseForScreenShot("Chromaotgram", 16);
```
brendanx67, line 299: So, copy metafile of just a specific chromatogram graph?

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
```diff
@@ -0,0 +1,431 @@
...
+            RunUI(() =>
+            {
+                // ReSharper disable once ConditionIsAlwaysTrueOrFalse
+                transitionSettingsUI.IonMobilityControl.IsUseSpectralLibraryIonMobilities = true;
+                transitionSettingsUI.IonMobilityControl.IonMobilityFilterResolvingPower = 50;
+                transitionSettingsUI.IonMobilityControl.WindowWidthType =
+                    IonMobilityWindowWidthCalculator.IonMobilityWindowWidthType.resolving_power;
+            });
+            //   �	The Transition Settings form should now look like this:
+            PauseForScreenShot("Transition Settings: IonMobility", 20);            //The Transition Settings should now look like:
```
brendanx67, line 360: Use IFormView for Transition Settings - Ion Mobility tab.

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
```diff
@@ -0,0 +1,431 @@
...
+            OkDialog(transitionSettingsUI, transitionSettingsUI.OkDialog);
+            doc = WaitForDocumentChange(doc);
+
+            //   The results must now be reimported with the newly applied IMS settings.
+            //   �	From the Edit menu, click Manage Results.
+            //   �	Click the Re-import button. �*� Should appear to the left of F_A_018.
+            //   �	Select M_A_001 in the Manage Results view.
+            //   �	Click the Re-import button. �*� Should appear to the left of M_A_001.
+            //   �	Click the OK button.
+            doc = SkylineWindow.Document;
```
brendanx67, line 373: There is a disposable object you can put in a using() clause for short blocks that cause a change like this.

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
```diff
@@ -0,0 +1,431 @@
...
+            RunUI(() =>
+            {
+                const string toolName = "LipidCreator";
+                hasLipidCreator = pick.HasTool(toolName);
+                if (hasLipidCreator)
+                {
+                    pick.SelectTool(toolName);
+                }
+            });
+            PauseForScreenShot(hasLipidCreator ? "LipidCreator in tool store" : "LipidCreator is not installed, do that first then take screen shot", 22);
```
brendanx67, line 418: I don't think this has anything to do with LipidCreator installation on the current computer. This would be missing if the ToolStore did not have LipidCreator in it. So, I think you need to clarify this message. I also think you want to put this check inside "IsPauseForScreenShots" to avoid any problems that might occur with this form trying to contact the website during TeamCity and nightly testing.

pwiz_tools/Skyline/TestPerf/TestPerf.csproj
```diff
@@ -135,7 +135,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="AreaCVHistogramQValueAndRatioTest.cs" />
-    <Compile Include="DdaTutorialTest.cs" />
+    <Compile Include="DdaTutorialTest.cs" />
```
brendanx67, line 138: Can't see why this is changing.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -1025,7 +1025,7 @@ public bool IsPauseForCoverShot
 
         public bool IsSavingCoverShots
         {
-            get { return /* CoverShotName != null*/ false; }
+            get { return IsPauseForCoverShot && CoverShotName != null; }
```
brendanx67, line 1028: I think this would only get used inside a PauseForCoverShot() call, right? Which should only happen when IsPauseForCoverShots is true.

brendanx67: Fine to leave it this way, though.

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
```diff
@@ -0,0 +1,415 @@
...
+            get { return true; }
+        }
+
+
+        [TestMethod]
+        [Timeout(int.MaxValue)] // These can take a long time
+        public void TestSmallMoleculeLibrariesTutorial()
+        {
+//            IsPauseForScreenShots = true;
+            RunPerfTests = true;
```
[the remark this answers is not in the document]
bspratt: I'd already fixed that, but hey, I should add that to the code inspection test!

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -1158,14 +1158,25 @@ public void PauseForCoverShot()
             Thread.Sleep(1000); // Give windows time to repaint
             var coverSavePath = GetCoverShotPath();
             ScreenshotManager.TakeNextShot(SkylineWindow, coverSavePath, ProcessCoverShot);
+            string coverSavePath2 = null;
```
brendanx67, line 1161: I am going to need to understand what you are doing here and why. I have used what existed before to create all necessary covershots, and I am not sure this represents an improvement. It looks like you can't get through without pausing the test, which seems like a bad idea, because it would require me to sit around and tend the machine while it is taking 60+ covershots. (20 x English, Japanese, Chinese)

pwiz_tools/Skyline/Controls/Startup/ActionTutorial.cs
bspratt: There are several other tutorial data sets that don't have a top level directory - I suspect that this code was the cause of some exception reports about "an item with this key has already been added", which the zip library may throw in this case.

pwiz_tools/Skyline/Controls/Startup/ActionTutorial.cs
bspratt: DirectoryEx.SetSkylineWorkingDirectories() ?

pwiz_tools/Skyline/Util/DataSourceUtil.cs
```diff
@@ -130,6 +130,37 @@ public static string GetSourceType(FileInfo fileInfo)
             }
         }
 
+        /// <summary>
```
brendanx67, line 133: Why not use IsDataSource(string path)? Because you are trying to figure this out for something in a ZIP file? Let's not do this. It's not really valid, because to know if a directory is a data source, you need to look inside it. Doing this correctly for a ZIP file would be more complicated, and you shouldn't need to do this if you organize the tutorial ZIP file consistent with all other tutorial ZIP files.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
bspratt: I'll add a comment "// So we don't roll into negative territory when SetSkylineWindowSize adds margins"

brendanx67 (changes requested):
Great to see the progress on this. Please address a few issues before merging.

pwiz_tools/Skyline/Controls/Startup/ActionTutorial.cs
bspratt: But according to Brendan such data sets are ill-formed, so I'll back this out and update the data set as you suggest. As to the others, I guess they're not causing trouble so I'll leave them alone.

pwiz_tools/Skyline/Controls/Startup/StartPage.cs
bspratt: Good catch, thanks - I'll see whether this is needed at all, as you suggest.

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
bspratt: >> avoid any problems that might occur with this form trying to contact the website during TeamCity and nightly testing
yeah, that's what I was thinking about - I'll make sure this is actually dealing with it properly

pwiz_tools/Skyline/TestPerf/TestPerf.csproj
bspratt: It's a line ending change. 
image

pwiz_tools/Skyline/Controls/Startup/StartPage.cs
bspratt: Yeah, just cruft. I tore out all the ActionTutorial.TutorialType stuff.

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
bspratt: right, thanks.

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
bspratt: right, thanks.

pwiz_tools/Skyline/TestPerf/SmallMolLibrariesTutorialTest.cs
bspratt: right, thanks.

pwiz_tools/Skyline/Util/DataSourceUtil.cs
bspratt: Yeah, already torn out since it was only to deal with a malformed zip file.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
bspratt: Perhaps this is a misnomer then:
image
Let's make that "Take cover shots" and not have it pause.

pwiz_tools/Skyline/Util/DataSourceUtil.cs
brendanx67: When I think about the ultimate UI for this, which it seems like we are not engaged in designing, I think it would be nice to be able to have the option to either pause at the moment of taking the covershot, which makes it easier to tweak, e.g. adjust the layout, File > Save, and grab the .view file, or just take screenshots in bulk. I guess I personally transitioned from producing covershots to taking them all in bulk, and kinda rushed to get back to other work once this largish undertaking was complete. So, as you are pointing out, I didn't leave it in an ideal condition for the next single covershot to be added.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
brendanx67: When I think about the ultimate UI for this, which it seems like we are not engaged in designing, I think it would be nice to be able to have the option to either pause at the moment of taking the covershot, which makes it easier to tweak, e.g. adjust the layout, File > Save, and grab the .view file, or just take screenshots in bulk. I guess I personally transitioned from producing covershots to taking them all in bulk, and kinda rushed to get back to other work once this largish undertaking was complete. So, as you are pointing out, I didn't leave it in an ideal condition for the next single covershot to be added.

pwiz_tools/Skyline/TestPerf/TestPerf.csproj
brendanx67: The new line ending appears to me to be in error, i.e. just LF and not CRLF?

bspratt: Yes, an error - except that the old one is the erroneous one. The revised/correct one is on the left.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
bspratt: So another radio button "Update cover shots"?

brendanx67: I guess the way it was intended to function was that it would only pause if you had not set the CoverShotName, which gets set in the test constructor. So, if you wanted to pause the test, you would comment out the CoverShotName, much the way you would set IsPauseForScreenShots = true to pause for screenshots when not using SkylineTester. I think this is still a fine implementation, provided it can be better commented and understood by new developers.

brendanx67: You can rename it and even move it if you want. Maybe "Covershot mode" after "Demo mode"

brendanx67: I don't think we need UI for deciding whether to pause or take a snapshot. I started thinking it would be pausing, hence most of the "pause" use and then decided that would be way too much work to do for all 20 x 3 languages x 2 formats. So, I guess the main intent for how you leave your test when you commit is fully automated without any pause. When you are developing it, you likely want to comment out its name and just have it pause in order to be able to tweak the layout, as I expect to do at some point for your test. Since, that feels more like a development time thing, I think it is fine to require the developer to comment out the name.

brendanx67: If you care enough, you can change all the language including the code to be clearer. e.g. "IsCoverShotMode" etc. Where covershot mode should pause if there is no CoverShotName and otherwise automatically take and store the screenshot at files based on the name.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -1789,11 +1798,25 @@ public void RestoreCoverViewOnScreen(bool hasSavedView = true)
...
+        }
+
+        // Set the Skyline window size, and center it in the screen to have the best chance of not needing to move it before Alt-PtrSc
+        public void SetSkylineWindowSize(int width, int height)
+        {
             RunUI(() =>
             {
-                var skylineSize = new Size(1200 + 14, 800 + 7);
                 var screenRect = Screen.FromControl(SkylineWindow).Bounds;
+                var skylineSize = new Size(Math.Min(screenRect.Width, width + marginW), Math.Min(screenRect.Height,  height + marginH)); // So we don't roll into negative territory when SetSkylineWindowSize adds margins
```
brendanx67, line 1819: Probably would prefer to fail the test, if these go negative. Or better yet, fail if the screenRect is not 1920x1080 and 100% scaling. This is an assumption of taking cover shots that I did not put into code, because I was the only person taking them and I knew they needed to be done on a 1920x1080 screen at 100% scaling.

brendanx67 (approved):
Looks fine. One minor comment. It would be great if the end result of the merge could keep the tutorial changes separate from the IsCoverShotMode changes, which make it harder to see what changed for the tutorial itself.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
bspratt: Easy enough to fail if screen is too small. Getting screen scaling turns out to be a mess, I don't think it's worth the trouble, especially given that other parts of our dev process already demand this be at 100%.

bspratt: I'll put the CoverShot related changes on their own branch then merge that.

brendanx67: Yes. I think just fail early in covershot mode, if the screen is not 1920x1080. No guarantees you would get a valid covershot without that exact screen resolution. Probably unlikely anyone would use greater than 100% at that resolution either.

bspratt: >>  fail early in covershot mode, if the screen is not 1920x1080

OK, will do

PR #1235, opened 2020-09-22
Prevent ArgumentException in FontSpec.Remake where font tries to get …

nickshulman:
…created with infinite size.

(Reported on exception web).


brendanx67 (approved):
Thanks, Nick.

PR #1241, opened 2020-09-25
Skyline: refactor our ParseEnum method to use Enum.TryParse() instead of Enum.Parse() in a try/catch block.

bspratt:
When debugging, "Exceptions Settings">"Break When Thrown">(all possible exceptions) is a great way to quickly locate a test exception. Employing exceptions as normal flow of control severely limits that feature's usefulness.

We should always use things like TryParse in favor of Parse when the language provides them. Exceptions should be exceptional.

Also, fix the typo in DetectionGraphType default value that was causing this to trip in the first place.

### DIFF
--- a/pwiz_tools/Skyline/Properties/Settings.Designer.cs
+++ b/pwiz_tools/Skyline/Properties/Settings.Designer.cs
@@ -922,7 +922,7 @@ namespace pwiz.Skyline.Properties {
 
         [global::System.Configuration.UserScopedSettingAttribute()]
         [global::System.Diagnostics.DebuggerNonUserCodeAttribute()]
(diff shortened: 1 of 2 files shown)


brendanx67 (approved):

nickshulman (commented):
Settings.Designer.cs is supposed to be automatically generated from Settings.settings. We should never be making changes to this file directly.

It looks like the problem is that Rita added some things to Settings.Designer.cs on 8/10 without changing Settings.settings. We should add the missing things to Settings.settings.

(We should fix this Settings.settings problem in the Release branch too).

bspratt:
Good point - I'm going to back out the Settings.Designer.cs change and let Rita deal with that seperately.

PR #1242, opened 2020-09-25
Add some things to Settings.settings that had been manually added to Settings.Designer.cs

nickshulman:
Settings.Designer.cs should never be edited by hand.
Visual Studio generates it from the file Settings.settings whenever Settings.settings is modified.

Visual Studio provides an editor for Settings.settings, but it's sometimes easier to edit the XML in Settings.settings directly.
If you have a complicated setting value type that cannot be represented by the XML in Settings.settings, then you can add the setting to Settings.cs, but never Settings.Designer.cs.

(Also, removed setting "DetectionsFontSize" since it was not being used, and I could not figure out what default value it should have).


bspratt:
Seems good, you even corrected "detection" to be "detections".

PR #1247, opened 2020-09-29
Skyline: refactor Adduct.TryParse so it does not employ exceptions

bspratt:
Adduct.TryParse is routinely used in searching for adduct descriptions in imported transition lists. Unfortunately its use of exceptions in normal flow of control makes it another case of "unexceptional exceptions" which complicate the use of the debugger to capture actual exceptions of interest. So, a small refactor to make it work without exceptions.


brendanx67 (approved):
Looks good to me.

PR #1249, opened 2020-09-30
Add iRT support for CLI import peptide search

kaipot:


pwiz_tools/Skyline/CommandLine.cs
```diff
@@ -1731,10 +1731,21 @@ private bool ImportSearch(CommandArgs commandArgs)
         private bool ImportSearchInternal(CommandArgs commandArgs, ref SrmDocument doc)
         {
             var progressMonitor = new CommandProgressMonitor(_out, new ProgressStatus(String.Empty));
+            IrtStandard irtStandard = null;
+            if (!string.IsNullOrEmpty(commandArgs.IrtStandardName))
+            {
+                irtStandard = Settings.Default.IrtStandardList.FirstOrDefault(standard => Equals(standard.Name, commandArgs.IrtStandardName));
```
nickshulman, line 1737: IrtStandardList is a XmlMappedList, so I believe you can do:
irtStandard = Settings.Default.IrtStandardList[commandArgs.IrtStandardName];

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/CommandLine.cs
kaipot: It looks like that will throw an exception if it isn't in the list, though

nickshulman: Oh, right, good point.

PR #1250, opened 2020-09-30
* fixed MsDataFileImpl.GetChromatogram() to convert time arrays to minutes

chambm:
* fixed MsDataFileImpl.GetChromatogram() to convert time arrays to minutes rather than assuming they're always in minutes

@brendanx67  @nickshulman My first attempt at this got the math wrong, but no tests failed (at least in TestData and TestFunctional, haven't tried TestPerf). Any ideas which test I should augment to check that the time units are correct?

### DIFF
--- a/pwiz_aux/msrc/utility/vendor_api/Shimadzu/ShimadzuReader.cpp
+++ b/pwiz_aux/msrc/utility/vendor_api/Shimadzu/ShimadzuReader.cpp
@@ -255,6 +255,10 @@ class ShimadzuReaderImpl : public ShimadzuReader
                         continue;
                     }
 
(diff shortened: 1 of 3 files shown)


brendanx67:
Search for tests that include Shimadzu.lcd files, right? I wouldn’t expect there to be many, and I expect you wrote some or all of them.

chambm:
I was thinking that because I got the math wrong in my time conversion that even the non-Shimadzu tests should have failed if they were checking the times, but I didn't think about the minutes<->minutes conversion being correct because it's just multiplying by 1 either way. So yeah, I'll look for a Shimadzu-centric test.

chambm:
OK I found a local LCD file that failed like Pawel's did. Probably only affects DDA files. I committed a fix for it in this PR.

PR #1252, opened 2020-10-01
Added protein turnover tool

alinmarsh:


nickshulman:
This looks great. Can you also add the source code for the C# project that was used to build "ProteinTurnoverArgCollector.dll"?

There is something weird going on with the line endings in a few files including "info.properties". It looks like your files are using just Carriage Return (which shows up in the diff as "^M"). They should use the standard Windows CR/LF.

alinmarsh:
Added the source code and mnemonics. Still working on line endings

nickshulman (approved):

PR #1264, opened 2020-10-08
Add "Simple Ratios" checkbox to Peptide/Molecule Settings 

nickshulman:
&gt; QuantificationSettings

When Simple Ratios is checked, Skyline no longer tries to pair transitions between light and heavy. Individual transitions have no ratio to heavy. The precursor ratio is calculated as the simple sum of light transition areas over heavy transition areas.

(Feature requested by Will)

This is what the checkbox looks like on the Quantification Tab:
image

Can we think of better text to put on the checkbox other than "Simple precursor ratios"?
Can the tooltip text be better?
_If checked, ratios between precursors will be calculated as the ratio of the sum of transition areas.
Transitions from the two precursors will not be paired with each other for the purpose of handling missing peaks or calculating ratio dot products._


brendanx67:
I think this checkbox should be moved up to under the "Normalization method:" dropdown list. They seem tightly associated to me. Aren't you talking about the ratio used in the normalization method? Where it is now, it just seems wedged in and possibly associated with units.

brendanx67 (approved):
Okay to merge.

PR #1276, opened 2020-10-15
Reduce size of class ChromKey from 136 bytes to 80 bytes.

nickshulman:
This change reduces the size of the class ChromKey by:
1. Changing the storage type of enums ChromSource and ChromExtractor to "byte".
2. Removing OptionalMidTime and OptionalCenterOfGravityTime
3. Changing storage OptionalStartTime and OptionalEndTime from two "Nullable<double>" to instead two "double" and one boolean.

This is what the size of these things are now (as output by new test "TestChromKeySize"):

TestContext   Messages: |   |  
-- | -- | --
Type:   `pwiz.Skyline.Model.Results.ChromKey` | Object size: 80 |  
Field Name | Type | Field Size
`_optionalMinTime` | `System.Double` | 8
`_optionalMaxTime` | `System.Double` | 8
`_hasOptionalTimes` | `System.Boolean` | 1
`<Target>k__BackingField` | `pwiz.Skyline.Model.Target` | 8
`<Precursor>k__BackingField` | `pwiz.Common.Chemistry.SignedMz` | 8
`<IonMobilityFilter>k__BackingField` | `pwiz.Skyline.Model.DocSettings.IonMobilityFilter` | 8
`<Product>k__BackingField` | `pwiz.Common.Chemistry.SignedMz` | 8
`<CollisionEnergy>k__BackingField` | `System.Single` | 4
`<ExtractionWidth>k__BackingField` | `System.Single` | 4
`<Source>k__BackingField` | `pwiz.Skyline.Model.Results.ChromSource` | 1
`<Extractor>k__BackingField` | `pwiz.Skyline.Model.Results.ChromExtractor` | 1
`<HasCalculatedMzs>k__BackingField` | `System.Boolean` | 1
`<HasScanIds>k__BackingField` | `System.Boolean` | 1
`<IsEmpty>k__BackingField` | `System.Boolean` | 1
BaseType:   `pwiz.Common.SystemUtil.Immutable` | Object size: 24 |  
BaseType: `System.Object` | Object size: 24 |  

We could certainly do more if we wanted to.
The "Target" and "Precursor" (and maybe a few more things) could be factored into a different class since they are always the same across all of the ChromKey's from a single target.


pwiz_tools/Skyline/Model/Results/ChromHeaderInfo.cs
```diff
@@ -1915,10 +1905,19 @@ public class ChromKey : IComparable<ChromKey>
...
         public bool HasScanIds { get; private set; }
         public bool IsEmpty { get; private set; }
-        public double? OptionalMinTime { get; private set; }
-        public double? OptionalMaxTime { get; private set; }
-        public double? OptionalCenterOfGravityTime { get; private set; } // Only used in SRM, to help disambiguate chromatograms with same Q1>Q3 but different retention time intervals
-        public double? OptionalMidTime { get; private set; }
+
+        public double? OptionalMinTime
+        {
+            get
```
bspratt, line 1911: inconsistent formatting, FWIW

pwiz_tools/Skyline/Test/ObjectSizeTest.cs
```diff
@@ -0,0 +1,66 @@
...
+            StringWriter stringWriter = new StringWriter();
+            TypeInspector.DumpTypeAndFields(type, stringWriter);
+            try
+            {
+                TestContext.WriteLine("{0}", stringWriter);
+            }
+            catch (NotImplementedException)
+            {
+                // TestRunner TestContext does not implement WriteLine.
+                // Do not output anything in that case
```
bspratt, line 62: why not use Console.WriteLine here?

nickshulman: The idea was to not clutter up the output in SkylineNightly.

bspratt:
Seems good. Probably should do some performance comparisons though?

pwiz_tools/Skyline/Test/ObjectSizeTest.cs
bspratt: So it's a feature! Perhaps the comment should say as much.

brendanx67:
So, when I count in my head, I see 6 x 8-byte fields = 48 + 24 base object = 72 bytes. How then do all the rest 2 x 4 + 6 x 1 = 14 bytes fit into the remaining 8 bytes to become 80 bytes? Unless the 24 base is not included and then 6x8 + 2x4 + 6x1 = 62 becomes 80, in which case this is not very well packed and should be rearranged to achieve a size of 64 bytes.

nickshulman:
> So, when I count in my head, I see 6 x 8-byte fields = 48 + 24 base object = 72 bytes. How then do all the rest 2 x 4 + 6 x 1 = 14 bytes fit into the remaining 8 bytes to become 80 bytes? Unless the 24 base is not included and then 6x8 + 2x4 + 6x1 = 62 becomes 80, in which case this is not very well packed and should be rearranged to achieve a size of 64 bytes.

The class "object" is 24 bytes, but, for some reason, 8 of those bytes are unused.
So, if you have a class derived from "object" which has 8 bytes or less of fields in it, then that derived class will also be 24 bytes.

So, it's really 16 + 62 = 78.

bspratt:
And you can still shave a few bytes with Brendan's boolean bitfield idea. 

Also I think you could move optional min,max into an object which would usually be a net savings if they are usually unpopulated.

brendanx67:
I am not so sure my bitfield can save any bytes due to packing. You could put all single-byte values into a 4-byte bitfield, but you would still get 4 bytes of padding instead of 2. I do think we should create an OptionalTime class with the 2 doubles and just have a reference to that, which is null when there are no optional times. This would save 8 bytes (and increase the padding from 2 to 3 bytes), and can't possibly be a performance consideration when Target and IonMobilityFilter are bigger and more heavily used classes.

PR #1277, opened 2020-10-15
Add progress bar to the RT regression form

rita-gwen:


pwiz_tools/Skyline/Model/DocSettings/Prediction.cs
```diff
@@ -27,6 +27,7 @@
 using pwiz.Common.Collections;
 using pwiz.Common.DataAnalysis;
 using pwiz.Common.SystemUtil;
+using pwiz.Skyline.Controls.Graphs;
```
brendanx67, line 30: You can't use a control class inside the model. Can you just use a base class or an interface or move CustomCancellationToken inside of the Model?

brendanx67 (changes requested):
Just one thing you need to change. Otherwise, looks pretty straightforward.

pwiz_tools/Skyline/Model/DocSettings/Prediction.cs
rita-gwen: Is it OK to use a base class, like ZedGraphControl or System.Windows.Forms.UserControl?

rita-gwen: BTW, I started looking for examples and found that AreaCVRefinementData has a lot of references to pwiz.Skyline.Controls.Graphs.

brendanx67: Ouch. Good point. We probably need to push through a cleanup pass. I found 26 instances of "using pwiz.Skyline.Controls" under Model. I wish we had a similar prefix for UI code. Also 26 instances of "using System.Windows.Forms". Probably we need to add these checks to Brian Pratt's new code inspection tool and push through the clean-up. Code reviews are obviously not enough, without automated enforcement.

brendanx67: No, nothing under System.Windows.Forms and not ZedGraphControl either.

rita-gwen: Fix pushed, waiting for tests to complete.

PR #1281, opened 2020-10-16
Fix fragment tolerance always using precursor tolerance units

chambm:
:facepalm:

This is causing users to think they are using a wide tolerance for fragment ions (like 1 Da) but it's being treated as 1 ppm if they're using ppm for parent tolerance.


nickshulman (approved):
Looks great. This would be great to merge into the 20.2 branch as well.

PR #1300, opened 2020-10-28
Initial work on skylineBatch

alinmarsh:


brendanx67:
Looks like you may have designed your forms on a monitor with scaling other than 100% (e.g. 150% or 200%). You should remove the image scaling lines that are causing the CodeInspection test to fail and review all your forms on a monitor with 100% scaling, which will likely require you to reduce your resolution. All our tutorial screenshots are taken at 1920x1080 with 100% scaling.

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/MainForm.cs
```diff
@@ -0,0 +1,543 @@
...
+        {
+            var numLines = textBoxLog.Lines.Length;
+            const int buffer = SkylineBatchLogger.MaxLogLines / 10;
+            if (numLines > SkylineBatchLogger.MaxLogLines + buffer)
+            {
+                textBoxLog.ReadOnly = false; // Make text box editable. This is required for the following to work
+                textBoxLog.SelectionStart = 0;
+                textBoxLog.SelectionLength =
+                    textBoxLog.GetFirstCharIndexFromLine(numLines - SkylineBatchLogger.MaxLogLines);
+                textBoxLog.SelectedText = string.Empty;
```
nickshulman, line 386: Usually you would set textBoxLog.Text to whatever you want the text to be, instead of using "SelectedText" to edit the text that is already in there.
(The danger I foresee with manipulating the text like this is that the TextBox might disagree with you about how many characters are there if there is anything unusual about the line ending characters).

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/MainSettings.cs
```diff
@@ -0,0 +1,162 @@
...
+
+        private enum Attr
+        {
+            TemplateFilePath,
+            AnalysisFolderPath,
+            DataFolderPath,
+            ReplicateNamingPattern,
+        };
+
+        public static MainSettings ReadXml(XmlReader reader)
```
nickshulman, line 117: Did you know that you effectively get an implementation of "ReadXml" and "WriteXml" for free just by adding the [XmlRoot] attribute at the top of the class. You only need to implement IXmlSerializable if you need to have your own custom implementation, which we do in Skyline because we need to be able to handle older versions of the way that we used to write out the XML. If you don't care about backwards or forwards compatibility you can usually skip implementing ReadXml and WriteXml.

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/Program.cs
```diff
@@ -0,0 +1,202 @@
...
+                {
+                    Application.Exit();
+                }
+            });
+
+            using (var mutex = new Mutex(false, $"University of Washington {AppName()}"))
+            {
+                if (!mutex.WaitOne(TimeSpan.Zero))
+                {
+                    MessageBox.Show($@"Another instance of {AppName()} is already running.", $@"{AppName()} Error", MessageBoxButtons.OK, MessageBoxIcon.Error);
```
nickshulman, line 72: I don't think you should prevent more than one instance of SkylineBatch running at the same time.
Does anything bad happen if I run more than one instance of SkylineBatch at the same time? Are they all saving their stuff to the same place?

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/Properties/Settings.cs
```diff
@@ -0,0 +1,133 @@
...
+                return list;
+            }
+            set {
+                this["ConfigList"] = value; // Not L10N
+            }
+        }
+
+        [UserScopedSetting]
+        [DefaultSettingValue("false")]
+        public bool MinimizeToSystemTray {
```
nickshulman, line 53: You might find it easier to define simple properties like this in "Settings.settings" using Design View in Visual Studio. You only need to add properties to "Settings.cs" if they require special code for reading or writing the value.

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/SkylineBatchConfig.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            MainSettings.Validate();
+            ReportSettings.Validate();
+           
+        }
+
+        public override string ToString()
+        {
+            Validate();
```
nickshulman, line 150: Validate should not be called from ToString, since ToString is not supposed to change anything.
Also, ToString() should not throw exceptions unless absolutely necessary.

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/SkylineBatchConfigForm.cs
```diff
@@ -0,0 +1,303 @@
...
+
+        public SkylineBatchConfigForm(IMainUiControl mainControl, SkylineBatchConfig config, ConfigAction action, bool isBusy)
+        {
+            _action = action;
+            _initialCreated = config?.Created ?? DateTime.MinValue;
+            _newReportList = new List<ReportInfo>();
+
+            _mainControl = mainControl;
+            _isBusy = isBusy;
+            InitializeComponent();
```
nickshulman, line 54: Unless it's actually necessary to do stuff first, "InitializeComponent" should be the very first line of this constructor.

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/packages.config
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
nickshulman, line 1: Does this file do anything? My Visual Studio seems to be ignoring it, but I wonder whether it would be able to fix my System.Collections.Immutable dll missing problem.

nickshulman (commented):
This looks really good.
My Visual Studio wasn't able to find System.Collections.Immutable, so I haven't actually run anything.
I think you should add System.Collections.Immutable.dll to the git project somewhere. It's possible that we should figure out how to use nuget, but usually we just check any extra dll's into the git project.

I have a bunch of comments on the code here. Nearly all of these comments are optional suggestions or things that you might not have thought about.

PR #1311, opened 2020-11-04
Allow showing Calibration Curve normalized value on Peak Area graph

nickshulman:
The right-click menu on the Peak Area graph will allow you to choose to display the concentration calculated using the calibration curve.

image

There are two new menu items on the Normalize To context menu: "Default" and "Calibration Curve". "Default" is always available but the "Calibration Curve" menu item only shows up if you have chosen a Regression Fit at "Peptide Settings" > Quantification.

"Default" means to use whatever Normalization Method is specified for the currently selected peptide. That option is more important on the Area CV Histogram graph, where you are showing data from multiple peptides. You need an option that says to use each peptide's individual normalization method.

image


pwiz_tools/Skyline/Model/Results/DocNodeChromInfo.cs
```diff
@@ -525,9 +518,11 @@ public bool IsGoodPeak(bool integrateAll)
 
         public float? GetRatio(int index)
         {
-            return index != RATIO_INDEX_GLOBAL_STANDARDS
-                ? _ratios[index]
-                : _ratios[_ratios.Count - 1];
+            // TODO: RatioIndex
```
brendanx67, line 521: Not clear what the TODO is about. Clarify and make TODO(nicksh). Also, why keep the ternary expression commented out?

pwiz_tools/Skyline/Test/CommandLineAssayImportTest.cs
```diff
@@ -268,7 +269,9 @@ private static void ValidateIrtAndLibrary(SrmDocument docAfter)
                     foreach (var nodeTran in nodePep.TransitionGroups.SelectMany(g => g.Transitions))
                     {
                         if (!nodeTran.HasLibInfo)
-                            Assert.Fail("Missing library info from {0} - {1}", nodePep, nodeTran.GetDisplayText(new DisplaySettings(nodePep, false, 0, 0)));
+                            Assert.Fail("Missing library info from {0} - {1}", nodePep, nodeTran.GetDisplayText(new DisplaySettings(
```
brendanx67, line 272: Seems overly complicated new code compared with supplying nodePep and false, 0, 0

pwiz_tools/Skyline/Test/FindNodeTest.cs
```diff
@@ -49,7 +50,7 @@ public void RunTestFindNode(RefinementSettings.ConvertToSmallMoleculesMode asSma
 
             SrmDocument doc = CreateStudy7Doc();
             doc = new RefinementSettings().ConvertToSmallMolecules(doc, TestDirectoryName, asSmallMolecules);
-            var displaySettings = new DisplaySettings(null, false, 0, 0); //, ProteinDisplayMode.ByName);
+            var displaySettings = new DisplaySettings(new NormalizedValueCalculator(doc), null, false, 0, NormalizeOption.RatioToFirstStandard(doc.Settings)); //, ProteinDisplayMode.ByName);
```
brendanx67, line 53: Seems overly complicated new code compared with supplying nodePep and false, 0, 0

pwiz_tools/Skyline/TestFunctional/NormalizeToCalibrationCurveTest.cs
```diff
@@ -0,0 +1,189 @@
...
+                            {
+                                AssertEx.IsTrue(double.IsNaN(value) || Equals(value, PointPairBase.Missing));
+                            }
+                        }
+                    });
+                }
+            }
+        }
+
+        private T FindGraphPane<T>() where T : class
```
brendanx67, line 175: Maybe put on AbstractFunctionalTest[Ex]?

pwiz_tools/Skyline/TestTutorial/ExistingExperimentsTutorialTest.cs
```diff
@@ -632,7 +632,7 @@ private void DoStudy7Test()
 
         private static void NormalizeGraphToHeavy()
         {
-            AreaGraphController.AreaView = AreaNormalizeToView.area_ratio_view;
+            AreaGraphController.AreaNormalizeOption = NormalizeOption.TOTAL;
```
brendanx67, line 635: TOTAL (area_percent_view) seems different from area_ratio_view.

brendanx67 (approved):
Nothing major. Just a few questions/suggestions. Clean up and merge.

nickshulman:
Thanks for the feedback, Brendan.

Oops. I had forgotten about the TODO comments that I had in there. They all indicated something which I did not want to commit to master at all. (We should come up with something that causes an inspection error such as "//DONTMERGE".)

There are a couple of things that we should think about before I merge this to master:
1. This change makes it so that we never use the ratio lists on TransitionChromInfo and TransitionGroupChromInfo. Those ratios are still there so that I can have my assert during tests where I verify that the numbers in those lists are the same as what I calculate from scratch with "NormalizedValueCalculator".

2. The SequenceTree uses NormalizedValueCalculator to calculate the ratios that it displays. I have tried to make sure that it is just as fast as it used to be. **Can you point me at a really large document which has heavy labels in it? I want to make sure that TransitionTreeNode.GetResultsText has not gotten any slower with my changes.**

3. When you change which label type you are normalizing to in the Area Replicate Graph Pane, it changes the text that gets displayed on the transitions in the Targets tree. Currently, when you choose Global Standards, it continues to display whatever the last ratio that you were displaying. Now that we have more options, like Median normalization and Calibration Curve, we should figure out what we should display in the Targets tree. Would people want to see the median normalized peak area for transitions in the Targets tree, in the same way that we show them the ratio to heavy?

PR #1315, opened 2020-11-09
Fix for the DetectionPlot test failure

rita-gwen:


pwiz_tools/Skyline/TestFunctional/DetectionsPlotTest.cs
```diff
@@ -78,8 +78,9 @@ protected override void DoTest()
             RunUI(() =>
             {
                 propDialog.SetQValueTo(0.003f);
+                propDialog.OkDialog();
             });
-            OkDialog(propDialog, propDialog.OkDialog);
+            WaitForClosedForm(propDialog);
```
brendanx67, line 83: This is functionally equivalent to the code it replaces.

pwiz_tools/Skyline/Controls/Graphs/DetectionsGraphController.cs
```diff
@@ -153,7 +153,7 @@ public class Settings
             public static float QValueCutoff
             {
                 get => Properties.Settings.Default.DetectionsQValueCutoff;
-                set => Properties.Settings.Default.DetectionsQValueCutoff = value;
+                set => Properties.Settings.Default.DetectionsQValueCutoff = value;
```
brendanx67, line 156: Probably not worth making this line-ending change, since it is the only change to this file.

pwiz_tools/Skyline/TestFunctional/DetectionsPlotTest.cs
```diff
@@ -91,29 +92,31 @@ protected override void DoTest()
             RunUI(() =>
             {
                 propDialog.SetQValueTo(0.001f);
+                propDialog.OkDialog();
             });
-            OkDialog(propDialog, propDialog.OkDialog);
+            WaitForClosedForm(propDialog);
```
brendanx67, line 97: Again. This doesn't actually change anything.

pwiz_tools/Skyline/TestFunctional/DetectionsPlotTest.cs
```diff
@@ -174,29 +179,29 @@ protected override void DoTest()
         private void AssertDataCorrect(DetectionsPlotPane pane, int refIndex, float qValue, bool record = false)
         {
             DetectionPlotData data = null;
-            WaitForConditionUI(() => (data = pane.CurrentData) != null 
-                                           && data.QValueCutoff == qValue
+            WaitForConditionUI(() => pane.CurrentData != null
```
brendanx67, line 182: Lots of unnecessary white space changes above. Is this the critical change? Using CurrentData everywhere instead of assigning it in the first logical expression and then using the assigned value?

pwiz_tools/Skyline/Controls/Graphs/DetectionToolbarProperties.cs
```diff
@@ -129,10 +129,10 @@ private void cmbTargetType_SelectedIndexChanged(object sender, EventArgs e)
         public void SetQValueTo(float qValue)
         {
             if (qValue == .01f)
-                rbQValue01.Select();
+                rbQValue01.Checked = true;
```
brendanx67, line 132: Is this the critical change? It does seem wrong to use Select() instead of Checked = true. Not sure why that would pass sometimes and not others, though.

brendanx67 (changes requested):
Not super clear what the fix is. Looks like a lot of white space changes and other changes that I would expect not to make any difference.

pwiz_tools/Skyline/Controls/Graphs/DetectionToolbarProperties.cs
rita-gwen: The only way I was able to reproduce the timeout failure was to focus on another window while the test was running and that traced down to using Select() method. I cannot be 100% sure, but I ran this test in a loop for an hour without any exceptions.

rita-gwen:
Sorry about the whitespace changes. I added and removed a lot of tracing statements while debugging and Visual Studio is very trigger-happy to re-format the surrounding code.

pwiz_tools/Skyline/TestFunctional/DetectionsPlotTest.cs
```diff
@@ -174,29 +179,29 @@ protected override void DoTest()
...
             DetectionPlotData data = null;
-            WaitForConditionUI(() => (data = pane.CurrentData) != null 
-                                           && data.QValueCutoff == qValue
+            WaitForConditionUI(() => pane.CurrentData != null 
+                                           && pane.CurrentData.QValueCutoff == qValue
                                            && DetectionPlotData.GetDataCache().Status == DetectionPlotData.DetectionDataCache.CacheStatus.idle,
                 () => $"Retrieving data for qValue {qValue}, refIndex {refIndex} took too long.");
             WaitForGraphs();
-            Assert.IsTrue(data.IsValid);
+            WaitForCondition(() => pane.CurrentData.IsValid);
```
rita-gwen, line 187: This is the fix for the assertion failure that Nick tried to fix. The problem was that WaitForGraphs() doesn't guarantee that the data computation is complete. I've seen this in the RT regression tests as well.

pwiz_tools/Skyline/TestFunctional/DetectionsPlotTest.cs
rita-gwen: Saves one Invoke() call. My initial theory was that two invoke calls in quick succession might occasionally be executed out of order.  It proved to be incorrect, but since it's slightly more efficient than the original code I decided to keep it.

PR #1325, opened 2020-11-19
Fix TestBackgroundEventThreads

nickshulman:


pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -689,7 +689,15 @@ public static void WaitForClosedForm(Form formClose)
...
                         var formCloseClassName = System.ComponentModel.TypeDescriptor.GetClassName(formClose);
-                        var formCloseText = formClose.Text;
+                        string formCloseText;
+                        try
+                        {
+                            formCloseText = formClose.Text;
+                        }
+                        catch
+                        {
+                            formCloseText = "@@ERROR@@";
```
bspratt, line 699: Seems a bit dramatic? Maybe just "@@(could not retrieve form text)@@"

bspratt (changes requested):

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -689,7 +689,15 @@ public static void WaitForClosedForm(Form formClose)
...
                         var formCloseClassName = System.ComponentModel.TypeDescriptor.GetClassName(formClose);
-                        var formCloseText = formClose.Text;
+                        string formCloseText;
+                        try
+                        {
+                            formCloseText = formClose.Text;
+                        }
+                        catch
+                        {
+                            formCloseText = "@@(could not retrieve form text)@@";
```
brendanx67, line 699: Probably not worth the extra effort, but you could also set a flag that would cause you to try to formClose.Invoke() to get its text while on its own thread, rather than while on the main window thread.

PR #1326, opened 2020-11-20
Skyline/work/20201120 heat maps

nickshulman:
This pull request adds a "Cluster" button (which is now displayed as an icon that is supposed to look like a colorful heat map) to the toolbar on top of the Document grid and other grids.
image

The cluster button works particularly well when you are looking at data where a numeric column has been pivoted, such as when you have checked "Pivot Replicate Name" in the view editor.

If you want to have more control over what values are getting clustered, you can click the "Advanced" button:
image
 
which brings up the Advanced Clustering Editor:
image

Also, if you are looking at a Group Comparison, there is a new built-in view called "Clustered":
image

The "Clustered" view in the group comparison automatically clusters on the Z-score of the replicate abundances.
image

(Clicking the multi-colored cluster button also automatically switches you to the "Clustered" view).

It also puts colored things on the ends of the dendrogram indicating the BioReplicate, Cohort, and Protein and Peptide.

You can also view the heat map as a graph:
image


alinmarsh:
I'm not using DendrogramControl anymore, but was hesitant to delete it. Do we need dendrograms for any thing else?

PR #1334, opened 2020-11-30
Add Skyline support for DDA search of DIA files via DiaUmpire

chambm:
* added ConverterSettingsControl for DiaUmpire
- replaced "Perform DDA Search" radio button with "Input file type" combobox so user can specify whether input files are DIA or DDA (or search results)
- fixed bug reading isolation window offsets from new WIFF2 SDK; we were using the absolute isolation m/z lower and upper bound instead of the lower and upper offset (a value relative to the isolation target m/z)
* fixed "Handle not initialized" bug from WIFF2 destructor when terminating process forcefully (like SeeMS does), which apparently may run the static destructors before non-statics
* updated KeyValueGridDlg to take a parent window and use checkboxes for boolean variables
* switched DiaUmpire to a one-spill-file-per-DIA-window model, where spill files are stored in TEMP; scan metadata is merged/sorted after all windows are processed, but scan data is read on the fly when requested by SpectrumList_DiaUmpire::spectrum()
* moved TemporaryFile from IDPicker::Merger to pwiz::util (Filesystem.cpp/hpp)
* fixed peak picker filter not providing correct centroid/profile metadata for detail levels less than full data
* changed TemporaryDirectory so the default "~SK" prefix can be changed; I made the MSAmanda temp directories all share a common directory under TEMP/~SK_MSAmanda
* updated non-DiaUmpire audit logs
* fixed DiaUmpire::Config::maxThreads not being initialized by parameterless ctor
* moved DiaUmpireTest to SpectrumList_DiaUmpireTest


chambm:
@brendanx67 @nickshulman I think a code review is in order before merge to master.

pwiz_tools/Skyline/Model/AbstractDdaSearchEngine.cs
```diff
@@ -36,13 +36,13 @@ public abstract class AbstractDdaSearchEngine : IDisposable
...
         public abstract Bitmap SearchEngineLogo { get; }
-        protected MsDataFileUri[] SpectrumFileNames { get; set; }
+        public MsDataFileUri[] SpectrumFileNames { get; protected set; }
         protected string[] FastaFileNames { get; set; }
 
         /// <summary>
         /// A string or numeric variable with a default value and, for numeric types, min and max values.
         /// </summary>
-        public class Setting
+        public class Setting : IAuditLogObject
```
nickshulman, line 45: I am not an expert on audit log objects, but I was under the impression that if a class had a self-describing ToString method, it would not need to implement IAuditLogObject.

pwiz_tools/Skyline/Model/DdaSearch/MSAmandaSearchWrapper.cs
```diff
@@ -202,15 +204,22 @@ private void InitializeEngine(CancellationTokenSource token, string spectrumFile
...
+            //Console.WriteLine($@"{Settings.ReportBothBestHitsForTD}       {Settings.CombineConsideredCharges}        {Settings.WriteResultsTwice}       {Settings.ForceTargetDecoyMode}");
             mzID.Settings = Settings;
             SearchEngine = new MSAmandaSearch(helper, _baseDir.DirPath, _outputParameters, Settings, token);
             SearchEngine.InitializeOutputMZ(mzID);
             Settings.LoadedProteinsAtOnce = (int) AdditionalSettings[MAX_LOADED_PROTEINS_AT_ONCE].Value;
             Settings.LoadedSpectraAtOnce = (int) AdditionalSettings[MAX_LOADED_SPECTRA_AT_ONCE].Value;
         }
     
-        public override bool Run(CancellationTokenSource tokenSource)
+        public override bool Run(CancellationTokenSource tokenSource, IProgressStatus status)
```
nickshulman, line 219: This method should really be taking a CancellationToken instead of a CancellationTokenSource (since it is only interesting in asking whether the job has been cancelled and does not need the ability to request cancellation).

That would require changing MSAmandaSearch so that it also took a CancellationToken instead of CancellationTokenSource.

pwiz_tools/Skyline/SettingsUI/FullScanSettingsControl.cs
```diff
@@ -1,1110 +1,1126 @@
...
+        private ImportPeptideSearchDlg.Workflow? _lastPeptideSearchWorkflow;
+        public void ModifyOptionsForImportPeptideSearchWizard(ImportPeptideSearchDlg.Workflow workflow, bool libIonMobilities)
+        {
+            var settings = _documentContainer.Document.Settings;
+
+            if (_lastPeptideSearchWorkflow == workflow)
+                return;
+            _lastPeptideSearchWorkflow = workflow;
+
+            // reset controls to default (without reconstructing them like Initialize() would do)
```
nickshulman, line 904: Usually for something like this we would replace the control on the form with a new one that we construct. Is there a reason you would not want to do this here?

nickshulman (commented):
Looks pretty good.
You might want to take everything in "DiaUmpire.cs" in ProteoWizardWrapper and spread it out across multiple source files in its own folder.
The source file "MsDataFileImpl.cs" is horribly laid out with a lot of things as inner classes as if everyone who added to this class though everything needed to be in the same source file.

pwiz_tools/Skyline/Model/DdaSearch/MSAmandaSearchWrapper.cs
chambm: I agree but since this requires a change in the MSAmanda DLLs and it's not specifically relevant to DIA-Umpire, let's leave that to a separate change set.

pwiz_tools/Skyline/SettingsUI/FullScanSettingsControl.cs
chambm: I'm not actually sure this is needed after Kaipo's earlier change with MakeFullScanSettingsControl(). I'll see if it works without it. But I had been trying to call Iniitialize() instead of this and it was not working like I expected (forgot the exact problem though).

pwiz_tools/Shared/Common/SystemUtil/FormUtil.cs
```diff
@@ -60,7 +60,26 @@ public static DialogResult ShowDialog(Control owner, Form dialog)
             }
             return dialogResult;
         }
-        
+
+        /// <summary>
+        /// Moves the control's X and Y coordinates according to the X and Y values.
+        /// </summary>
+        public static void Offset(this Control control, int x = 0, int y = 0)
```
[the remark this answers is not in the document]
chambm: Do you mean use the new extension function I added below this one? Because the framework doesn't have a Point function that offsets a point and returns the new point. But yes I agree this function could be briefer by piggybacking off the below one.

chambm:
Also can a few other devs please try running the TestTtofDiaUmpireFullTutorial test? I have noticed some very sporadic crashes but haven't been able to pinpoint the cause or repro conditions.

pwiz_tools/Shared/Common/SystemUtil/FormUtil.cs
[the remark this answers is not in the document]
nickshulman: Oops! You're right. I was the one who was confused.

chambm:
> You might want to take everything in "DiaUmpire.cs" in ProteoWizardWrapper and spread it out across multiple source files in its own folder.
Are you really referring to `DiaUmpire.cs`? Because the CS wrapper seems pretty brief to me. The `Config` class is a nested class so to move that out it would either have to be a partial class or un-nested, right?

> The source file "MsDataFileImpl.cs" is horribly laid out with a lot of things as inner classes as if everyone who added to this class though everything needed to be in the same source file.
While I pretty much agree, to be fair the vast majority of that file is consumed by MsDataFileImpl class itself. So separating the helper classes into separate files would only help so much.

pwiz_tools/Skyline/FileUI/PeptideSearch/DDASearchControl.cs
```diff
@@ -74,11 +73,30 @@ protected virtual void UpdateTaskbarProgress(TaskbarProgress.TaskbarStates state
...
+
+            var newEntry = new ProgressEntry(DateTime.Now, message);
             _progressTextItems.Add(newEntry);
             txtSearchProgress.AppendText($@"{newEntry.ToString(showTimestampsCheckbox.Checked)}{Environment.NewLine}");
-            UpdateTaskbarProgress(TaskbarProgress.TaskbarStates.Normal, status.PercentComplete);
-            progressBar.Value = status.PercentComplete;
+
+            if (status.IsError)
+            {
+                Program.ReportException(status.ErrorException);
```
nickshulman, line 84: You should use "MessageDlg.ShowWithException" here instead.

Program.ReportException is only for things that are actual bugs where we want the user to submit a bug report to the exception web site.

nickshulman (commented):
Looks really good. Just a couple of comments.

We might want to change this to radio buttons with tooltips instead of having the help text in the combo box options.

image

chambm:
> Looks really good. Just a couple of comments.
> 
> We might want to change this to radio buttons with tooltips instead of having the help text in the combo box options.
> 
> image

What do you think about Nick's suggestion @brendanx67  ?

pwiz_tools/Skyline/FileUI/PeptideSearch/DDASearchControl.cs
chambm: So `ShowWithException` doesn't allow the user to report the error with a button press? DDA support is still pretty new: I'm thinking that for the near future we should allow users to report these exceptions easily. Perhaps we can just add catches for the common non-bug errors that get reported here, and handle those with `ShowWithException`, but let unexpected errors fall through to `ReportException`?

chambm:
> You might want to take everything in "DiaUmpire.cs" in ProteoWizardWrapper and spread it out across multiple source files in its own folder.

Do you recommend moving the nested Config class to a separate DiaUmpireConfig file? I agree the IterationListenerToMonitor nested class should move to its own file; it's not specific to DiaUmpire at all.

pwiz_tools/Skyline/FileUI/PeptideSearch/DDASearchControl.cs
nickshulman: You certainly could show a message box with multiple buttons on it, telling them to push one of the buttons if they think the error was a bug (and pushing that button would call Program.ReportException to bring up the ReportErrorDlg). There is no place where we currently do that in the code, but it would be a reasonable thing to do.

You would not want the user to see the ReportErrorDlg without that other message box first. The information on the ReportErrorDlg is presented in a way that is very hard for users to read, so even when the error message is straightforward (e.g. "The file already exists"), users will typically not be able to figure out what went wrong.

chambm:
@brendanx67 @nickshulman Should we make this an integration branch for a few days before merging? Do you want to review yourself, @brendanx67 ?

pwiz_tools/Skyline/FileUI/PeptideSearch/ConverterSettingsControl.cs
```diff
@@ -0,0 +1,314 @@
...
+                (value, setting) => setting.Validate(value));
+        }
+
+        private void UpdateSettingIfNonDefault<T>(IDictionary<string, AbstractDdaSearchEngine.Setting> settingStore,
+            string settingName, T newValue, T defaultValue) where T : IEquatable<T>
+        {
+            if (newValue.Equals(defaultValue))
+                return;
+
+            dynamic newValueDynamic = newValue;
```
chambm, line 272: Doing constructor overload resolution at runtime with a generic parameter required this dynamic voodoo, which required adding the Microsoft.CSharp reference. Any problem with that @nickshulman ?

nickshulman:
There's nothing wrong with doing it that way.
I can't think of any reason that we would need this code to target an
earlier version of the C# language.

In general, we probably should try to make use of as many language features
as possible so that we learn more.
It's rather unnecessary in this case since the value eventually gets
converted to an object. (AbstractDdaSearchEngine.Setting could have a
constructor that takes an object).
-- Nick

chambm:
> (AbstractDdaSearchEngine.Setting could have a constructor that takes an object). 

The typed constructors ensure that the min/max values are the same type as the default/current value. That could be done at runtime with an object/object/object constructor of course, but with overloads you get compile-time checking for most usages.

pwiz_tools/Skyline/TestPerf/DiaUmpireTutorialTest.cs
```diff
@@ -0,0 +1,922 @@
...
+        private void RunTest()
+        {
+//            IsPauseForScreenShots = true;
+//            RunPerfTests = true;
+//            IsPauseForCoverShot = true;
+            CoverShotName = "DIA-SWATH with DiaUmpire";
+
+            RunFunctionalTest();
+
+            Assert.IsFalse(IsRecordMode);   // Make sure this doesn't get committed as true
```
nickshulman, line 334: This assert here fails if you check "Record Audit Logs" in SkylineTester (and the test will not get run in subsequent languages).

If you want to keep this assert here, I think you need to change "IsRecordMode" so that it does something other than return "IsRecordAuditLogForTutorials".

pwiz_tools/Skyline/TestPerf/DiaUmpireTutorialTest.cs
```diff
@@ -0,0 +1,922 @@
...
+            else
+            {
+                var ambiguousDlg = ShowDialog<MessageDlg>(recalibrateMessage.ClickNo);
+                RunUI(() => AssertEx.Contains(ambiguousDlg.Message,
+                    Resources.BiblioSpecLiteBuilder_AmbiguousMatches_The_library_built_successfully__Spectra_matching_the_following_peptides_had_multiple_ambiguous_peptide_matches_and_were_excluded_));
+                OkDialog(ambiguousDlg, ambiguousDlg.OkDialog);
+            }
+            OkDialog(addIrtDlg, addIrtDlg.OkDialog);
+
+            var peptidesPerProteinDlg = ShowDialog<PeptidesPerProteinDlg>(() => {}, 600000); //importPeptideSearchDlg.ClickNextButton(), 600000);
```
nickshulman, line 635: You should use "WaitForOpenForm" here instead of "ShowDialog".

"ShowDialog" will fail if the dialog happens to already be showing.

nickshulman (approved):
Looks good to me.

chambm:
Ah crap. Forgot to squash. Let me see if I can fix that.

chambm:
@brendanx67  @nickshulman I force pushed to master to squash all those commits. Hopefully I didn't f it up but of course I'll fix it if I did. I noticed a new option in the branch protection options: `Require linear history` which would prevent merging PRs with merge commits. Should we enable that? It's merge commits that cause the wonky lines in the TortoiseGit log.

PR #1336, opened 2020-12-01
Skyline: fix some issues with background loading of ion mobility libraries (.imsdb files)...

bspratt:
…, including potential timing issues that could prevent a full loading of the dB or spurious complaints from audit logger about unlogged document changes. The root problem was the original implementation not following the library/librarySpec pattern.  TestIonMobility() was modified to demonstrate the issue and verify the fix.

Not reported by any user: problem discovered while working on FAIMS improvements.

### DIFF
--- a/pwiz_tools/Skyline/Model/IonMobility/IonMobilityLibraryManager.cs
+++ b/pwiz_tools/Skyline/Model/IonMobility/IonMobilityLibraryManager.cs
@@ -17,6 +17,7 @@
  * limitations under the License.
  */
 using System.Collections.Generic;
+using pwiz.Skyline.Model.DocSettings;
 using pwiz.Skyline.Model.DocSettings.Extensions;
 using pwiz.Skyline.Util;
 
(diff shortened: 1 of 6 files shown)


pwiz_tools/Skyline/Model/IonMobility/IonMobilityLibraryManager.cs
```diff
@@ -124,9 +125,11 @@ private IonMobilityLibrary LoadIonMobilityLibrary(IDocumentContainer container,
         private static IonMobilityLibrary GetIonMobilityLibrary(SrmDocument document)
         {
             if (document == null)
```
nickshulman, line 127: I would have probably done:
            return document?.Settings?.TransitionSettings?.IonMobilityFiltering?.IonMobilityLibrary 
                   ?? IonMobilityLibrary.NONE;

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/Model/IonMobility/IonMobilityLibraryManager.cs
bspratt: That is tidy. I'll make the change and rerun tests before merging.

PR #1338, opened 2020-12-02
Move "fragmentIons not in combo box" message out of resx

chambm:
* moved "fragmentIons not in combo box" message out of resx to an inline verbatim string; it is intended to catch programming errors


brendanx67 (changes requested):
I would rather just leave it alone for 20.2 and fix this only in master. We are only fixing really serious issues in 20.2. Regardless of the probability of breaking anything it is always non-zero. So, we only take any risk with important fixes in patch releases.

PR #1339, opened 2020-12-03
Skyline: fix a problem with manual renaming after pasting a protein sequence into Targets tree

bspratt:
fix for issue https://skyline.ms/announcements/home/support/thread.view?rowId=49107 in which:

  - the user pastes a protein sequence into the Targets tree, it gets the default name "sequence1"
  - the user types in a name to replace "sequence1" while background protein metadata search is still underway
  - after a few seconds the metadata search completes, and the displayed name mysteriously reverts to "sequence1"

N.B. as this is timing dependent, and our automated tests are mandated to not require internet access, writing a test for this fix (i.e. adding timings to the fake web lookup system) proved to be tricky and finally deemed not worth the effort for a fairly obscure problem.

Reported by Gabriele Philipps


nickshulman (approved):
Looks good to me.

A different way that you could have fixed this might have been change _processedNodes so that instead of it's key being the node.Id.GlobalIndex, it included the node's protein metadata. That way, if the node's metadata changed before the search was completed, it would look like no search had been done, and you'd have to send out a new request.

bspratt:
That is a more general and future proof approach (handles, for example, allowing the user to manually edit other metadata beyond just the name). I'll try that before making a final merge.

bspratt:
Eh, upon reflection I think this is enough of a corner case that it's probably best to just move on. If we ever do start allowing direct manipulation of protein metadata in the document grid we can revisit this.

PR #1346, opened 2020-12-07
Allow negative CE in EditCustomMoleculeDlg

kaipot:


bspratt (commented):
I think we still want to leave it blank for zero values.

kaipot:
> I think we still want to leave it blank for zero values.

Good call, I fixed that.

PR #1347, opened 2020-12-09
Fix Mascot fixed mods in idconvert and BiblioSpec

chambm:
* fixed BiblioSpec treating Mascot fixed terminal mods as not overridden by variable terminal mods
* fixed IdentData (idconvert) not overriding Mascot fixed mods with variable mods


kaipot (approved):
Looks good to me

chambm:
I think the new code ignores some complexities with Mascot mods. Exclusive mods (put in by quantitation methods like SILAC), query-level mods (presumably from spectral library search), multiple mods per site when MULTI_SITE_MODS=1, are all cases we probably break on. I really, really wish Mascot Parser just provided a function to enumerate mods on a peptide.

PR #1349, opened 2020-12-10
Workaround for first IDE build after a clean quickbuild of Skyline

chambm:
* added workaround for first IDE build messing up DLL file copies after a clean quickbuild of Skyline

If you want to build for running in the IDE, use the pwiz_tools/Skyline//Build target.

Tthe existing pwiz_tools/Skyline//Skyline.exe target will continue to leave Skyline files in Release directory.

The Build target will clean them which works around the file copy problem.


bspratt (approved):
Seems good, thanks.

PR #1351, opened 2020-12-11
* added TargetName check in UpdatePrositConfiguration.bat

chambm:
* added TargetName check in UpdatePrositConfiguration.bat so building official "Skyline" isn't allowed if PrositConfig_production.xml is missing

Brendan mentioned moving the batch file logic to bjam, which is definitely possible, but this seemed simpler. (Actually I first tried to put it in the SetupDeployProject script but that didn't work because it runs after the bjam action builds Skyline.)


nickshulman (commented):
I am not sure I understand this change, but I think we should be doing this check regardless of whether it's Skyline or Skyline-Daily.

Brendan would never want to release an official Skyline-Daily which did not have support for Prosit.

chambm:
Good point. The --official flag to quickbuild currently changes the Skyline TargetName from Skyline-daily to Skyline, but that too is somewhat misnamed. I suppose there should be 2 flags, one for --official and one for --skyline-release (--release would be too close to variant=release which is a bjam C++ feature that can be abbreviated as just 'release'). Then this script should be checking for the --official flag rather than the TargetName. And then TargetName would changed based on --skyline-release, and this script wouldn't care about TargetName.

chambm:
OK, I implemented the logic described above. So, assuming you have a quickbuild wrapper that builds Skyline like `skyline-build.bat` (or `b64.bat`, whatever):
- Build development, Skyline-daily build:  `skyline-build.bat`
- Build development, Skyline build: `skyline-build.bat --skyline-release`
- Build production, Skyline-daily build: `skyline-build.bat --official` (won't work without PrositConfig_production.xml)
- Build production, Skyline build: `skyline-build.bat --official --skyline-release` (won't work without PrositConfig_production.xml)

- Build production, Skyline-daily admin installer: `skyline-build.bat --official pwiz_tools/Skyline/Executables/Installer//setup.exe`
- Build production, Skyline build: `skyline-build.bat --official --skyline-release pwiz_tools/Skyline/Executables/Installer//setup.exe`

nickshulman (approved):

chambm:
The final version uses SKYLINE_BRANCH to decide whether the TargetName will be Skyline or Skyline-daily. The --official flag will check that Prosit_production.xml is available or the build will fail.

PR #1354, opened 2020-12-12
Move Refine menu items into a separate class "RefineMenu"

nickshulman:
I was thinking that the SkylineWindow class is too big, and we should move the menu items their associated code into separate files.

So, I was thinking we would make a UserControl called "RefineMenu", and Skyline's constructor would create it, and ask for the menu items that need to go on the Refine menu.

So, here's what it looks like in the designer:
image

The Skyline design view has a "Refine" menu item with no children.

In SkylineWindow constructor, it constructs as RefineMenu and calls:
refineToolStripMenuItem.DropDownItems.AddRange(RefineMenu.DropDownItems.ToArray());

Does this seem like a reasonable thing to be doing?
We can then apply this same technique to moving other chunks of menu-associated functionality into separate classes.


bspratt:
I've also often thought that was getting to be a pretty big file. I like this move.

brendanx67 (approved):
Looks good to me. Definitely a good idea to find ways to break this massive file up before our tools just give up and force us to do the work.

PR #1355, opened 2020-12-13
Fix for the duplicate audit log hash problem

rita-gwen:


nickshulman:
I think you should make it so that AuditLogEntry has a DocumentFormat on it, which it uses to decide how the hash should be calculated.
AuditLogEntry.ChangeDocumentFormat(DocumentFormat newFormat) would return a new AuditLogEntry whose parent has also had its format changed, and has the hash calculated in the manner appropriate for that format.

So, then, instead of having to pass the DocumentFormat up using XmlTextWriterWithState you would just be calling WriteXml on an object which already knew what format it was supposed to be.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -300,24 +328,36 @@ public static bool ReadFromXmlTextReader(XmlTextReader reader, out string logged
...
-                result.AuditLogEntries.Enumerate().Where(entry => !entry.Hash.SkylAndActualHashesEqual()).ToArray();
 
-            if (docFormat != null && (!result.RootHash.SkylAndActualHashesEqual() || modifiedEntries.Length > 0))
+            // recalculate the actual hash for the current document format
+            var getHashes = new Func<AuditLogEntry, KeyValuePair<AuditLogHash, AuditLogEntry>>(e =>new KeyValuePair<AuditLogHash, AuditLogEntry>(e.Hash, e));
+            if (!DocumentFormat.CURRENT.Equals(result.FormatVersion))
+                getHashes = new Func<AuditLogEntry, KeyValuePair<AuditLogHash, AuditLogEntry>>(
+                    e => new KeyValuePair<AuditLogHash, AuditLogEntry>(new AuditLogHash(e, e.Hash.SkylHash, docFormat), e));
+
+            var hashes = result.AuditLogEntries.Enumerate().Select(getHashes);
```
nickshulman, line 351: Can you change this to a "foreach" with an if statement inside of it?
I think  that would be much easier to undertand compared to having "getHashes" being a Func<AuditLogEntry...>.

nickshulman (commented):
I think my comment about adding DocumentFormat to AuditLogEntry was not helpful.
I think the actual thing to do long term is remove the Hash from AuditLogEntry, and only calculate it during loading and saving, but doing so might be more work than is appropriate for this pull request.

rita-gwen:
I created an issue to track the suggested fix: https://skyline.ms/issues/home/issues/details.view?issueId=765

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
rita-gwen: Yes, I rewrote this piece, it was a bit clumsy.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -300,24 +328,38 @@ public static bool ReadFromXmlTextReader(XmlTextReader reader, out string logged
...
-            // We can't just always ignore non-existent hashes, otherwise people could just delete the hash elements
+            // Root and entry-level hashes validation
+            // We can't ignore non-existent hashes, otherwise people could just delete the hash elements
             // and get Skyline to successfully load the audit log
-            var modifiedEntries =
-                result.AuditLogEntries.Enumerate().Where(entry => !entry.Hash.SkylAndActualHashesEqual()).ToArray();
 
-            if (docFormat != null && (!result.RootHash.SkylAndActualHashesEqual() || modifiedEntries.Length > 0))
+            // recalculate the hashes for the current document format
+            var hashes = result.AuditLogEntries.Enumerate().Select(e =>
```
nickshulman, line 346: I think this would be easier to debug if you were calculating the hash and checking whether it was valid in the same pass through the loop.

So, you would have:

            var hashes = new List<AuditLogHash>();
            var entriesWithIncorrectHash = new List<AuditLogEntry>();
            foreach (var entry in result.AuditLogEntries.Enumerate())
            {
                AuditLogHash hash;
                if (DocumentFormat.CURRENT.Equals(docFormat))
                    hash = entry.Hash;
                else
                    hash = new AuditLogHash(entry, entry.Hash.SkylHash, docFormat);
                if (!hash.SkylAndActualHashesEqual())
                {
                    entriesWithIncorrectHash.Add(entry);
                }
                hashes.Add(hash);
            }

nickshulman (approved):
Looks good.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
```diff
@@ -1457,8 +1510,12 @@ public byte[] GetAuditLogHash()
                 encodedBytes.Add(enc.GetBytes(SkylineVersion));
                 encodedBytes.Add(GetTimeStampBytesForHash());
                 encodedBytes.Add(enc.GetBytes(TimeZoneOffset.TotalHours.ToString(CultureInfo.InvariantCulture)));
-
-                // Avoid heap thrash performance issue by carefully allocating hash buffer size
+                if (Parent != null && !Parent.IsRoot && docFormat >= DocumentFormat.SEQUENTIAL_LOG_HASH)
+                {
+                    encodedBytes.Add(Convert.FromBase64String(Parent.Hash.ActualHash.HashString));
```
nickshulman, line 1515: This line causes a StackOverflowException in TestBlibMoleculeLists in the 32 bit build.

nickshulman (changes requested):

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
rita-gwen: Code updated.

pwiz_tools/Skyline/Model/AuditLog/AuditLogEntry.cs
rita-gwen: Fix committed.

nickshulman:
These changes have been rolled into #1385 which has been merged to master.

PR #1358, opened 2020-12-17
* fixed ImportFastaControl not checking for existence of filepath

chambm:
* fixed ImportFastaControl not checking for existence of filepath before letting user continue, and in DDA search mode, disabled multiline FASTA control even before Browse button is clicked, and made Clear button not restore to multiline mode
* added a generic string-only override for ImportFastaHelper.ShowFastaError


brendanx67 (approved):
Surprising to find PasteDlg and ImportFastaControl related in this way, but if they are, this seems like a reasonable fix.

chambm:
Yeah I was a little surprised to find ImportFastaHelper in PasteDlg as well. Kind of makes sense given its FASTA pasting history though.

PR #1364, opened 2020-12-18
SkylineNightly: exercise Turkish instead of French on pass2+ tests (F…

bspratt:
…rench gets a thorough testing in pass1, and both exercise the same decimal format), and add Turkish to the nightly perf test language rotation


pwiz_tools/Skyline/TestRunner/Program.cs
```diff
@@ -731,9 +732,20 @@ private static DirectoryInfo GetSkylineDirectory()
                 runTests.Log("# Pass 2+: Run tests in each selected language.\r\n");
             }
 
+            // As fr and tr exercise pretty much the same things (numeric formats, specifically),
+            // and as pass1 already gives fr a thorough check, switch to testing just tr in pass 2+
+            if (asNightly && pass >= 2 && 
+                inUseLanguages.Any(l => l.StartsWith("fr")) && 
+                inUseLanguages.Any(l => l.StartsWith("tr")))
+            {
+                var listLanguages = inUseLanguages.ToList();
```
nickshulman, line 741: I would have probably done:
                inUseLanguages = inUseLanguages.Where(l => !l.StartsWith("fr")).ToArray();

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/TestRunner/Program.cs
```diff
@@ -731,9 +732,20 @@ private static DirectoryInfo GetSkylineDirectory()
                 runTests.Log("# Pass 2+: Run tests in each selected language.\r\n");
             }
 
+            // As fr and tr exercise pretty much the same things (numeric formats, specifically),
+            // and as pass1 already gives fr a thorough check, switch to testing just tr in pass 2+
+            if (asNightly && pass >= 2 && 
+                inUseLanguages.Any(l => l.StartsWith("fr")) &&
```
brendanx67, line 738: How about use AreEqual(inUseLanguages, qualityLanguages)? Otherwise, I think this could get a bit strange if I am using the tests tab and choose all 5 languages to have it stop running French on pass 2. We are really talking about the Nightyl/Quality case here.

brendanx67 (changes requested):
This needs to only happen for Nightly/Quality runs, and not arbitrary Tests tab runs where I may have chosen both French and Turkish.

pwiz_tools/Skyline/TestRunner/Program.cs
bspratt: If you're thinking of the case where "Run full quality pass" is ticked on the Tests tab, it's already ignoring your language selections, along with the "Off screen" checkbox, because that's what a Quality run does.  Only your test selections are honored.

Ticking "Run full quality pass"  on the Tests tab also introduces special Quality run behavior such as running perftests in one language only, where the language depends of the day of the year. 

Probably the actual fix here is to disable the "offscreen" and "Language" controls when "Run full quality pass" is ticked. I'd also move that and the "Run demo mode" over to the left, vertically aligned with all the others. It's too easy to miss where it is right now.

brendanx67: How about just:

if (qualityMode && pass >= 2)

I also find myself wondering if we should just remove "tr" instead of "fr", since "tr" will be covered by pass 1 now. I don't have a strong opinion there, though.

I do just want this behavior only when we are using the set of languages calculated as "qualityLanguages" and never any other time.

brendanx67: I don't think the other changes you mention warrant our attention right now.

bspratt: >> since "tr" will be covered by pass 1 now

I guess I missed that part of the plan - at the moment pass 1 is still fr only. I thought that was the logic in dropping fr for pass 2+.

bspratt: OK, as discussed this now just adds tr to pass 1, and to the language rotation for perftests.

It also points out several tests that break under tr locale. Do we want to fix those in this branch before merging to trunk?

brendanx67: Yes. We always want to avoid creating an unnecessary emergency by breaking the nightly tests, when we have any way of helping that.

bspratt:
Everything passes under tr now, between these changes and Matt's MSAmanda fix.

PR #1367, opened 2020-12-22
Fixes for CalibrateIrtDlg

kaipot:


kaipot:
Made a few changes after the demo over zoom (table allows rows to be deleted but otherwise is readonly)

pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
```diff
@@ -782,14 +822,21 @@ public void SetFixedPoints(int one, int two)
             comboMaxPeptide.SelectedIndex = two;
         }
         #endregion
+
+        private void gridViewCalibrate_CellFormatting(object sender, DataGridViewCellFormattingEventArgs e)
+        {
+            if (e.Value is double value && double.IsNaN(value))
+            {
+                e.Value = !double.IsNaN(value) ? value.ToString(Formats.RETENTION_TIME, CultureInfo.CurrentCulture) : string.Empty;
```
bspratt, line 830: Isn't it the case that double.IsNaN will always be false here?

pwiz_tools/Skyline/SettingsUI/PeptideGridViewDriver.cs
```diff
@@ -46,6 +46,7 @@ protected PeptideGridViewDriver(DataGridViewEx gridView, BindingSource bindingSo
         }
 
         protected bool AllowNegativeTime { get; set; }
+        protected bool AllowMissingTime { get; set; }
```
bspratt, line 49: Maybe a comment explaining why this is here?

bspratt (changes requested):
Just a couple of notes, seems good and you've already demo'd for us.

pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
kaipot: Good catch, fixed

pwiz_tools/Skyline/SettingsUI/PeptideGridViewDriver.cs
kaipot: Added comment

PR #1374, opened 2020-12-29
Skyline: remove chromatogram load timeout code that was specific to a vendor reader which has since been updated

bspratt:
We probably should have made this change when Thermo updated their reader. As it is, it interferes with code coverage testing so I'm removing it.


chambm:
If that doesn't fix it, you may be able to speed up the test cycle by copying the build config and changing it to only run the failing test (assuming it still fails when run by itself).

brendanx67 (approved):
Looks good to me. Thanks for tracking down the challenging bugs in printf varargs use and variable initialization.

pwiz_tools/BiblioSpec/src/ProteinPilotReader.cpp
```diff
@@ -487,20 +494,24 @@ void ProteinPilotReader::addElement(double& mass, string element, int count){
...
 void ProteinPilotReader::addMod(){
     // first check to see if we already have one for this mod
     map<string, double>::iterator found = modTable_.find(curMod_.name);
     if( found != modTable_.end() && //if it was found and has different mass
         found->second != modTable_[curMod_.name] ){
         throw BlibException(false, "Two entries for a modification named %s,"
-                            "one with delta mass %d and one with %d.",
-                            found->second, modTable_[curMod_.name]);
+                            "one with delta mass %f and one with %f.",
+                            curMod_.name, found->second, modTable_[curMod_.name]);
```
chambm, line 509: I was having trouble reproducing this error with /RTCs; I had to attach a debugger to understand why the error wasn't triggering. Turns out that, ahem, `found->second` is always equal to modTable_[curMod_.name] because we got the found itr by searching modTable_ for that name. :) It should be `found->second != curMod_.deltaMass`. So now I'm wondering why this error would EVER trigger with the current logic. Floating point instability perhaps? The uninitialized values usually have extreme exponents.

chambm: It's amazing how long you can stare at code and not see seemingly obvious errors... different floating point behavior under VM may explain why we saw the error on TC but not our bare metal machines.

PR #1376, opened 2020-12-30
Fix ImportPeptideSearchDlg height

kaipot:
Here we only increase the height of the ImportPeptideSearchDlg if the built library contains ion mobilities.
(Side note: we can't decrease the height when selected from the start page even though the workflow group box is hidden, because it cuts off stuff on the Full Scan settings page.)


brendanx67 (approved):
Thanks, Kaipo.

PR #1382, opened 2021-01-05
* added SkylineBatch to Skyline admin installer

chambm:
* fixed SkylineBatchTest to work from Skyline's build directory as well as its own build dir (i.e. when it's run from IDE)


chambm:
@alinmarsh @brendanx67 The build agent doesn't have R installed. But Skyline has an RInstaller class. Why doesn't SkylineBatch use that?

alinmarsh:
> @alinmarsh @brendanx67 The build agent doesn't have R installed. But Skyline has an RInstaller class. Why doesn't SkylineBatch use that?

My understanding is RInstaller is more set up for finding and installing a specific version of R. SkylineBatch finds **all** installed versions of R to allow the user to customize the version they want to use for each individual script. However, SkylineBatch will display an error if there are no R installations so maybe I could use RInstaller in this case?

brendanx67:
Ali is right. In this case, we assume that the user of SkylineBatch has set up their own R with whatever script they want to run and they just want to point SkylineBatch at the right location to run it. We are not trying to offer R installation help, as we do with external tools. This is supposed to be a replacement for a set of batch scripts (*.bat files) that I have used a lot and even posted with webinars on the Skyline website. I have generally worked with my own R scripts in R Studio while working with these batch scripts.

chambm:
This is ready to review/merge then. The SkylineBatchTest has a bjam target but it isn't run by default or on TeamCity.

pwiz_tools/Skyline/Executables/Installer/Product-template.wxs
```diff
@@ -114,12 +114,20 @@
         <RemoveFolder Id="ApplicationProgramsFolder" On="uninstall"/>
         <RegistryValue Root="HKCU" Key="Software\MacCossLabUW\Skyline64" Name="installed" Type="integer" Value="1" KeyPath="yes"/>
       </Component>
+      <Component Id="ApplicationBatchShortcut64">
+        <Shortcut Id="ApplicationBatchMenuShortcut32" Name="$(var.ProductName) Batch" Target="[#SkylineBatch_exe]"/>
```
nickshulman, line 118: "$(var.ProductName) Batch" ends up appearing on the Start Menu as:
Skyline-Daily (64 bit) Batch
I think we should change this to:
"$(var.Skyline.TargetName) Batch (64)"
which I think would end up looking like:
Skyline-Daily Batch (64 bit)

nickshulman (commented):
Looks pretty good.

I had to add the files "SkylineRunner.exe" and "SkylineDailyRunner.exe" so that the build would work. We should decide on some place in the source tree that those files should actually be checked in, and AutoQC and SkylineBatch should both include those files as Links their respective .csproj files.

PR #1383, opened 2021-01-06
Skyline: our custom code inspection test now treats .resx files with nonstandard AutoScaleDimensions as an error. Each developer should fix this, on this branch, as assigned at https://docs.google.com/spreadsheets/d/1D-B3VlfLnUBh5NuQfUIT-zGIIwWmKGYrtKK8-H5ubPs/edit#gid=0

bspratt:
They should all look like this:

  <data name="$this.AutoScaleDimensions" type="System.Drawing.SizeF, System.Drawing">
    <value>6, 13</value>
  </data>


brendanx67 (changes requested):
Looks too symplistic. You will need to look at each of these forms on a 100% monitor. Simply flipping the scaling like this will result in non-standard dimensions for buttons and potentially worse problems.

bspratt:
I suppose you're right - I obviously just did a quick search and replace while I was waiting for some tests to run. If we want this taken care of properly we're going to need to mandate that folks fix the code they wrote - I don't have the bandwidth to look at and mess with all 30+ instances. One way to achieve closure here is to just make this an error so that everybody sees that there's a problem, then harangue the team until it stops failing. I know you don't like to see nightlies with any errors at all but I do think this is the best social engineering approach.

PR #1387, opened 2021-01-07
Skyline (20.2): Fix "isolation" text in Chinese RESX files and update…

brendanx67:
… DIA/SWATH tutorial screenshots
- Remove italicized text in Chinese RESX files


brendanx67:
Decided to stay with the original Chinese translation for "isolation" in use in the UI for years.

PR #1390, opened 2021-01-08
Skyline: Add UI for setting AbstractFunctionalTest.PauseStartPage to SkylineTester

brendanx67:


bspratt (changes requested):
I was initially confused by "StartPage" since that's what we call the window that normally pops up when you open Skyline. Perhaps PauseStartPage to match what it's setting in AbstractFunctionalTest?

Otherwise seems fine, I'm sure you've given it a good manual test since you developed it for your own work.

PR #1392, opened 2021-01-11
Deconvolute precursor area, display deconvoluted chromatograms and a few other things

nickshulman:
This pull request has a few things that are eventually going to be useful for MSstats:
New columns in Document Grid:
**TransitionResult.TransitionResultIsQuantititative** : almost the same as the user-settable column "Transition.Quantitative", but it's false if the transition is MS2 and the acquisition method is "DDA".
**TransitionResult.TransitionResultIsMs1**: true if the transition came from an MS1 scan
**PrecursorResult.DeconvolutedPrecursorArea**: isotope deconvoluted ms1 area.

Also added View>Transform menu item to show deconvoluted chromatograms:
Here is what some chromatograms look like normally:
image
And then you can click the "Deconvolute Isotopes" on the right-click menu:
image
And then it looks like this:
image
It's a pretty subtle change.

Here's how the deconvoluted areas compare with the traditional total areas:
image


pwiz_tools/Skyline/Controls/Graphs/GraphChromatogram.cs
```diff
@@ -2060,6 +2081,88 @@ private bool IsQuantitative(TransitionDocNode transitionDocNode)
             return transitionDocNode.IsQuantitative(DocumentUI.Settings);
         }
 
+#if false
```
brendanx67, line 2084: If you really want to keep this, add a comment explaining it.

pwiz_tools/Skyline/Menus/ChromatogramContextMenu.resx
```diff
@@ -117,9 +117,12 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <metadata name="contextMenuChromatogram.TrayLocation" type="System.Drawing.Point, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
+  <metadata name="modeUIHandler.TrayLocation" type="System.Drawing.Point, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
```
brendanx67, line 120: Changes to this file seem a bit gratuitous for adding one menu item.

pwiz_tools/Skyline/Menus/ViewMenu.resx
```diff
@@ -124,6 +124,43 @@
     <value>155, 17</value>
   </metadata>
   <assembly alias="System.Drawing" name="System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
+  <data name="menuStrip1.Location" type="System.Drawing.Point, System.Drawing">
```
brendanx67, line 127: Again, wow, Visual Studio sure does modify RESX files a lot for small changes.

pwiz_tools/Skyline/Model/Databinding/Entities/ColumnToolTips.resx
```diff
@@ -294,6 +294,9 @@ values.</value>
     <value>Declustering Potential for the precursor ion according to
 instrument/vendor specific default declustering potential equation within Skyline</value>
   </data>
+  <data name="DeconvolutedPrecursorArea">
+    <value>The observed MS1 area adjusted based on the predicted isotope distribution of the precursors</value>
```
brendanx67, line 298: Explain more clearly. e.g. Does it only impact overlapping isotope distributions like deuterated Leucine, or does it impact even a single precursor? What is it doing? A regression of some sort, yes? Is it a regression on the log values? or the linear values? I assume the isotope distribution is the independent value and the measured are the dependent value? And then you use the resulting equation measured-area = isotope-proportion * b1 + b0?

pwiz_tools/Skyline/Model/Databinding/Entities/ColumnToolTips.resx
```diff
@@ -1276,6 +1279,12 @@ using the Results Grid.</value>
   <data name="TransitionResult">
     <value>Transition Result</value>
   </data>
+  <data name="TransitionResultIsMs1">
+    <value>Whether this result was obtained from an MS1 scan.</value>
+  </data>
+  <data name="TransitionResultIsQuantitative">
+    <value>Whether this result is treated as quantitative, based on the Quantitative property of the Transition and the Full Scan Acquisition Method.</value>
```
brendanx67, line 1286: Explain more clearly. Maybe, "the user-modifiable Quantitative property" ... "Full Scan Acquisition Method, where MS/MS fragment ions are non-quantitative for DDA." Or something that makes it clearer why this property is needed and what really it does.

pwiz_tools/Skyline/Model/Databinding/Entities/PrecursorResult.cs
```diff
@@ -96,6 +98,68 @@ public void ChangeChromInfo(EditDescription editDescription, Func<TransitionGrou
...
+                    }
+                }
+
+                if (!myIndex.HasValue)
+                {
+                    return null;
+                }
+
+                var minMz = deconvolutionKeys.Min(key => key.TransitionGroupDocNode.PrecursorMz.Value);
+                // TODO(nicksh): decide the window width based on Full Scan settings
```
brendanx67, line 156: Not much time to do this left. I think we need to do a pass like the VC++ team used to do to expunge all "TODO" comments from the code and make them either "CONSIDER" (if there is an issue that might have meaning but may not be necessary to handle ever), or "FUTURE" if it is something that would be "nice to do some day" (but maybe never).

pwiz_tools/Skyline/Model/Databinding/Entities/TransitionResult.cs
```diff
@@ -116,6 +116,22 @@ public string Note
...
+            {
+                return Transition.DocNode.IsQuantitative(SrmDocument.Settings);
+            }
+        }
+
+        public bool TransitionResultIsMs1
+        {
+            get
+            {
+                return SrmDocument.Settings.GetChromSource(Transition.DocNode) == ChromSource.ms1;
```
brendanx67, line 131: Probably this should be != ms2, especially since you widened SIM to be <1000 m/z. It is very possible that MS1 spectra show up with ChromSource.sim.

pwiz_tools/Skyline/Properties/Resources.resx
```diff
@@ -11317,12 +11317,21 @@ Example: Crosslink to PEPTIDE: T [4]</comment>
   <data name="SmallMoleculeTransitionListReader_GetMoleculeTransitionGroup_Inconsistent_molecule_description" xml:space="preserve">
     <value>Inconsistent molecule description</value>
   </data>
+  <data name="DiaUmpireDdaConverter_Run_Starting_DIA_Umpire_conversion" xml:space="preserve">
```
brendanx67, line 11320: I am a little suspicious of all of the changes in this file because DIA-Umpire and DDASearchControl changes show up. You may be duplicating entries in this file. I had that problem with an auto-merge I did after Matt's commit.

pwiz_tools/Skyline/Properties/Settings.settings
```diff
@@ -761,5 +761,11 @@
     <Setting Name="DetectionsShowLegend" Type="System.Boolean" Scope="User">
       <Value Profile="(Default)">True</Value>
     </Setting>
+    <Setting Name="DisplayDeconvolutedChromatograms" Type="System.Boolean" Scope="User">
+      <Value Profile="(Default)">False</Value>
+    </Setting>
+    <Setting Name="AreaCVRatioIndex" Type="System.String" Scope="User">
```
brendanx67, line 767: Why would you be adding this value?

pwiz_tools/Skyline/Skyline.csproj
```diff
@@ -5289,162 +5293,162 @@
   <ItemGroup>
     <PublishFile Include="BiblioSpec.pdb">
       <Visible>False</Visible>
+      <PublishState>Include</PublishState>
```
brendanx67, line 5296: Revert all these Include, True changes to avoid fighting over this ordering with other developers.

pwiz_tools/Skyline/app.config
```diff
@@ -234,9 +234,15 @@
             <setting name="AreaGraphType" serializeAs="String">
                 <value>replicate</value>
             </setting>
+            <setting name="DetectionGraphType" serializeAs="String">
```
brendanx67, line 237: Why would you need to change so much in this file? Have others failed to update it?

pwiz_tools/Skyline/app.config
```diff
@@ -719,14 +776,17 @@
...
             </setting>
+            <setting name="InstalledVersion" serializeAs="String">
+                <value />
+            </setting>
         </pwiz.Skyline.Properties.Settings>
     </applicationSettings>
   <system.data>
     <DbProviderFactories>
       <remove invariant="System.Data.SQLite"/>
-      <add name="SQLite Data Provider" invariant="System.Data.SQLite" description=".Net Framework Data Provider for SQLite" type="System.Data.SQLite.SQLiteFactory, System.Data.SQLite,
```
brendanx67, line 727: Leave this block as it was.

brendanx67 (changes requested):
Some comments. Mostly not super important. I don't see any tests. This kind of change can't be merged without tests.

pwiz_tools/Skyline/Model/Databinding/Entities/ColumnToolTips.resx
nickshulman: Good point. I'll try to come up with a better sentence.

If we are fitting only one isotope distribution, then it is a least squares linear regression through the origin. The X coordinates are the predicted distribution values (normalized so that the total distribution sums to 1), and the Y coordinates are the observed peak areas. The fitted slope is the number that gets reported as the DeconvolutedPrecursorArea.
There is no y-intercept term. This happens to be appropriate because the Y coordinates were the background subtracted areas.

When fitting multiple overlapping isotope distributions, it's a least squares fit. It's described in the 1966 paper Least Squares Analysis and Simplification of Multi-Isotope Mass Spectra J. I. Brauman.

brendanx67: Okay, I get it. Y = mx1 + mx2 + mx3 where x1 + x2 + x3 = 1 means that Y = m. But, yeah, I think you need to mention "linear regression through the origin" in the description. I think most readers would assume "linear regression" implies "least squares". Maybe even reference "(Brauman, Anal. Chem. 1966)" for overlapping isotope distributions. That gets me there with a Google search.

nickshulman:
It's pretty hard for me to find a case where "DeconvolutedPrecursorArea" is sufficiently different from "TotalAreaMs1" to really make a difference. I think I should delay trying to get this work committed to master.
I think the only things that we should do this version are:
TransitionResult.TransitionResultIsQuantititative
TransitionResult.TransitionResultIsMs1

These are things that I think MSstats should start using. MSstats can continue to calculate the precursor area by summing up the MS1 transitions areas.
I will open a new pull request with just those two new TransitionResult properties.

PR #1395, opened 2021-01-11
Give the user an error "The file appears to be corrupted" if the head…

nickshulman:
…er at the end of the .skyd file has impossible values.

An AutoQC user had a .skyd file that was corrupted.
They were getting an error message "Array dimensions exceeded supported range".

With this change, they will get an error message which looks like this:
image

(Is that the best wording for this message? Should it say anything more? Should it tell them to delete the file?)


vagisha:
Yes, I think letting the user know that the only way to recover is to delete the .skyd file and reimport the raw data would be helpful.  
Thanks a lot for adding a check for corrupted data!

vagisha (approved):

brendanx67 (approved):
One minor text change. Looks good.

PR #1396, opened 2021-01-12
Fix "AutoScaleDimensions" on several .resx files.

nickshulman:


bspratt:
I was kinda thinking we'd do these on the branch that treats them as errors, but that's not super important. 

And probably not even a very good idea if that branch is very long lived, so maybe this is better.

brendanx67:
Sorry to piggyback on your PR, but I figured it would save some time to role in the RESX files I edited during the Monday meeting.

Thanks for pushing this forward, Nick and Brian.

brendanx67:
Once all of these changes are completed, however, someone should test how the forms look in all 3 languages using SkylineTester. It should not take more than 15 minutes using the Forms tab.

bspratt:
In that case I'm going to move these changes to the branch, where it will be easier to figure out which dialogs need that final manual testing step. Nick, please don't merge this - nor destroy it quite yet

bspratt (changes requested):
Given that we want to do a manual check of all these changes, I'm moving these to the branch that deals with this issue. Please do not merge this to master, nor deleted it quite yet.

PR #1398, opened 2021-01-12
Make it so that "ModeUIExtender" does not implement ISupportInitialize

nickshulman:
Remove modeUIHandler ISupportInitialize calls from .designer.cs files.

ModeUIExtender does not make use of any of the ISupportInitialize functionality.
I think we should remove it since it adds unnecessary complexity to our .designer.cs files.

Visual Studio Designer adds the BeginInit and EndInit calls whenever it notices that the thing supports ISupportInitialize. Eventually, those calls will get added to all of our .designer.cs files. But, if we remove that interface from this class now, we can prevent all those calls from making it into the code base.


bspratt:
Sounds good. I was noticing that it created extra churn when updating the AutoScaleDimensions stuff.

bspratt (approved):

PR #1399, opened 2021-01-12
Skyline/work/20210111 inspect code for model dependency on UI

bspratt:
This checks for code in the model namespace that uses UI namespace (Skyline..*UI, Skyline.Controls, Skyline.Alerts).

I've picked off some of the easier refactorings.


nickshulman (approved):
Looks good to me.
I probably would have moved "ViewLibraryPepInfo" into Model instead of moving "ViewLibraryPepInfoList" into SettingsUI, but this way works too.

brendanx67 (approved):
I leave it up to you and Nick on the ViewLibraryPepInfoList. It will be nice to have this check and work toward enforcing it fully.

bspratt:
I'm going to go with Nick's suggestion.

PR #1408, opened 2021-01-14
Increase nightly hang threshold time and only allow 4 hours if hang occurs 9am-5pm

kaipot:


kaipot:
We should also detect if a hang ends (i.e. the log file starts getting written to again) and send a notification / reset the ending time.
However if someone has taken manual action that causes the log file to be written to, the end time should remain extended (or be extended even further) to prevent SkylineTester from being killed while someone is debugging. I'm not sure if there is a good way to distinguish whether a hang has ended naturally or due to manual action.

nickshulman (approved):
Looks good to me.

PR #1417, opened 2021-01-18
* added support for BiblioSpec to read .mzid.gz files

chambm:
* added support for BiblioSpec to read .mzid.gz files (pwiz can read them just like uncompressed files)
* changed MSAmandaWrapper's expected output filepath to the .mzid.gz file
* fixed ImportPeptideSearchDlg continuing to import library even when DDA search fails

@nickshulman Does the ImportPeptideSearchDlg change make sense to you? It didn't make sense to me to run CloseWizard() if the build library step failed (without cancellation). It makes more sense to me to let the user go back and try again, or let the user cancel explicitly from the wizard.

### DIFF
--- a/pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
+++ b/pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
@@ -621,8 +621,6 @@ namespace pwiz.Skyline.FileUI.PeptideSearch
                     if (!ImportFastaControl.ImportFasta(ImportPeptideSearch.IrtStandard))
                         return;
 
-                    ImportPeptideSearch.SearchEngine.Dispose();
-
                     WizardFinish();
                     return;
 
@@ -1039,6 +1037,10 @@ namespace pwiz.Skyline.FileUI.PeptideSearch
         {
             // Close file handles to the peptide search library
             ImportPeptideSearch.ClosePeptideSearchLibraryStreams(Document);
+
+            // Dispose DDA SearchEngine if it is initialized
+            ImportPeptideSearch.SearchEngine?.Dispose();
+
             DialogResult = result;
         }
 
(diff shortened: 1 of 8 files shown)


pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -1043,6 +1037,10 @@ public void CloseWizard(DialogResult result)
         {
             // Close file handles to the peptide search library
             ImportPeptideSearch.ClosePeptideSearchLibraryStreams(Document);
+
+            // Dispose DDA SearchEngine if it is initialized
+            ImportPeptideSearch.SearchEngine?.Dispose();
```
nickshulman, line 1042: This does not seem like the right place to put this line.
If the user closes the wizard by clicking the X button on the dialog, I think this code will not run.

Usually we put cleanup code like this in OnFormClosed.

nickshulman (commented):
It certainly sounds like an improvement to keep user in the wizard if there was an error.
Kaipo is the real expert on this wizard.

kaipot (approved):
Looks good to me, except as Nick mentioned we want to make sure SearchEngine gets disposed. Maybe OnFormClosed should call CloseWizard.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
chambm: Does the same logic also apply to ClosePeptideSearchLibraryStreams? But I agree it should definitely run if the user closes the wizard, so does Kaipo's suggestion work?

chambm: I guess I was assuming that the Cancel button and the X button would lead to the same result (why wouldn't they, at least from the user's perspective?)

nickshulman: Yes, ClosePeptideSearchLibraryStreams should go in OnFormClosed as well.

CloseWizard really should just be the one line:
DialogResult = result;
(or we could get rid of the function entirely and have the caller do that)

This code was written in 2012, and I think we did not know what we were doing back then.

PR #1418, opened 2021-01-19
Try to fix ImportResultsControl exception

kaipot:


brendanx67:
Kaipo, how have you tested this? In email, you said you could not reproduce it. I hope you have at least created a way to reproduce this case and tested to ensure that your new code works as expected. Never a good idea to simply code a solution you expect to work but cannot test.

kaipot:
> Kaipo, how have you tested this? In email, you said you could not reproduce it. I hope you have at least created a way to reproduce this case and tested to ensure that your new code works as expected. Never a good idea to simply code a solution you expect to work but cannot test.

Yeah, I only tested the "normal" code paths to make sure they hadn't changed and then the unexpected case by controlling execution with the debugger. But I guess in the unexpected state, other things might've changed as well besides the control that causes the exception.
I'll look into the code some more to see how to reproduce the bug so I can add a proper test for it. I noticed something weird which may have allowed the wizard to get into this state, related to DDA search like you mentioned.

kaipot:
It looks like you can go back to the first page with DDA search. The changes fixed the exception in this case but I noticed some other weird UI bugs if the user goes back to the first page and changes the workflow which I am looking into.

nickshulman (approved):
Looks good to me

PR #1424, opened 2021-01-25
Skyline: fix a problem (exception thrown) when importing results with…

bspratt:
… an ion mobility library when a precursor ion appears more than once in a document.

Per https://skyline.ms/announcements/home/support/thread.view?rowId=49618

Reported by user "Juan C. Rojas E."


brendanx67 (approved):
Great to get real-world feedback. Thanks for working through this on the support board

pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
```diff
@@ -1318,7 +1318,7 @@ public LibraryIonMobilityInfo GetIonMobilities(LibKey[] targetIons, MsDataFileUr
                     foreach (var ion in targetIons)
```
nickshulman, line 1318: I would have done:
foreach (var ion in targetIons.Distinct())
and skipped the dict.ContainsKey check below.

nickshulman (approved):

PR #1425, opened 2021-01-26
SkylineTester: pull perftest datasets from PanoramaWeb instead of Skyline.ms

bspratt:


chambm (approved):
Looks good to me, especially factoring out the URL to a function.

PR #1428, opened 2021-01-29
Allow specifying more than one crosslink between peptides.

nickshulman:
This change allows users to specify more than one crosslink between peptides.
The way that peptides are represented in the Document Tree is now a lot different. Now, a crosslinked peptide is a flat list of one or more peptides, and a flat list of one or more crosslinkers, where each crosslinker has two CrosslinkSite objects in it. A CrosslinkSite is an integer representing which peptide, and an integer representing the amino acid index.

(The old way of representing it was a tree, where nodes were identified by ModificationSite objects, which were modification name and amino acid index. The new way is much simpler)

As before, you bring up the crosslink editor by choosing a crosslink modification in the Edit Modifications dialog:
image

The Edit Linked Peptides editor has one grid for all of the crosslinks, and one grid for all of the linked peptides. Initially, the Linked Peptides grid is blank, and the Crosslinks grid allows you to choose which pairs of amino acids in the primary peptide are connected by a looplink.
image

If you want to crosslink to a new peptide, you can type its sequence into the grid below:
image

If there are any rows in the Linked Peptides grid, then two more columns appear in the upper grid which allow you to specify which peptides are linked together:
image

After you return to the Edit Modifications dialog, the crosslinks that you have chosen are displayed as modifications on the main peptide.
image


brendanx67:
I think it would help, if you could somehow highlight the linked AA in the peptide sequences in the Crosslinks grid. If you can't do rich text bolding and coloring as they will eventually show up in the Targets view, then maybe at least do something in ASCII text like put brackets around the linked AA.

1:KNIC[K]KNACK - 5:K

Other possible ways of bracketing in ASCII include:

{K}
(K)
<K>
>K<

Or even casing:

1:KNICkKNACK - 5:K

nickshulman:
> I think it would help, if you could somehow highlight the linked AA in the peptide sequences in the Crosslinks grid. If you can't do rich text bolding and coloring as they will eventually show up in the Targets view, then maybe at least do something in ASCII text like put brackets around the linked AA.

Were you thinking something like this?:
image

PR #1433, opened 2021-02-02
* added MutedTotalHandleLeakTestNames and MutedUserGdiHandleLeakTestN…

chambm:
@brendanx67 Looks like I already did try to mute it in release but missed a change to the BelowThresholds function which probably prevented it from working.


brendanx67 (approved):
Ah. Makes sense. Thanks for fixing and moving the mechanism into master for future use.

chambm:
@bspratt Should we expect code inspection to fail for release right now? It should be testing this branch merged into the 20_2 branch, not master.  We don't normally run code inspection on the release branch, right?

bspratt:
Those should not be failing. But in this case it's due to a TC upgrade changing the rules, rather than a change in the code. I'll see about quieting that.

bspratt:
>> We don't normally run code inspection on the release branch, right?
Yes, that's how TC is configured, come to think of it. I guess this is a different set of rules. Still no big deal to tweak this, I think.

PR #1434, opened 2021-02-02
Skyline: clean up a couple of issues with the new transition list import column picker

bspratt:
* remove quotes around user's header names
*  grey out text in ignored columns
*  add tooltips to columns 

Also try to fix an apparent race condition in tests where flow of control goes through the form.

image


nickshulman (approved):
Looks good to me.

PR #1440, opened 2021-02-05
Change the text color on the DataGridView cells to white when the background is dark

nickshulman:
Currently, the data grid view always lightens the background colors of the cells so that it is always readable with either black text color (default) or blue text color (for the hyperlinks). The colors look washed out like this:
image

This change makes it so that the grid uses the background colors that it was intended to, and changes the text color to white if the background color is dark:
image

This has potential, but I think I'm doing some things wrong which causes it to not be aesthetically pleasing.
I think my way of deciding whether a background color is "dark" is maybe not correct (I copied it from Stack Overflow). Also, the threshold at which we decide to change the text color should maybe depend on whether the text color is currently Black (for regular text) or Blue (for hyperlinks). Also, maybe the blue text should become a different color when we change it.


brendanx67:
I guess the simple question to start with is, why not just use the washed-out colors on both sides of the row headers? It makes sense that you want colors to match between the cells and the dendrograms, but why not just use the light colors on both?

nickshulman:
> I guess the simple question to start with is, why not just use the washed-out colors on both sides of the row headers? It makes sense that you want colors to match between the cells and the dendrograms, but why not just use the light colors on both?

If we were going to do that, then we should probably take a harder look at the method I wrote "LightenColor" which currently replaces each RGB component with the average of that component with 255. I am not sure whether that technique preserves the visual distinctness of the colors.

I downloaded a list of 1024 visually distinct colors from StackOverflow (in "pwiz.Common.Colors.ColorPalettes.LARGE_PALETTE"). I am not sure how well they remain visually distinct after I lighten them.

Also, the heat map graph itself uses the full strength colors. I am not sure whether we would want that graph to use washed out colors as well. If not, then we would want to make sure that my "LightenColor" algorithm changes the colors in a way that a human make sense to a human, since I always wanted the datagridview to server as a color key for the heat map.

brendanx67:
Maybe worth noting that Don wrote code to generate visually distinct colors at the peptide level for the multi-peptide graphs. I am not sure how light they are, but it probably also makes sense to use those colors in the rows, especially when the rows represent peptides. That way the colors would match the ones in the Targets view and multi-peptide graphs. My sense is that those colors are lighter than the palette you are using, but I don't really know whether they are light enough to show black text on top of.

brendanx67:
But, yes, I agree consistency is key here, and we should try for as much consistency as possible. If that requires white text, then this seems like a good start. Are the grid blues and reds as dark as the heatmap? Dark blue and dark red would require white text also, but maybe the heatmap also never gets that dark.

nickshulman:
> Are the grid blues and reds as dark as the heatmap? Dark blue and dark red would require white text also, but maybe the heatmap also never gets that dark.

If I don't use washed out colors on the grid, there are a few spots where the red gets dark enough to switch to white text (although we could change the threshold number at which it switches to white, and it would probably still look fine). The blue never gets that dark.
image

nickshulman:
I merged this pull request into 1492.

PR #1448, opened 2021-02-11
Retain isolation scheme in Import Peptide Search

kaipot:


pwiz_tools/Skyline/TestTutorial/DiaTutorialTest.cs
```diff
@@ -327,7 +327,7 @@ protected override void DoTest()
             RestoreViewOnScreen(27);
             PauseForScreenShot<GraphChromatogram>("Chromatogram graph metafile", 26);
 
-            RunUI(SkylineWindow.SelectedNode.Expand);
+            RunUI(() => SkylineWindow.SelectedNode.Expand());
```
brendanx67, line 330: This change seems irrelevant and is not changing what the test does. Probably should not be included.

brendanx67 (approved):
Looks good, as long as the DiaTutorialTest change is reverted.

pwiz_tools/Skyline/SettingsUI/FullScanSettingsControl.cs
```diff
@@ -953,7 +953,8 @@ public void ModifyOptionsForImportPeptideSearchWizard(ImportPeptideSearchDlg.Wor
 
                 ProductMassAnalyzer = PrecursorMassAnalyzer;
 
-                if (workflow == ImportPeptideSearchDlg.Workflow.dia && Settings.Default.IsolationSchemeList.Count > 1)
+                if (workflow == ImportPeptideSearchDlg.Workflow.dia && Settings.Default.IsolationSchemeList.Count > 1 &&
+                    settings.TransitionSettings.FullScan.IsolationScheme == null)
                 {
                     comboIsolationScheme.SelectedIndex = 1;
```
brendanx67, line 959: Also, you should add a comment here to clarify what this is doing. i.e. What does 1 mean?

kaipot:
> Looks good, as long as the DiaTutorialTest change is reverted.

Sorry I forgot to mention, this was a minor bug I ran into, when running the test in debug mode it caused a cross thread exception when it wasn't called from the anonymous function.

PR #1449, opened 2021-02-11
- fixed BiblioSpec's MaxQuant reader not applying terminal label-type mods

chambm:
Reported by Ida Suppanz

Tested as working on Ida's data (as well as after tweaking their mqpar.xml to use a C-terminal label instead of N-terminal).

### DIFF
--- a/pwiz_tools/BiblioSpec/src/MaxQuantReader.cpp
+++ b/pwiz_tools/BiblioSpec/src/MaxQuantReader.cpp
@@ -751,7 +751,6 @@ void MaxQuantReader::addFixedMods(vector<SeqMod>& v, const string& sequence, con
     // iterate over sequence
     for (int i = 0; i < (int)sequence.length(); i++)
     {
-        // check for label mods
         boost::range::insert(v, v.end(), getFixedMods(sequence[i], i + 1, modsAnywhere));
         if (i == 0)
             boost::range::insert(v, v.end(), getFixedMods(sequence[i], i + 1, modsAnyNTerm));
--- a/pwiz_tools/BiblioSpec/src/MaxQuantReader.cpp
+++ b/pwiz_tools/BiblioSpec/src/MaxQuantReader.cpp
@@ -682,7 +682,7 @@ void MaxQuantReader::storeLine(MaxQuantLine& entry)
 
     try
     {
-        addModsToVector(curMaxQuantPSM_->mods, entry.modifications, entry.modifiedSequence);
+        addModsToVector(curMaxQuantPSM_->mods, entry.modifications, entry.modifiedSequence, entry.sequence);
     }
     catch (const MaxQuantWrongSequenceException& e)
     {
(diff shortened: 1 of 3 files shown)


pwiz_tools/BiblioSpec/src/MaxQuantReader.cpp
```diff
@@ -869,7 +869,13 @@ void MaxQuantReader::addLabelModsToVector(vector<SeqMod>& v, const string& rawFi
                                    "raw file '%s'.",
                                    labelingState, labels->labelingStates.size(), rawFile.c_str());
     }
-    const vector<const MaxQuantModification*>& labelMods = labels->labelingStates[labelingState].mods;
+
+    const auto& labelState = labels->labelingStates[labelingState];
+    const auto& labelMods = labelState.mods;
```
nickshulman, line 874: If the user has an N-terminal amino acid which applies to a particular amino acid, that modification is going to get applied to all of the occurrences of that amino acid, even the ones that are not on the N-terminal, right?

To do this 100% correctly, you would have to separately consider the modifications for all the applicable positions depending on where you are in the peptide sequence.

You would need to duplicate the code in "addModsToVector":
```
            boost::range::insert(v, v.end(), getFixedMods(modSequence[i], i + 1 - modsTotalLength, modsAnywhere));
            if (i == 0)
                boost::range::insert(v, v.end(), getFixedMods(modSequence[i], i + 1 - modsTotalLength, modsAnyNTerm));
            else if (i + 1 == sequenceLength)
                boost::range::insert(v, v.end(), getFixedMods(modSequence[i], i + 1 - modsTotalLength, modsAnyCTerm));

            if (i > 0)
                boost::range::insert(v, v.end(), getFixedMods(modSequence[i], i + 1 - modsTotalLength, modsNotNTerm));
            if (i + 1 < sequenceLength)
                boost::range::insert(v, v.end(), getFixedMods(modSequence[i], i + 1 - modsTotalLength, modsNotCTerm));
```

Is there a reason that it's not worth doing that? Like, are we assuming that people are not going to combine location specificity with amino acids specificity on their label modifications like this?

chambm: The new code adds labels that are N or C terminal (position) with no site specificity. The ones with site specificity should already be added by the for loop (without any regard to position). Do you have a particular position/site combination you think wouldn't be supported here?

chambm: Oh, I see what you mean. Yes, the current for loop is potentially too inclusive because it doesn't consider position.

pwiz_tools/BiblioSpec/src/MaxQuantReader.cpp
```diff
@@ -731,12 +731,48 @@ void MaxQuantReader::addDoublesToVector(vector<double>& v, const string& valueLi
...
+    vector<const MaxQuantModification*> modsProteinNTerm;
+    vector<const MaxQuantModification*> modsProteinCTerm;
+    */
+
+    for (const auto& mod : modsAnyNTerm) { if (mod->sites.empty()) v.insert(v.begin(), SeqMod(1, mod->massDelta)); }
+
+    // iterate over sequence
+    for (int i = 0; i < (int)sequence.length(); i++)
+    {
+        // check for label mods
```
nickshulman, line 754: I think this comment does not apply anymore.

nickshulman (approved):

PR #1451, opened 2021-02-12
Use the PaneProgressBar on the heat map and pca plots

nickshulman:
This change now makes it so that the PaneProgressBar gets used by RTScheduleGraphPane. Here's what it looks like as it is updating a change that I made to the list of window widths:
image

Here are some things that I would like feedback on:
I added a new class "ProgressHandler" which can be constructed from just a CancellationToken and an Action&lt;int>.
Q. Do we really need a new class which does the same thing as "IProgressMonitor"?
A. Nick's answer: I tried using IProgressHandler in ClusterDataSet.PerformClustering, and it would have been pretty complicated code to pass an IProgressStatus object to the IProgressHandler to indicate that the percent complete had changed.
Q. Should UpdatePercentComplete take a double instead of an integer
A. Nicks's answer: I think it probably should (or at least there should be an overload which takes an integer). I could imagine a case where integer granularity between 1 and 100 would not be fine enough.

I made some changes to PaneProgressBar.
I changed the coordinates form CoordType.PaneFraction to CoordType.XPaneFractionYChartFraction. This simplified the code a lot because now the Y-coordinate zero is right below the Chart Title. I currently am positioning the progress bar 10% of the way below that title (.1f):
`        private PointF _barLocation = new PointF(1f/3, .1f);
`

Q. Does anybody have any opinion about what the Y-coordinate of that progress bar should be?

Here are some more screenshots:
image

image


brendanx67:
Maybe the implementation could be updated to support unknown progress, like the Windows progress bar where it just bounces back and forth, indicating that processing is happening, but it is not possible to indicate how close it is to complete.

pwiz_tools/Skyline/Controls/Databinding/DataboundGridControl.cs
```diff
@@ -663,149 +663,22 @@ private void heatMapContextMenuItem_Click(object sender, EventArgs e)
             ShowHeatMap();
         }
 
-        public bool ShowHeatMap()
+        public ClusterInput GetClusterInput()
```
brendanx67, line 666: I tend to prefer using "Get" for things that don't create new objects, and prefixes like "Create" or "Calc" for things that do.

pwiz_tools/Skyline/Controls/Databinding/DataboundGridControl.cs
```diff
@@ -663,149 +663,22 @@ private void heatMapContextMenuItem_Click(object sender, EventArgs e)
...
-                        points.Add(point);
-                        iCol++;
-                    }
-                }
-            }
-            var graphResults = new ClusterGraphResults(clusteredResults.RowDendrogramData?.DendrogramData, rowHeaders, columnGroups, points);
-            var heatMapGraph = new HierarchicalClusterGraph()
+        public bool ShowHeatMap()
+        {
+            var heatMapGraph = new HierarchicalClusterGraph
```
brendanx67, line 673: They are dockable now, yes, but it looks like we still just create a new one whenever one is shown, and I don't see any code to preserve them in the layout, if I do File > Save. Do they get properly preserved in the .sky.view file and restored? Do other group comparison windows?

pwiz_tools/Skyline/Controls/Databinding/DataboundGridControl.cs
```diff
@@ -964,13 +837,12 @@ private void pCAToolStripMenuItem_Click(object sender, EventArgs e)
 
         public void ShowPcaPlot()
         {
-            var resultsTuple = GetClusteredResults();
-            var pcaPlot = new PcaPlot()
+            var pcaPlot = new PcaPlot
```
brendanx67, line 840: Same as above. Do these get preserved in the .sky.view and restored, and how do we ensure that we only get one in a specific context? i.e. If I execute exactly the same command to show this plot 5 times, will I get 5 of the same graphs? Do they now update properly based on a document change? If so, it seems undesirable to be making new copies.

pwiz_tools/Skyline/Controls/Graphs/PaneProgressBar.cs
```diff
@@ -31,46 +31,41 @@ public class PaneProgressBar : IProgressBar
         readonly LineObj _left = new LineObj()
         {
             IsClippedToChartRect = true,
-            Line = new Line() { Width = 4, Color = Color.Green, Style = DashStyle.Solid },
-            Location = new Location(0, 0, CoordType.PaneFraction),
+            Line = new Line { Width = 4, Color = Color.Green, Style = DashStyle.Solid },
+            Location = new Location(0, 0, CoordType.XPaneFractionYChartFraction),
```
brendanx67, line 35: It would be better to be able to place this precisely aligned to the top of the graph the way we do for the RT stick annotations (ID and Predicted) in the chromatogram graphs.

brendanx67 (changes requested):
Mostly looks good. Glad to see the fix to RTScheduleGraphPane and growing use of the new in-graph progress indicator.

Some suggestions and a bigger question about saving and restoring Cluster and PCA plots. In the group comparison context they should behave like the bar and volcano plots. Do those save and restore correctly. I know they only create a single copy of themselves.

nickshulman:
> Do those save and restore correctly?

No, the work to make these graphs behave like proper dockable windows is going to happen in a different branch.
My plan is to make it so that there is only one heat map: if you ask for a heat map from either a Document Grid or a Fold Change Grid, and there is already a heat map showing, then the grid will become the new data provider for the heat map.
In the .sky.view file, the heat map will remember which grid window was its owner.
I haven't figured out all of the details of how that's going to work.

nickshulman:
> shouldn't IsCanceled check the CancellationToken instead of just returning false?
Oops. Good catch. I'll fix this.
Yes, I needed to use this class for a call to "BrukerTimsTofMethodExporter.GetScheduling" from RTScheduleGraphPane.

> It would be better to be able to place this precisely aligned to the top of the graph the way we do for the RT stick annotations (ID and Predicted) in the chromatogram graphs.
The problem with positioning things using pixels the way that ChromGraphItem.AddRetentionTimeAnnotation does is that you have to reposition the items every time the chart size changes. So, for ChromGraphItem, those things get added to the GraphPane during the ZedGraphControl.OnPaint method (MSGraphPane.drawLabels).
I was hoping to avoid that with this PaneProgressBar. Can we come up with a position in terms of chart fraction or the other sort of automatically adjusting coordinates so we don't need to be listening to the resize events?

brendanx67:
I suppose 20% from the top is fine for positioning.

PR #1454, opened 2021-02-16
Handle Import Peptide Search case with no results files

kaipot:


pwiz_tools/Skyline/FileUI/PeptideSearch/ImportFastaControl.cs
```diff
@@ -304,6 +306,13 @@ private static bool HasPrecursorTransitions(SrmDocument doc)
 
         public bool ImportFasta(IrtStandard irtStandard)
         {
+            if (!IsImportingResults && AutoTrain)
+            {
+                MessageDlg.Show(WizardForm,
+                    Resources.ImportFastaControl_ImportFasta_Cannot_automatically_train_mProphet_model_since_no_results_files_are_being_imported__Uncheck_this_option_in_order_to_continue_);
```
brendanx67, line 312: I think you can just show a message with just an OK button when the checkbox is checked and then uncheck it immediately. Just so the user understands why it cannot be checked. They could also back up and add results files, right?

brendanx67 (changes requested):
I would prefer the message about mProphe to appear immediately when the user attempts to check the checkbox, and then the checkbox is unchecked. That way they are told why they can't have this option, but nothing is required on their part to continue, and if they really want, they can use the Back button to add results and enable checking the checkbox.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportFastaControl.cs
brendanx67: Also, just bad form to tell a user to do something manually that you could do for them automatically. Even as implemented, it would be better to ask the user if they are sure they want to continue without the specified mProphet model than to tell them to uncheck the checkbox manually. But, I think what I am suggesting is better than that. It is like what we do when a user attempts to choose "Scheduled" in the method export form, but they lack enough retention time prediction information to achieve the scheduling.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -544,7 +544,17 @@ private void NextPage()
                             CloseWizard(DialogResult.Cancel);
                         }
 
-                        if (ImportResultsControl.ResultsFilesMissing)
+                        var anyResults = ImportResultsControl.FoundResultsFiles.Any();
+                        if (!anyResults)
+                        {
+                            if (MessageBox.Show(this,
```
brendanx67, line 550: Also, these need to be changed to MultiButtonMessageDlg() to support writing tests for this case. We should get this added to Brian Pratt's static code analysis tool. We don't really want any use of MessageBox in the Skyline code, because it can't be dismissed in an automated test, which kind of proves the other message is not covered in an automated test.

brendanx67 (changes requested):
Also, need to change the way messages are shown to not use MessageBox.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportFastaControl.cs
kaipot: Okay, I wasn't really sure how to implement it but your suggestion sounds good. I fixed that behavior.

brendanx67 (approved):

PR #1455, opened 2021-02-16
Add average cycle time

kaipot:
From https://skyline.ms/announcements/home/support/thread.view?rowId=49861

### DIFF
--- a/pwiz_tools/Skyline/Model/Databinding/Entities/TransitionResult.cs
+++ b/pwiz_tools/Skyline/Model/Databinding/Entities/TransitionResult.cs
@@ -85,6 +85,17 @@ namespace pwiz.Skyline.Model.Databinding.Entities
         [Format(NullValue = TextUtil.EXCEL_NA)]
         public int? PointsAcrossPeak { get { return ChromInfo.PointsAcrossPeak; } }
 
+        [Format(Formats.RETENTION_TIME, NullValue = TextUtil.EXCEL_NA)]
+        public double? AverageCycleTime
+        {
+            get
+            {
+                return StartTime.HasValue && EndTime.HasValue && PointsAcrossPeak.HasValue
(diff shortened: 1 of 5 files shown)


pwiz_tools/Skyline/Model/Databinding/Entities/TransitionResult.cs
```diff
@@ -85,6 +85,17 @@ public void ChangeChromInfo(EditDescription editDescription, Func<TransitionChro
         [Format(NullValue = TextUtil.EXCEL_NA)]
         public int? PointsAcrossPeak { get { return ChromInfo.PointsAcrossPeak; } }
 
+        [Format(Formats.RETENTION_TIME, NullValue = TextUtil.EXCEL_NA)]
+        public double? AverageCycleTime
+        {
+            get
+            {
+                return StartTime.HasValue && EndTime.HasValue && PointsAcrossPeak.HasValue
```
nickshulman, line 93: You can simplify this by doing:
                return (EndTime - StartTime) * 60 / PointsAcrossPeak;

nickshulman (commented):
I wonder if it should be called "CycleTimeAcrossPeak", so that it's clear that you're only looking at the cycle time across the peak, and not the average cycle time across the entire chromatogram.

pwiz_tools/Skyline/Model/Databinding/Entities/TransitionResult.cs
kaipot: That's useful, I didn't know that works with nullables.

kaipot:
> I wonder if it should be called "CycleTimeAcrossPeak", so that it's clear that you're only looking at the cycle time across the peak, and not the average cycle time across the entire chromatogram.

That's a better name, I've updated it.

brendanx67 (approved):
Looks good.

PR #1458, opened 2021-02-17
Prevent the "Peptide"->"Molecule" translation from happening on AlertDlg

nickshulman:
Prevent the "Peptide"->"Molecule" translation from happening on AlertDlg

If anyone needs that translation done, it needs to be done before the message is passed to AlertDlg. The AlertDlg has no way of knowing which parts of the message contain user content (filenames, etc) and should therefore never be translated.

(When importing peak boundaries, users sometimes see a message about the column "MoleculeModifiedSequence" being required. This bug was most recently mentioned on a support request from Marek)


bspratt (changes requested):
This seems like a pretty big hammer. Shouldn't we instead be identifying the uses where the translation behavior isn't wanted, and calling a variant that doesn't do the translation?

nickshulman:
Can you find an example of a place in Skyline where we show a message box, and the "Peptide -> Molecule" conversion is necessary and correct?

If the user presses the "Copy" button on the message box, the thing that gets copied to the clipboard is actually the text of the message before the conversion, so we would need to fix that if this is the behavior we actually want.

The thing is, it's really only appropriate to do the "Peptide -> Molecule" conversion on string templates: That is, the strings that we can find in "Resources.resx", which might contain things like "{0}", and which you might call "string.Format". Once those substitutions have happened, it's really too late to do this sort of substitution. The sorts of bugs it leads to are ones where Skyline says "There was an error opening file MyMoleculeDocument.sky", and the user has no idea what Skyline is talking about, since they know that the file they asked Skyline to open was called "MyPeptideDocument.sky".

Usually, when we have the potential for not knowing what state a string is in, the right thing to do is to create a new class. Like, back when we could never remember whether the strings we were passing around were real file paths, or were things with a "|" at the end of them returned by "SampleHelp.EncodeSampleId". We eventually got those sorts of bugs to stop happening by introducing the class "MsDataFilePath".

There's probably something we can do in terms of introducing a class called "StringTemplate", but I haven't figure out the details of how that would work.

bspratt:
I wonder if we couldn't identify uses of strings that are amongst our
resources, and selectively apply the translation to the parts that aren't
between braces?

bspratt:
So if we see "The peptide 'magic peptide' is already in this list" and we find a resource "The peptide '{0}' is already in this list", it comes out as "The molecule 'magic peptide' is already in this list".

nickshulman:
We should revisit this next version.

PR #1459, opened 2021-02-17
Skyline/work/20210217 skyline batch patch

alinmarsh:
Small fixes for adding a configuration to an empty list. Would like to make a patch release

### DIFF
--- a/pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/SkylineBatchConfigForm.cs
+++ b/pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/SkylineBatchConfigForm.cs
@@ -52,7 +52,8 @@ namespace SkylineBatch
             _newReportList = new List<ReportInfo>();
 
             _mainControl = mainControl;
-            _initialEnabled = config?.Enabled ?? false;
+            if (config != null) 
+                _initialEnabled = config.Enabled;
             _isBusy = isBusy;
 
             _canEditSkylineSettings = !Installations.HasLocalSkylineCmd;
@@ -151,9 +152,9 @@ namespace SkylineBatch
             {
                 openDialog.InitialDirectory = Path.GetDirectoryName(textSkylinePath.Text);
             }
-            catch (ArgumentException)
+            catch (Exception)
             {
-                // pass
+                // Pass: Exception is thrown if the string is empty or an incorrectly formatted file path
             }
             if (openDialog.ShowDialog()== DialogResult.OK)
                 textSkylinePath.Text = openDialog.FileName;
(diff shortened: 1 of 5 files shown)


pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/SkylineBatchConfigForm.cs
```diff
@@ -146,7 +147,14 @@ private void btnSkylineFilePath_Click(object sender, EventArgs e)
         {
             OpenFileDialog openDialog = new OpenFileDialog();
             openDialog.Filter = TextUtil.FILTER_SKY;
-            openDialog.InitialDirectory = Path.GetDirectoryName(textSkylinePath.Text);
+            try
+            {
+                openDialog.InitialDirectory = Path.GetDirectoryName(textSkylinePath.Text);
+            }
+            catch (ArgumentException)
```
brendanx67, line 154: Maybe make this Exception. I can imagine this causing exceptions other than ArgumentException, and it seems like we would always just want to fall back to something else or whatever the default is. Maybe improve the comment "pass" to explain better what will happen, e.g. "Accept the default directory"

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/SkylineBatchConfigForm.cs
```diff
@@ -282,8 +290,11 @@ private void gridReportSettings_SelectionChanged(object sender, EventArgs e)
         private void SetInitialSkylineSettings(SkylineBatchConfig config)
         {
             if (!_canEditSkylineSettings) return;
+            if (config != null)
+                _skylineTypeControl = new SkylineTypeControl(config.UsesSkyline, config.UsesSkylineDaily, config.UsesCustomSkylinePath, config.SkylineSettings.CmdPath);
+            else
+                _skylineTypeControl = new SkylineTypeControl(false, false, true, "C:\\Program Files\\Skyline");
```
brendanx67, line 296: It would also be good to have a comment here explaining when this can happen. It seems like you are defaulting to using a custom path pointing to C:\Program Files\Skyline regardless of whether it is there or not. That seems like it needs explaining.

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/SkylineBatchConfigForm.cs
```diff
@@ -52,7 +52,7 @@ public SkylineBatchConfigForm(IMainUiControl mainControl, SkylineBatchConfig con
             _newReportList = new List<ReportInfo>();
 
             _mainControl = mainControl;
-            _initialEnabled = config.Enabled;
+            _initialEnabled = config?.Enabled ?? false;
```
brendanx67, line 55: Could also just use

if (config != null)
    _initialEnabled = config.Enabled

Because the default values is always false. Not quite as fancy on the syntax and requiring 2 lines, but to me simpler to read.

brendanx67 (changes requested):
Some stylistic suggestions and one substantive change in exception handling.

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/SkylineBatchConfigForm.cs
alinmarsh: Added the comment. I realized I'm using a different constructor in my current version, so I replaced that as well. I'm trying not to change too much for the patch but I think this makes more sense.

PR #1461, opened 2021-02-19
Fix bug preventing save command on new line

alinmarsh:
Fix bug preventing save command on new line


brendanx67 (changes requested):
Looks like it should work. Probably we should add a test that proves it does.

PR #1462, opened 2021-02-20
Allow "File > Share" on documents that have not been saved yet

nickshulman:
This change makes it so that you can do "File > Share" on a document that has not been saved yet.
If the document is dirty, the user gets prompted to save, but they are allowed to say "No", and still proceed with the Share.

If the document is not dirty, then there is an additional option in the "What format do you want to save the file as" dialog "Current Saved File". Here's what the options look like when I am trying to share a document which is the very old 1.9 format:
image

If the current saved file is an official version of Skyline, it will call it what it is "Skyline 3.6", but if it's an intermediate or very old version, it gets called "Version 3.61" or whatever.


nickshulman:
I ran some speed tests

PR #1464, opened 2021-02-23
Skyline: Fix case causing PathTooLongException to reach the exception web page

brendanx67:
- Show the message to the user instead to allow them to shorten the path


nickshulman:
Looks good to me.
Before "SkylineWindow.HasFileToOpen" was added, we used to see a PathTooLongException in "SkylineWindow.OpenFile". It looks like your fix here will also prevent us from getting hitting the same error later on.

PR #1465, opened 2021-02-24
Skyline: extend our custom code inspection to detect forgotten uses of PauseTest()…

bspratt:
…, which can stall our nightly tests. The idea is that any unexpected use of PauseTest() will fail the TeamCity build, so it never makes it to Nightly. And, it's a reminder on your local build to clean that up before a commit. (I recently found myself nearly committing this error on a branch, it's a common enough problem that this seems like a useful check.)

N.B. this initial commit includes an intentional example of this error, to verify that the desired effect is achieved in TeamCity.

### DIFF
--- a/libraries/boost-build/src/tools/common.jam
+++ b/libraries/boost-build/src/tools/common.jam
@@ -774,9 +774,9 @@ if --verbose in [ modules.peek : ARGV ]
     {
         actions hard-link
         {
-            $(RM) "$(<)" 2$(NULL_OUT) $(NULL_OUT)
+            IF EXIST "$(<)" $(RM) "$(<)" 2$(NULL_OUT)
             $(LN) "$(<)" "$(>)" 2$(NULL_OUT) $(NULL_OUT)
-            IF ERRORLEVEL 1 $(CP) "$(>)" $(WINDOWS-CP-HACK) "$(<)" $(NULL_OUT)
+            IF ERRORLEVEL 1 $(CP) "$(>)" $(WINDOWS-CP-HACK) "$(<)"
         }
     }
     else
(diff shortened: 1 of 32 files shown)


brendanx67 (approved):
Looks good. Though maybe not intended to leave AuditLogTutorialTest ignoring this inspection.

bspratt:
It appears to be in that test with intention - do you think we should just remove it?

brendanx67:
I am not following. You mean the commented-out PauseTest()? If that is all we are talking about, then, yes, just remove it. I somehow thought your detection excluded line comments and I don't see a functional PauseTest() in AuditLogTutorialTest

bspratt:
I'm thinking of this one: 
image

Which isn't really necessary since the test will block anyway, I expect.

bspratt:
(and you are correct, it does ignore commented-out occurrences.)

bspratt:
Actually when I think about it,  it won't block - it will just come back with nothing entered. So yes, I think this one does need to remain.

brendanx67:
Okay. Sure. That seems necessary. Not sure why I didn't see that in my
search.

brendanx67:
Though, hmmm... Now that we have a difference, it would probably be better to add a more specific method for this step named something like PauseForManualTutorialStep(), which just calls PauseTest(). That way you wouldn't have to turn off this check for the entire file, just because there is a need for a pause that is not pausing for a screenshot.

bspratt:
I was musing on something like that, too. I'll implement that.

Actually I think the right change is to replace PauseTest() with
PauseForManualTutorialStep() and remove the call to ShowDialog. The
password dialog should not be closed programmatically.

bspratt:
Meh, I just strengthened the message to make it clear that the test is
expecting to do the work closing the password dialog, and the user should
only close the  PauseForManualTutorialStep()  window .

brendanx67 (approved):
Works for me

PR #1470, opened 2021-03-01
Fix irtdb docxml bug

kaipot:


kaipot:
@nickshulman This is a bug where changing the settings on the document caused a hang if there were many variable modifications. I originally tried to fix it how I asked you about earlier (changing all modifications to explicit), but it caused problems later when trying to import the document. I think this is a better way to fix the problem.

nickshulman (approved):

PR #1472, opened 2021-03-02
Skyline: Replace uses of MessageBox.Show() with our own MessageDlg.Show() …

bspratt:
…for better automated test and and small molecule interface support. Add a check to our custom code inspection so that no new uses of MessageBox.Show() are introduced in future.


pwiz_tools/Skyline/Alerts/AlertDlg.cs
```diff
@@ -181,11 +182,31 @@ private void btnMoreInfo_Click(object sender, EventArgs e)
...
+                    focusButton = buttons.Count == 3 ? buttons[1] : buttons.Count == 2 ? buttons[0] : null;
+                    break;
+                case MessageBoxDefaultButton.Button3:
+                    focusButton = buttons.FirstOrDefault();
+                    break;
+            }
+
+            if (focusButton != null)
+            {
+                focusButton.Select();
```
nickshulman, line 209: Should we be setting the "AcceptButton" property on the form instead?
The AcceptButton is the button that gets pressed if the user presses Enter while the focus is on some other control such as a textbox. That's the standard way of designating the "Default" button on a form, but it might not make a difference on AlertDlg since AlertDlg only has buttons on it.

pwiz_tools/Skyline/FileUI/ImportResultsDlg.cs
```diff
@@ -477,7 +477,7 @@ private IEnumerable<int> ChooseSamples(string dataSource, IEnumerable<string> sa
                 KeyValuePair<string, MsDataFileUri[]>[] namedPaths = DataSourceUtil.GetDataSourcesInSubdirs(dirRoot).ToArray();
                 if (namedPaths.Length == 0)
                 {
-                    MessageBox.Show(this,
+                    MessageDlg.Show(this,
                         string.Format(Resources.ImportResultsDlg_GetDataSourcePathsDir_No_results_found_in_the_folder__0__, dirRoot),
                         Program.Name);
```
nickshulman, line 482: We should not pass Program.Name.

pwiz_tools/Skyline/Alerts/AlertDlg.cs
```diff
@@ -44,21 +44,22 @@ public AlertDlg() : this(@"Alert dialog for Forms designer")
...
             Message = message;
             btnMoreInfo.Parent.Controls.Remove(btnMoreInfo);
-            Text = Program.Name;
+            Text = title ?? Program.Name;
             toolStrip1.Renderer = new NoBorderSystemRenderer();
         }
 
-        public AlertDlg(string message, MessageBoxButtons messageBoxButtons) : this(message)
+        public AlertDlg(string message, MessageBoxButtons messageBoxButtons, string title = null,
+            MessageBoxDefaultButton defaultButton = MessageBoxDefaultButton.Button1) : this(message, title)
```
nickshulman, line 60: I know "MessageBoxDefaultButton" is the standard way of designating the default button in a MessageBox, but could we use "DialogResult" as the way to designate it in AlertDlg? The caller definitely knows which DialogResult they are intend to be the default button, but they might not be so sure about whether that's the first or second button.

nickshulman (commented):
The way that we currently have it, if the user wants a simple "OK" message (or "OK" with "More Info"), then they can use "MessageDlg". If they want a couple of buttons, they can use "MultiButtonMessageDlg". If they want anything more complicated (such as changing the Title on the dialog), they need to construct an AlertDlg, set some properties on it, and show it inside of a using block.

The advantage of doing things that way is that the constructors and other methods do not get cluttered with extra arguments that are only used in a few cases. The disadvantage is that it takes several lines to show an AlertDlg if you are doing something nonstandard.

pwiz_tools/Skyline/Alerts/AlertDlg.cs
bspratt: Excellent feedback, as usual. I'll look into these. 

I was going for minimal widespread code changes but it is perhaps just as well to make it clear that this doesn't behave exactly the way a MessageBox does. So yes, I should have swapped in MultiButtonMessageDlg where appropriate.

>> The advantage of doing things that way is that the constructors and other methods do not get cluttered with extra arguments that are only used in a few cases. 
If the objection is just that the ctor has a lot of default arguments, it's not a bother to create multiple ctors.

>> The disadvantage is that it takes several lines to show an AlertDlg if you are doing something nonstandard.
That seems like a fair sized disadvantage, especially when "nonstandard" encompasses things that are pretty normal for MessageBox use. They made the API for a reason, surely?

pwiz_tools/Skyline/Alerts/AlertDlg.cs
bspratt: Done, thanks.

nickshulman (commented):
Looks great except for my two new comments which you are welcome to ignore and check in.

pwiz_tools/Skyline/Alerts/AlertDlg.cs
```diff
@@ -44,21 +44,30 @@ public AlertDlg() : this(@"Alert dialog for Forms designer")
         {
         }
 
-        public AlertDlg(string message)
+        public AlertDlg(string message) : this(message, null)
+        {
+        }
+
+        public AlertDlg(string message, string title)
```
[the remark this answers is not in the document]
bspratt: There are one or two, yes.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -1048,9 +1048,7 @@ public static void PauseForManualTutorialStep(string description = null)
         public static void PauseTestUI(string description = null)
         {
             if (!Program.SkylineOffscreen)
-                MessageBox.Show(description ?? string.Empty, @"Test paused on UI thread",
-                    MessageBoxButtons.OK,
-                    MessageBoxIcon.Information);
+                MessageDlg.Show(SkylineWindow, description ?? string.Empty, @"Test paused on UI thread, click OK to resume");
```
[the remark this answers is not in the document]
bspratt: Good point, thanks.

PR #1476, opened 2021-03-05
Enforce filename extension when the user is prompted to browse for a missing library

nickshulman:
This change makes it so that a user is only allowed to choose a file with the same extension as the file that they are looking for when they are asked to browse for a missing library when opening a Skyline document.

A couple of notes:
1. Browsing for a missing library is the only way I know of that you can get a non-.blib file added to the document as a library. However, I have not actually found a user who is willing to say that they've even seen the MissingFileDlg, much less admitted that they used it to choose an invalid file. There is probably some other way to get such invalid files added to the document as a library, but I have no idea what that would be.
2. This change makes it so that MissingFileDlg forces the user to choose a file with the same extension as the file they are looking for. It does this by comparing the filename extension of the chosen file with FileHint. A better way to do this would probably be to make a class called "SpecType" where instances of the SpecType class (e.g. "SpecTypes.BLIB") enforce the rules about filename extension. Then, a SpecType would be passed in to MissingFileDlg.


pwiz_tools/Skyline/TestFunctional/SynchSiblingsTest.cs
```diff
@@ -66,8 +66,9 @@ protected override void DoTest()
             Assert.AreEqual(pepCount, SkylineWindow.Document.PeptideCount);
             RunUI(() => SkylineWindow.NewDocument());
             Assert.AreEqual(0, SkylineWindow.Document.PeptideCount);
-            RunDlg<MissingFileDlg>(() => SkylineWindow.OpenFile(documentPath),
-                dlg => dlg.OkDialog(movedPath));
+            var missingFileDlg = ShowDialog<MissingFileDlg>(() => SkylineWindow.OpenFile(documentPath));
+            RunDlg<AlertDlg>(()=>missingFileDlg.ChooseFile(documentPath), alertDlg=>alertDlg.OkDialog());
```
brendanx67, line 70: Should validate that it has the right message too.

brendanx67 (approved):
Looks good to me. On suggestion on the test.

PR #1480, opened 2021-03-06
Fix iRT tests

kaipot:


pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
```diff
@@ -823,6 +823,8 @@ public void SetFixedPoints(int one, int two)
             comboMaxPeptide.SelectedIndex = two;
         }
 
+        public int NumFixedPointPeptides => Math.Min(comboMinPeptide.Items.Count, comboMaxPeptide.Items.Count);
```
nickshulman, line 826: I think this should be called "NumFixedPointOptions" or "NumFixedPointPeptideOptions"

The number of fixed point peptides is always 2. This is the number of peptides the user is allowed to choose between.

nickshulman (approved):

pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.cs
kaipot: > I think this should be called "NumFixedPointOptions" or "NumFixedPointPeptideOptions"
> 
> The number of fixed point peptides is always 2. This is the number of peptides the user is allowed to choose between.

Good point, renamed now.

PR #1485, opened 2021-03-10
Fix irt tests

kaipot:
So I think the actual problem was that the test was doing stuff with the form, but sometimes before the Form.Shown event handler had executed.
I'm not 100% sure here but I ran it a bunch of times on Brian's machine and didn't get any failures with this fix.

### DIFF
--- a/pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.Designer.cs
+++ b/pwiz_tools/Skyline/SettingsUI/Irt/CalibrateIrtDlg.Designer.cs
@@ -297,8 +297,6 @@
             this.MinimizeBox = false;
             this.Name = "CalibrateIrtDlg";
             this.ShowInTaskbar = false;
-            this.Load += new System.EventHandler(this.OnLoad);
-            this.Shown += new System.EventHandler(this.CalibrateIrtDlg_Shown);
             ((System.ComponentModel.ISupportInitialize)(this.bindingSourceStandard)).EndInit();
             ((System.ComponentModel.ISupportInitialize)(this.gridViewCalibrate)).EndInit();
             this.grpRegressionEquation.ResumeLayout(false);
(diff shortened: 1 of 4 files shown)


nickshulman (approved):
Looks good to me.

I wonder whether this could have been fixed by moving the stuff that's in the "Shown" event and have it happen instead in the "Load" event.

The code and the end of "System.Windows.Forms.Form.OnLoad" calls "OnShown" via a BeginInvoke. I wonder if our test code is sneaking in between the OnLoad and OnShown events and that's what's causing all these problems.

brendanx67:
OnShown is definitely dangerous, since our tests stop waiting as soon as
the form is visible. Much the way we have to be careful not to do anything
that depends on the code that runs after Form.Show() returns when we have
only blocked until the form disappears.

kaipot:
I must've originally had it in the Load event, but needed it to happen after the ModeUIAwareFormHelper does some stuff in the Load event. I guess it could probably done by add another OnLoad handler programatically (multiple handlers get executed synchronously in the order they are subscribed?)

nickshulman:
> I must've originally had it in the Load event, but needed it to happen after the ModeUIAwareFormHelper does some stuff in the Load event. I guess it could probably done by add another OnLoad handler programatically (multiple handlers get executed synchronously in the order they are subscribed?)

It sounds like you should just override "OnLoad" and add your code after the call to "base.OnLoad()".
(Also, you should get rid of the method that is already there: "CalibrateIrtDlg.OnLoad(object, EventHandler)". That code is no longer necessary since we DataGridViewEx already takes care of that)
In general, it is better to override the "OnXxx" methods on a form, instead of attaching event handlers (it's a little more efficient).

I believe event handlers get put into a hash table, so the order in which they get called is kind of random.

nickshulman (approved):

PR #1488, opened 2021-03-11
Add a progress message for preprocessing step of import of large Assay Libraries and Transition Lists

bspratt:
Also, allow import files with apparently missing critical columns to show in the interactive column picker.


pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
```diff
@@ -429,24 +430,43 @@ private void CheckEssentialColumn(Tuple<int, string> column)
...
-            MissingEssentialColumns = new List<string>();
-            CheckEssentialColumn(new Tuple<int, string>(columns.PeptideColumn, Resources.ImportTransitionListColumnSelectDlg_PopulateComboBoxes_Peptide_Modified_Sequence));
-            CheckEssentialColumn(new Tuple<int, string>(columns.PrecursorColumn, Resources.ImportTransitionListColumnSelectDlg_PopulateComboBoxes_Precursor_m_z));
-            CheckEssentialColumn(new Tuple<int, string>(columns.ProductColumn, Resources.ImportTransitionListColumnSelectDlg_PopulateComboBoxes_Product_m_z));
-            var docNew = _docCurrent.ImportMassList(_inputs, Importer, null,
-                _insertPath, out testSelectPath, out testIrtPeptides, out testLibrarySpectra,
-                out testErrorList, out testPeptideGroups);
-            if (testErrorList.Any())
+            SrmDocument docNew = null;
+            var errorCheckCanceled = false;
```
nickshulman, line 436: I think this needs to be initialized to "true".
That way, it will only get set to "false" if it makes it to the end of the LongWaitDlg action.
(otherwise, it might end up "false" if ImportMassList decides to throw an OperationCancelledException.

bspratt: Makes sense, thanks.

brendanx67:
I tried this and I was surprised at how little progress it actually showed. I got a message "Reading transition list" for about 10 seconds and no movement on the progress bar. Then "Inspecting peptide sequence information" for another 20 seconds and again no movement on the progress bar until it suddenly jumped to 100%. This is with a 6.9 million line file. I think we may want to find a way to use StreamReaderWithProgress. Need to extract a base class that doesn't do hashing, which Tobi added.

Anyway, still not feeling like this is ready for prime-time... though it is an improvement over what we currently have released to Skyline-daily.

bspratt:
Can you point me at that 6,9 million line file?

chambm:
Is that a line endings conflict? We need a GitHub action linting line endings... relying on client-side autoCRLF setting is too error prone. I think I turned mine off yesterday when I was trying to fix a local commit which was showing on GitHub as every line as changed due to a line ending change (I'm not sure how that happened in the first place since I had autoCRLF on though).

PR #1489, opened 2021-03-11
Skyline: Improve transition list format detection

brendanx67:
- detect international formats regardless of the current culture

### DIFF
--- a/pwiz_tools/Skyline/Model/Import.cs
+++ b/pwiz_tools/Skyline/Model/Import.cs
@@ -315,6 +315,12 @@ namespace pwiz.Skyline.Model
             Separator = separator;
         }
 
+        public MassListInputs(IList<string> lines)
+        {
+            InitFormat(lines);
+            _lines = lines;
+        }
+
         public IList<string> ReadLines()
         {
             return _lines ?? (_lines = _inputFilename != null ? ReadLinesFromFile() : ReadLinesFromText());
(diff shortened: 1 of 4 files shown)


brendanx67:
Added a new test to test all the various formats and ran that test in SkylineTester under all possible current cultures.

brendanx67:
This should impact the SmallMoleculeTransitionListCSVReader also. Maybe the code in that file:
            // Skip over header line to deduce decimal format
            string line = csvText.Count > 1 ? csvText[1] : header;
            MassListImporter.IsColumnar(line, out formatProvider, out separator, out columnTypes);
            // Double check that separator - does it appear in header row, or was it just an unlucky hit in a text field?
            if (!header.Contains(separator))
            {
                // Try again, this time without the distraction of a plausible but clearly incorrect seperator
                MassListImporter.IsColumnar(line.Replace(separator,'_'), out formatProvider, out separator, out columnTypes);
            }

Should change to use MassListInputs

pwiz_tools/Skyline/Model/Import.cs
```diff
@@ -362,8 +362,21 @@ private void InitFormat(IList<string> inputLines)
                 char sep;
                 IFormatProvider provider;
                 Type[] columnTypes;
-                if (!MassListImporter.IsColumnar(inputLines[0], out provider, out sep, out columnTypes))
+                string inputLine = 0 < inputLines.Count ? inputLines[0] : string.Empty;
+                if (!MassListImporter.IsColumnar(inputLine, out provider, out sep, out columnTypes))
+                {
                     throw new IOException(Resources.SkylineWindow_importMassListMenuItem_Click_Data_columns_not_found_in_first_line);
+                }
+                // If there are no numbers in the first column, try the second. Without numbers the format provider may not be correct
```
nickshulman, line 370: I think this comment supposed to say "first row" instead of "first column"

nickshulman (approved):
Looks good to me.

bspratt (approved):
Looks good. I like the improved modularity.

PR #1490, opened 2021-03-12
Fix failure in TestImportHighPrecTransitionList and TestPrecursorIon …

nickshulman:
…in code coverage builds on TeamCity.

The problem is that PasteTransitionListSkipColumnSelect returns as soon as the column select dialog goes away.
However, there is still some work that gets done inside of a LongWaitDlg.

There might be a better way to fix this, where PasteTransitionListSkipColumnSelect would wait until the operation was completely finished, but I couldn't think of a way to do that.


bspratt (approved):
Makes sense, thanks for the fix.

PR #1494, opened 2021-03-16
Fix false detection of hang when transitioning from standard time to DST

kaipot:


nickshulman (commented):
I think this is going to have the same problem when we set the clocks back an hour on November 7.

Imagine last write time is 1:45am (before we set the clocks back) (8:45am UTC).
Add an hour to that, and it's still 1:45am (after we set the clocks back) (9:45am UTC).
If you compare that time to 1:50am, it's going to look like there was a hang (8:50am UTC).

Here's some code I wrote to test this:
```
            var lastWriteTimeUtc = new DateTime(2021, 11, 7, 8, 45, 0, DateTimeKind.Utc);
            var signalTimeUtc = new DateTime(2021, 11, 7, 8, 50, 0, DateTimeKind.Utc);
            var lastTimeLocal = lastWriteTimeUtc.ToLocalTime();
            var signalTimeLocal = signalTimeUtc.ToLocalTime();
            var hourLaterLocal = lastTimeLocal.ToUniversalTime().AddMinutes(60).ToLocalTime();
            Console.Out.WriteLine("Last Time {0}", lastTimeLocal);
            Console.Out.WriteLine("Hour later {0}", hourLaterLocal);
            Console.Out.WriteLine("Signal Time {0}", signalTimeLocal);
            // This assert fails, which means that it looks like a hang even though it shouldn't be.
            Assert.IsTrue(signalTimeLocal < hourLaterLocal);

```

I think you need to compare the two times both in local time and in UTC time, and say that it's a hang only if it looks like a hang in both time zones.

brendanx67:
Or, really, just change the number from 60 to 90 and call it good. The likelihood that something will wait for 30 minutes over the daylight saving time change and not really be a hang seems slim enough that we can live with that risk.

brendanx67:
This was only the problem it was, because you chose 60 minutes and the clocks change by 60 minutes.

kaipot:
> I think this is going to have the same problem when we set the clocks back an hour on November 7.
> 
> Imagine last write time is 1:45am (before we set the clocks back) (8:45am UTC).
> Add an hour to that, and it's still 1:45am (after we set the clocks back) (9:45am UTC).
> If you compare that time to 1:50am, it's going to look like there was a hang (8:50am UTC).
> 
> Here's some code I wrote to test this:
> 
> ```
>             var lastWriteTimeUtc = new DateTime(2021, 11, 7, 8, 45, 0, DateTimeKind.Utc);
>             var signalTimeUtc = new DateTime(2021, 11, 7, 8, 50, 0, DateTimeKind.Utc);
>             var lastTimeLocal = lastWriteTimeUtc.ToLocalTime();
>             var signalTimeLocal = signalTimeUtc.ToLocalTime();
>             var hourLaterLocal = lastTimeLocal.ToUniversalTime().AddMinutes(60).ToLocalTime();
>             Console.Out.WriteLine("Last Time {0}", lastTimeLocal);
>             Console.Out.WriteLine("Hour later {0}", hourLaterLocal);
>             Console.Out.WriteLine("Signal Time {0}", signalTimeLocal);
>             // This assert fails, which means that it looks like a hang even though it shouldn't be.
>             Assert.IsTrue(signalTimeLocal < hourLaterLocal);
> ```
> 
> I think you need to compare the two times both in local time and in UTC time, and say that it's a hang only if it looks like a hang in both time zones.

Good point, I didn't think about that. I fixed it now.

nickshulman (approved):

PR #1495, opened 2021-03-17
Skyline: Fix case importing a transition list

brendanx67:
- ModificationMatcher explicit mods prevent loss mods


nickshulman (approved):
Looks good to me.

Loss-only modifications currently show up in the "Peptide Modified Sequence Full Names" column in the Document Grid. I should probably fix that.

brendanx67:
Yep. Fix that. We don't highlight those AAs with bold font in the Targets view either.

PR #1496, opened 2021-03-17
Fix NRE using Insert Peptides dialog to insert a peptide which has both a neutral loss and an implicit modification.

nickshulman:
This seems to fix the NRE that I reported earlier today.

I will keep testing this fix, and will probably commit it along with a different fix that I am working on which makes it so that "PeptideModifiedSequenceFullNames" does not include neutral loss modifications.


brendanx67:
I can see how this might work to solve the NRE, but does it work correctly on your PEPTIDECK case, leaving the C modified with Carbamidomethyl (C)? I had to do more than just solve the NRE to get this working. I also chose to add these matches to the Matches table in the ModificationMatcher instead of what you did. I will send you my diffs which I am running tests on now. When that completes, I will switch to your branch and try your new test.

nickshulman:
Oops. You're right, Brendan. The Cysteine doesn't end up getting applied.
Your fix sounds like it reflects a better understanding of how ModificationMatcher is supposed to work.

PR #1497, opened 2021-03-18
Skyline: Fix a NRE in Edit > Insert > Peptides found by Nick with neutral losses

brendanx67:
- It seems my last PR caused some problems in wider use than our automated tests
- New automated test added to cover more cases, including the one that caused the NRE


nickshulman (approved):
Looks good to me.

PR #1498, opened 2021-03-18
Prevent neutral losses from showing up in the Targets tree or "Peptid…

nickshulman:
…eModifiedSequenceFullNames" in the Document Grid.

The Targets tree and the Document Grid both make use of the "ModifiedSequence" class to figure out which modifications are on any particular amino acid. The static method "ModifiedSequence.GetModifiedSequence" figures out which explicit and implicit modifications are applied to each amino acid. This method will now strip out all zero-mass modifications since we do not want them to be displayed anywhere.

I'll write some unit tests for this before I commit it to master.


brendanx67:
I haven't looked at this closely yet, but let's try to get it into both 20.2 and master. The way we currently display a peptide with both loss-only modifications and a heavy precursor is pretty broken.

PR #1504, opened 2021-03-24
Add "Library Ion Mobility" to the Document Grid

nickshulman:
(feature requested by Sue)

This change adds "Library Ion Mobility" to the Document Grid, under "Precursors":
image

Here's what it looks like in the Document Grid:
image


brendanx67:
Good questions. Maybe ask John F., Erin, and Sue. I get the sense that people are trying to move to CCS becoming the more important value for IMS. So, I would think you would want to include it if it is present, and you might want to make it the first value. Currently, it does seem like the values are ordered based on the likelihood they will be present.

Do these only show up for a .imsdb like Sue is working with, or do you also support when the values come from a .blib (or soon a .clib)?

bspratt:
> Do these only show up for a .imsdb like Sue is working with, or do you also support when the values come from a .blib (or soon a .clib)?

I've thought about the idea of a provenance field for IM information, which would hold the name of the library or indicate that it comes from an explicit value. It would make things less mysterious when displayed here.

pwiz_tools/Skyline/Model/Databinding/Entities/IonMobilityObject.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                parts.Add(IonMobilityValue.Value.ToString(Formats.IonMobility));
+            }
+
+            if (_units != eIonMobilityUnits.none)
+            {
+                parts.Add(IonMobilityUnits);
+            }
+
+            // Consider: maybe include CollisionCrossSection and HighEnergyOffset if present?
```
bspratt, line 53: Definitely include these, yes.

pwiz_tools/Skyline/Model/Hibernate/Formats.cs
```diff
@@ -40,6 +40,8 @@ public static class Formats
         public const string Mz = "0.####";
         public const string SamplingTime = "0.00";
         public const string OneOverK0 = "0.####";
+
+        public const string IonMobility = "0.#####";
```
bspratt, line 44: This is probably more digits than needed - perhaps rename OneOverK0 to IonMobility instead

PR #1512, opened 2021-03-29
Fix transition settings in Import Peptide Search

kaipot:


brendanx67 (changes requested):
This does not feel that clear what you are doing and why. You are also going to need to add some amount of change to the tests to test this new code and why it is important.

pwiz_tools/Skyline/FileUI/PeptideSearch/TransitionSettingsControl.cs
```diff
@@ -349,11 +352,30 @@ public TransitionSettings GetTransitionSettings(Form parent)
...
+                : TransitionLibraryPick.filter;
+            if (CountFinder != null && pick == TransitionLibraryPick.filter)
+                pick = TransitionLibraryPick.all;
             var libraries = new TransitionLibraries(ionMatchTolerance, minIonCount, ionCount, pick);
             Helpers.AssignIfEquals(ref libraries, settings.Libraries);
 
             return new TransitionSettings(settings.Prediction, filter, libraries, settings.Integration, instrument, settings.FullScan, settings.IonMobilityFiltering);
         }
+
+        private IEndCountFragmentFinder CountFinder => TransitionFilter.GetEndFragmentFinder(
```
brendanx67, line 366: This needs a comment on exactly what it is doing and how it is used.  You may want a less terse expression that allows you to better comment the choices for when the values currently in the form are used and when the value from the current document is used.

pwiz_tools/Skyline/FileUI/PeptideSearch/TransitionSettingsControl.cs
```diff
@@ -349,11 +352,30 @@ public TransitionSettings GetTransitionSettings(Form parent)
...
+            TransitionFilter.GetEndFragmentNameFromLabel(IonFilter ? IonRangeTo : DocumentTransitionSettings.Filter.FragmentRangeLast.Label)
+        ) as IEndCountFragmentFinder;
+
+        private void comboRangeTo_SelectedIndexChanged(object sender, System.EventArgs e)
+        {
+            var countFinder = CountFinder;
+            if (countFinder != null)
+            {
+                IonCount = countFinder.Count;
```
brendanx67, line 375: This is a tricky bit of code. Even though, it may all work out, I think you should not use IonCount as much as you do, since in each reference it is doing int.Parse() of the text value, which you set at the top. So, it is unlikely to throw a FormatException. But, you probably should assign to an integer local variable at the top, and then use that below.

int ionCount = countFinder.Count;
IonCount = ionCount;

pwiz_tools/Skyline/FileUI/PeptideSearch/TransitionSettingsControl.cs
```diff
@@ -349,11 +352,30 @@ public TransitionSettings GetTransitionSettings(Form parent)
                 settings.Instrument.MaxInclusions, settings.Instrument.MinTime, settings.Instrument.MaxTime);
             Helpers.AssignIfEquals(ref instrument, settings.Instrument);
 
-            TransitionLibraryPick pick = settings.Libraries.Pick != TransitionLibraryPick.none ? settings.Libraries.Pick : TransitionLibraryPick.all;
+            var pick = settings.Libraries.Pick != TransitionLibraryPick.none && DocumentSettings.PeptideSettings.Libraries.LibrarySpecs.Any(lib => !lib.IsDocumentLibrary)
```
brendanx67, line 355: Add a comment explaining this conditional. I have to admit that "none" and "all" are not totally clear names.

pwiz_tools/Skyline/FileUI/PeptideSearch/TransitionSettingsControl.cs
```diff
@@ -349,11 +352,30 @@ public TransitionSettings GetTransitionSettings(Form parent)
                 settings.Instrument.MaxInclusions, settings.Instrument.MinTime, settings.Instrument.MaxTime);
             Helpers.AssignIfEquals(ref instrument, settings.Instrument);
 
-            TransitionLibraryPick pick = settings.Libraries.Pick != TransitionLibraryPick.none ? settings.Libraries.Pick : TransitionLibraryPick.all;
+            var pick = settings.Libraries.Pick != TransitionLibraryPick.none && DocumentSettings.PeptideSettings.Libraries.LibrarySpecs.Any(lib => !lib.IsDocumentLibrary)
+                ? settings.Libraries.Pick
+                : TransitionLibraryPick.filter;
+            if (CountFinder != null && pick == TransitionLibraryPick.filter)
```
brendanx67, line 358: Comment here too. Even though I understand this one, it still deserves a quick comment.
// If IonRangeTo is a specific ion count, then the range restriction needs to be ignored in the
// presence of a spectral library for choosing ions.

brendanx67 (commented):
Added some line-by-line comments.

chambm:
The peptide count from TestDiaSearchFixedWindows dropped significantly. Is that because the defaults changed?

brendanx67:
Yeah. Let's figure out how the settings would need to change to get back to the original values, understand why they changed, and then decide whether to keep the changes to the defaults or accept the reduced count.

kaipot:
In this case the default changed from TransitionLibraryPick.all to TransitionLibraryPick.filter (since all ignores the ion range settings). I'll look into these files specifically to see exactly why the peptides dropped so much.

brendanx67:
Yes. It is worth trying just switching from ion 3 to ion 1 which would be expected to have the same effect. But, I would not expect the addition of y1, y2, b1, and b2 to have a net negative impact. This is just one way to test we understand the change that is causing the effect we see.

brendanx67:
Any progress on this?

kaipot:
> 
> 
> Any progress on this?

I don't think the effect is as dramatic as the numbers make it look, it is mostly just coming from one missing peptide that occurs many times.

[out-all.txt](https://github.com/ProteoWizard/pwiz/files/6534503/out-all.txt)
[out-filter.txt](https://github.com/ProteoWizard/pwiz/files/6534502/out-filter.txt)
[out-diff.txt](https://github.com/ProteoWizard/pwiz/files/6534501/out-diff.txt)

brendanx67:
Maybe you could add a "remove repeated peptides" into the test to make this more obvious.

kaipot:
> Any progress on this?

I don't think the effect is as dramatic as the numbers make it look, it is mostly just coming from one peptide missing 

> Maybe you could add a "remove repeated peptides" into the test to make this more obvious.

Yeah, I tried using the remove repeated peptides option and the numbers go from 16 peptides, 17 precursor, 153 transitions (with TransitionLibraryPick.all) to 14 peptides, 15 precursor, 135 transitions (with TransitionLibraryPick.filter). Does that seem reasonable?
I'll modify the test to use that option, that's a good idea.

PR #1513, opened 2021-03-29
Fix for dda search with same filenames

kaipot:


chambm:
What's the problem being fixed? Same filenames in different paths?

brendanx67:
Do we need to add a test?

kaipot:
> What's the problem being fixed? Same filenames in different paths?

Yeah, it will cause an exception since the dictionary key for the SpectrumSourceFiles since the key is the filename.

> Do we need to add a test?

I'll add this case to DdaSearchTest.

pwiz_tools/Skyline/TestFunctional/TestFunctional.csproj
```diff
@@ -436,6 +436,7 @@
     <None Include="ConfigureToolsDlgTest.zip" />
     <None Include="CopyPasteTest.zip" />
     <None Include="CrosslinkingTest.zip" />
+    <None Include="DdaSearchTest.zip" />
```
chambm, line 439: We could have saved some repo space by copying the file in the test rather than duplicating it in the zip file, but without rewriting the git history it's too late now. Not worth using BFG for 1.3MB.

chambm (approved):

pwiz_tools/Skyline/TestFunctional/TestFunctional.csproj
chambm: @brendanx67 What does including these zips in the project file accomplish? I'm wondering why something wasn't failing because it wasn't already included.

brendanx67: It doesn't actually accomplish much, I guess, since we always keep them from being copied. I personally like to be able to see them in the project viewer, which is why I do it. I feel they really are part of the project. There is some utility in being able to turn on full file view in the Project viewer and consider files that are not added to the project as something worth looking at. Sometimes they turn out to be dead code, which should be removed from the Git source tree. If not adding things becomes normal, it is much harder to use this feature.

PR #1517, opened 2021-03-31
Allow PeptideShaker mzid files to lookup spectra by spectrum title

kaipot:
https://skyline.ms/announcements/home/support/thread.view?rowId=50606

### DIFF
--- a/pwiz_tools/BiblioSpec/src/MzIdentMLReader.cpp
+++ b/pwiz_tools/BiblioSpec/src/MzIdentMLReader.cpp
@@ -219,6 +219,11 @@ void MzIdentMLReader::collectPsms(map<DBSequencePtr, Protein>& proteins) {
                             curPSM_->specName = idStr;
                         }
                         break;
+                    case PEPTIDESHAKER_ANALYSIS:
+                        curPSM_->specName = result.hasCVParam(MS_spectrum_title)
+                            ? result.cvParam(MS_spectrum_title).valueAs<string>()
+                            : idStr;
+                        break;
                     default:
                         curPSM_->specName = idStr;
                         break;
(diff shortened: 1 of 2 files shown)


nickshulman (approved):
Looks good to me.

chambm:
I'm not too sure about this. MGF does not require spectrum title to be unique. That's why we use index to point into MGF. I would want to be sure that PeptideShaker chokes on an MGF that has non-unique titles, because then we'd be reasonably certain to never expect a PeptideShaker mzIdentML for an MGF with non-unique titles. Even then there's no guarantee that they won't change PeptideShaker later to handle non-unique titles.

kaipot:
> I'm not too sure about this. MGF does not require spectrum title to be unique. That's why we use index to point into MGF. I would want to be sure that PeptideShaker chokes on an MGF that has non-unique titles, because then we'd be reasonably certain to never expect a PeptideShaker mzIdentML for an MGF with non-unique titles. Even then there's no guarantee that they won't change PeptideShaker later to handle non-unique titles.

Hmm, didn't know that about MGF. What if we keep using index for MGF but titles for other formats?

chambm:
PeptideShaker ONLY works with MGF AFAIK: http://compomics.github.io/projects/peptide-shaker#mzidentml-support

> PeptideShaker can load results from virtually any identification algorithm in the mzIdentML format as long as the minimal peptide to spectrum match information is present in the file.
> 
> The following is required:
> 
> Spectrum file format has to be mgf.
> Each PSM has a score or e-value as a PSM score CV term.

kaipot:
Weird, on the support request the user shared mzid/mzml results from PeptideShaker. I just tried using msconvert to convert the mzml to an mgf and that seems to work (with indices).

kaipot:
Seems like this was a peptideshaker workflow issue rather than a BiblioSpec issue

PR #1518, opened 2021-03-31
Add "TransitionResultIsQuantitative" and "TransitionResultIsMs1" columns to the Document Grid

nickshulman:
This pull request has the parts from Pull Result #1392 which I would like the next version of MSstats to use.

MSstats needs a more robust way of telling which transitions are not-quantitative.
Also, I think we should give them a more reliable column to look at to decide whether a Transition is MS1 or MS2.

So, the columns "Transition Result Is Quantitative" and "Transition Result Is MS1" are going to be added to the Document Grid.

(For the other parts of #1392 involving deconvoluting isotopes, it was too hard to find any data where it made a compelling difference.)


brendanx67 (approved):
Looks good to me.

PR #1519, opened 2021-03-31
Fix "Loss-only modifications may not be explicit." error

nickshulman:
Now that users are able to add neutral loss modifications to their peptides using the "Modify Peptide" dialog, the error "Loss-only modifications may not be explicit" no longer applies.

(user "jmr" reported this error with a crosslink modification which had zero mass and some neutral losses, but this error can be reproduced with simple neutral loss modifications)

Also, fix an error "Protein sequence may not be empty" that can happen in the Edit Linked Peptides dialog if the focus is in the new row of the peptide sequences grid when you press "OK" on the Edit Linked Peptides dialog when the focus is in an uncommitted new row in the DataGridView.
(also reported by jmr)

### DIFF
--- a/pwiz_tools/Skyline/EditUI/EditLinkedPeptidesDlg.cs
+++ b/pwiz_tools/Skyline/EditUI/EditLinkedPeptidesDlg.cs
@@ -470,9 +470,16 @@ namespace pwiz.Skyline.EditUI
             for (int i = 0; i < _peptideRows.Count; i++)
             {
                 var peptideRow = _peptideRows[i];
-                if (i == _peptideRows.Count - 1 && string.IsNullOrEmpty(peptideRow.Sequence))
+                if (string.IsNullOrEmpty(peptideRow.Sequence))
                 {
-                    continue;
+                    if (i == _peptideRows.Count - 1)
+                    {
+                        // ignore the last row if it is blank
+                        continue;
+                    }
+                    MessageDlg.Show(this, Resources.PasteDlg_ListPeptideSequences_The_peptide_sequence_cannot_be_blank);
+                    SetGridFocus(dataGridViewLinkedPeptides, i, colPeptideSequence);
+                    return;
                 }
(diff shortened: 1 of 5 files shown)


brendanx67 (changes requested):
Maybe better to revert all the extra changes to Resources.Designer.cs. Most of them are white space, and two depend on where a string gets truncated in the header comment. Not so sure that the next developer won't just end up setting these back. I have seen other changes recently that were limited to just the strings being changed in that file.

nickshulman:
> Maybe better to revert all the extra changes to Resources.Designer.cs. Most of them are white space, and two depend on where a string gets truncated in the header comment. Not so sure that the next developer won't just end up setting these back. I have seen other changes recently that were limited to just the strings being changed in that file. Otherwise, it looks good.

nickshulman:
> Maybe better to revert all the extra changes to Resources.Designer.cs. Most of them are white space, and two depend on where a string gets truncated in the header comment. Not so sure that the next developer won't just end up setting these back. I have seen other changes recently that were limited to just the strings being changed in that file. Otherwise, it looks good.

I suspect that might be happening because I recently used the "Help > Check for Updates" menu item in Visual Studio. I am running Visual Studio version 15.9.34.
What version are other people running?
The dates of each patch release can be found on this page:
https://docs.microsoft.com/en-us/visualstudio/releasenotes/vs2017-relnotes

nickshulman:
> Looks like I haven't done that in a while. I am using 15.9.23. Maybe we can have everyone do that after we release 21.1 and until then just try avoid unexpected changes to RESX files, i.e. revert the parts you did not intend.

I have reverted the lines where the text of the comments changed.
The remaining changes are cases where the file in master has LF instead of CRLF.
It looks like those incorrectly terminated lines were introduced with the DIA-Umpire checkin earlier this week. (It was complicated to track down, since git blame does not tell you who changed a line ending: only who changed a line)

PR #1521, opened 2021-04-02
Pre-build step to alphabetize our resx files

nickshulman:
This PR contains a program "SortRESX.exe" which can rewrite .resx files so that all of the elements are sorted by the "name" attribute.
I also made it so that it removes a few tags such as "...TrayLocation" and "&gt;&gt;....Type" which often cause a lot of annoying merge conflicts (such as when the name of the Assembly changes from "Skyline-Daily" to "Skyline").

We could run this tool on the source tree as part of a Skyline pre-build step. (It takes less than a second to traverse the entire tree if no .resx files need to be modified, or a few seconds if all of the .resx files need to be changed.).

In this pull request, I made it so that it only ran on the .resx files in "pwiz_tools\Skyline\EditUI" so that you can get a sense of what changes it would make to .resx files.

Would doing this to the .resx files make the process of localization easier? Would there be value in doing this before we send the files off to be localized, or is this something that should wait until we start working on Skyline 21.2.


chambm:
Perhaps we could do this as a precommit hook? Those are client-side so they can't be enforced, but we could have a script which would create the hook for new developers/clones. And it'd be pretty obvious if someone commits a PR which which didn't use the hook.

brendanx67:
I am all for making this fairly automatic. Unsure how much this would impact the localization process the first time. Hopefully not a lot, but I could imagine reasons why it might require extra work for the first release containing these changes.

PR #1524, opened 2021-04-04
Fix IndexOutOfRangeException which can happen if using "Modify" menu …

nickshulman:
…item on a molecule and changing its chemical formula which results in some precursor transitions no longer existing in the mass distribution.

reported by Chris Ashwood on exception web.

Chris ran into this same exception back in January, so it looks like we forgot to fix it then.
This is the best fix that I could think of: we just skip any transitions whose mass indexes are no longer possible with the new formula.


bspratt:
Looks great. This has been on my to do list but obviously I never got to it.

bspratt (approved):

PR #1530, opened 2021-04-08
Fix problem where DDA search incorrectly adds isotope modifications to the document as structural modifications.

nickshulman:
Here is a possible fix for the bug I sent mail about where the perform DDA search wizard incorrectly adds isotope modifications to PeptideModifications.StaticModifications. (It's really confusing when this happens and it took me a while to figure out what was wrong with my document)

It might be that there are some other things that we should fix about this. Specifically, it's not clear that the set of modifications in the document should change at all based on what the users check or uncheck on the "Add Modifications" page of the DDA search wizard. If the wizard does not add a modification to the document, then I imagine that the modification will be a "explicit" modification, which might work perfectly fine with the search results.

I am happy to write a functional test for this since it might be tricky to make sure we get the behavior right.


chambm (approved):
Looks much nicer than the IsDDASearch approach.

PR #1531, opened 2021-04-11
Skyline: Always output a message beginning with "Error:" when returning a non-zero exit code from CommandLine.

vagisha:


brendanx67 (approved):
I made a few changes myself, after realizing that SkylineRunner is distributed as a single EXE file, and we are not going to add resource DLLs to that distribution. So, for the foreseeable future it cannot be localized.

I think will probably merge this today or early tomorrow to get the RESX file changes into Skyline for translation with 21.1. But, we really need to make sure we have a way of testing this before releasing a new SkylineRunner. I did not feel at all certain my changes would work flawlessly.

Also, I don't like the brittleness of StartsWith("Error:") in SkylineRunner, which seems like it would only work if SkylineRunner is called without --timestamp or --memstamp, which add line prefixes. I think that is fine for AutoQC Loader and Skyline Batch, because they write the timestamps themselves, but it is a bit weird that adding --timestamp will turn off exit error codes.

All this can be addressed in a follow-up PR.

brendanx67 (approved):
Seems good enough to merge to master. Though it will need more testing before releasing a new SkylineRunner.exe. Merging to master now will allow us to get the new string resources into translation for 21.1.

brendanx67:
Or put the code in a separate class and .cs file in SkylineRunner and compile that in both SkylineRunner and TestData. Maybe check with Nick on what he would recommend.

PR #1532, opened 2021-04-11
Skyline: A first pass at making the Barcelona PRM on Fusion Lumos tutorial

brendanx67:
an official Skyline tutorial


brendanx67:
I updated the Word doc for this tutorial created by the PRBB team for Skyline 20.2, which we will be teaching with for Skyline Online 2021. I did not fully update it to meet the Skyline Tutorial Style Guidelines, nor did I automate it, but these are things we should do in the coming months. This has been a popular tutorial for Skyline classes since it was introduced in November, 2017, and it was then presented as Webinar 17 in January, 2018.

brendanx67:
I think the best next step might be for Nat to give it a pass and create a version that meets the style guidelines, with the existing screenshots I just took. Then (or concurrently) someone needs to write the test to make sure it doesn't get broken, and automate taking screenshots so that we can get it into our queue for translation to Chinese and Japanese.

brendanx67:
This work moved to PR #1962

PR #1533, opened 2021-04-13
Add a View > Targets menu item for non-proteomic UI modes.

bspratt:
Visible in Mixed and Small molecule modes only. Enabled only when the Targets window is not open.

Also, hide the View > Modifications menu item for small molecule UI mode.

This fixes:

Issue 791 Inconsistencies in Small Molecule UI mode make it difficult to reopen a closed Targets window
"From Brendan:
Bug found teaching Skyline Online. In small molecule mode, View > Proteins > etc. gets removed but not replaced with anything, which makes it confusing how to get the Targets window back if you close it. We should replace it with a View > Targets menu item that brings up the Targets view.

Also, we seem to leave View > Modifications > etc. in small molecule mode, when it doesn't apply to small molecules. It also brings up the Targets window. So, that was my easiest solution to getting the Targets window back after closing it, but the menu item should not be there. So we should be removing View > Proteins and Modifications and replacing them with View > Targets."


bspratt:
I figure the only likely design controversy here is the enable/disable of the menu item based on the Targets window visibility. I think it's good that users are made aware of its availability, but it would be confusing if it was left enabled when clicking on it would produce no action.

Some screen shots (note the UI modes)

image
image
image
image
image

brendanx67:
Don't disable it when it is visible, just put a check next to it, and when you choose it and it is visible, it should still set the focus to the Targets view. It is an "Activate" command, and not a Show/Hide command, but we do put a check by it when it is visible to let the user know that Skyline thinks it is visible. So, if it is not immediately apparent, the user might want to search for it in a situation where there is a lot of screen real estate and the window may be floating and partially off screen.

nickshulman (approved):
Looks good to me

bspratt:
So:
image
image
image
image
image

pwiz_tools/Skyline/Menus/ViewMenu.cs
```diff
@@ -827,6 +832,7 @@ public void DocumentUiChanged()
         }
         public void ViewMenuDropDownOpening()
         {
+            viewTargetsMenuItem.Checked = Settings.Default.ShowPeptides;
```
brendanx67, line 835: It looks like the targets window can be shown without setting Settings.Default.ShowPeptides to true, which should get fixed before relying on it in this way. Look for uses of SkylineWindow.ShowSequenceTreeForm().

brendanx67 (changes requested):
I think the line to set Settings.Default.ShowPeptides needs to move in the function SkylineWindow.ShowSequenceTreeForm.

pwiz_tools/Skyline/Menus/ViewMenu.cs
bspratt: Huh, ShowSequenceTreeForm(false) appears to very carefully not change the value of Settings.Default.ShowPeptides, so I guess I don't really understand what that value means, but it seems it's not a reliable indicator of that window's visibility.

I'll set that checkmark value to reflect the targets window being non-null and visible.

brendanx67: Sounds safest. I am pretty sure we are not either hooking when the user clicks the X button to close the Targets window. So, yeah, not reliable to use the Settings value as an indicator of Targets view visibility.

bspratt:
OK, I'll back out anything that directly messes with that settings value.

brendanx67:
That sounds like more than I was suggesting. I just want you to put back 2 lines in your last commit. Using the View menu to show the Targets view changes the value today, and it should change the value for your View > Targets menu item. However, calls to SkylineWindow.ShowSequenceTreeForm(true) should not change the Settings.Default.ShowPeptides value, as it has not in the past. Your change to rely on visibility for the menu check mark was the right fix.

bspratt:
Agreed, I just meant to say no *new* manipulations. I think you'll find the current code meets your expectations, including having View>Targets available in all UI modes per our side conversation.

PR #1534, opened 2021-04-13
Add ability to download files for configurations

alinmarsh:


pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/ConfigRunner.cs
```diff
@@ -200,6 +241,92 @@ public void WriteBatchCommandsToFile(CommandWriter commandWriter, int startStep)
...
+            var allFiles = server.GetServerFiles;
+            var fileNames = allFiles.Keys;
+            var existingDataFiles = Directory.GetFiles(mainSettings.DataFolderPath).Length;
+            var dataFilter = new Regex(mainSettings.Server.DataNamingPattern);
+            var downloadingFilesEnum =
+                from name in fileNames
+                where dataFilter.IsMatch(name)
+                select name;
+            var downloadingFiles = downloadingFilesEnum.ToList();
+            if (downloadingFiles.Count == existingDataFiles && !Program.FunctionalTest) return;
```
nickshulman, line 260: It looks like you're assuming that all of the files in the folder are ones that you were supposed to download (so that if the number of files in the folder match, you can assume that you're done).

Usually there are ways for bogus files to end up in there so that's not a safe assumption to make.

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/ConfigRunner.cs
```diff
@@ -36,15 +44,17 @@ public class ConfigRunner : IConfigRunner
         private readonly object _lock = new object();
         private RunnerStatus _runnerStatus;
         private readonly ProcessRunner _processRunner;
-        private List<string> _batchCommandsToLog;
+        private string _batchFile;
+
+        private List<CancellationTokenSource> _runningCancellationTokens;
```
nickshulman, line 49: You only need one CancellationTokenSource, since you never want to cancel a subset of the things that are happening.

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/ConfigRunner.cs
```diff
@@ -200,6 +241,92 @@ public void WriteBatchCommandsToFile(CommandWriter commandWriter, int startStep)
...
+                var fileName = downloadingFiles[i];
+                if (triesOnFile == 0)
+                {
+                    _logger.Log(string.Format(Resources.ConfigRunner_DownloadData__0____1__of__2__, fileName, i + 1, downloadingFiles.Count));
+                }
+                
+                var localFileDestination = Path.Combine(mainSettings.DataFolderPath, fileName);
+                var currentFileSize = (double) allFiles[fileName].Size;
+
+                if (File.Exists(localFileDestination) && Math.Abs(new FileInfo(localFileDestination).Length - currentFileSize) < 0.0001)
```
nickshulman, line 284: Why are you comparing these values as doubles instead of long integers?

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/ConfigRunner.cs
alinmarsh: Good point. It checks that the files exist and are the correct size now.

pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/ConfigRunner.cs
alinmarsh: I forgot I casted to the double, oops

pwiz_tools/Skyline/Executables/SharedBatch/LaunchBatch/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// </summary>
+        [STAThread]
+        static void Main(string[] args)
+        {
+            Application.EnableVisualStyles();
+            Application.SetCompatibleTextRenderingDefault(false);
+
+            var appReferencePath = args[0];
+
+            // join remaining args (will be more than one if filepath has spaces)
```
nickshulman, line 30: This doesn't sound right. Normally paths with spaces appear as a single argument value.
If they are coming in as more than one argument, then it usually means that someone else was supposed to put quotes around it.
Trying to reconstruct the original value by adding spaces in between the arguments that you got is not going to work if the path happened to have two spaces in a row in it.

alinmarsh: That makes sense, I didn't have enough escaped quotes when I added it to the registry. Fixed it now and the path comes in as one argument, thanks!

PR #1537, opened 2021-04-15
Fix NRE that can occur doing "Remove RT Outliers" if other Retention …

nickshulman:
…Time graphs (e.g. peptide comparison RT) are visible in addition to the Retention Time Linear Regression graph.

(Reported by Aaron on the exception web)

It used to be that the code would use the first graph window in "_listGraphRetentionTime", which often would not even be a linear regression graph. This change makes it so that it looks for the RTLinearRegressionGraphPane.
I am thinking that this is probably not the correct fix either. That is, there might be more than one RTLinearRegressionGraphPane showing (maybe both "Score to Run" and "Run to Run") and I really need to find the one that the user right-clicked on.

I am planning on changing this so that it looks instead for the active window, and finds the graph that way.


brendanx67 (approved):

nickshulman:
It is possible to have both the "Score to Run" and "Run to Run" linear regression graph panes visible at the same time, so this fix is going to be re-engineered to make it so that the Remove Outliers command gets sent to the correct window.
I will create a new pull request once I figure out the correct fix.

PR #1539, opened 2021-04-16
Always use comma separators in invariant exported reports

alinmarsh:


nickshulman:
In the Skyline graphical user interface, users can choose either "CSV" or "TSV" in the SaveFileDialog.

The code to figure out which column separator to use based on what they chose in the SaveFileDialog is in SkylineViewContext.Export(Control, ViewInfo):
```
                char separator = saveFileDialog.FilterIndex == 2
                    ? TextUtil.SEPARATOR_TSV
                    : TextUtil.GetCsvSeparator(DataSchema.DataSchemaLocalizer.FormatProvider);

```
Wouldn't we want to do the same thing here?

brendanx67:
So, you would prefer this?

            var schemaLocalizer = reportInvariant
                ? DataSchemaLocalizer.INVARIANT
                : SkylineDataSchema.GetLocalizedSchemaLocalizer();
            var viewContext = DocumentGridViewContext.CreateDocumentGridViewContext(_doc, schemaLocalizer);
            if (reportColSeparator != TextUtil.SEPARATOR_TSV)
                TextUtil.GetCsvSeparator(schemaLocalizer.FormatProvider);

It is non-trivial to use exactly the ternary expression you have provided, because the reportColSeparator is calculated in CommandArgs based on --report-format=<csv | tsv> where --report-invariant may be specified after this on the command-line.

        public static readonly Argument ARG_REPORT_FORMAT = new DocArgument(@"report-format",
            new []{ARG_VALUE_CSV, ARG_VALUE_TSV},
            (c, p) => c.ReportColumnSeparator = p.IsValue(ARG_VALUE_TSV)
                    ? TextUtil.SEPARATOR_TSV
                    : TextUtil.CsvSeparator);
        public const string ARG_VALUE_CSV = "csv";
        public const string ARG_VALUE_TSV = "tsv";
        public static readonly Argument ARG_REPORT_INVARIANT = new DocArgument(@"report-invariant",
            (c, p) => c.IsReportInvariant = true);

nickshulman:
Oops. I misunderstood. I thought this change was going to use commas as the separator column separator when the language was French.
What you have here looks correct.

brendanx67:
I think what is in the PR now is easier to read, but maybe add a comment.

// Make sure invariant report format uses a true comma if a tab separator was not specified.

PR #1541, opened 2021-04-19
Skyline: Fixed File > Import > Transition List for small molecules

brendanx67:
- Broken during the implementation of the column picker for peptide transition list import
(reported by Skyline Online)

Initial draft. Needs an explicit test to prove and avoid breaking again.


bspratt:
OK, new test is in place. This all looks good to me.

PR #1543, opened 2021-04-19
Full scan viewer menu updates

rita-gwen:
Reproducing the menu options available in the library match window in the full scan viewer.


rita-gwen:
Added a menu option to sync the m/z scale between the library match and full scan graphs.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -1022,11 +1063,18 @@ private void btnIsolationWindow_Click(object sender, EventArgs e)
 
         private void graphControl_ContextMenuBuilder(ZedGraphControl sender, ContextMenuStrip menuStrip, Point mousePt, ZedGraphControl.ContextMenuObjectState objState)
         {
-            ZedGraphHelper.BuildContextMenu(graphControl, menuStrip, true);
+            //ZedGraphHelper.BuildContextMenu(graphControl, menuStrip, true);
+            menuStrip.Tag = this;
```
nickshulman, line 1067: I think it's better to construct a new menu item with an event handler which has all of the variables that it needs, rather than trying to store that information in the "Tag" property of anything.
The problem with using Tag like this is that it's really hard for someone to track down which code is going to use that Tag.

pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
```diff
@@ -1603,4 +1625,9 @@ public SelectedSpectrumEventArgs(SpectrumDisplayInfo spectrum, bool isUserAction
         public SpectrumDisplayInfo Spectrum { get; private set; }
         public bool IsUserAction { get; private set; }
     }
+    public sealed class MzRange
```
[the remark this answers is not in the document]
rita-gwen: I made the set accessors private, that should do it.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -451,6 +453,25 @@ private void CreateIonMobilityHeatmap()
...
+        {
+            public TransitionGroupDocNode precursor;
+            public ImmutableList<IonType> types;
+            public ImmutableList<int> charges;
+            public ImmutableList<Adduct> rankAdducts;
+            public ImmutableList<IonType> rankTypes;
+
+            public int CompareTo(RankingContext other)
+            {
+                var res = ReferenceEquals(precursor.Id, other.precursor.Id) && types.SequenceEqual(other.types)
```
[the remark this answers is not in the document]
rita-gwen: Fixed.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -451,6 +453,25 @@ private void CreateIonMobilityHeatmap()
...
+            public ImmutableList<int> charges;
+            public ImmutableList<Adduct> rankAdducts;
+            public ImmutableList<IonType> rankTypes;
+
+            public int CompareTo(RankingContext other)
+            {
+                var res = ReferenceEquals(precursor.Id, other.precursor.Id) && types.SequenceEqual(other.types) 
+                          && charges.SequenceEqual(other.charges) 
+                          && rankAdducts.SequenceEqual(other.rankAdducts) && rankTypes.SequenceEqual(other.rankTypes);
+                return res ? 0 : 1;
```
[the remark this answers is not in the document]
rita-gwen: Fixed.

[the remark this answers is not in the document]
nickshulman: This hasn't changed. Did you forget to push your changes?

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
rita-gwen: There is no clean way to destroy a menu item after it is created. Usually it leads to GDI handles leak, as we discussed. Fortunately, there is a property on the ContextMenuToolStrip pointing to the control that initiated the context menu. I will just use that instead.

nickshulman:
I believe that the "Synchonize m/z axis" menu item is supposed to have a checkmark next to it and change a boolean setting. When the setting is on, then any time the user zooms in or out on either the Full Scan graph or the Library Match graph, the other graph immediately updates to match it.
In this way, the menu item would behave similar to the "Synchronize Zooming" menu item on the chromatogram graph.

The way you it is implemented now, the user has to click the "Sync m/z scale" menu item in order to get the other window to update.

I think the way that the chromatograms synchronize their zooming is a good example to follow. In the function "SkylineWindow.graphChromatogram_ZoomAll", Skyline iterates over all of the windows that need to be synchronized to each other, skipping the one which fired the event, and tells each of them to zoom to the new range. That event gets fired by "graphControl_zoomEvent" in GraphChromatogram.cs.

rita-gwen (commented):
Nick, could you look though the last commit for the dynamic sync m/z implementation?  The tests are passing, hopefully I should be able to merge this PR this week.

pwiz_tools/Skyline/Skyline.resx
```diff
@@ -3915,4 +3915,16 @@
   <data name="&gt;&gt;$this.Type" xml:space="preserve">
     <value>pwiz.Skyline.Util.FormEx, Skyline, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null</value>
   </data>
+  <data name="synchMzScaleToolStripMenuItem.Size" type="System.Drawing.Size, System.Drawing">
+    <value>193, 22</value>
+  </data>
+  <data name="synchMzScaleToolStripMenuItem.Text" xml:space="preserve">
+    <value>Sync m/z scale</value>
```
brendanx67, line 3922: We use the full English term "Synchronize" for "Synchronize Zooming" on the chromatogram graphs. Stay consistent. Also it should be "Synchronize m/z Scale". You can't capitalize "m/z" (when possible it should be italicized), but otherwise menu items are Title Case.

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -907,6 +907,28 @@ private void charge4MenuItem_Click(object sender, EventArgs e)
         {
             _graphSpectrumSettings.ShowCharge4 = !_graphSpectrumSettings.ShowCharge4;
         }
+        private void synchMzScaleToolStripMenuItem_Click(object sender, EventArgs e)
+        {
+            if (_graphFullScan == null || !_graphFullScan.Visible || _graphSpectrum == null || !_graphSpectrum.Visible) return;
```
brendanx67, line 912: This code needs work. 1. we don't use single line "if (condition) return;" These should be 2 lines.

We generally use "as" instead of "if (x is Type) t = (Type) x;

Assigning senter to source seems like it could be done using ?? followed by your other expression.

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -907,6 +907,28 @@ private void charge4MenuItem_Click(object sender, EventArgs e)
...
+            IMzScaleCopyable target = _graphSpectrum;
+            if (source is GraphSpectrum)
+                target = _graphFullScan;
+            target?.SetMzScale(source.Range);
+        }
+
+        public void SynchMzScale(bool direction = false)
+        {
+            if(direction)
+                synchMzScaleToolStripMenuItem_Click(_graphSpectrum, EventArgs.Empty);
```
brendanx67, line 928: We don't use "if(x)" but "if (x)" instead. We don't call event handler methods directly, but introduce a function that the event handler calls and call that from the other function.

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -1038,7 +1064,7 @@ void GraphSpectrum.IStateProvider.BuildSpectrumMenu(bool isProteomic, ZedGraphCo
             menuStrip.Items.Insert(iInsert++, toolStripSeparator14);
 
             // Need to test small mol
-            if (isProteomic)
+            if (isProteomic && control is GraphSpectrum)
```
brendanx67, line 1067: Too much use of "is GraphSpectrum". Better to use polymorphism, i.e. put something on your interface that both implementations can answer to distinguish behavior.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -93,6 +95,7 @@ public GraphFullScan(IDocumentUIContainer documentUIContainer)
             spectrumBtn.Visible = false;
             filterBtn.Visible = false;
             lblScanId.Visible = false; // you might want to show the scan index for debugging
+            toolBar.Visible = false;
```
brendanx67, line 98: Not sure about this change. Does it lead to a sudden appearance of the toolbar after the spectrum loads? That's a big layout change. Why is it necessary now?

pwiz_tools/Skyline/TestFunctional/FullScanGraphTest.cs
```diff
@@ -157,6 +157,13 @@ protected override void DoTest()
             TestScale(529, 533, 0, 50);
             ClickChromatogram(33.06, 68.8, PaneKey.PRECURSORS);
             TestScale(452, 456, 0, 300);
+
+            RunUI(() => SkylineWindow.ShowGraphSpectrum(true));
```
brendanx67, line 161: I don't see how this tests the feature beyond that it doesn't crash. You should be checking for matching coordinates, right?

brendanx67 (changes requested):
Doesn't look ready to merge to me.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.resx
```diff
@@ -486,10 +462,16 @@
...
+  </data>
+  <data name="&gt;&gt;toolStripSeparator1.Type" xml:space="preserve">
+    <value>System.Windows.Forms.ToolStripSeparator, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </data>
   <data name="&gt;&gt;$this.Name" xml:space="preserve">
     <value>GraphFullScan</value>
   </data>
   <data name="&gt;&gt;$this.Type" xml:space="preserve">
-    <value>pwiz.Skyline.Util.DockableFormEx, Skyline, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null</value>
+    <value>pwiz.Skyline.Util.DockableFormEx, Skyline-daily, Version=20.2.1.468, Culture=neutral, PublicKeyToken=null</value>
```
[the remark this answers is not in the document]
rita-gwen: Corrected.

pwiz_tools/Skyline/Skyline.resx
rita-gwen: Corrected.

pwiz_tools/Skyline/SkylineGraphs.cs
rita-gwen: Corrected.

pwiz_tools/Skyline/SkylineGraphs.cs
rita-gwen: Corrected.

pwiz_tools/Skyline/SkylineGraphs.cs
rita-gwen: Corrected.

pwiz_tools/Skyline/TestFunctional/FullScanGraphTest.cs
rita-gwen: Added scale assertions for the sync in both directions.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
rita-gwen: I thought that showing a toolbar with some active controls in an empty window could be confusing to the user. There is no functional need for this change. Reverted.

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -908,6 +908,35 @@ private void charge4MenuItem_Click(object sender, EventArgs e)
             _graphSpectrumSettings.ShowCharge4 = !_graphSpectrumSettings.ShowCharge4;
         }
 
+        private void SynchMzScaleToolStripMenuItemClick(object sender)
+        {
+            if (_graphFullScan == null || !_graphFullScan.Visible
+                                       || _graphSpectrum == null || !_graphSpectrum.Visible)
+                return;
+            Settings.Default.SyncMZScale = synchMzScaleToolStripMenuItem.Checked;
```
nickshulman, line 916: After this line you should do:
if (!Settings.Default.SyncMZScale)
   return;

It probably won't make a difference, since the graph axes are going to already have been synced with each other, but Skyline should not try to sync them one last time when the user is turning off synchronization.

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -908,6 +908,35 @@ private void charge4MenuItem_Click(object sender, EventArgs e)
...
         }
 
+        private void SynchMzScaleToolStripMenuItemClick(object sender)
+        {
+            if (_graphFullScan == null || !_graphFullScan.Visible
+                                       || _graphSpectrum == null || !_graphSpectrum.Visible)
+                return;
+            Settings.Default.SyncMZScale = synchMzScaleToolStripMenuItem.Checked;
+            //testing support
+            var source = sender as IMzScaleCopyable ??
```
nickshulman, line 918: The "//testing support" comment only applie to the "as IMzScaleCopyable" part of this statement.
You should probably make this into a full "if ... else" statement so that you can comment the two parts of the if statement separately.

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -908,6 +908,35 @@ private void charge4MenuItem_Click(object sender, EventArgs e)
...
+                         (IMzScaleCopyable)(synchMzScaleToolStripMenuItem.Owner as ContextMenuStrip)?.SourceControl?.Parent?.Parent;
```
nickshulman, line 919: I think you should do:
(IMzScaleCopyable)(synchMzScaleToolStripMenuItem.Owner as ContextMenuStrip)?.SourceControl?.FindForm()

The problem with "Parent?.Parent" is that it would stop working if someone redesigned the form so that the ZedGraphControl was inside of an extra Panel or something.

nickshulman (commented):
I would recommend that you have a function called "ListMzScaleCopyables":
```
        private IEnumerable<IMzScaleCopyable> ListMzScaleCopyables()
        {
            if (_graphFullScan != null && _graphFullScan.Visible)
            {
                yield return _graphFullScan;
            }

            if (_graphSpectrum.Visible && _graphSpectrum.Visible)
            {
                yield return _graphSpectrum;
            }
        }
```

Then, whenever you want to synchronize with a particular Source window. Then you can do something like this:
```
            foreach (var target in ListMzScaleCopyables())
            {
                if (!ReferenceEquals(target, source)) {
                    target.SetMzScale(source.Range);
                }
            }
```

I think this ends up being easier than checking whether "source" is "_graphFullScan" or "_graphSpectrum".
Also, this would make it easier in the future if we ever wanted to increase the set of graphs that are having their X-axes synchronized. For instance, we might want to make it so that the graph on the View > Spectral Libraries window would get synchronized. Or we might want to allow the user to show more than one GraphFullScan at the same time.

pwiz_tools/Skyline/SkylineGraphs.cs
rita-gwen: This was an if statement originally but has been changed to inline ?? by Brendan's request. Could you guys decide between yourself which one is better and let me know your verdict.

rita-gwen: This was the original Brendan's comment:
> Assigning senter to source seems like it could be done using ?? followed by your other expression.

brendanx67:
I resolved the place where Nick and I disagreed and did a bit more clean-up, and added a TODO. I still see any number of bugs when I play with synchronize m/z axis in the UI. This still does not feel ready to merge.

Rita, let's definitely do a UI review and look at the things I am trying in my testing.

PR #1561, opened 2021-04-28
Skyline: Improve command-line agreement between Error: logging and fa…

brendanx67:
…ilure exit codes


alinmarsh:
Looks good!
image
This was the case where it used to continue to export reports. Now it stops as expected.

alinmarsh (approved):

PR #1562, opened 2021-04-28
Overloads for ChromatogramList functions on non-MSVC builds

chambm:
* added overloads for several ChromatogramList functions on non-MSVC builds


bspratt (commented):
This all seems fine, but I'm confused as to why TC isn't barking about this already?

chambm:
Yeah, I'm confused as to why all the GCC builds aren't failing without this. I know I only made the change because some Linux build I tried gave an error without it, but I don't remember which settings caused it. I don't see any harm merging this though.

bspratt:
Agreed, go ahead and pull the trigger or I can if you prefer.

PR #1569, opened 2021-04-30
Fix failure in "FullScanFilterTest" when running all tests within Visual Studio.

nickshulman:
SkylineRunnerErrorDetectionTest was causing (by means of the "--culture" commandline argument) LocalizationHelper.CurrentUICulture to get changed and not changed back, which caused subsequent tests to fail.

The "--culture" command line argument is only necessary if the test needs to test the commandline in a new process (in which case it does not matter if it cleans itself up).
Since SkylineRunnerErrorDetectionTest is doing its work by calling into the commandline code in this process, it can just use "LocalizationHelper.CallWithCulture" to cause the culture to be set to what it needs to.

The FullScanFilterTest fails if you do "Run All Tests" in Visual Studio. It's also been failing in the Debug code coverage build on TeamCity.

### DIFF
--- a/pwiz_tools/Skyline/TestData/CommandLineTest.cs
+++ b/pwiz_tools/Skyline/TestData/CommandLineTest.cs
@@ -3065,6 +3065,11 @@ namespace pwiz.SkylineTestData
             TestDetectError(true, true, ci);   // both timestamp and memstamp
         }
 
+        /// <summary>
+        /// Tests that "IsErrorLine" works when the commandline is invoked in a particular culture.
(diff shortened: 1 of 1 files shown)


vagisha (approved):
Thank you so much for the explanation! Can you please add this as a comment in the test?

brendanx67 (commented):
Should we remove --culture then? I think it was added just to support this test.

nickshulman:
No, "--culture" is used in all of the other commandline tests which launch
the commandline as a separate process. Check out the callers of
GetProcessStartInfo.

PR #1570, opened 2021-04-30
Fix problem where Audit Log (.skyl) file does not get included in .sk…

nickshulman:
…y.zip

(Reported by Brendan)

### DIFF
--- a/pwiz_tools/Skyline/Model/SrmDocumentSharing.cs
+++ b/pwiz_tools/Skyline/Model/SrmDocumentSharing.cs
@@ -249,7 +249,7 @@ namespace pwiz.Skyline.Model
             }
             else
             {
-                zip.AddFile(DocumentPath);
+                AddDocumentAndAuditLog(zip, DocumentPath);
             }
             Save(zip);
         }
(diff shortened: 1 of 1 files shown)


pwiz_tools/Skyline/Model/SrmDocumentSharing.cs
```diff
@@ -378,6 +378,12 @@ private void SaveDocToTempFile(ZipFileShare zip)
             string entryName = GetDocumentFileName();
             string docFilePath = Path.Combine(EnsureTempDir().DirPath, entryName);
             Document.SerializeToFile(docFilePath, docFilePath, ShareType.SkylineVersion ?? SkylineVersion.CURRENT, ProgressMonitor);
+            AddDocumentAndAuditLog(zip, docFilePath);
+        }
+
+        private void AddDocumentAndAuditLog(ZipFileShare zip, string docFilePath)
+        {
+            zip.AddFile(docFilePath);
```
brendanx67, line 386: This makes it look like we add the docFilePath twice. Can that be right?

brendanx67 (commented):
Unclear on why it looks like we add the docFilePath twice.

PR #1576, opened 2021-05-04
Fix problem with BlibSearch where no matches are found if the spectru…

nickshulman:
…m charges are negative.

(Reported by Chris)


chambm:
At a glance I'd say you're right, this should be done in a sign-sensitive way. If "negative" charges are going to match, the charges should be negative in the `vector<int>` returned by `BiblioSpec::getPossibleCharges()`. I suspect it'll be as simple as flipping the sign when reading spectra in (based on their polarity; the actual charge state in mzML should never be negative), but I'll look at it tomorrow.

chambm:
Who is using BlibSearch? As far as I can tell the output file renaming is broken: it always writes a `.report.tmp` file instead of  `.report` (it's supposed to rename from the former to the latter). The tests have kept passing because the test inputs directory has the expected filenames. Not sure why that is but I'll rename the test outputs so they have to be regenerated for tests to pass.

nickshulman:
The only person I know of who is using BlibSearch is Chris Ashwood in this support request:
https://skyline.ms/announcements/home/support/thread.view?rowId=50988

The MacCoss lab used to have BlibSearch as part of the standard peptide search pipeline that we ran all the time, but that was probably more than five years ago.

I was wondering why the output file was being created with such a strange name.

chrashwood:
Hi Nick and Matthew,

Thank you both for working on this. I would like to contribute some code to improve functionality to better support small molecule blib files. Thanks to Nick's fix regarding negative polarity, I should be able to work on the BlibSearch code once merged. Is this feasible?

nickshulman:
Chris,

You can get started working on this now without anyone's permission. Click the "Fork" button in the top right corner of the github page to create your own copy of the repository. You will have a copy of all of the branches that are in this repository, including the branch that contains this fix.
You can commit your changes to your repository, and when you want those changes to be integrated into the original ProteoWizard repository you can come back here and create a pull request from your for to this repository.

Instructions for how to build Skyline are here:
https://skyline.ms/wiki/home/software/Skyline/page.view?name=HowToBuildSkylineTip

The only difference for building BiblioSpec would be that you would execute the command:
`qb64 pwiz_tools\BiblioSpec\src`

chambm:
I suspect that search-decoy test was only passing before because of the rename issue. The output file it actually tested against is in the inputs tarball from 2014. Will look at it hopefully tomorrow.

nickshulman:
I am closing this since pull request #1584 includes these changes and we should just commit that pull request when it is ready.

chambm:
OK but don't delete the branch because I retargeted 1584 onto this branch.

PR #1579, opened 2021-05-07
Add iRT standards to document when importing peptide search with automatic/cirt

kaipot:


chambm:
I'm getting an error `A protein sequence may not contain the character '[' at 4` when one of the CiRT peptides includes a mod (e.g. [C+57]). Can you check it when one of the standards chosen has a mod?

kaipot:
> I'm getting an error `A protein sequence may not contain the character '[' at 4` when one of the CiRT peptides includes a mod (e.g. [C+57]). Can you check it when one of the standards chosen has a mod?

Good catch, it'll be a problem trying to add peptides to a document if it contains an unknown modification. Not sure if we can automatically create the modifications either since we don't necessarily know if it is a structural or isotope modification. Any suggestions?

brendanx67:
Use modification matcher and make sure it solves everything we know of. Leave it at that until we receive a new user complaint. Do we know of anything besides C[+57] (structural) and K[+8]/R[+10] (isotope)? You could even use a hack to solve for those with a TODO to create a more general solution. We really need to get past the current blocking issue.

kaipot:
> Use modification matcher and make sure it solves everything we know of. Leave it at that until we receive a new user complaint. Do we know of anything besides C[+57] (structural) and K[+8]/R[+10] (isotope)? You could even use a hack to solve for those with a TODO to create a more general solution. We really need to get past the current blocking issue.

Okay, I'll use modification matcher. That will fix the C+57 case which would be most of them. The case of automatic iRT where it doesn't match any known standard could be a problem but it shouldn't happen often.

chambm (approved):
It now works with my upcoming PASEF tutorial.

chambm:
Found a bug in this. It's adding the iRT counts after the "Import FASTA - This operation has created the following targets" dialog. So the final document counts don't match the status bar counts after finishing the wizard. Can the automatic iRTs be added before the dialog gets its counts?

kaipot:
> Found a bug in this. It's adding the iRT counts after the "Import FASTA - This operation has created the following targets" dialog. So the final document counts don't match the status bar counts after finishing the wizard. Can the automatic iRTs be added before the dialog gets its counts?

I'll fix this in a separate PR.

PR #1584, opened 2021-05-10
Fix for reporting query retention times and library retention times now added to report from blib libraries

chrashwood:
- Query retention times were being reported as 0 instead of the actual value
- Library/reference spectrum retention times were not being recorded in the report file.


chrashwood:
Hi Matt,
The context is that with a negative mode mzML (containing small molecules) as the querySpectra and a small molecule blib (generated in Skyline from an SSL) as the refSpectra, the retention times are 0. Tabs are replaced with spaces, sorry about that.

chambm:
`the refSpectra retention times were 0.` did you mean the querySpectra then?

chambm:
I'm going to merge this into the pwiz branch and work on test failures from there.

PR #1586, opened 2021-05-12
Add tests for SelectIrtStandardDlg and UseCurrentCalculatorDlg

kaipot:


brendanx67 (changes requested):
One minor coding style change. Otherwise, it looks good and I tested that SkylineTester can now show these forms for translation review.

PR #1588, opened 2021-05-12
Code inspection tweaks - please review TeamCity for new failures in Shared/Common

bspratt:
- extend code inspection to Skyline/../Shared/Common
- only run code inspection once
- add antivirus exclusion test to the new run-once list

### DIFF
--- a/pwiz_tools/Skyline/Test/CodeInspectionTest.cs
+++ b/pwiz_tools/Skyline/Test/CodeInspectionTest.cs
@@ -299,7 +299,10 @@ namespace pwiz.SkylineTest
 
             foreach (var fileMask in allFileMasks)
             {
-                foreach (var filename in Directory.GetFiles(root, fileMask, SearchOption.AllDirectories))
+                var filenames = Directory.GetFiles(root, fileMask, SearchOption.AllDirectories).ToList();
+                filenames.AddRange(Directory.GetFiles(Path.Combine(root, @"..", @"Shared", @"Common"), fileMask, SearchOption.AllDirectories));
+
+                foreach (var filename in filenames)
                 {
                     if (Equals(filename, thisFile))
                     {
--- a/pwiz_tools/Skyline/TestRunner/Program.cs
+++ b/pwiz_tools/Skyline/TestRunner/Program.cs
@@ -82,6 +82,9 @@ namespace TestRunner
             {"TestInstrumentInfo", new ExpandedLeakCheck(LeakCheckIterations * 2, true)}
         };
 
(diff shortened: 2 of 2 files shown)


brendanx67 (changes requested):
Looks good, but seems to require more to pass the tests.

bspratt:
> Looks good, but seems to require more to pass the tests.

It's the same set of inspections. All newly detected issues are in code which was not formerly inspected.

brendanx67:
Well, still to merge, those need to be made warnings or they would need to be fixed. Am I wrong that some are causing errors that would break CI and nightly test runs? As usual, it is never okay to break the tests, even for new things we find that are "incorrect". The transition to full lockdown needs to be managed.

bspratt:
I wasn't going to merge until the newly detected issues are fixed.  The code inspection test is already noisy as heck with failures that we're "temporarily" tolerating, adding to that seems pointless.

brendanx67:
That also solves the noted issue: "seems to require more to pass the tests" Somehow it needs to be made to pass the tests before it can be merged.

bspratt:
@nickshulman has indicated that the newly flagged issues are in his code, so hopefully he'll get to those soon, and fix them in this branch. When it's green on TC again we'll merge.

pwiz_tools/Shared/Common/DataBinding/Controls/QuickFilterForm.cs
```diff
@@ -189,7 +189,11 @@ public virtual void ShowErrorMessage(Exception e)
             }
             else
             {
-                MessageBox.Show(this, e.Message);
+                // If there is no ViewContext (which should never happen), fall back to using the system message box to show the
+                // error message.
+                // Put a line break in the function call so that code inspection passes
+                MessageBox
+                    .Show(this, e.Message);
             }
```
bspratt, line 197: There's a mechanism for calling out exceptions to the rule, let me do this one properly before we merge.

bspratt: That's done, I'll assume the various resx changes are good as they are automated and extensive.

nickshulman (approved):
I was thinking we could instead make it so that MessageBox.Show is a warning generated by a custom inspection in Skyline.sln.DotSettings.

CodeInspectionTest only really needs to be used for files that Resharper refuses to inspect for us (i.e. our .designer.cs files).

Either way works well.

bspratt:
If there's a way to express all the things we're looking for, and all the exceptions, that is the preferable option. That's not my impression though.

PR #1597, opened 2021-05-20
Workaround intermittent errors in SpectrumList_DiaUmpireTest

chambm:


brendanx67:
Should we put a timebomb on this to make sure we don't just leave it this way indefinitely?

chambm:
Maybe, but this change didn't solve the problem. I'm pretty perplexed how it's getting intermittent logic errors when it's running only a single thread! Memory corruption seems like the only explanation to me but running with the debug CRT doesn't pick up anything. Sigh...

PR #1600, opened 2021-05-21
When adding small molecule library values to a document, just ignore any fragment peaks with absurdly small or large m/z values …

bspratt:
… (this was prompted by a user with an msp file that contained a fragment of m/z = 0.3, the other 4.19 million lines were fine). Formerly was an unhandled exception.


brendanx67:
The need for this makes me wonder if the small molecule transition assignment code is respecting the TransitionInsturment parameters. I would expect these to cause any peaks outside the min to max range (default 50 to 1500) to be ignored entirely. So, I wouldn't expect MS/MS peaks with the m/z values described to have any impact.

bspratt:
I'll double check that - but the problem here was with a hardcoded limit in the molecule ctor Validate() call, before any other rules can be applied.

bspratt:
Yes, it filters on instrument range as expected when adding transitions to the document from the library.

pwiz_tools/Skyline/Model/AbstractModificationMatcher.cs
```diff
@@ -499,7 +499,14 @@ public PeptideDocNode CreateDocNodeFromSettings(LibKey key, Peptide peptide, Srm
             var transitionsUnranked = new List<DocNode>();
             foreach (var peak in spectrum.Peaks)
             {
-                transitionsUnranked.Add(TransitionFromPeakAndAnnotations(key, nodeGroupMatched, fragmentCharge, peak, null));
+                try
```
brendanx67, line 502: This call should be paying attention to settings and the minimum value that causes the exception reported to initiate this fix. For peptide fragment annotation, we definitely consider the settings, and since we do not rank fragments outside the instrument range. This code also strikes me as odd that you wouldn't just create the precursor and then use a precursor.ChangeSettings(Settings, diff ?? SrmSettingsDiff.ALL) to materialize all of the transitions based on the settings. That way you only write the code once to materialize transitions based on settings.

brendanx67 (changes requested):
This may be a usable bandaid for 21.1, but it should have a TODO and issue posted to more thoroughly revise this code. That you end up trying to materialize an invalid transition and need to catch a failure of the constructor seems like an indicator of poor design.

PR #1606, opened 2021-05-24
Include iRT standards in counts on final wizard dialog

kaipot:


brendanx67 (approved):

chambm:
The test failures suggest that the change works as expected. Can you update the expected counts in the tests and at the same time verify that the status bar shows the same numbers as the updated expected counts?

kaipot:
> The test failures suggest that the change works as expected. Can you update the expected counts in the tests and at the same time verify that the status bar shows the same numbers as the updated expected counts?

I did some quick tests of it - can you confirm that it works for the case you were seeing as well? I'll merge after that.

chambm:
Yes, status bar numbers are consistent with counts dialog with CiRT.

chambm:
Ack, wait, the CiRT auto-add doesn't work now. I get the `The calculator DIA-PASEF-tutorial requires all 15 of its standard peptides to be in the targets list in order to determine a regression.` error again.

chambm:
You should be able to change one of the old (QE/TTOF) tests to CiRT to repro as before.

kaipot:
> You should be able to change one of the old (QE/TTOF) tests to CiRT to repro as before.

Can you share a diff? I'm not able to reproduce it with either CiRT or automatic on the DiaTtofTutorial.

chambm:
OK so it happens with the small dataset but not the big one (importing only `target_protein_sequences.fasta` vs `napedro_3mixed_human_yeast_ecoli_20140403_iRT_reverse.fasta` . So I guess the CiRT peptides that get picked for the small dataset are not in `target_protein_sequences.fasta`, but they are in the full FASTA. So the iRT group gets added in the former case but not the latter. But for some reason later on in the import process it thinks the CiRT peptides are NOT in the document. :/ Does this make sense to you? 

I can't give you a repro yet because I'm still uploading the DIA-PASEF tutorial zip, but I would expect it to also happen on the DiaTtofFullSearchTutorial (and not on DiaTtofTutorial).

chambm:
OK yes it does repro on DiaTtofFullSearchTutorial (after setting IrtStandard to CIRT_SHORT).

kaipot:
Yeah, I can reproduce it with the other dataset, looking into it now.

kaipot:
It turned out that if we add the iRT standards before showing the dialog, some of the refinement stuff the dialog does can end up removing them (and of course if we add them after, the counts shown on the dialog don't include them). So we have to be a bit careful about the order things happen, should be fixed now.

chambm:
Note that TeamCity only runs the Ttof version. The other 3 versions of that test are going to need updating too. I'm running them now.

brendanx67:
It is a big time commitment, but when making changes that could impact tutorial log files, you really need to run everything under Tutorials and all test in Perf that end with "Tutorial" in 3 languages (English, Chinese, Japanese).

chambm:
So far this change doesn't affect audit logs; other than localization issues, wouldn't only an audit-log-matching failure in one language indicate the need to run and update all the other languages? It occurs to me now that perhaps that's not right in this case. Should adding iRT peptides independent of the FASTA result in an audit log entry? I can see it either way (because it's part of the wizard so as long as the wizard execution is audited the iRT adding can be considered implied).

brendanx67:
I guess maybe not. This change in ordering was intended to fix what the UI says the document will contain. What actually happened to the document did not change. I probably just have audit logs on the brain from doing the translation updates. Carry on.

chambm:
Looks like only the non-full-search tutorials changed counts.

For TestDiaQeDiaUmpireTutorial:
                TargetCounts = new[] { 14, 185, 215, 1289 },
                FinalTargetCounts = new[] { 10, 185, 215, 1289 },

For TestDiaTtofDiaUmpireTutorial:
                TargetCounts = new[] { 14, 213, 277, 1661 },
                FinalTargetCounts = new[] { 10, 213, 277, 1661 },

kaipot:
Thanks Matt, I updated the tutorials with those numbers. Shall I go ahead and merge this then?

chambm:
I just force-ran the perf tests. If all tests pass, it's ready.

chambm (approved):

PR #1614, opened 2021-05-26
Improvements for Waters SONAR Handling:

bspratt:
new Waters SDK
SONAR filtering is now based on m/z extraction tolerance, rather than pretending to be IMS
Full Scan graph is SONAR aware, 3D view is in terms of quadrupole range instead of drift time

Also:
Skyline wasn't able to access 3-array ion mobility data in mzML format


pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.cpp
```diff
@@ -79,13 +80,13 @@ SpectrumList_IonMobility::SpectrumList_IonMobility(const msdata::SpectrumListPtr
...
         units_ = IonMobilityUnits::drift_time_msec;
     }
-    else // reading an mzML conversion?
+    else // reading an mzML conversion
     {
         if (inner->size() == 0)
             return;
 
         // See if first scan has any ion mobility data
-        SpectrumPtr spectrum = inner->spectrum(0, false);
```
chambm, line 88: Isn't it pretty common for the first Waters scan to be lockmass? I could easily see that being non-IMS.

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.cpp
```diff
@@ -98,6 +99,14 @@ SpectrumList_IonMobility::SpectrumList_IonMobility(const msdata::SpectrumListPtr
                 units_ = IonMobilityUnits::compensation_V;
             else if (!scan.userParam("drift time").empty()) // Oldest known mzML drift time style
                 units_ = IonMobilityUnits::drift_time_msec;
+            for (int b = 0; b < spectrum->binaryDataArrayPtrs.size(); b++)
+            {
+                if (spectrum->binaryDataArrayPtrs[b]->hasCVParam(CVID::MS_raw_ion_mobility_array))
```
chambm, line 104: I did this a little more robustly in the MBI PR: https://github.com/ProteoWizard/pwiz/pull/1484/files#diff-03d7cf0af6b548cf602001d8c4932f32ab89c6be17131f85efdaa32738f1dc35

I think you can copy that bit here. You'll get the unit as well, although another else branch is needed to support the new SONAR scheme (not using ion mobility cvParam). For some reason I didn't do the change to hasCombinedIonMobility, was that one of the reasons it was acting up in Skyline?

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.cpp
```diff
@@ -170,15 +182,31 @@ PWIZ_API_DECL double SpectrumList_IonMobility::ccsToIonMobility(double ccs, doub
...
 
-PWIZ_API_DECL std::pair<int, int> SpectrumList_IonMobility::sonarMzToDriftBinRange(int function, float precursorMz, float precursorTolerance) const
+///
+/// Deal with Waters SONAR equipment, which filters an m/z range using its ion mobility hardware and reports the data as if it were ion mobility
+///
+PWIZ_API_DECL bool SpectrumList_IonMobility::isWatersSonarData() const
 {
-    if (equipment_ != IonMobilityEquipment::WatersSonar)
-        throw runtime_error("SpectrumList_IonMobility::sonarMzToDriftBinRange] function only works on Waters SONAR data");
+    return equipment_ == IonMobilityEquipment::WatersSonar;
```
chambm, line 190: Definitely need the extra else branch above to set the equipment_ parameter for the new SONAR scheme, or else SONAR mzMLs certainly won't work with this function.

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.hpp
```diff
@@ -41,8 +41,8 @@ class PWIZ_API_DECL SpectrumList_IonMobility : public msdata::SpectrumListWrappe
     static bool accept(const msdata::SpectrumListPtr& inner);
     virtual msdata::SpectrumPtr spectrum(size_t index, bool getBinaryData = false) const;
 
-    // N.B this should agree with the enum IONMOBILITY_TYPE in pwiz_tools\BiblioSpec\src\BlibUtils.h
-    enum class IonMobilityUnits { none, drift_time_msec, inverse_reduced_ion_mobility_Vsec_per_cm2, compensation_V };
+    // N.B this order starting with none=0 should agree with the enum IONMOBILITY_TYPE in pwiz_tools\BiblioSpec\src\BlibUtils.h
+    enum class IonMobilityUnits {waters_sonar = -1, none, drift_time_msec, inverse_reduced_ion_mobility_Vsec_per_cm2, compensation_V };
```
chambm, line 45: Not a big deal, but is a waters_sonar entry really needed here? If equipment is SONAR, then the reader doesn't need to check units since it will know it needs to handle that special SONAR array scheme...

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.cpp
```diff
@@ -134,7 +143,10 @@ PWIZ_API_DECL bool SpectrumList_IonMobility::canConvertIonMobilityAndCCS(IonMobi
 PWIZ_API_DECL bool SpectrumList_IonMobility::hasCombinedIonMobility() const
 {
     if (sl_ == nullptr)
-        return false;
+    {
+        // Not a native reader, but mzML may have 3-array representation
+        return has_mzML_combined_ion_mobility_;
```
chambm, line 148: Can't we just check equipment/units here instead of having an extra bool?

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.cpp
bspratt: Might be better to check the first several, yes.

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
```diff
@@ -467,10 +471,18 @@ PWIZ_API_DECL void SpectrumList_Waters::getCombinedSpectrumData(int function, in
     auto mzItr = &mz[0], intensityItr = &intensity[0], driftTimeItr = &driftTime[0];
     for (int scan = 0; scan < numScansInBlock; ++scan)
     {
-        double dt = rawdata_->GetDriftTime(function, scan);
+        double dt;
+        if (config_.reportSonarBins)
```
chambm, line 475: I'm not sure about this new reader config parameter. Shouldn't we just change the SONAR representation to the new recommend mzML representation (2 arrays for lower and upper quad m/z instead of 1 array for drift time)? Then you shouldn't need to make thousands of CLI calls to sonarBinToMzRange, right?

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.cpp
bspratt: I was trying to evaluate the effect of SONAR filtering by comparing it to mzML from a current msconvert, which produces a drift time representation. Loading that into Skyline with no IMS filtering enabled was the plan, but Skyline wouldn't read it at all because the pwiz layer was only recognizing 3-array data in native formats.

I'm not yet worrying about proper SONAR representation in mzML, initial focus is on raw file handling so I can get something to Waters. But yeah, that will need to be addressed.

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.cpp
bspratt: Agreed. As I say, not yet worrying about that for mzML files.

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.hpp
bspratt: What it really points out is that "IonMobilityUnits" is really a misnomer - should be "AdditionalSeperationType" or somesuch.

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.cpp
bspratt: Huh, in the MBI PR code is a "MS_raw_ion_mobility_array" data array really a marker for drift time, specifically?

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.cpp
bspratt: I agree it feels clunky, but what combination would that be? Same equipment and units could have either representation in an mzML doc.

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.cpp
```diff
@@ -98,6 +99,14 @@ SpectrumList_IonMobility::SpectrumList_IonMobility(const msdata::SpectrumListPtr
                 units_ = IonMobilityUnits::compensation_V;
             else if (!scan.userParam("drift time").empty()) // Oldest known mzML drift time style
                 units_ = IonMobilityUnits::drift_time_msec;
+            for (int b = 0; b < spectrum->binaryDataArrayPtrs.size(); b++)
```
chambm, line 102: If you don't need b as a counter/output variable, you can do range-based for (the modern C++ equivalent of C#'s foreach):
`for (const auto& binaryArrayPtr : spectrum->binaryDataArrayPtrs)`

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
bspratt: From Skyline's point of view, it's much nicer to say "I've determined and cached the integer bin range of interest for each precursor, so just give me those bin numbers rather than making me deal with two values per bin."

Also those upper/lower quad m/z values are necessarily going to be fictions - there's no hard cutoff. The current implementation seems more, what, truthful.

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.cpp
bspratt: Nice, will do.

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.cpp
chambm: Yes, raw ion mobility array is only drift time.

pwiz/analysis/spectrum_processing/SpectrumList_IonMobility.cpp
chambm: You're right, nevermind.

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
chambm: So you're making one sonarBinToMzRange call per precursor? How long does it take for 100k precursors?

bspratt: That's right, and using that bin number with Skyline's existing "ion mobility" filtering capability for chromatogram extraction, where bin number is the filtered value rather than ion mobility. Minimal code changes required on the Skyline side.

Hadn't thought to time that call specifically - why do you ask?

chambm: Just seems like a lot of ferrying back and forth through the CLI layer. But if it's not a noticeably long time then I expect you're right and using bin number directly makes sense.

bspratt: The alternative is allocating, copying, and passing 1/3 again as much data per each of bazillions of measurements (mz,intensity,quadLow,quadHigh vs mz/intensity/bin) and checking twice as many values in the yet to be written filtering code. That just doesn't feel like a win, But I can't back that up with data, no.

bspratt: Yeah, but it turns out we've never actually needed to handle the 3 array IMS representation in mzML so I'm not too worried about 4-array SONAR in mzML.

chambm: That's sound logic for 3 array IMS: I don't see that anybody's complained about it and it's been around for some time. But given Hans' long-standing interest in proper SONAR representation in mzML, should we run this decision by Hans? If he says that eventually he expects Skyline to support 4-array SONAR mzML, that may have implications for design choices now (e.g. do we want Skyline to have 2 filtering schemes, or do we want to have some kind of translation function in SpectrumList_IonMobility so Skyline can always treat it as bin numbers).

bspratt: If it really became an issue I'd probably just write some code to inspect the file and reverse engineer the bin setup. Sure, it's a performance hit, but mzML vs raw is already a performance hit so I'm not too bothered. But I think the dynamic is going to be the same with these large files as with other large files - folks will generally just avoid conversion to mzML when Skyline can read the original.

bspratt:
Here are some notes I put together for Waters, includes screenshots of revised heatmap for SONAR
[SONAR in Skyline.pptx](https://github.com/ProteoWizard/pwiz/files/6657616/SONAR.in.Skyline.pptx)

brendanx67:
Looks cool. It would be nice to see an MS/MS 3D spectrum on slide 4. It would be nice to see the full extracted transition chromatograms for both precursors and products (probably split graph) for the peptide IAGDY... Maybe before the slide showing "no chromatograms". Though, no chromatograms should have blank space to the left of the targets, not red dots. The red dots indicate chromatograms with no signal, not a lack of chromatograms.

bspratt:
Useful comments, thanks - made me realize that I'm not happy with the default zoom on the MS/MS (high energy mSE, really) scans. Not a SONAR issue per se but worth looking at.

And, I'm unclear on the difference between "no signal" and "no chromatogram".

bspratt:
image

bspratt:
OK, now its:
image

brendanx67:
Looks good. You appear to have captured the important difference between "unmeasured" and "measured but absent"

PR #1618, opened 2021-05-27
Add diaPASEF tutorial

chambm:
* added draft diaPASEF tutorial document
* added diaPASEF tutorial tests (TestDiaPasefTutorial and TestDiaPasefFullSearchTutorial); the data is not uploaded yet
* improved error message "The calculator {0} requires all of its standard peptides in order to determine a regression." to include count and list of missing peptides
* changed GraphChromatogram.TestFullScanSelection to return a string describing error instead of simple false (this only seems to be used in AbstractFunctionalTestEx.CheckFullScanSelection as an assert)
* moved ParseIrtProperties() to TestFunctional


chambm:
@bspratt  @brendanx67 Error is due to a `PauseTest("Click on and record appropriate ChromatogramClickPoint")` guarded by a conditional on IsRecordMode. I left it in because it's useful to pause the test in order to record a new chromatogram point. Should I just add a simple alias for PauseTest that doesn't trip the check? Or is that too hacky and abusable? I'm not sure how easy a "disable once" kind of thing would be for the CodeInspection test. Or maybe there's an easy way to just pick an appropriate point programmatically?

nickshulman:
@chambm I don't think there should be any "PauseTest" call in there, even if "IsRecordMode" is on.
Instead, the coordinates where the click happens should be hard coded in each test, and not be part of what gets written to the console when IsRecordMode is true.
In general, recording a new set of expected values should not require any human interaction or judgement, since the person who needs to update these values is not going to know where to click.

If you really wanted a PauseTest, you could call "PauseAndContinueForm.Show", but I think you really don't want to do that here.

By the way, when I try to run TestDiaPasefTutorial on my computer I get an error that it can't download http://skyline.ms/tutorials/DIA-PASEF-small.zip

chambm:
> I don't think there should be any "PauseTest" call in there, even if "IsRecordMode" is on.
Instead, the coordinates where the click happens should be hard coded in each test, and not be part of what gets written to the console when IsRecordMode is true.

The hard-coded coordinates are tested when IsRecordMode==false. I can't hard-code coordinates when I don't know the coordinates in advance, which is the case when adding a new variant of this tutorial, or even changing the peptide used as an example. It makes no sense to me to test hard-coded coordinates for one assertion but then print many other expected values to console; if any new expected values are printed to console, almost all of them should be. However, it would make sense to only ask to record new coordinates if clicking the existing coordinates doesn't work (throws an exception).

I agree that instead of using a PauseTest, it would be better to encode the criteria (rather than the coordinates) for which point to click into the test, and try to get the coordinates automatically. But I didn't know how to implement that quickly and figured it wasn't worth the time. Likewise, having my IsRecordMode run fail because the hard-coded coordinates were wrong was also a waste of time. So pausing to tell the runner to record a new coordinate set made the most sense. But on review, the pause message should probably say "click a point near the peak of the left-side chromatogram pane", because "appropriate" does assume knowledge about how the other points were picked.

> In general, recording a new set of expected values should not require any human interaction or judgement, since the person who needs to update these values is not going to know where to click.

I would agree with this by adding "whenever practical" after "in general". :) The improved click message I mentioned above avoids the problem with the human not knowing where to click, right? It's basically just telling them to do something I didn't know how to automate and it's such a low frequency thing that it wasn't worth figuring out.

bspratt:
If you do decide that leaving it there is proper, add this comment to the
line with the PauseTest() on it
 // Purposely using MessageBox here
so our custom code inspection test doesn't flag it.

I've just revised the code inspection error message to explain how
this works.

chambm:
> If you do decide that leaving it there is proper, add this comment to the line with the PauseTest() on it // Purposely using MessageBox here so our custom code inspection test doesn't flag it. I've just revised the code inspection error message to explain how this works.

Thanks, that's good to know! I decided to go with Nick's suggestion of using `PauseAndContinueForm.Show`, but I also improved the message and only pause if the chromatogram point is wrong (and IsRecordMode==true of course).

chambm:
@nickshulman DIA-PASEF-small.zip is up on skyline.ms now

PR #1619, opened 2021-05-28
Export spectral library improvements (use best instead of first; z-score; move UI code out of model)

kaipot:


brendanx67:
Need to fix tests.

nickshulman (approved):

PR #1620, opened 2021-05-28
Fix exception that occurs when irtdb file does not exist

kaipot:


nickshulman:
Usually the only way to fix an exception like this is to catch it when it happens.
If you are getting an exception when the file does not exist, then there will always be some other sort of exception that can happen if the file happens to be corrupted, or if you don't have permissions to read the file.
I am not familiar with the exception that you are trying to fix here? Would it be possible to add a try/catch around the spot where the exception is happening?

pwiz_tools/Skyline/Skyline.cs
```diff
@@ -76,6 +76,7 @@
 using pwiz.Skyline.ToolsUI;
 using pwiz.Skyline.Util;
 using pwiz.Skyline.Util.Extensions;
+using DatabaseOpeningException = pwiz.Skyline.Model.Irt.DatabaseOpeningException;
```
brendanx67, line 79: This seems like an oversight. I don't see that you are actually using this exception type.

kaipot:
> Usually the only way to fix an exception like this is to catch it when it happens.
> If you are getting an exception when the file does not exist, then there will always be some other sort of exception that can happen if the file happens to be corrupted, or if you don't have permissions to read the file.
> I am not familiar with the exception that you are trying to fix here? Would it be possible to add a try/catch around the spot where the exception is happening?

You're right, it's better to catch other exeptions like the database being corrupted. This is an exception that happens when closing peptide settings, it tries to check if it needs to add iRT standards to the document. I think it's better in this case to just ignore rather than show an error.

pwiz_tools/Skyline/Skyline.cs
```diff
@@ -2175,9 +2176,17 @@ private void ShowPeptideSettingsUI(IWin32Window parent, PeptideSettingsUI.TABS?
...
             var calc = RCalcIrt.Calculator(Document);
-            if (calc == null || !File.Exists(calc.DatabasePath))
+            if (calc == null)
                 return;
-            calc = calc.Initialize(null) as RCalcIrt;
+            try
+            {
+                calc = calc.Initialize(null) as RCalcIrt;
+            }
+            catch (Exception)
```
brendanx67, line 2185: Should we really be returning silently? I would much prefer that the user see a message about what is going on that may allow them to fix the problem, (e.g. file missing, access restriction, database corruption, etc.)

We don't want the unhandled exception form, but we do want to show them some information. Code like this doesn't necessarily reduce support. It may cause more confusion, and because the exception may be system dependent, it may be very challenging for us to debug by asking for file to be sent to us.

brendanx67 (changes requested):
I would like to see a demo of what happens when this failure occurs and a test that covers it. We should be showing the user an error message. The test could pretty easily cover 2 cases: 1) the file is removed, 2) the file is corrupted (just a bogus text file that is not an irtdb).

brendanx67 (approved):

PR #1628, opened 2021-06-04
Fix order by m/z method export bug when multiple precursors have the same m/z

kaipot:


brendanx67 (changes requested):
This needs a test to prove it fixes the case presented in the bug report.

brendanx67 (approved):

PR #1629, opened 2021-06-04
Issue 790: Edit > Insert > Peptides lacks the ability to add specific precursors

nickshulman:
This change allows users to put charge indicators (e.g. "+++") at the end of their peptide sequences in the Edit Insert Peptides dialog.

image


pwiz_tools/Skyline/EditUI/PasteDlg.cs
```diff
@@ -444,57 +459,48 @@ private SrmDocument AddPeptides(SrmDocument document, bool validating, ref Ident
             return document;
         }
 
-        private List<string> ListPeptideSequences()
+        /// <summary>
+        /// Returns a list of tuples with the peptide sequences and optional charges.
+        /// The returned list may contain nulls, and will have exactly the same number of rows as gridViewPeptides.
+        /// </summary>
+        private List<Tuple<string, Adduct>> ListPeptideSequences()
```
brendanx67, line 466: I really hate Tuple, especially as a return value like this. It makes the calling code with Item1 and Item2 require a look at the function returning the Tuple to know what these are. I much prefer defining a struct with named properties.

pwiz_tools/Skyline/EditUI/PasteDlg.cs
```diff
@@ -407,8 +414,16 @@ private SrmDocument AddPeptides(SrmDocument document, bool validating, ref Ident
...
-                                                                                                  SrmSettingsDiff.ALL);
+                {
+                    nodePepNew = matcher.GetModifiedNode(pepModSequence, fastaSequence);
+                    var settings = document.Settings;
+                    if (!adduct.IsEmpty)
+                    {
+                        settings = settings.ChangeTransitionFilter(f =>
+                            f.ChangePeptidePrecursorCharges(new[] {adduct}));
+                    }
+                    nodePepNew = nodePepNew.ChangeSettings(settings, SrmSettingsDiff.ALL);
```
brendanx67, line 425: It seems like this would give you a peptide node with auto-select set to true, even when its precursor does not match the true settings. I think you may need to check that the document settings are not already just the adduct (e.g. peptide is charge 2 and the settings are only charge 2). Then if that is not the case, you need to change the settings and remember to set auto-select to false on the peptide.

pwiz_tools/Skyline/EditUI/PasteDlg.cs
```diff
@@ -444,57 +459,48 @@ private SrmDocument AddPeptides(SrmDocument document, bool validating, ref Ident
...
-        private List<string> ListPeptideSequences()
+        /// <summary>
+        /// Returns a list of tuples with the peptide sequences and optional charges.
+        /// The returned list may contain nulls, and will have exactly the same number of rows as gridViewPeptides.
+        /// </summary>
+        private List<Tuple<string, Adduct>> ListPeptideSequences()
         {
-            List<string> listSequences = new List<string>();
-            for (int i = gridViewPeptides.Rows.Count - 1; i >= 0; i--)
+            List<Tuple<string, Adduct>> listSequences = new List<Tuple<string, Adduct>>();
```
brendanx67, line 468: Really prefer var over stating the type twice.

pwiz_tools/Skyline/EditUI/PasteDlg.cs
```diff
@@ -444,57 +459,48 @@ private SrmDocument AddPeptides(SrmDocument document, bool validating, ref Ident
...
                         Column = colPeptideSequence.Index,
                         Line = i,
-                        Message = Resources.PasteDlg_ListPeptideSequences_The_structure_of_this_crosslinked_peptide_is_not_supported_by_Skyline
+                        Message = Resources.PasteDlg_ListPeptideSequences_The_peptide_sequence_cannot_be_blank
                     });
                     return null;
                 }
-
-                listSequences.Add(peptideSequence);
+                listSequences.Add(Tuple.Create(peptideSequence, adduct));
```
brendanx67, line 503: Why not just use sequenceAdductTuple? I don't see that the values get changed anywhere. If you used a struct instead of a Tuple, the code would likely read clearly enough without needing to create the named variables.

pwiz_tools/Skyline/TestFunctional/PastePeptidesWithChargesTest.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            RunFunctionalTest();
+        }
+
+        protected override void DoTest()
+        {
+            // List of peptide sequences, charges, and protein names to paste in the Paste Peptides grid
+            var peptideChargeProteins = new[]
+            {
+                Tuple.Create("GVDC[+57]QEVSQEK", 2, "sp|Q61687|ATRX_MOUSE"),
```
brendanx67, line 47: Too much repetition of cases. You should add at least one case that will produce more complicated charge indicators.

PEPTIDER, +4
PEPTIDER+++++

Both of these should work, but they are not covered by your test.

pwiz_tools/Skyline/TestFunctional/PastePeptidesWithChargesTest.cs
```diff
@@ -0,0 +1,105 @@
...
+            foreach (var tuple in peptideChargeProteins)
+            {
+                var unmodifiedSequence = FastaSequence.StripModifications(tuple.Item1);
+                var peptideDocNodes =
+                    SkylineWindow.Document.Molecules.Where(pep => pep.Peptide.Sequence == unmodifiedSequence).ToList();
+                Assert.AreEqual(1, peptideDocNodes.Count);
+                var peptideDocNode = peptideDocNodes[0];
+                Assert.AreEqual(1, peptideDocNode.TransitionGroupCount);
+                var precursor = peptideDocNode.TransitionGroups.First();
+                Assert.AreEqual(tuple.Item2, precursor.PrecursorCharge);
```
brendanx67, line 101: Add a check that charge 2 precursors are auto-select (since that is the default setting), and non-charge 2 precursors are not auto-select. You might also repeat the test with the document settings changed to only charge 3 precursors and check that produces the right auto-select settings, and also with multiple charges (e.g. "2, 3, 4") and check that all resulting peptides are not auto-select.

brendanx67 (changes requested):
Nice start. I like that the crosslink handling moves into a separate containing function, which makes the AddPeptides() function, which is already a bit long, easier to read.

Had some suggestions and requested changes. The test needs to be improved.

brendanx67:
This got me wondering if we really should do it this way, despite the fact that I may have suggested it, and it is the way that a direct paste of a peptide sequence works. It seems like Edit > Insert > Peptides seems to be trying to use a grid to take all the syntactical magic out of pasting a list of peptides into Skyline. So, why not maintain that in this case? It seems more consistent to add radio buttons below the grid for "Peptides" and "Precursors", the way we have "Peptides" and "Small molecules" in Edit > Insert > Transition List.

For Edit > Insert > Peptides, the "Peptides" choice would present a grid like the one today, and the "Precursors" choice would add a "Charge" or "Precursor Charge" column after the "Peptide Sequence" column.

pwiz_tools/Skyline/TestFunctional/PastePeptidesWithChargesTest.cs
nickshulman: Is there any other place in Skyline where the user can type in "PEPTIDER, +4" and have it work?
There are places where Skyline displays charges that way, but I cannot find any code in Skyline that knows how to parse that.

nickshulman:
Thanks for the feedback, Brendan,
I still have more work to do on this, which is why it's staying in "Draft" state. For one thing, I know I need to make sure that Skyline does something reasonable if the user inserts the same peptide more than once with different charges. I need to make sure that Skyline gives them only one peptide with the union of all of the charges that they have specified in their insert list.

> I really hate Tuple, especially as a return value like this. It makes the calling code with Item1 and Item2 require a look at the function returning the Tuple to know what these are. I much prefer defining a struct with named properties.

I have now added a new class called "ChargedSequence" which is a peptide sequence plus an adduct. I was hoping that such a class already existed, but I could not find any. I did find an unused class called "PeptideCharge" in RefinementSettings.cs, and I have removed that unused class.
I was kind of hoping that I could use the class "LibraryKey" instead of having to define the new class "ChargedSequence", but I really needed the original string to eventually pass to ModificationMatcher, so defining a new class seemed the better way to go here.

> It seems like this would give you a peptide node with auto-select set to true, even when its precursor does not match the true settings. I think you may need to check that the document settings are not already just the adduct (e.g. peptide is charge 2 and the settings are only charge 2). Then if that is not the case, you need to change the settings and remember to set auto-select to false on the peptide.

I have now made it so that the if a charge has been specified the peptide gets created with "AutoManageChildren" false. I am thinking that it would not be good to have AutoManageChildren depend on whether the selected charge happens to match the current Precursor Charges settings. No strong opinion either way.

> It seems more consistent to add radio buttons below the grid for "Peptides" and "Precursors", the way we have "Peptides" and "Small molecules" in Edit > Insert > Transition List.

That sounds like a lot of work.

PR #1633, opened 2021-06-07
fix a problem with Skyline processing Waters IMS data where the lockmass functions do not have IMS data

bspratt:
The problem comes down to two different abbreviation schemes within the same data file, as the lockmass channel does not have IMS data. This prevented Skyline from identifying the lock mass channel, and thus from ignoring it.


bspratt:
I don't think this will bother any other pwiz consumer?

chambm (changes requested):
Sorry but I don't think this is a correct solution. The non-combined form of spectra should always, always conform to a valid nativeID format. That's `function=x process=x scan=x` for Waters. For combined spectra, which are really merged versions of individual spectra, the nativeID should start with `merged=<index>`, and the rest of the id is not so rigid (because clearly there's no way to get back to a native spectrum solely from the id).

What about just filtering out the (assumed) lockmass function with a SpectrumList_Filter so Skyline doesn't even see it?

bspratt:
>> What about just filtering out the (assumed) lockmass function with a SpectrumList_Filter so Skyline doesn't even see it?

Good idea, only problem is determining which function that is. Skyline is messing that up because the abbreviation code assumes that the file is all combined mode, or not at all.

bspratt:
OK, now properly dealing with the fact that we can't depend on all scans in a file to use the same id encoding scheme - in particular with 3-array IMS representation in Waters files where the lockmass function has no IM dimension at all.

pwiz/data/msdata/MSData.cpp
```diff
@@ -1091,21 +1091,27 @@ PWIZ_API_DECL size_t SpectrumList::findAbbreviated(const string& abbreviatedId,
     if (empty()) return size();
 
     // "sample=1 period=1 cycle=123 experiment=2" splits to { sample, 1, period, 1, cycle, 123, experiment, 2 }
-    string firstId = spectrumIdentity(0).id;
-    bal::split(actualTokens, firstId, bal::is_any_of(" ="));
-
-    if (actualTokens.size() != abbreviatedTokens.size() * 2)
+    for (size_t s = size(); s-- > 0;) // Some files contain mixed scan types - e.g. Waters where lockmass scans may be non-IMS and others IMS, so we may need to search a bit to find a format match
```
chambm, line 1094: Why search backward? Did you check that this iteration doesn't have any significant performance impact? Is there a non-negligible chance that Skyline will look for a format that won't match any spectrum and thus will iterate the entire spectrumIdentity list and still not get a result?

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -594,7 +594,7 @@ protected virtual SpectrumList SpectrumList
                             {
                                 if (GetMsLevel(spectrum) == 1)
                                 {
-                                    var function = MsDataSpectrum.WatersFunctionNumberFromId(id.abbreviate(spectrum.id), HasCombinedIonMobilitySpectra);
+                                    var function = MsDataSpectrum.WatersFunctionNumberFromId(id.abbreviate(spectrum.id), spectrum.id.Contains(MERGED_TAG));
```
chambm, line 597: Hmm, maybe `HasCombinedIonMobilitySpectra && spectrum.id.Contains(MERGED_TAG)`. Because a lot of merged spectra won't be 3-array IMS spectra.

pwiz/data/msdata/MSData.cpp
bspratt: Backward only because the lockmass stuff tends to be the first thing in the file. There is a chance that yes, this will fail, and expensively, but presumably only once before Skyline throws its hands up. Performance would be less of a concern if the search used a dictionary instead of a list walk.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
bspratt: Makes sense, done

chambm (changes requested):
Should we have a test for this case? You could just remove the IMS array from some other Waters test and check that it still processes ok. I just checked that you can just delete the .cdt file from the lockmass function (usually func003) and pwiz will treat it as non-IMS.

bspratt:
Good plan, I'll do that.

pwiz_tools/Skyline/TestPerf/PerfImportResultsWatersIMSTest.cs
```diff
@@ -47,7 +47,7 @@ public class ImportWatersIMSTest : AbstractFunctionalTest
         public void WatersIMSImportTest()
         {
             Log.AddMemoryAppender();
-            TestFilesZip = GetPerfTestDataURL(@"PerfImportResultsWatersIMS.zip");
+            TestFilesZip = GetPerfTestDataURL(@"PerfImportResultsWatersIMSv2.zip"); // v2 has _func003.cdt file removed, to test our former assumption that lockmass would have IMS data if other functions did and vice verse
```
chambm, line 50: Does this leave other tests still testing that the lockmass-with-IMS test still works?

bspratt: Hmm, good point. I could set it to use one data set or the other depending on day-of-year.

chambm: You can delete the new dataset and instead (conditional on day of year) rename the func003.cdt file before the test and restore it afterward. Might have to handle restoring the file at the beginning if a previous run failed to restore it though.

bspratt: I think this is just easier. If that file somehow didn't get restored it would screw up any machine also running the release branch tests.

PR #1636, opened 2021-06-08
fix a problem with an interrupted ion mobility background loader not finishing the job

bspratt:


pwiz_tools/Skyline/Model/IonMobility/IonMobilityLibraryManager.cs
```diff
@@ -50,7 +50,12 @@ public override void ClearCache()
 
         protected override bool StateChanged(SrmDocument document, SrmDocument previous)
         {
-            return !ReferenceEquals(GetIonMobilityLibrary(document), GetIonMobilityLibrary(previous));
+            var libPrevious = GetIonMobilityLibrary(previous);
```
brendanx67, line 53: Hmmm... This kind of just says that any document change will restart your load. What if there is no subsequent document change? This seems to indicate a design flaw if this is necessary. Let's talk through the issue and how it is caused.

bspratt: Perhaps the logic change really belongs in ChromatogramManger.IsReadyToLoad().

bspratt: OK, in conversation you made the point that if the IM library class is properly immutable, then my change should not be necessary. And, it is properly immutable - successfully loading the db into memory does cause a document change. I think the issue I was hitting (under a boggy debug build) was that the chromatogram extractor wasn't being told to wait for the imsDB to load into memory, the way it's told to wait for spectral libs to load.

brendanx67 (approved):

pwiz_tools/Skyline/Model/IonMobility/IonMobilityLibraryManager.cs
brendanx67: That does look better to me.

PR #1645, opened 2021-06-17
Fix a problem where fullscan graph is displaying fragment details for the currently selected transition, but the user then uses the Target tree to remove that node from the document.

bspratt:
I haven't actually been able to repro, but a user has been helpful in reporting and provides this explanation:

"The error occurs after various 'selection' operations: in this case it was de-selecting a peptide. I attached a screenshot and the error message. The error doesn’t occur directly after changing a selection but rather on the next click that would navigate to another peptide in the browser on the left.  Usually it is possible to continue working in the project."

Skyline Bug 210616 Marc Sylvester Screenshot GUI

Skyline version: 21.1.0.146-9b577b3b8 (64-bit)
Installation ID: 4ea34116-1060-40d2-97ba-f947ffa1b294
Exception type: NullReferenceException
Error message: Der Objektverweis wurde nicht auf eine Objektinstanz festgelegt.

--------------------

System.NullReferenceException: Der Objektverweis wurde nicht auf eine Objektinstanz festgelegt.
   bei pwiz.Skyline.Controls.Graphs.GraphFullScan.CreateSingleScan() in C:\proj\skyline_21_1_x64\pwiz_tools\Skyline\Controls\Graphs\GraphFullScan.cs:Zeile 644.
   bei pwiz.Skyline.Controls.Graphs.GraphFullScan.CreateGraph() in C:\proj\skyline_21_1_x64\pwiz_tools\Skyline\Controls\Graphs\GraphFullScan.cs:Zeile 300.
   bei pwiz.Skyline.Controls.Graphs.GraphFullScan.UpdateUI(Boolean selectionChanged) in C:\proj\skyline_21_1_x64\pwiz_tools\Skyline\Controls\Graphs\GraphFullScan.cs:Zeile 877.
   bei pwiz.Skyline.SkylineWindow.UpdateGraphPanes(Object sender, EventArgs e) in C:\proj\skyline_21_1_x64\pwiz_tools\Skyline\SkylineGraphs.cs:Zeile 756.
   bei System.Windows.Forms.Timer.OnTick(EventArgs e)
   bei System.Windows.Forms.Timer.TimerNativeWindow.WndProc(Message& m)
   bei System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
Exception caught at: 
   bei System.Windows.Forms.Application.ThreadContext.OnThreadException(Exception t)
   bei System.Windows.Forms.Timer.TimerNativeWindow.OnThreadException(Exception e)
   bei System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
   bei System.Windows.Forms.UnsafeNativeMethods.DispatchMessageW(MSG& msg)
   bei System.Windows.Forms.UnsafeNativeMethods.DispatchMessageW(MSG& msg)
   bei System.Windows.Forms.Application.ComponentManager.System.Windows.Forms.UnsafeNativeMethods.IMsoComponentManager.FPushMessageLoop(IntPtr dwComponentID, Int32 reason, Int32 pvLoopData)
   bei System.Windows.Forms.Application.ThreadContext.RunMessageLoopInner(Int32 reason, ApplicationContext context)
   bei System.Windows.Forms.Application.ThreadContext.RunMessageLoop(Int32 reason, ApplicationContext context)
   bei pwiz.Skyline.Program.Main(String[] args) in C:\proj\skyline_21_1_x64\pwiz_tools\Skyline\Program.cs:Zeile 307.


brendanx67:
I can't reproduce getting the exception window, but I can pretty easily reproduce getting the NRE by showing the FullScanGraph, turning on Rita's new MS/MS spectrum annotation mode, deleting the peptide it is looking at, using the arrow buttons to advance through spectra in the file. Pretty quickly I get a message form:
image

This does not happen when GraphFullScan is not in spectrum annotation mode.

brendanx67 (changes requested):
Can you try the test scenario I explained and improve this change to at least graph the spectrum without any annotation? The current fix removes all peaks from the plot. So, you get just an empty plot. In contrast, when you turn off spectrum annotation, everything works based on the original selection, even though it has been deleted. I don't know that feel this is ideal, but at least it is still showing spectra, and not a blank graph.

bspratt:
OK, I can repro and have made the change requested. I still see annotations though, which is strange. I should probably kick this to Rita.

brendanx67 (approved):
I changed this back closer to what you had but just show an all gray spectrum when the annotation button is pressed and the peptide is not found. The way you had it we just reverted to the way the graph is rendered when the button is not pressed, but that seems strange without also unpressing the button and giving the user an explanatory message.

bspratt:
I note that "unmatched" is left explicitly untranslated - this is considered a corner case, presumably?

bspratt:
I guess that's not text that ever appears in the UI.

I can still get a strange display, though:
image

PR #1647, opened 2021-06-17
Save column locations only when they produce valid transitions

henrytsanford:
Previously the column memory would save and then assign column locations that would produce totally invalid transition lists. For instance, it would save sets of column locations missing essential columns like precursor m/z.  

Now column locations are only saved if there is at least one valid transition on the list.


bspratt (approved):
Looks solid, there's a test for the work, and you've shown me a live demo. Let's make sure the issues database gets updated to note which issues this fixes, or fixes in part.

brendanx67 (commented):
Definitely good not to save columns that were not used to import a transition list successfully. I am curious whether we also discard saved column positions when it can be determined that the columns do not work with a newly pasted transition list. There is a fair amount of intelligence built into picking columns for peptides. It may be infrequent that saved positions will work better than automatic picking.

bspratt:
You're suggesting only invoking the column picker when the original largely-but-not-always successful column identification fails to work? Or invoking the column picker with the results of a successful identification instead of the stored one?

Either way it seems like it would be frustrating if the original code *thought* it had it right but it didn't. Less so if the user has a chance to override, of course.

henrytsanford:
Currently we do not discard saved column positions if they do not work with a newly pasted transition list. Brendan, would you like me to work on changing that before merging this, or should it be added to the issues list?

bspratt (changes requested):
Looks good! Go ahead and merge.

bspratt:
Oh wait, no, some of the TeamCity builds are unhappy. We'll need to sort that out before merging.

henrytsanford:
I was going to add the feature Brendan requested (discarding the saved column positions if they do not work on a newly pasted transition list) before I merge. Until I add that feature the test I recently pushed will fail.

bspratt:
OK. Best not to push a test that you know will fail, though - it attracts unwanted attention.

henrytsanford:
Sorry about that - will keep it in mind for the future

pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
```diff
@@ -162,6 +162,18 @@ private void PopulateComboBoxes()
             // It's not unusual to see lines like "744.8 858.39 10 APR.AGLCQTFVYGGCR.y7.light 105 40" where protein, peptide, and label are all stuck together,
             // so that all three lay claim to a single column. In such cases, prioritize peptide.
             columns.PrioritizePeptideColumn();
+
```
bspratt, line 165: comments are good!

pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
```diff
@@ -174,29 +186,57 @@ private void PopulateComboBoxes()
...
+        {
+            // Save the detected columns so if the saved columns are invalid we can revert back
+            var detectedColumns = new List<Tuple<int, string>>();
+            var numColumns = Importer.RowReader.Lines[0].ParseDsvFields(Importer.Separator).Length;
+            for (int i = 0; i < numColumns; i++)
+            {
+                detectedColumns.Add(new Tuple<int, string>(i, ComboBoxes[i].Text));
+            }
+
+            // Change the column positions to the saved columns so we can check if they produce valid transitions
```
bspratt, line 204: make this a method and use twice

pwiz_tools/Skyline/TestFunctional/PasteTransitionListTest.cs
```diff
@@ -95,26 +79,37 @@ protected override void DoTest()
             RunUI(() =>
             {
                 var peptideBoxes = peptideTransitionList.ComboBoxes;
+                peptideBoxes[14].SelectedIndex = 0;
                 // Changes the contents of a combo box
```
bspratt, line 83: this isn't a very helpful commnent - *why* are we changing it?

pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
```diff
@@ -171,32 +184,67 @@ private void PopulateComboBoxes()
...
-                SetComboBoxText(columns.PrecursorColumn, Resources.ImportTransitionListColumnSelectDlg_PopulateComboBoxes_Precursor_m_z);
-                SetComboBoxText(columns.ProductColumn, Resources.ImportTransitionListColumnSelectDlg_PopulateComboBoxes_Product_m_z);
-                SetComboBoxText(columns.ProteinColumn, Resources.ImportTransitionListColumnSelectDlg_PopulateComboBoxes_Protein_Name);
-                SetComboBoxText(columns.FragmentNameColumn, Resources.ImportTransitionListColumnSelectDlg_PopulateComboBoxes_Fragment_Name);
-                // Commented out for consistency because there is no product charge column
-                // SetComboBoxText(columns.PrecursorChargeColumn, Resources.ImportTransitionListColumnSelectDlg_PopulateComboBoxes_Precursor_Charge);  
+                lines.Add(Importer.RowReader.Lines[i]);
+            }
+            // Try importing that list to check for errors
+            MassListInputs input = new MassListInputs(lines);
```
brendanx67, line 211: Could be done with a single Linc expression.

pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
```diff
@@ -171,32 +184,67 @@ private void PopulateComboBoxes()
...
+
+            var allError = ReferenceEquals(insertionParams.Document, _docCurrent);
+            // If all transitions are errors, reset the columns to the detected columns
+            if (allError)
+            {
+                SetColumnPositions(detectedColumns);
             }
         }
+        // Returns the current column positions as a list of tuples
+        private List<Tuple<int, string>> CurrentColumnPositions()
```
brendanx67, line 226: Think I prefer a struct with field names to a Tuple<> here, especially with the comment below explaining Item1 and Item2. Better to have self-documenting code when possible.

brendanx67 (changes requested):
Suggested some changes. I think it is probably best to go through this on Zoom and clean it up a bit together.

pwiz_tools/Skyline/Properties/Settings.cs
```diff
@@ -539,40 +539,22 @@ public List<string> CustomImportTransitionListHeaders
...
+                    CustomImportTransitionListColumnTypesList = list;
                 }
-                return (int)this[@"CustomImportTransitionListColumnCount"];
+                return (List <string>)this[@"CustomImportTransitionListColumnTypesList"];
             }
             set
             {
-                this[@"CustomImportTransitionListColumnCount"] = value;
+                this[@"CustomImportTransitionListColumnTypesList"] = value;
             }
```
bspratt, line 558: This should really be one string used three times, rather than three different strings, e.g.
var CustomImportTransitionListColumnTypesList = @"CustomImportTransitionListColumnTypesList";
this avoids the risk of one of the three copies having a typo.
BUT the rest of the file looks this way, and it's better to follow local convention. So this is fine, but yuck.

brendanx67 (approved):

pwiz_tools/Skyline/TestFunctional/PasteTransitionListTest.cs
```diff
@@ -87,6 +89,13 @@ protected override void DoTest()
             // Clicking the OK button should save the column locations
             OkDialog(peptideTransitionList, peptideTransitionList.OkDialog);
 
+            // Verify that the correct columns were saved in the settings
+            var expectedColumns = new [] {protName, peptide, precursor, ignoreColumn, decoy, product, ignoreColumn, fragName};
+            // expectedColumns is the same length as targetColumns for comparison
+            var actualColumns = 
+                Settings.Default.CustomImportTransitionListColumnTypesList.GetRange(0, expectedColumns.Length);
```
bspratt, line 96: I don't get why you aren't just taking the whole Settings.Default.CustomImportTransitionListColumnTypesList for comparison?

pwiz_tools/Skyline/TestUtil/AbstractFunctionalTestEx.cs
```diff
@@ -435,6 +435,16 @@ public void ChangeGroupComparison(Control owner, string name, Action<EditGroupCo
             });
         }
 
+        /// <summary>
+        /// Create a new document and wait for it to load
+        /// </summary>
+        /// <param name="forced"></param>
+        public void LoadNewDocument(bool forced)
```
bspratt, line 442: Handy!

pwiz_tools/Skyline/TestFunctional/PasteTransitionListTest.cs
henrytsanford: The last fifteen or so columns of the transition list I am inspecting here are all "ignore column" and although they are also saved in the settings, I thought comparing them might be redundant. Looking at it now though, that could be the simpler solution

bspratt (commented):
ignore this, typed in wrong area

pwiz_tools/Skyline/TestFunctional/PasteTransitionListTest.cs
bspratt: Settings.Default.CustomImportTransitionListColumnTypesList

bspratt: Yeah, I'd think that they should be the same length?

PR #1653, opened 2021-06-22
Team city code inspection fixes.

nickshulman:
Add reference to Google.Protobuf.dll to TestData.csproj to fix new code inspection errors that have shown up on TeamCity.


pwiz_tools/Skyline/ToolsUI/ToolInstallUI.cs
```diff
@@ -109,6 +109,7 @@ public InstallZipToolHelper(Control parent, InstallProgram installProgram)
                 return OverwriteOrInParallel(_parent, toolCollectionName, toolCollectionVersion, reportList, foundVersion, newCollectionName);
             }
 
+            // ReSharper disable once MemberHidesStaticFromOuterClass
```
bspratt, line 112: This is the only one that worries me a bit. It would be good if Brendan signed off on it.

pwiz_tools/Skyline/EditUI/UniquePeptidesDlg.cs
```diff
@@ -364,10 +364,10 @@ private void QueryPeptideProteins(ILongWaitBroker longWaitBroker)
                             var peptideDocNode = tuple.Item2;
                             HashSet<Protein> proteins = new HashSet<Protein>();
                             var peptideGroupDocNode = PeptideGroupDocNodes.First(g => ReferenceEquals(g.PeptideGroup, peptideGroup));
-                            List<Protein> proteinsForSequence;
-                            if (sequenceProteinsDict.TryGetValue(peptideDocNode.Peptide.Target.Sequence, out proteinsForSequence))
+                            // ReSharper disable once ConditionIsAlwaysTrueOrFalse
```
brendanx67, line 367: Should the function above be .FirstOrDefault() instead? Otherwise, why check for null and tell ReSharper to ignore the fact that it can never be null?

pwiz_tools/Skyline/ToolsUI/ToolInstallUI.cs
brendanx67: Could we also fix this by renaming the delegate something like "InstallProgramFunc"? Is that what it is complaining about? I guess I might prefer that.

brendanx67 (changes requested):
Two places I would like to understand better.

pwiz_tools/Skyline/EditUI/UniquePeptidesDlg.cs
bspratt: Makes sense. It was even sillier before, though - checking for null after it had already been used.

bspratt:
I'm reading up on that...
https://www.jetbrains.com/help/resharper/MemberHidesStaticFromOuterClass.html

bspratt:
I think I'll leave this one to you and Nick.

pwiz_tools/Skyline/EditUI/UniquePeptidesDlg.cs
brendanx67: Either way, it was a no-op before and still a no-op. So, either it should just be removed or if we think PeptideGroupDocNodes could be empty, the FirstOrDefault(). I don't claim to know whether the list can be empty or not.

bspratt: Yeah, I went with FirstOrDefault, doesn't hurt to be safety minded. If you and/or Nick can resolve the delegate thing we'll be back to green on TC code inspections.

pwiz_tools/Skyline/Skyline.sln.DotSettings
```diff
@@ -88,6 +88,7 @@
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=SuggestUseVarKeywordEverywhere/@EntryIndexedValue">DO_NOT_SHOW</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=SuggestUseVarKeywordEvident/@EntryIndexedValue">DO_NOT_SHOW</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=TabsAndSpacesMismatch/@EntryIndexedValue">SUGGESTION</s:String>
+	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=TabsAreDisallowed/@EntryIndexedValue">WARNING</s:String>
```
nickshulman, line 91: There are now a lot of warnings about tabs. I don't think you meant to turn "TabsAreDisallowed" into a warning.

pwiz_tools/Skyline/ToolsUI/ToolInstallUI.cs
brendanx67: After reviewing, the necessary churn is not worth the fix. It is not hurting anything. Let's just go with Nick's idea to disable this warning for this one case.

pwiz_tools/Skyline/Skyline.sln.DotSettings
bspratt: I did, actually, because I didn't think we had any, and don't want any, and it doesn't throw off any warnings for me locally. Not critical, of course, OK to remove - but strange...

bspratt: OK, reverted.

PR #1655, opened 2021-06-24
Recent update to TeamCity server now requires TLS 1.3, we were hard c…

bspratt:
…oded for 1.2. The current recommendation from MSFT for future-proofing HTTPS (https://docs.microsoft.com/en-us/dotnet/framework/network-programming/tls) is don't specify TLS levels at all, let the OS decide. But we worry that this will mess up Win7 and Win8 installs, so we continue to specify explicitly.

N.B. SkylineNightlyShim continues to target .NET 4.5.1 - as we are continuing to hard-code TLS we may as well leave the shim as primitive as possible.

### DIFF
--- a/pwiz_tools/Skyline/SkylineNightly/Nightly.cs
+++ b/pwiz_tools/Skyline/SkylineNightly/Nightly.cs
@@ -525,8 +525,11 @@ namespace SkylineNightly
 
         private void DownloadSkylineTester(string skylineTesterZip, RunMode mode)
         {
-            // Make sure we can negotiate with HTTPS servers that demand TLS 1.2 (default in dotNet 4.6, but has to be turned on in 4.5)
-            ServicePointManager.SecurityProtocol |= (SecurityProtocolType.Tls | SecurityProtocolType.Tls11 | SecurityProtocolType.Tls12);
+            // The current recommendation from MSFT for future-proofing HTTPS https://docs.microsoft.com/en-us/dotnet/framework/network-programming/tls
+            // is don't specify TLS levels at all, let the OS decide. But we worry that this will mess up Win7 and Win8 installs, so we continue to specify explicitly
+            var Tls13 = (SecurityProtocolType)12288; // From decompiled SecurityProtocolType - compiler has no definition for some reason
+            ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls | SecurityProtocolType.Tls11 | SecurityProtocolType.Tls12 | Tls13;
+
             using (var client = new WebClient())
             {
                 client.Credentials = new NetworkCredential(TEAM_CITY_USER_NAME, TEAM_CITY_USER_PASSWORD);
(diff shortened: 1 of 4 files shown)


bspratt:
Hopefully I can get Trey to roll back to TLS 1.2 for 48 hours, then the fleet of test machines will be able to grab a revised SkylineNIghtlyShim.exe and move forward from there.

pwiz_tools/Skyline/SkylineNightly/Nightly.cs
```diff
@@ -525,8 +525,11 @@ public void StartLog(RunMode runMode)
 
         private void DownloadSkylineTester(string skylineTesterZip, RunMode mode)
         {
-            // Make sure we can negotiate with HTTPS servers that demand TLS 1.2 (default in dotNet 4.6, but has to be turned on in 4.5)
-            ServicePointManager.SecurityProtocol |= (SecurityProtocolType.Tls | SecurityProtocolType.Tls11 | SecurityProtocolType.Tls12);
+            // The current recommendation from MSFT for future-proofing HTTPS https://docs.microsoft.com/en-us/dotnet/framework/network-programming/tls
+            // is don't specify TLS levels at all, let the OS decide. But we worry that this will mess up Win7 and Win8 installs, so we continue to specify explicitly
+            var Tls13 = (SecurityProtocolType)12288; // From decompiled SecurityProtocolType - compiler has no definition for some reason
```
nickshulman, line 530: It would probably be better to do:
const SecurityProtocolType Tls13 = (SecurityProtocolType)12288

chambm:
How evil would it be to use uint.MaxValue (i.e. all bits turned on)?

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/SkylineNightly/Nightly.cs
bspratt: You're pretty much always right, but why is that better?

(And, any idea why I'm having to do this at all?)

bspratt:
> How evil would it be to use uint.MaxValue (i.e. all bits turned on)?
Pretty sure that would throw:
https://docs.microsoft.com/en-us/dotnet/api/system.net.servicepointmanager.securityprotocol?view=net-5.0#System_Net_ServicePointManager_SecurityProtocol
NotSupportedException
The value specified to set the property is not a valid SecurityProtocolType enumeration value.

pwiz_tools/Skyline/SkylineNightly/Nightly.cs
nickshulman: I was just thinking that "const" is better than "var" because it you end up with one less local variable (although the local variable gets optimized away, so it doesn't matter).

I have no idea why you need to do this. Someone sort of asked a question about this on Stack Overflow but none of the answers address why the symbol is not available.
https://stackoverflow.com/questions/47269609/system-net-securityprotocoltype-tls12-definition-not-found

PR #1657, opened 2021-06-25
Fixes a problem where we were not identifying transition list columns…

bspratt:
… as numeric if the values were in quotes.

Reported by user Whitney Stutts

### DIFF
--- a/pwiz_tools/Skyline/Model/Import.cs
+++ b/pwiz_tools/Skyline/Model/Import.cs
@@ -1969,18 +1969,13 @@ namespace pwiz.Skyline.Model
 
         private static bool TrySplitColumns(string line, char sep, out string[] columns)
         {
-            columns = line.Split(sep);
+            columns = line.ParseDsvFields(sep);
             return columns.Length > 1;
         }
 
         private static Type GetColumnType(string value, IFormatProvider provider)
         {
             double result;
-            var quote = @"""";
-            if (value.StartsWith(quote) && value.EndsWith(quote))
-            {
-                value = value.Substring(1, value.Length - 2);
-            }
             if (double.TryParse(value, NumberStyles.Number, provider, out result))
                 return typeof(double);
             else if (FastaSequence.IsExSequence(value))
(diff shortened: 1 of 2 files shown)


pwiz_tools/Skyline/Model/Import.cs
```diff
@@ -1976,6 +1976,11 @@ private static bool TrySplitColumns(string line, char sep, out string[] columns)
         private static Type GetColumnType(string value, IFormatProvider provider)
         {
             double result;
+            var quote = @"""";
```
brendanx67, line 1979: Looks like my fault for sure in trying to handle international inputs regardless of locale. I think the correct fix is actually to get rid of TrySplitColumns and use ParseDsvFields instead. It is what gets used in the end, and it already has logic for dealing with quoted fields.

brendanx67 (changes requested):
Prefer a different fix, using ParseDsvFields instead of TrySplitColumns, unless that turns out not to work for some reason.

pwiz_tools/Skyline/Model/Import.cs
bspratt: Can we go with this patch for now and enter an issue to do the more thorough fix later?

brendanx67: It is not a lot of work to try what I suggested. It is like changes to 3 lines. If it passes all existing tests and your new test, it is a way better change that would fix other potential bugs.

bspratt: Fair enough.

bspratt: Actually I think it's sufficient to change just one line within TrySplitColumns rather than replacing it- just change line.Split with line.ParseDsvColumns.  This accommodates the data set that alerted us to the issue, and the previously modified test.

brendanx67 (approved):
Good point. It's only a single-line change. Very tidy.

PR #1671, opened 2021-07-05
Add mass error to the full scan viewer.

rita-gwen:


pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
```diff
@@ -732,6 +732,9 @@ private SpectrumGraphItem MakeGraphItem(SpectrumDisplayInfo spectrum, SpectrumNo
                 rankAdducts,
                 rankTypes,
                 null);
+
+            var isHighRes = !Equals(settings.TransitionSettings.FullScan.ProductMassAnalyzer,
```
nickshulman, line 736: GraphSpectrum is used for displaying spectra from a library. The spectra in the library were often acquired on a completely different instrument than the raw files in the current document, so you should not be looking at TransitionSettings to decide whether the library spectrum is high resolution or not.

I think the spectrum should always display the mass error, even if the spectrum is low resolution. Skyline does not calculate the mass error when extracting chromatograms for low resolution data, but if the user has asked to see the mass error in this graph, Skyline should show it.

nickshulman (commented):
It looks like the GraphFullScan never displays the mass errors for MS1 spectra. How difficult would it be to add that?

brendanx67:
We definitely want to be able to display mass errors for MS1 and MS/MS even when not annotating them fully.

pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
rita-gwen: I removed the high res logic to show mass error for all types of spectra.

brendanx67: I don't know. Should it show mass error in PPM for low-resolution spectra? I would have thought low resolution should show mass error in m/z units. I guess we can leave it in PPM and wait to see what users think.

rita-gwen:
I pushed the initial implementation for mass error display in non-annotated mode, but there is a couple of things to discuss still. 
First, I feel that using SpectrumFilterPair class to calculate the mass error would be an unnecessary overcomplication. It would also affect performance since SpectrumFilterPair aggregates the ion mobility data even though it is already aggregated by the full scan viewer. So, to avoid code duplication I refactored the mass error calculation into a separate utility class. Hope this is acceptable.
Second, in MS/MS mode mass errors shown are different in annotated and not annotated modes for the selected products (those in the document tree). This is because in annotated mode it is calculated using a single matched peak and in not annotated mode it is averaged over the extraction window. I am not sure if this is acceptable, but do not see a way to avoid it.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -726,6 +743,14 @@ private void  CreateSingleScan()
                     var item = new SpectrumItem(pointLists[i], GetTransitionColor(transition), _msDataFileScanHelper.ScanProvider.Transitions[i].Name, 2);
                     var curveItem = _graphHelper.GraphControl.AddGraphItem(GraphPane, item, false);
                     curveItem.Label.IsVisible = false;
+                    if (massErrors != null)
+                    {
+                        var errorAccumulator =
+                            new SpectrumFilterPair.MeanErrorAccumulator(true, ChromExtractor.summed,
+                                transition.ProductMz);
+                        pointLists[i].ForEach(pt => errorAccumulator.AddPoint(pt.X, pt.Y));
```
nickshulman, line 751: I think you need to go back to the original spectrum in order to know all of the mz's and intensities that are within this transition's extraction window.
pointLists[i] is going to be missing some values if the extraction windows for the transitions overlap with each other

Those pointLists arrays are appropriate for deciding what color to make the lines on the graph, but for actually calculating the mass error I think you need to go back to the original spectrum.

nickshulman (commented):
I think the "MeanErrorAccumulator" class looks great.
I would probably have called it "IntensityAccumulator". You might even want to eventually make it its own top-level class in its own source file.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -734,17 +749,37 @@ private void  CreateSingleScan()
...
 
-                // Create curve for all points to provide shading behind stick graph.
-                if (_msDataFileScanHelper.MsDataSpectra.Length > 0 && !_msDataFileScanHelper.MsDataSpectra[0].Centroided)
+            if (Settings.Default.ShowFullScanMassError)
+            {
+                massErrors = new double[_msDataFileScanHelper.ScanProvider.Transitions.Length];
+
+                //create and initialize a map of transition,accumulator pairs
+                var meanErrorsMap =
+                    new Dictionary<Identity, IntensityAccumulator>(_msDataFileScanHelper.ScanProvider.Transitions.Length);
```
nickshulman, line 768: The "Equals" method on Identity usually does not do what you want.
You should either use "IdentityEqualityComparer" when you construct this Dictionary, or you should use "IdentityPath" as the key in this dictionary.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -600,6 +612,8 @@ public override int GetHashCode()
                         _transitionIndex = new int[]{};
                 }
 
+                var isHighRes = !Equals(settings.TransitionSettings.FullScan.ProductMassAnalyzer,FullScanMassAnalyzerType.qit);
```
nickshulman, line 615: I think this line is no longer necessary.

pwiz_tools/Skyline/Model/Results/ScanProvider.cs
```diff
@@ -40,6 +40,12 @@ public class TransitionFullScanInfo
         public double? ExtractionWidth;
         public IonMobilityFilter _ionMobilityInfo;
         public Identity Id;  // ID of the associated TransitionDocNode
+
+        public bool MatchMz(double mz)
```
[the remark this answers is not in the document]
rita-gwen: The original logic in GraphFullSpectrum was this:
!(mz <= transition.ProductMz.Value - transition.ExtractionWidth/2 || mz > transition.ProductMz.Value + transition.ExtractionWidth/2))
which is the same I currently have in the MatchMz() method. If you think this is a bug I'll fix it.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
rita-gwen: You mean like this: 
`                var meanErrorsMap =
                    new Dictionary<Identity, IntensityAccumulator>(_msDataFileScanHelper.ScanProvider.Transitions.Length, new IdentityEqualityComparer<Identity>());
`

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
rita-gwen: Right. I wonder how it passed code inspection.

rita-gwen:
And I was wondering if I should show the mass error in the ion mobility heat map mode. It is technically possible, but I'll have to aggregate the spectrum along the IM dimension, so there will be some performance penalty. Not sure it's worth it.

pwiz_tools/Skyline/Model/Results/IntensityAccumulator.cs
```diff
@@ -0,0 +1,41 @@
+﻿namespace pwiz.Skyline.Model.Results
+{
+    public class IntensityAccumulator
```
brendanx67, line 3: File and class must have the same name.

pwiz_tools/Skyline/Model/Results/ScanProvider.cs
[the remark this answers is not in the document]
brendanx67: The two should match, as Nick points out. This is a very slight difference <= and < are virtually identical with double values. I would have thought the filter would be "mz >= ProductMz.Value - ExtractionWidth / 2 &&
mz <= ProductMz.Value + ExtractionWidth / 2", but I can see that Nick is right and I must have made it left inclusive and right exclusive. Better add a comment here explaining that the logic was to match SpectrumFilterPair.FilterSpectrumList so that this discussion is not forever lost.

pwiz_tools/Skyline/Model/Results/SpectrumFilterPair.cs
```diff
@@ -265,40 +265,23 @@ public ExtractedSpectrum FilterQ3SpectrumList(MsDataSpectrum[] spectra, bool use
                     }
 
                     // Add the intensity values of all peaks that pass the filter
-                    double totalIntensity = extractedIntensities[targetIndex]; // Start with the value from the previous spectrum, if any
-                    double meanError =  highAcc ? meanErrors[targetIndex] : 0;
-                    for (int iNext = iPeak; iNext < mzArray.Length && mzArray[iNext] < endFilter; iNext++)
+                    var accumulator = new IntensityAccumulator(highAcc, Extractor, targetMz)
```
brendanx67, line 268: This is so performance-critical that I am going to need to see a perf test on a large file the proves extraction using this new mechanism is not slower. It shouldn't be, but I hate to admit the number of times "shouldn't be" and "actually not" have turned out to differ.

brendanx67 (changes requested):
Mostly looking pretty good now. It needs a test, and you need to prove that you didn't slow down large-scale extraction. Other than that, only very minor changes.

pwiz_tools/Skyline/Model/Results/SpectrumFilterPair.cs
rita-gwen: I ran DriftTimePredictionTutorial test on my machine on 2_21 and on this branch. The elapsed time for import results operation was almost identical: 3:08 and 3:10 min. Is this large enough? I can run it in a loop for a while to gather some statistic and tighten the confidence interval.

pwiz_tools/Skyline/Model/Results/IntensityAccumulator.cs
rita-gwen: Fixed

pwiz_tools/Skyline/Model/Results/SpectrumFilterPair.cs
rita-gwen: MatchMz logic is fixed as well.

brendanx67: Try the DiaSwathFull tutorial tests. The drift time tests will be dominated by spectrum loading. What you need is something where a lot of extraction gets done on each spectrum, i.e. extraction to spectrum point retrieval is high. That is not the case for the tutorial you chose. If the DiaSwathFull tests look good, I may still want one more test withsomething like the PanHuman library, i.e. something with just a lot, a lot of targets.

rita-gwen: This is SkylineTester output from TestDiaTtofFullSearchTutorial: `
Skyline 2.21
[14:42]   1.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.46/154.8 MB, 90/578 handles, 416 sec.
[14:49]   2.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.86/160.8 MB, 92/581 handles, 415 sec.
[14:56]   3.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.86/164.1 MB, 91/578 handles, 413 sec.
[15:03]   4.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.86/165.6 MB, 91/568 handles, 413 sec.
[15:10]   5.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.86/164.5 MB, 91/563 handles, 409 sec.
[15:17]   6.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.86/167.4 MB, 92/592 handles, 412 sec.
[15:24]   7.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.86/171.5 MB, 92/578 handles, 431 sec.
**Mean=415.6 std=7.2**

PR branch
[13:51]   1.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.46/153.6 MB, 89/576 handles, 419 sec.
[13:58]   2.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.85/151.9 MB, 91/581 handles, 419 sec.
[14:05]   3.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.85/156.3 MB, 91/571 handles, 423 sec.
[14:12]   4.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.86/165.2 MB, 91/586 handles, 420 sec.
[14:19]   5.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.85/165.0 MB, 90/562 handles, 424 sec.
[14:26]   6.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.85/165.8 MB, 90/572 handles, 423 sec.
[14:33]   7.0   TestDiaTtofFullSearchTutorial                  (en)   0 failures, 7.85/157.8 MB, 91/576 handles, 419 sec.
**Mean=421.0 std=2.2**
`
Looks pretty similar to me. Where can I find those PanHuman library files?

rita-gwen: I recorded import times in the TestDiaTtofFullSearchTutorial as Brendan requested for 2.21 (288 +-5 sec) and this branch (286+-3 sec).  No significant difference.

PR #1676, opened 2021-07-07
Classify transition lists by first attempting to read them as proteomics transition lists

henrytsanford:
Previously we would search for headers particular to small molecule transition lists in order to determine if a transition list was peptide or small molecule. With this change, we try to create a peptide row reader using the first 100 lines of the pasted text. In order to create a peptide row reader, we must find the matching amino acid sequence and precursor m/z columns, among other things. If we can successfully create a peptide row reader, we assume it is a peptide transition list. 

If we are unable to create a peptide row reader, it falls back on our existing logic and looks at the headers. 

This also fixes a small redundancy where the same transition list would be classified as peptide or small molecule twice in the course of one import.


pwiz_tools/Skyline/Model/Import.cs
```diff
@@ -373,29 +373,35 @@ private void InitFormat(IList<string> inputLines)
         {
             if (FormatProvider == null)
             {
-                char sep;
-                IFormatProvider provider;
-                Type[] columnTypes;
-                string inputLine = 0 < inputLines.Count ? inputLines[0] : string.Empty;
-                if (!MassListImporter.IsColumnar(inputLine, out provider, out sep, out columnTypes))
+                if (!TryInitFormat(inputLines, out var provider, out var sep))
```
bspratt, line 376: Could probably use some comments through here.

pwiz_tools/Skyline/Model/Import.cs
```diff
@@ -412,19 +418,31 @@ public class MassListImporter
 
         private int _linesSeen;
 
-        public MassListImporter(SrmDocument document, MassListInputs inputs)
+        // This constructor is only suitable for investigating the peptide-vs-small molecule nature of inputs
+        // CONSIDER(henryS) Arguably that code should be split out into its own class
```
bspratt, line 422: Brendan, is it worth the churn?

pwiz_tools/Skyline/Model/Import.cs
```diff
@@ -412,19 +418,31 @@ public class MassListImporter
...
 
         public SrmDocument Document { get; private set; }
         public MassListRowReader RowReader { get; private set; }
-        public SrmSettings Settings { get { return Document.Settings; } }
+        public SrmSettings Settings { get; private set; }
         public MassListInputs Inputs { get; private set; }
         public IFormatProvider FormatProvider { get { return Inputs.FormatProvider; } }
         public char Separator { get { return Inputs.Separator; } }
-        public bool IsSmallMoleculeInput { get; private set; }
+        public SrmDocument.DOCUMENT_TYPE InputType { get; private set; }
```
bspratt, line 445: comment this, something like // Peptide vs Smallmol vs None (meaning Unknown)

pwiz_tools/Skyline/Model/SrmDocument.cs
```diff
@@ -1482,11 +1482,13 @@ private SrmDocument MergeMatchingPeptidesUserInfo(IList<PeptideGroupDocNode> pep
             return docNew;
         }
 
-        public MassListImporter PreImportMassList(MassListInputs inputs, IProgressMonitor progressMonitor, bool tolerateErrors)
+        public MassListImporter PreImportMassList(MassListInputs inputs, IProgressMonitor progressMonitor, bool tolerateErrors, DOCUMENT_TYPE inputType = DOCUMENT_TYPE.none)
```
bspratt, line 1485: a little explanatory comment header here would be nice

pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -1871,7 +1871,7 @@ public void ImportMassList(string fileName)
             }
         }
 
-        public void ImportMassList(MassListInputs inputs, string description, bool assayLibrary)
+        public void ImportMassList(MassListInputs inputs, string description, bool assayLibrary, SrmDocument.DOCUMENT_TYPE inputType = SrmDocument.DOCUMENT_TYPE.none)
```
bspratt, line 1874: General need for comment headers throughout, for the next coder that comes along

bspratt (changes requested):
It strikes me (and this was true before your changes) that there's an awful lot of logic in EditMenu.cs that would be useful to the commandline version of Skyline. We should probably investigate that and unify the code. That is, move that logic to Import.cs, probably.

Otherwise see individual review comments, which are mostly about wanting more comments.

pwiz_tools/Skyline/Model/SrmDocument.cs
henrytsanford: I added a comment describing the method and the parameter I added. Is this the comment format you are looking for? I am not sure what the best comment style is when I only want to comment on one of the parameters.

bspratt: If you don't want to go the full Monty on the comment, then just put it on its own line and comment that:
       /// <summary>
        /// Return a mass list import if the progress monitor is not cancelled and we are able to read the document
        /// </summary>
       public MassListImporter PreImportMassList(MassListInputs inputs, IProgressMonitor progressMonitor, bool tolerateErrors,
            DOCUMENT_TYPE inputType = DOCUMENT_TYPE.none) // "None" means "don't know if it's peptides or small molecules, go figure it out".

bspratt (approved):
OK, you've verified that there's not an issue with model and UI code intermingling. Sounds like Brendan is OK with getting this into master.

PR #1677, opened 2021-07-08
Allow filtering for targets when building spectral library from existing blib files

kaipot:


pwiz_tools/BiblioSpec/src/BlibBuilder.cpp
```diff
@@ -566,72 +580,77 @@ int BlibBuilder::readSequences(set<string>** seqSet, bool modified)
...
+        {
+            // get modification
+            size_t endIdx = sequence.find(']', i + 1);
+            if (endIdx != string::npos)
             {
-                // get modification
-                size_t endIdx = sequence.find(']', i + 1);
-                if (endIdx != string::npos)
+                ++i;
+                istringstream deltaMassExtractor(sequence.substr(i, endIdx - i));
```
chambm, line 612: How many rows are you likely to be parsing? If it's anything significant, you should use lexical_cast<double> here (which is overloaded to use strtod, much more efficient than istringstream).

pwiz_tools/BiblioSpec/src/BlibBuilder.cpp
```diff
@@ -361,11 +361,25 @@ int BlibBuilder::transferLibrary(int iLib,
...
         progress->increment();
 
-        int spectraId = sqlite3_column_int(pStmt, 0);
+        bool skip = false;
+        if (targetSequences != NULL || targetSequencesModified != NULL) {
+            // filtering targets
+            bool noFilter =
+                // don't filter if targetSequences is not null and it contains the unmodified sequence
+                (targetSequences != NULL &&
+                 targetSequences->find(boost::lexical_cast<string>(sqlite3_column_text(pStmt, 1))) != targetSequences->end()) ||
```
chambm, line 370: Pull these columns into named variables to make the code easier to follow.

chambm:
I wish you could force diff to work at the function block level. :) It seems we're missing a test for this case. You're wanting to run BlibBuild with a blib input and the -u or -U option to filter for only certain peptides? The test we have uses msms.txt input. Can you add one for blib input? Presumably that one would fail without this change?

chambm (approved):

PR #1682, opened 2021-07-13
* fixed using declaration for bnw::system causing VS2019 error

chambm:
* fixed using declaration for bnw::system causing VS2019 ambiguity compile error even with all calls fully qualified (need to switch to a linting technique instead)
* fixed unnecessary inclusion of gzio.c in zlib
* added `pwiz::util::getline()` for convenient handling of DOS-style line endings on *nix; replaced most uses of `std::getline()` with `pwiz::util::getline()` (except where byte counting is important)

@bspratt The line ending thing plagued me when I ran pwiz tests on GCC (via Linux Docker containers running on my Windows 10 PC) on a git repo I checked out on Windows (so auto-CRLF setting gave the non-binary files CRLF endings). I was annoyed at being annoyed by this issue in 2021, so I fixed it everywhere in pwiz. Other than the byte-counting caveat I've documented, can you think of any negative consequences for us of silently dropping `\r` from the end of newline-delimited lines?

### DIFF
--- a/Jamroot.jam
+++ b/Jamroot.jam
@@ -351,6 +351,9 @@ project pwiz
         <toolset>clang:<cxxflags>-Wno-unused-local-typedef
         <toolset>darwin:<cxxflags>-Wno-unused-local-typedef
 
+        # avoid unused parameter warnings (especially annoying for stub functions)
+        <toolset>clang:<cxxflags>-Wunused-parameter
+
         # avoid C++1x deprecation warnings
         <toolset>gcc:<cxxflags>-Wno-deprecated-declarations
         <toolset>clang:<cxxflags>-Wno-deprecated-declarations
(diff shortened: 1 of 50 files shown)


bspratt:
I can't imagine an issue with this behavior - the windows behavior of getline() already strips the /r along with the /n.

But I do wish the function name wasn't overloaded this way - a maintainer might easily be confused by this. Could we call it something like unix_getline() or os_agnostic_getline()? OK, those are terrible, how about safe_getline()?

chambm:
I considered renaming the function instead of overloading but because I gave it the exact same interface as `std::getline`, I couldn't convince myself it was necessary. It does have slightly different behavior by design though, so your concern is enough to tip the balance. `getlineDieCarriageReturnDie`! (Pwiz function names are usually camel case.) `safe_getline` seems too vague. I like `os_agnostic_getline` but it's long. How about `getlinePortable`?

bspratt:
If we can't have getlineDieCarriageReturnDie() then I'll settle for getlinePortable().

bspratt:
So we know from TC success that this code is fine for VS2017 and VS2019 alike. But if we update TC to VS2019 the we'll break the current release branch, yes? Or will we have both 2017 and 2019 available on those images?

chambm:
VS2017 is installed on the updated image so that shouldn't be an issue.

bspratt (approved):
Looks good to me. :shipit: :

chambm:
Maybe we should hold off on VS2019 until right before we branch 21.2? That doesn't mean this can't be merged now, just that the builds will keep using VS2017 until then. Let's see what Brendan says when he gets back. There's no rush on the switch I think but we don't want the dev docs to have to worry about building with 2 different VS versions (except people responsible for patching older releases).

bspratt:
That's probably quite some time from now - I'm pretty sure Brendan wanted us just to pull the trigger on this. As it doesn't prevent the use of VS2017 I can't see the harm in doing it now.

chambm:
Yes like I said I agree we can merge this now, I just wouldn't push anybody to switch to VS2019 until later (at least, if they think they might be making cherry picks to the release branch).

PR #1683, opened 2021-07-13
Fix mismatch between color of annotations and lines on small molecule library explorer graph

henrytsanford:
In the library explorer spectrum graph, ranked small molecule fragments are colored differently than unranked small molecule fragments. However, the annotations were being colored as if all small molecule fragments were unranked. This resulted in a mismatch between the color of the lines on the graph and the color of the annotations. 

This change fixes the mismatch by considering the rank of small molecule fragments when assigning the color of annotations.


bspratt (approved):
Code looks good to me, and I've seen your demo of the fix as well. NIce to see a test added, as well.

PR #1685, opened 2021-07-19
Move Shimadzu method export out of Skyline and update libraries

kaipot:


pwiz_tools/Skyline/Executables/BuildMethod/BuildShimadzuMethod/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+
+// General Information about an assembly is controlled through the following
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("BuildShimadzuMethod")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("BuildShimadzuMethod")]
+[assembly: AssemblyCopyright("Copyright ©  2021")]
```
bspratt, line 13: Should this be "Copyright © University of Washington 2021"?

pwiz_tools/Skyline/Skyline.csproj
```diff
@@ -3324,6 +3325,7 @@
       <Link>unimod.obo</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <None Include="Method\Shimadzu\BuildShimadzuMethod.exe.config" />
```
nickshulman, line 3328: This should be changed so that the ".exe.config" gets copied to the output directory and included in the installers.
Skyline now ships with all of the .exe.config files  since they tell Windows which .Net version the exe needs. If the .exe.config is missing, Skyline forces the user to install .Net 3.5.

PR #1687, opened 2021-07-20
Feature/20210716 msconvert sonar mzml output

bspratt:
Improved msconvert handling for Waters SONAR data conversion to mzML, including the new 4 array compact representation (enabled by the --combineIonMobilitySpectra commandline switch).


chambm:
Give until the middle of next week to review this please. 😄

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
```diff
@@ -230,26 +230,42 @@ PWIZ_API_DECL SpectrumPtr SpectrumList_Waters::spectrum(size_t index, DetailLeve
...
+        else
+        {
+            rawdata_->Info.GetPrecursorMassRange(ie.function, ie.scan, minMZ, maxMZ); // Get the quadrupole filter range this spectrum
+        }
+    }
+    else
+    {
+        rawdata_->Info.GetAcquisitionMassRange(ie.function, minMZ, maxMZ);
+    }
     scan.scanWindows.push_back(ScanWindow(minMZ, maxMZ, xUnit));
```
chambm, line 248: As mentioned in my (tardily sent) email, scan window should definitely be the min and max m/z possible in the spectrum itself (that's the TOF analyzer for SONAR, not quadrupole). It should have nothing to do with isolation. For MS1 SONAR in the uncombined representation, we can't use the isolation window element (same problem we have with plain MSe), so I think we should just put a couple of userParams in the root result spectrum for that representation. We won't (can't) add them for the combined representation because that's what the extra 2 arrays are for.

chambm (changes requested):
Other than my comment it all looks good. Was updating the CV straight-forward enough? I'll add pepXML support for the new enzyme.

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
bspratt: CV update was pretty smooth, except for a term that had been renamed and required a little manual intervention.

Can I leave the userParam stuff to you? I think you have a much clearer vision of what you want there.

chambm:
@bspratt Do you have a _sonar.inf file? We have this SONAR_Short.raw file but no _sonar.inf file to go with it and I'm not sure how to generate one. The calibration would probably be fine as an identity operation (no change between calibrated and uncalibrated masses). [This Waters page](https://support.waters.com/KB_Inf/MassLynx/WKB202040_How_to_create_post-acquisition_a_SONAR_calibration_file) claims to have a template inf file as an attachment but I can't see it.

bspratt:
[VERSION#1]
<---- version must be first line

[DATE]
Acquired Date=21-Dec-2020
<---- comment
Acquired Time=03:25:08
[INSTRUMENT]
Quad Resolution=15.0
Collision Cell Pressure=3.0
Collision Cell Gradient=4.0
Ion Energy=0.2
Post Filter=2
Resolution Mode=Resolution
Pusher Period=85
[SONAR#1]
Scan Time=0.1
Start M/Z=400.0
End M/Z=1000.0
Quadrupole Window=20.0
Calibration=400,-1.1422,2.5378,7.1251
Calibration=1000,207.3578,211.2178,216.0451

PR #1688, opened 2021-07-21
Write transition CE and DP to the protobuf

vagisha:
For this Panorama support request: https://panoramaweb.org/home/support/announcements-thread.view?rowId=6233

### DIFF
--- a/pwiz_tools/Skyline/Model/TransitionDocNode.cs
+++ b/pwiz_tools/Skyline/Model/TransitionDocNode.cs
@@ -575,6 +575,9 @@ namespace pwiz.Skyline.Model
 
         public double? GetCollisionEnergy(SrmSettings settings, PeptideDocNode nodePep, TransitionGroupDocNode nodeGroup)
         {
+            if (ExplicitValues.CollisionEnergy.HasValue)
+                return ExplicitValues.CollisionEnergy; // Explicitly imported, overrides any calculation
+
             double? ce = null;
             TransitionPrediction predict = settings.TransitionSettings.Prediction;
             var optimizationMethod = predict.OptimizedMethodType;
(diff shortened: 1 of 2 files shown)


pwiz_tools/Skyline/Model/TransitionDocNode.cs
```diff
@@ -555,9 +556,90 @@ public SkylineDocumentProto.Types.Transition ToTransitionProto(SrmSettings setti
...
+            {
+                if (!ce.HasValue)
+                {
+                    ce = OptimizationStep<CollisionEnergyRegression>.FindOptimizedValue(settings,
+                        nodePep, nodeGroup, this, optimizationMethod, ceRegression,
+                        SrmDocument.GetCollisionEnergy);
+                }
+            }
+
+            if (ExplicitValues.CollisionEnergy.HasValue)
```
nickshulman, line 612: Usually this would go at the beginning of the function, and the function would return the explicit value early, instead of going through the other code paths of calculating it.
(probably doesn't make any difference, so this way is fine too)

nickshulman (approved):
Looks good to me.
The function "AssertEx.Serializable(SrmDocument)" tries to make sure that the document is the same regardless of what "CompactFormatOption" it was saved with. Unfortunately, since these properties only get written out, and get ignored when reading in, this was not detected when it compared the serialized and deserialized documents. I can't think of a good way to improve that assert, so we might just have to remember to be careful when adding write only attributes in the future.

pwiz_tools/Skyline/Model/TransitionDocNode.cs
vagisha: Thanks.  I moved it to the beginning.

PR #1689, opened 2021-07-21
Fix TestLibraryExplorerAsSmallMolecules

henrytsanford:
Only test small molecule annotations after switching to a small molecule library


bspratt (approved):
Looks good. One of those "why did this ever work at all?" things, we've all been there.

PR #1695, opened 2021-07-23
It seems some Lab Solutions versions don't write TIC information, so be tolerant of it being missing

bspratt:
That is to say, Shimadzu .lcd  chromatogram files may not contain TIC values.

reported by user Motonao Nakao (on the msconvert support board)


chambm:
Only thing I would add here is a // TODO for logging the exception when I finally get around to adding logging support in pwiz. :)

chambm (approved):

PR #1697, opened 2021-07-26
Selection of centroided vs profile peaks in Full Spectrum Viewer.

rita-gwen:


pwiz_tools/Skyline/app.config
```diff
@@ -771,6 +771,9 @@
             <setting name="SyncMZScale" serializeAs="String">
                 <value>False</value>
             </setting>
+            <setting name="FullScanPeakType" serializeAs="String">
```
rita-gwen, line 774: I left it as Peak Type because Scan Type is already taken for MS1/MS2 selection. Let me know if you have any other suggestions for the name.

pwiz_tools/Skyline/Model/Results/ScanProvider.cs
```diff
@@ -261,10 +295,14 @@ public void Dispose()
         {
             lock (this)
             {
-                if (_dataFile != null)
+                foreach (var key in _dataFileCentroidedMap.Keys.ToList())
```
nickshulman, line 298: I am not sure I understand what is happening in this function.
Normally, you would call Dispose on all of the Value's in the Dictionary, and then clear the dictionary.
It looks instead like you are looking for one specific item in the dictionary, calling Dispose on that, and removing it from the Dictionary.
"Dispose" is supposed to be called on an object at the end of its life, so if "Dispose" is called on the ScanProvider, we would normally expect that this Dictionary would end up empty because the ScanProvider is no longer alive.

PR #1701, opened 2021-07-28
Perftests should get their data from S3 instead of PanoramaWeb when SKYLINE_DOWNLOAD_FROM_S3…

bspratt:
… environment variable is set to "1"

And should retry on the original URL when a file can't be found on S3.


bspratt:
Do we want to leave the "downloading..." messages in there?

PR #1702, opened 2021-07-28
BiblioSpec: add support for XTandem point mutations

kaipot:
https://skyline.ms/issues/home/issues/details.view?issueId=827

no real data to test with but works okay on a manually edited test file


chambm:
This branch/PR works for a PM search I did with X!Tandem, including a peptide with a PM from T->C and then the C+57.  But changing the peptide sequence to one that doesn't occur in any FASTA that a user might provide is a pretty big problem. For example it would totally break protein association. @brendanx67 @nickshulman Does Skyline have any explicit support for mutations? Should they be represented as a mod on the original AA?

PR #1704, opened 2021-07-29
Update DigitalRune.Windows.Docking.dll to something that was built wi…

nickshulman:
…th .Net 4.7.2

The source code for this dll now lives at:
https://github.com/uw-maccosslab/developers/tree/master/skylinedev/DigitalRune-Docking-Windows


brendanx67 (changes requested):
The original list of .NET 3.5 dependent DLLs you provided also included the JA and zh-CHS resource DLLs:

.\ja\DigitalRune.Windows.Docking.resources.dll
.\zh-CHS\DigitalRune.Windows.Docking.resources.dll

Please add these to the PR.

nickshulman:
> The original list of .NET 3.5 dependent DLLs you provided also included the JA and zh-CHS resource DLLs:
> 
> .\ja\DigitalRune.Windows.Docking.resources.dll
> .\zh-CHS\DigitalRune.Windows.Docking.resources.dll
> 
> Please add these to the PR.

Excellent point. I did not realize those dll's were there. They had not been updated since they were first added in 2014, but I probably should have been updating them every time I checked in a new DigitalRune.Windows.Docking.dll.

brendanx67 (approved):

nickshulman:
I don't think there's any need to change these particular dll's at this time.
This pull request was based on a misunderstanding of how Windows decides that an application requires .Net 3.5.

chambm:
Are you saying .NET 3.5 DLLs will work on a machine without .NET 3.5 installed?

nickshulman:
> Are you saying .NET 3.5 DLLs will work on a machine without .NET 3.5 installed?

Yes, Skyline works fine on a clean Windows 10 machine without .Net 3.5 installed.
The executable is the thing that determines which version of .Net is necessary.
We figured out that the only reason that Windows was requiring the user to install .Net 3.5 when they tried to do one of the method exports is that were not included the ".exe.config" files for those method export executables in our install. If Windows tries to run a .Net .exe and there is no .exe.config, then Skyline will force the user to install .Net 3.5. But if the .exe.config is there and says that it uses .Net 4.x, then it does not matter what other dll's are there-- the thing will work without .Net 3.5.

chambm:
Curious. I wonder if that would work in the odd (and currently irrelevant to us) case where a DLL created its own AppDomain or hosted its own instance of the CLR.

PR #1706, opened 2021-07-30
Include tutorial audit log files in SkylineTester.zip

nickshulman:


brendanx67 (approved):

bspratt:
I'm curious - why this change? There are lots of tests that will fail if you don't have the source controlled data files downloaded.

nickshulman:
> I'm curious - why this change? There are lots of tests that will fail if you don't have the source controlled data files downloaded.

I wanted to know whether the tutorial tests would pass on a computer that did not have Visual Studio or .Net 3.5 installed. (I think it's just the ones that use Perocolator).
So far I have only run into a couple of other tests that fail, but I have not managed to run the whole test suite yet.
I don't think I've run into a test yet which is failing because the source tree is missing. The test "CodeInspection" passes if it sees that the source code directory is not there.

bspratt:
There are at least a couple dozen that pull their data from checked-in .zip
files. I don't remember why I singled out the code inspection test like
that.

nickshulman:
> There are at least a couple dozen that pull their data from checked-in .zip files. I don't remember why I singled out the code inspection test like that.

All .zip files from the source tree are included in SkylineTester.zip. The method "CreateZipInstallerWindow.FindZipFiles" finds them all.

brendanx67:
The original intent of the SkylineTester artifacts was that they could pass
all tests without needing the source tree. Don Marsh designed it this way.
Especially the tutorial tests and all of the tests that show forms on the
Forms tab, because we had intended to allow professional translation
testers to use the artifacts in testing their translations. In the end, I
have had to do the translation testing, because the companies we use are
not able to handle this even with SkylineTester.

bspratt:
Excellent, thanks for fixing the omission of the tutorial test data.

PR #1708, opened 2021-08-02
Fix ViewLibraryDlg handling of peptide with modified 'X'

kaipot:


nickshulman:
I don't understand what this pull request is fixing. I know the letters "B", "J", "X", and "Z" sometimes behave strangely when they are used in peptide sequences. What have they done this time?

kaipot:
> I don't understand what this pull request is fixing. I know the letters "B", "J", "X", and "Z" sometimes behave strangely when they are used in peptide sequences. What have they done this time?

Probably not very common but this change is to fix an exception that happens when viewing a spectral library containing an 'X' with a modification on it.

PR #1710, opened 2021-08-05
Enable writing CoV values for Thermo Fusion

kaipot:
https://skyline.ms/announcements/home/support/thread.view?rowId=52128


nickshulman (approved):
Looks good to me.

PR #1712, opened 2021-08-06
Fix for the spectrum comparison bug

rita-gwen:


brendanx67:
I will fix this by adding equality members to SpectrumDisplayInfo.

PR #1713, opened 2021-08-06
Performance improvements for handling large .MSP spectral library imports

bspratt:
I've done some profiling on Skyline as it deals with reading in a large .MSP file (small molecule stuff in particular).
 
NistLibSpec.cs: Use compiled regular expressions, and be more judicious about when they are used (i.e. don't check a line for RT value when RT value has already been located)
LibKeyModificationMatcher.cs: no need to mess with modifications in small molecule data
ViewLibraryPepInfo.cs and ViewLibraryPepInfoList.cs cache the string representation of libkeys for faster sort
Adduct.cs: we tend to see the same adduct description text again and again, so save parsing costs by keeping a threadsafe string->adduct dictionary
BioMassCalc.cs: eliminate use of regular expression in chemical formula parsing, and implement faster and more robust method of deriving non-labeled versions of formulas
RefineTest.cs: extend a test to ensure that we no longer confuse, for instance, "T" and "Ti" or "D" and "Db" when stripping labels from a formula


henrytsanford (approved):
These look valuable, especially avoiding changing modifications for non-peptide molecules. I don't see any conflicts with what I've written so far for the expanded library search.

bspratt:
I find that when merged to your branch, a very large .MSP file (485,796 entries in 12,471,966 lines) caches to .slc and displays in 37 seconds. Without these changes, it takes 72 seconds. There's still a dead spot where (I think) it's sorting for search purposes, but it's a little less painful. Still, that should probably be done only on demand, and have longwaitDlg with it.

You may wish to merge this branch into yours before, or even instead of, this making it into Master.

nickshulman (approved):
Looks good to me.

If you wanted, instead of having the "_knownAdducts" dictionaries as static variables in the "Adduct" class, you could instead have a class called "AdductParser" which you construct only when you know you are going to be parsing a large number of adducts. It would have a dictionary inside it, and would handle checking its dictionary before calling "Adduct.TryParse", and putting the result into its dictionary.

The advantage of this would be that the dictionary would disappear after you were done using the AdductParser. Also, you probably wouldn't care about AdductParser being thread safe, so you could use a regular Dictionary.
But, there are advantages to the way you've done it, too.

PR #1714, opened 2021-08-06
Synchronized integration

kaipot:


nickshulman:
The "Group By" does not work quite the way I would expect.
I am looking at the Rat Plasma dataset, and the behavior I would like is that if I adjust the peak boundaries on a Diseased replicate, it will change the peak boundaries for all of the other Diseased replicates, but leave the Healthy replicates unchanged. Likewise, if I change the peak boundaries on a Healthy replicate, it should leave the Diseased replicates unchanged.

So, I want Skyline to group my replicates by "Condition". I also want Skyline to synchronize the integration of all of the replicates. However, I do not want any synchronization to happen between groups.

This does not seem to be how it works.

Instead, when I choose Group by "Condition", I can choose whether "Diseased", "Healthy" or both are included, and the synchronization happens across all replicates that are included.

Does what I am saying make sense? I imagine users are specifically choosing the "Group By" attribute because it is something that divides the replicates into groups of things which need to be integrated differently from the things outside the group, but the same as the things inside of the group.

kaipot:
> The "Group By" does not work quite the way I would expect.
> I am looking at the Rat Plasma dataset, and the behavior I would like is that if I adjust the peak boundaries on a Diseased replicate, it will change the peak boundaries for all of the other Diseased replicates, but leave the Healthy replicates unchanged. Likewise, if I change the peak boundaries on a Healthy replicate, it should leave the Diseased replicates unchanged.
> 
> So, I want Skyline to group my replicates by "Condition". I also want Skyline to synchronize the integration of all of the replicates. However, I do not want any synchronization to happen between groups.
> 
> This does not seem to be how it works.
> 
> Instead, when I choose Group by "Condition", I can choose whether "Diseased", "Healthy" or both are included, and the synchronization happens across all replicates that are included.
> 
> Does what I am saying make sense? I imagine users are specifically choosing the "Group By" attribute because it is something that divides the replicates into groups of things which need to be integrated differently from the things outside the group, but the same as the things inside of the group.

I think I see what you mean. That should still be achievable (by selecting a single group, doing all integration for that group, and repeating for other groups). However I'm not sure if the behavior you described would work as well if someone wants to synchronize integration for 2 of 3 groups, for example.

PR #1715, opened 2021-08-10
Skyline: Fix bug where GraphSpectrum leaves its y-scale unchanged too long

brendanx67:
Unfortunately, this currently causes TestMethodEditTutorial to fail because it somehow locks the .blib file. I can't see why the equals comparison operation should cause .blib file locking. But, maybe it is just getting unluck with a race condition.


nickshulman:
I think the reason that the .blib file is getting opened an extra time is that "SpectrumInfo.Equals" causes the spectrum to be read from the database. So, this is the callstack where the .blib file is getting opened again:

```
 	Skyline-daily.exe!pwiz.Skyline.Util.PooledSqliteConnection.Connect() Line 41	C#
 	Skyline-daily.exe!pwiz.Skyline.Util.ConnectionPool.GetConnection(pwiz.Skyline.Model.Identity id, System.Func<System.IDisposable> connect) Line 229	C#
 	Skyline-daily.exe!pwiz.Skyline.Util.ConnectionId<System.Data.SQLite.SQLiteConnection>.Connection.get() Line 334	C#
 	Skyline-daily.exe!pwiz.Skyline.Util.PooledSqliteConnection.ExecuteWithConnection<pwiz.Skyline.Model.Lib.SpectrumPeaksInfo.MI[]>(System.Func<System.Data.SQLite.SQLiteConnection, pwiz.Skyline.Model.Lib.SpectrumPeaksInfo.MI[]> function) Line 106	C#
 	Skyline-daily.exe!pwiz.Skyline.Model.Lib.BiblioSpecLiteLibrary.ReadSpectrum(pwiz.Skyline.Model.Lib.BiblioLiteSpectrumInfo info) Line 1235	C#
 	Skyline-daily.exe!pwiz.Skyline.Model.Lib.CachedLibrary<pwiz.Skyline.Model.Lib.BiblioLiteSpectrumInfo>.LoadSpectrum(object spectrumKey) Line 1082	C#
 	Skyline-daily.exe!pwiz.Skyline.Model.Lib.BiblioSpecLiteLibrary.LoadSpectrum(object spectrumKey) Line 1225	C#
 	Skyline-daily.exe!pwiz.Skyline.Model.Lib.SpectrumInfoLibrary.SpectrumPeaksInfo.get() Line 2411	C#
> 	Skyline-daily.exe!pwiz.Skyline.Model.Lib.SpectrumInfo.Equals(pwiz.Skyline.Model.Lib.SpectrumInfo other) Line 2345	C#
 	Skyline-daily.exe!pwiz.Skyline.Model.Lib.SpectrumInfo.Equals(object obj) Line 2355	C#
 	Skyline-daily.exe!pwiz.Skyline.Controls.Graphs.GraphSpectrum.UpdateUI(bool selectionChanged) Line 1001	C#
 	Skyline-daily.exe!pwiz.Skyline.Controls.Graphs.GraphSpectrum.comboSpectrum_SelectedIndexChanged(object sender, System.EventArgs e) Line 1366	C#
```

So, the GraphSpectrum is hanging onto a SpectrumInfo that belongs to something which is not the current SrmDocument, and if it tells that thing to load the spectrum, the PooledSqliteConnection will never get closed because it does not belong to the current document.

nickshulman (approved):
Looks good enough to me.

The usual way to implement "Equals" for a lazy-loaded class like "SpectrumInfo" is that things would be equal if the key used to look them up in the database were the same. So, SpectrumInfoProsit and SpectrumInfoLibrary would each have their own overriding implementation of "Equals". And, the lazy-loaded parts like the actual peaks of the spectrum would not be looked at by "Equals".

Your way of making it so that the SpectrumInfoLibrary caches the value of "_peaksInfo" is probably going to improve performance in some scenario that we are not aware of, so that's a good thing.

brendanx67:
Good point. I am going to stick with this for now. It is already an improvement over what we had, and it solves the problem Rita originally changed this for. To implement what you are suggesting, it would also need to compare the library file path, which is not totally easy to get to in this structure. But I take your point. File path and lookup ID uniquely determine equality without needing to load the spectrum at all.

PR #1717, opened 2021-08-12
Issue 829: Change reports to report 1-based AA positions

nickshulman:
This change adds the columns "First Position" and "Last Position" to the document grid which show the 1-based amino acid position of the peptide in the protein.
This also changes the numbers that are displayed in the Targets tree so that they are 1-based instead of 0-based.
The columns "Begin Pos" and "End Pos" are now marked "Obsolete", which means that they only show up in the Report Editor if you push the "show all columns" button. 
If show all columns is on, here's what the column picker looks like:
image

Here's what the Peptides report in the Document Grid looks like:
image


brendanx67 (approved):
Looks good. I guess good that our audit logs are version stamped, given how much these values show up in them and the meaning will change between the last and next version.

PR #1719, opened 2021-08-13
Hide irrelevant menu items when FSV is not annotated.

rita-gwen:
Annotation-related items (like ion type or charge selection) in the context menu of the full scan viewer are not relevant when it is not in the annotation mode and should not be shown. This PR fixes that and also changes the annotations menu item text and toolbar button tooltip to "Show Peak Annotations"


brendanx67 (approved):
One small naming issue. Otherwise, looks good.

PR #1724, opened 2021-08-17
If we cannot tell if a transition list is proteomic or small molecule, decide based on the mode set in the UI

henrytsanford:


pwiz_tools/Skyline/Model/Import.cs
```diff
@@ -482,7 +482,8 @@ public bool PreImport(IProgressMonitor progressMonitor, ColumnIndices indices, b
             if (InputType == SrmDocument.DOCUMENT_TYPE.none)
             {
                 InputType =
-                    SmallMoleculeTransitionListCSVReader.IsPlausibleSmallMoleculeTransitionList(lines, Settings) ? SrmDocument.DOCUMENT_TYPE.small_molecules : SrmDocument.DOCUMENT_TYPE.proteomic;
+                    SmallMoleculeTransitionListCSVReader.IsPlausibleSmallMoleculeTransitionList(lines, Settings, 
+                       Program.ModeUI) ? SrmDocument.DOCUMENT_TYPE.small_molecules : SrmDocument.DOCUMENT_TYPE.proteomic;
```
bspratt, line 486: This feels wrong - Program.ModeUI in model code? Seems like it should be passed in. What do we do in the command line context?

pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
```diff
@@ -1919,12 +1919,12 @@ public int RowCount
             get { return Rows.Count; }
         }
 
-        public static bool IsPlausibleSmallMoleculeTransitionList(string csvText, SrmSettings settings)
+        public static bool IsPlausibleSmallMoleculeTransitionList(string csvText, SrmSettings settings, SrmDocument.DOCUMENT_TYPE modeUI = SrmDocument.DOCUMENT_TYPE.none)
```
bspratt, line 1922: again, model code, feels weird to see "UI" in a variable name. Maybe call it "defaultDocumentType"?

pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
```diff
@@ -1919,12 +1919,12 @@ public int RowCount
...
 
-        public static bool IsPlausibleSmallMoleculeTransitionList(string csvText, SrmSettings settings)
+        public static bool IsPlausibleSmallMoleculeTransitionList(string csvText, SrmSettings settings, SrmDocument.DOCUMENT_TYPE modeUI = SrmDocument.DOCUMENT_TYPE.none)
         {
-            return IsPlausibleSmallMoleculeTransitionList(MassListInputs.ReadLinesFromText(csvText), settings);
+            return IsPlausibleSmallMoleculeTransitionList(MassListInputs.ReadLinesFromText(csvText), settings, modeUI);
         }
 
-        public static bool IsPlausibleSmallMoleculeTransitionList(IList<string> csvText, SrmSettings settings)
+        public static bool IsPlausibleSmallMoleculeTransitionList(IList<string> csvText, SrmSettings settings, SrmDocument.DOCUMENT_TYPE modeUI = SrmDocument.DOCUMENT_TYPE.none)
```
bspratt, line 1927: Same thing here - feels funny to see "UI" in model code.

bspratt (changes requested):
Some question of what we should do when there is no UI, and fussy concerns about variable names

pwiz_tools/Skyline/Model/Import.cs
henrytsanford: In the command line context it defaults to proteomic mode. Perhaps I'll change it to pass in the modeUI in UI cases and pass in "proteomic" in command line uses. 

Another option for the command-line context could be to use the document type of the open document

bspratt: >> Another option for the command-line context could be to use the document type of the open document
I like it!

bspratt (changes requested):
Looking good - just a couple of minor quibbles.

pwiz_tools/Skyline/Model/SrmDocument.cs
```diff
@@ -1489,6 +1489,13 @@ private SrmDocument MergeMatchingPeptidesUserInfo(IList<PeptideGroupDocNode> pep
         /// <summary>
         /// Return a mass list import if the progress monitor is not cancelled and we are able to read the document
         /// </summary>
+        /// <param name="inputs"></param>
```
bspratt, line 1492: It's a little weird to leave these empty - we should at least put a half-assed description in there

bspratt (changes requested):
Down to a single quibble!

bspratt (approved):
Looks good to me! When you do merge it to master, make sure to note the commandline behavior as well as the UI behavior described in the PR title.

PR #1726, opened 2021-08-18
Disable MSTest timeouts for Skyline - they interfere with code coverage tests...

bspratt:
… and we have our own hang detection for nightly test runs so we don't really need them anyway.

The important change here is to the *-template.testsettings files, the other changes are just to remove crufty individual timeouts from test code.

Internal test issue, not reported by any user.


bspratt:
Team City Code Coverage tests work again with these changes.

bspratt:
Although, interestingly, we just had a successful TC run *without* these changes. I suppose the point, though, is to make that less surprising.

nickshulman (approved):
Looks good to me.

PR #1728, opened 2021-08-20
Fix error that happens if you set Instrument Min Time, but leave Instrument Max Time blank.

nickshulman:
Skyline will hit an error if you try to import results and only one of Min Time or Max Time has been given a value on Transition Settings > Instrument

(Reported by Liyan)

SpectrumFilterPair had an assert that if either minTime or maxTime was not null, they both had to be not null.
Unfortunately, this would not work if the user had set Instrument Min Time in the Transition Settings.

I think the assert was there because "ChromKey.ChangeOptionalTimes" does not take nullables as arguments.
Those optional times get used by ChromCacheBuilder.CompareMaxRetentionTime to decide the order to handle chromatograms.

We now only call ChromKey.ChangeOptionalTimes if both minTime and maxTime have values, which will be the case if the peptide or molecule had a real predicted retention time.

If SpectrumFilterPair's MinTime and MaxTime are coming from the Transition Instrument Settings, then there is no need to call ChromKey.ChangeOptionalTimes, since they won't actually affect the order in which the chromatograms need to be handled.


brendanx67 (commented):
Note that the changes you are undoing were introduced for performance improvement. I think you should leave _filterByTime and set it to true when either min or max is set, and otherwise do what you are doing. That way, the default state where neither is set is still as fast as the old code, and we accept that setting either min or max will slow down extraction in very large files.

nickshulman:
> Note that the changes you are undoing were introduced for performance improvement. I think you should leave _filterByTime and set it to true when either min or max is set, and otherwise do what you are doing. That way, the default state where neither is set is still as fast as the old code, and we accept that setting either min or max will slow down extraction in very large files.

What are you thinking could be changed about this code?

"_filterByTime" is no longer necessary since the code can check whether "MinTime" or "MaxTime" are null.
Previously, the backing fields ("_minTime" and "_maxTime") for MinTime and MaxTime were not nullable, and there was one boolean "_filterByTime" which determined whether they were both null. That had to change since Transition Instrument Settings make it possible for one but not the other to be null.

"_filterByTime" was also used to decide whether to call "ChromKey.ChangeOptionalTimes", but now that call happens if MinTime and MaxTime are both not null. We can't call ChangeOptionalTimes with null, so that check has to remain that way. Since Instrument Settings are the only way to get one but not both of MinTime and MaxTime to be null, and the instrument settings apply to all of the molecules in the document, failing to call ChangeOptionalTimes in this case is not going to affect the ChromKey sort order.

PR #1732, opened 2021-08-23
Set a timeout for git so that our nightly builds can retry instead of just hanging…

bspratt:
… when the connection to GitHub stalls out.


pwiz_tools/Skyline/SkylineTester/CommandShell.cs
```diff
@@ -356,6 +356,11 @@ private Process CreateProcess(string fileName, string workingDirectory = null)
             _process.StartInfo.Arguments = arguments;
             _process.StartInfo.StandardOutputEncoding = Encoding.UTF8; // So we can read Japanese from TestRunner's console
             _process.StartInfo.StandardErrorEncoding = Encoding.UTF8; // So we can read Japanese from TestRunner's console
+
+            // Configure git to fail if its https connection stalls out, so our retry logic can kick in
```
chambm, line 360: Did you test if it works by using some crazy values like 1gbit and 5 seconds?

bspratt: Yes, I should have mentioned that.

chambm (approved):

PR #1742, opened 2021-08-30
VS2019 build fixes for msvc_setup_script.bat

chambm:
This PR changes msvc_setup_script.bat to take an argument for platform toolset version (i.e. currently 141, 142, 143) so the version of Visual Studio tools is explicitly chosen. Previously, Boost.Build might be using one toolset and msvc_setup_script might set up a different one. (At one point msvc_setup_script was choosing the newer of 2017 and 2019, but recently it's been choosing the oldest of those two.) VS2022 will be supported as well if it follows the pattern and uses VS170COMNTOOLS as its environment variable.


bspratt (commented):
Looks like this wasn't any fun to implement... How much of this is just about specifying 64 vs 32 bit? Because I think we'd be pretty comfortable just assuming 64 in jamroot if it's not specified. Or maybe that wouldn't make any difference to the changeset?

chambm:
It was never really about 32 vs 64. The problem was I wanted the setup script to be told which VS version (platform toolset) to use, and then I had IDPicker which was depending on that script to bring rc.exe and editbin.exe into the PATH for build events. Since I needed to pass the script the platform version, I checked out the Visual Studio macros in a C++ project, which does have a PlatformToolset variable. But the C# projects turned out to not have that variable.  :rofl:  For most of these commits I didn't have a local machine with both 2017 and 2019 installed (only the TC build agents). But I finally gave in and installed 2017 on my 2019 machine, and have been testing locally. And some of the commits I only tested on 32-bit and not 64-bit. :joy:

bspratt:
What a tangled web we weave!

PR #1743, opened 2021-08-30
Show rdotp in the Replicate comparison peak area window.

rita-gwen:


pwiz_tools/Skyline/TestTutorial/ExistingExperimentsTutorialTest.cs
```diff
@@ -641,6 +651,22 @@ private static void NormalizeGraphToHeavy()
             SkylineWindow.UpdatePeakAreaGraph();
         }
 
+        private void VerifyRdotPLabels(string[] replicates, string[] rdotps)
```
nickshulman, line 654: This is not going to work right when you run the test in French where numbers are formatted differently. You are going to need to have a list of doubles, and call the method double.ToString(formatString, CultureInfo.CurrentCulture).

pwiz_tools/Skyline/TestTutorial/ExistingExperimentsTutorialTest.cs
```diff
@@ -641,6 +651,22 @@ private static void NormalizeGraphToHeavy()
...
         }
 
+        private void VerifyRdotPLabels(string[] replicates, string[] rdotps)
+        {
+            if (!Program.SkylineOffscreen)
+            {
+                var rdotpLabels = SkylineWindow.GraphPeakArea.GraphControl.GraphPane.GraphObjList.OfType<TextObj>().ToList()
+                    .FindAll(txt => txt.Text.StartsWith(@"rdotp")).Select((obj) => obj.Text).ToArray();
+
+                Assert.IsTrue(Enumerable.Range(0, replicates.Length).All(i =>
```
nickshulman, line 661: It would be better if you could make this into a for loop which loops over all of the replicates, and calls Assert for each replicate.
It's really hard to debug something like this when it fails because all you get it "Assert.IsTrue failed" and there's no indication of which replicate did it, and what the value was that was being tested.

pwiz_tools/Skyline/Controls/Graphs/AreaReplicateGraphPane.cs
```diff
@@ -822,6 +835,26 @@ private string GetDotProductResultsText(TransitionGroupDocNode nodeGroup, int in
...
                     }
                     break;
+                case AreaExpectedValue.ratio_heavy:
+                    var document = GraphSummary.DocumentUIContainer.DocumentUI;
+                    var normalizeOption = AreaGraphController.AreaNormalizeOption.Constrain(document.Settings);
+                    values = new float?[] { };
+                    if (normalizeOption.NormalizationMethod is NormalizationMethod.RatioToLabel ratioToHeavy)
+                    {
+                        var precursorNodePath = DocNodePath.GetNodePath(nodeGroup.Id, document);
+                        if (precursorNodePath.Peptide != null && nodeGroup.LabelType.Equals(IsotopeLabelType.light))
```
nickshulman, line 845: I think this should be "!nodeGroup.LabelType.Equals(ratioToHeavy.Label)"
"ratioToHeavy.Label" is the label that is going to appear in the numerator.
If nodeGroup.LabelType is different from ratioToHeavy.Label, then you want to display the rdotp.

pwiz_tools/Skyline/Controls/Graphs/AreaReplicateGraphPane.cs
```diff
@@ -489,7 +495,11 @@ public override void UpdateGraph(bool selectionChanged)
                     }
 
                     // Add area for this transition to each area entry
-                    AddAreasToSums(pointPairList, sumAreas);
+                    Func<double,double, double> aggregateFunc = (sums, y) => sums + y;
+                    if (BarSettings.Type == BarType.Cluster)
```
nickshulman, line 499: I do not understand what this code here is for. If the bar is a cluster then we want to take the Max of all the values instead of summing them? Sounds plausible, but maybe there should be a comment saying why we're doing this.

pwiz_tools/Skyline/Controls/Graphs/AreaReplicateGraphPane.cs
```diff
@@ -320,8 +323,11 @@ public override void UpdateGraph(bool selectionChanged)
                         if (parentGroupNode.HasLibInfo)
                             ExpectedVisible = AreaExpectedValue.library;
                     }
+                    if (ratioToLabel != null && !ratioToLabel.IsotopeLabelType.IsLight && isShowingMsMs)
```
nickshulman, line 326: Why is the "!ratioToLabel.IsotopeLabelType.IsLight" here? I would expect that the user could have their normalization either be "ratio to heavy" or "ratio to light", and everything would mostly behave the same but reversed.

pwiz_tools/Skyline/Controls/Graphs/AreaReplicateGraphPane.cs
rita-gwen: Yes, that is correct. For a cluster graph we need to use max instead of sum.

PR #1745, opened 2021-09-02
SkylineNightly: after 30 minutes of failing to download SkylineTester .lastFinished build,…

bspratt:
…  don't give up as we formerly did but instead go an additional 90 minutes of retrying, alternating between trying for .lastFinished vs .lastSuccessful builds.

By waiting to fail over to the potentially stale .lastSuccessful build, we get some testing done on relatively recent code while still being alerted to a problem thanks to the lower test count.

### DIFF
--- a/pwiz_tools/Skyline/SkylineNightly/Nightly.cs
+++ b/pwiz_tools/Skyline/SkylineNightly/Nightly.cs
@@ -322,12 +322,11 @@ namespace SkylineNightly
                         LogAndThrow("Unable to download SkylineTester");
                     }
 
-                    // After 30 minutes, start alternating between lastFinished and lastSuccessful builds
-                    var initialRetryExceeded = i > 30;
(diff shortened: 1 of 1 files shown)


chambm:
It makes sense to try lastFinished for a while before trying lastSuccessful, but it doesn't make sense to me to alternate between them in the second phase. What's the logic there?

chambm:
IMO it would make more sense to just fall back to lastSuccessful. If it turns out to be a connectivity problem, lastSuccessful is almost certainly going to be the same build as lastFinished. And if it was due to a build failure, then there are 2 possibilities: a new build is running and when it finishes successfully then lastSuccessful will be the same as lastFinished, or a new build isn't running and lastSuccessful will be the only working option.

bspratt:
Fair enough, Ill make the change.

chambm:
The container failure is fixed in #1744 and I'm working on the debug failures due to changing over to VS 2019 as the default TC toolset.

PR #1746, opened 2021-09-02
Fix transition list first line import bug

HenryEstberg:
A fix for a bug where the first line of a transition list was not imported in the case where the transition list did not contain a header row

### DIFF
--- a/pwiz_tools/Skyline/Util/Extensions/Text.cs
+++ b/pwiz_tools/Skyline/Util/Extensions/Text.cs
@@ -719,9 +719,9 @@ namespace pwiz.Skyline.Util.Extensions
             Initialize(reader, separator, hasHeaders);
         }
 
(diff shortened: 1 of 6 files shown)


pwiz_tools/Skyline/Util/Extensions/Text.cs
```diff
@@ -731,16 +731,17 @@ public void Initialize(TextReader reader, char separator, bool hasHeaders = true
...
+            if (columnPositions != null)
             {
                 userHeaders = TextUtil.TextSeparate(separator.ToString(), columnPositions);
                 // userHeaders = userHeaders.Replace(@" ", string.Empty);
                 _titleLine = userHeaders;
             }
             var fields = _titleLine.ParseDsvFields(separator);
             NumberOfFields = fields.Length;
-            if (!hasHeaders)
+            if (!hasHeaders && columnPositions == null)
```
henrytsanford, line 744: I'm a bit confused why this additional condition is necessary

henrytsanford (approved):
Looks good to me, I was just curious about one part of the header reading

PR #1748, opened 2021-09-07
Fix pasting header row bug

henrytsanford:
This fixes a bug introduced by [pull request #1676](https://github.com/ProteoWizard/pwiz/pull/1676) where pasting in the header row of a transition list without any of the associated data rows produces an index out of range exception instead of the error message that it used to.


bspratt (approved):
Looks good to me,

PR #1750, opened 2021-09-09
It's annoying that the BiblioSpec tests drop a bunch of tempfiles int…

bspratt:
…o the root of the source tree, this puts them in the same directory as the test executables.


chambm (approved):
Yeah that was annoying. I didn't even realize it was my change that had started doing that. Oops!

PR #1751, opened 2021-09-09
Issue 832: Improve ZedGraph handing of clipboard exceptions

nickshulman:
This change makes it so that regular "Copy" menu item on the ZedGraph context menu gets replaced with one that we implement in Skyline. Our implementation uses "ClipboardHelper.SetData", which will show the user an error message if the clipboard is locked by a different application.

The menu items on the ZedGraph context menu look the same as they always have:
image

I added a test "ZedGraphClipboardTest" which makes sure that the menu items put the right type of objects on the clipboard.
That test calls "ClipboardEx.UseInternalClipboard(false)" because the "Copy Metafile" menu item always has to use the system clipboard, because we have never figured out how to create a DataObject with a metafile in it, so it has to use the Win32 function SetClipboardData instead.
Also, the test uses the system clipboard in order to simulate the clipboard being locked by another application.

ZedGraphClipboardTest might be a little annoying if you are trying to use the computer while it is running, since it uses the real clipboard, and it also flashes the context menu on the screen to make sure it has the right items on it. I could not figure out how to show a context menu offscreen, so I think we have to live with that too.


bspratt:
>> ZedGraphClipboardTest might be a little annoying if you are trying to use the computer while it is running

Well, it wouldn't be the only test that messes with you even in "offscreen" mode, so we can probably live with that.

pwiz_tools/Skyline/EditUI/ZedGraphClipboard.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// </summary>
+        public static void AddToContextMenu(ZedGraphControl zedGraphControl, ContextMenuStrip contextMenuStrip)
+        {
+            var itemsToInsert = CreateCopyMenuItems(zedGraphControl);
+            for (int i = 0; i < contextMenuStrip.Items.Count; i++)
+            {
+                var item = contextMenuStrip.Items[i];
+                // We recognize the original Copy menu item by its name and tag, which are set in when 
+                // the menu item is created in "ZedGraphControl.contextMenuStrip1_Opening".
+                if (@"copy".Equals(item.Tag) && item.Name == @"copy")
```
bspratt, line 42: No L10N issues here, presumably?

pwiz_tools/Skyline/EditUI/ZedGraphClipboard.cs
```diff
@@ -0,0 +1,78 @@
...
+                {
+                    InsertMenuItems(contextMenuStrip.Items, i, itemsToInsert);
+                    contextMenuStrip.Items.Remove(item);
+                    return;
+                }
+            }
+            // Did not find "Copy" menu item to replace (should not happen).
```
bspratt, line 49: Do we maybe want some kind of alert for non-release versions?

pwiz_tools/Skyline/Properties/Resources.Designer.cs
```diff
@@ -365,7 +365,7 @@ public class Resources {
         }
```
bspratt, line 365: Unfortunate amount of churn here, hopefully a one time effect?

bspratt (commented):
Seems good. Pity that the tools fight us they way they do.

pwiz_tools/Skyline/Properties/Resources.Designer.cs
brendanx67: Unless you are sure this will not get switched back by the next developer, please revert the lines unrelated to your changes.

brendanx67 (approved):
Mostly looks good. Would like to see whitespace-only changes reverted.

pwiz_tools/Skyline/EditUI/ZedGraphClipboard.cs
```diff
@@ -0,0 +1,78 @@
...
+            // Add all of the copy menu items to the end of the context menu
+            contextMenuStrip.Items.AddRange(itemsToInsert.ToArray());
+        }
+
+        private static IEnumerable<ToolStripItem> CreateCopyMenuItems(ZedGraphControl zedGraphControl)
+        {
+            yield return new ToolStripMenuItem(Resources.ZedGraphClipboard_CreateCopyMenuItems_Copy, null, (sender, args) => CopyZedGraphImage(zedGraphControl));
```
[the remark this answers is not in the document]
nickshulman: > Are you positive we don't leak these? Always a bit nervous about anything that creates menu items on the fly every time a menu is shown. Most of our menu items get created exactly once.

I tried very hard to get this to leak. The test displays the context menu on the screen, and calls "PerformClick" on the menu item, and TestRunner never detected a leak.

PR #1752, opened 2021-09-10
Tweaks for MSFragger:

bspratt:
Prefer MGF input when both MGF and raw data are present.
  When MGF is not present, acknowledge its use of index as scan locator.

It's not clear that the second part of this change is useful, or even desirable - should we just insist on MGF?


pwiz_tools/BiblioSpec/src/PepXMLreader.cpp
```diff
@@ -235,19 +235,20 @@ void PepXMLreader::startElement(const XML_Char* name, const XML_Char** attr)
...
                parentAnalysisType_ = MSFRAGGER_ANALYSIS;
        }
 
        setSpecFileName(fileroot_.c_str(), extensions, dirs);
 
-       if ((analysisType_ == MSFRAGGER_ANALYSIS || parentAnalysisType_ == MSFRAGGER_ANALYSIS) && bal::iends_with(getSpecFileName(), ".mgf")) {
-           lookUpBy_ = NAME_ID;
-           specReader_->setIdType(NAME_ID);
+       if ((analysisType_ == MSFRAGGER_ANALYSIS || parentAnalysisType_ == MSFRAGGER_ANALYSIS)) {
+           SPEC_ID_TYPE idType = bal::iends_with(getSpecFileName(), ".mgf") ? NAME_ID : INDEX_ID;
```
chambm, line 249: This is probably breaking the non-TDF/non-MGF cases. For those cases, the correct idType gets set outside this block: so this block is only meant to override in the case of msfragger and MGF. I do think it's fine to insist on having the MGF if that was the input format though. Unlike the RAW/mzML case where they should be effectively interchangeable.

bspratt: So prefer MGF but not insist on it, and don't mess with the existing SPEC_ID_TYPE?

bspratt: >> I do think it's fine to insist on having the MGF if that was the input format though.
Not sure how to check that?

pwiz_tools/BiblioSpec/src/BuildParser.cpp
```diff
@@ -102,10 +102,10 @@ void BuildParser::setSpecFileName(
             }
             if (path.empty())
                 path = ".";
-            for (const auto& dir : bfs::directory_iterator(path)) {
-                bfs::path dirPath = dir.path();
-                string trialName = dirPath.filename().string();
-                for (const string& ext : extensions) {
+            for (const string& ext : extensions) { // Search for extensions in priority order
```
nickshulman, line 105: I am not sure (but I might be wrong) that it is a good idea to swap the nesting of the loops iterating over extensions and folders.

For most of the extensions, we really have no opinion which one is better. People might get confused as to why the .mz5 from the grandparent directory is being used instead of the .mzML file that is in the same folder as the pepxml.

If do need to make ".mgf" that much preferred over the raw files, it might be the right thing to do would be to prevent the raw file extensions not get added to the list if it's msfragger.

pwiz_tools/BiblioSpec/src/PepXMLreader.cpp
chambm: The only way I can see to even guess that the pepXML came from TDF is if the spectrum_query elements have ion_mobility. But of course that might change if MSFragger adds support for FAIMS. Still, for now I would say if we have ion_mobility but the spectrum file we found doesn't end in `calibrated.mgf`, then we should abort.

pwiz_tools/BiblioSpec/src/BuildParser.cpp
chambm: I agree with this. It's better to search all extensions in a single directory before moving to the parent directory, in the general case. And even raw file extensions should only be excluded for msfragger on a Bruker TDF.

pwiz_tools/BiblioSpec/src/PepXMLreader.cpp
chambm: Also, how about a test? It might be reasonable to move the existing Hela test files for msfragger TDF into a Hela%20QC_PASEF_Slot1-5_01_57_cutout_2min.d directory in the inputs directory. Maybe add a dummy analysis.tdf file. Then it should break without these changes, right? (with a dummy TDF it would break differently than with a real one, but I don't think that's important)

pwiz_tools/BiblioSpec/src/BuildParser.cpp
bspratt: Well, we already had code that appeared to depend on the order of the extensions list to prioritize one flavor of mgf over another, it got tripped up because the arbitrary order of files happened to present the .d before either of the .mgf. So I swapped the loop order within the directory. The directory structure traversal should be unchanged - we still exhaust "," before moving on to "..".  Or that's what I intended anyway.

pwiz_tools/BiblioSpec/src/PepXMLreader.cpp
bspratt: I don't know anything about MSFragger or any permutations of file formats it might employ - perhaps I should hand this off to you as the original implementer of this bit of code?

pwiz_tools/BiblioSpec/src/BuildParser.cpp
chambm: Ah, you're right Brian. You only changed the enumeration of entries in the current directory, not the enumeration of directories. The old way was definitely dependent on the arbitrary order of entries, as you said.

pwiz_tools/BiblioSpec/src/PepXMLreader.cpp
chambm: Sounds good, I'll take over.

PR #1762, opened 2021-09-18
Use <nobr> tags in Command Line documentation HTML instead of non-breaking hyphen &#8209;

nickshulman:
When Skyline is displaying the command-line documentation window ("Help > Documentation > Command Line"), Skyline replaces all of the "-" characters with "&amp;#8209;" (non-breaking hyphen "&#8209;") so that the argument names do not wrap when you look at them in the Documentation Viewer window.

This causes problems if you try to use Ctrl+F to find and you are looking for something like "--refine".
Also, if you select all and copy and paste into Notepad++, those non-breaking hyphens turn into boxes.

This change makes it so that the text remains the same, but those argument names get surrounded with "&lt;nobr>&lt;/nobr>".
image

Here's what a snippet of the HTML looks like:
`<tr><td><nobr>--refine-min-peptides</nobr>=&lt;integer&gt;</td><td>Proteins with fe`

I had to use the &lt;nobr> tag, because using &lt;span style="white-space: nowrap;"> did not work at all.

### DIFF
--- a/pwiz_tools/Skyline/CommandArgs.cs
+++ b/pwiz_tools/Skyline/CommandArgs.cs
@@ -2557,10 +2557,32 @@ namespace pwiz.Skyline
                 // ReSharper restore LocalizableElement
             }
 
+            // Regular expression for an argument: a hyphen surrounded by zero or more word characters
+            // (i.e. letters, numbers or punctuation connector) or hyphens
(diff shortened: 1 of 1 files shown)


brendanx67 (approved):
This has the added benefit that arguments can be selected and copy-pasted into scripts or directly to the command-line, which was also broken by the non-breaking hyphens.

PR #1763, opened 2021-09-18
Change ZedGraphClipboardTest to only use the System clipboard if "Run…

nickshulman:
…PerfTests" is true.

### DIFF
--- a/pwiz_tools/Skyline/TestFunctional/ZedGraphClipboardTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/ZedGraphClipboardTest.cs
@@ -68,40 +68,51 @@ namespace pwiz.SkylineTestFunctional
             ClickContextMenuItem(graphChromatogram.GraphControl, copyMenuText);
             Assert.IsTrue(HasClipboardFormat(DataFormats.Bitmap));
 
+            bool canUseSystemClipboard = RunPerfTests;
             // Switch to using the system clipboard for the rest of the test so that we can test the "Copy Metafile"
             // menu item as well as testing the message when the clipboard is locked.
-            ClipboardEx.UseInternalClipboard(false);
+            // Only use the system clipboard if RunPerfTests is true, because tests using the system clipboard can be annoying.
+            if (canUseSystemClipboard)
+            {
+                ClipboardEx.UseInternalClipboard(false);
+            }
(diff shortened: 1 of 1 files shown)


chambm:
I don't understand how but it looks like this PR broke the Windows x86 build. The ConsoleImportPeptideSearchTest is failing (which also triggers the uninstall test failure but that's derivative of any TestData test failing):
https://teamcity.labkey.org/viewLog.html?buildId=1543862&tab=buildResultsDiv&buildTypeId=bt19

I just reran an older PR (1739) that hasn't been updated to the latest master since this PR was merged and it still passes:
https://teamcity.labkey.org/viewLog.html?buildId=1545347&tab=queuedBuildOverviewTab

chambm:
Ping @nickshulman (not sure if you got a notification because the PR is merged)

nickshulman:
@chambm, isn't everything passing on master now?

chambm:
Well that is mysterious. Let's continue to observe. If it's really this intermittent then it could certainly have been an earlier PR.

chambm:
Still flaky but very mysterious that it's only flaky on TC:
https://teamcity.labkey.org/project.html?buildTypeId=bt19&tab=testDetails&testNameId=-5468656425461629408&projectId=ProteoWizard&order=START_DATE_DESC&page=1

Maybe because the nightly tests don't do x86? Should I just pause the x86 Skyline configs @brendanx67 ?

nickshulman:
What code is writing out the message "Cannot process request because the process (3604) has exited." to the log file? Are we outputting that somewhere in our code, or is the operating system doing this.

If I run the ConsoleImportPeptideSearchTest on my computer about 5,000 times, I can eventually get a NullReferenceException in  "CommandProgressMonitor.Wait". That failure does not look anything like what we are seeing on TeamCity, so I imagine it is not related to this failure.

chambm:
The `Cannot process request because the process (3604) has exited` error is from calling `Process.Kill()` on a process that's already exited. However that's usually delivered with an InvalidOperationException. It seems very likely that that's happening on the BlibBuild call but I can't see how it's swallowing the rest of the exception details but still delivering the message. The original failure, including that message, was before I added the try/finally with an extra `proc.Kill()`, so it shouldn't be that:
https://teamcity.labkey.org/viewLog.html?buildId=1556374&buildTypeId=bt19&tab=buildLog

PR #1764, opened 2021-09-22
Fix DiaUmpire issues (Skyline 21.1)

chambm:
* added missing iteration update for DiaUmpire assigning spectra to windows
* fixed DiaUmpire.GetConfigFromDiaUmpireOutput() to parse VariableWindows userParam (but it turns out it is being truncated in mz5 output so can't really use it in that case; i.e. it'll always look like the windows are different even if they're really the same)
* added msconvert parameters to improve DiaUmpire performance on vendor raw files (`--acceptZeroLengthSpectra --simAsSpectra --combineIonMobilitySpectra --filter "peakPicking true 1-"`)
* added finally block to ProcessRunner as a last ditch effort to kill a process
* added ImportingFromFile sentinel value for tests to check if DIA windows are still being detected
* added TestDiaUmpireWiffFile to test DIA Umpire runs directly on vendor data


pwiz_tools/Shared/Common/SystemUtil/ProcessRunner.cs
```diff
@@ -86,85 +86,94 @@ public class ProcessRunner : IProcessRunner
...
-                sbError.AppendFormat("\r\nCommand-line: {0} {1}\r\nWorking directory: {2}{3}", processPath,
-                // ReSharper restore LocalizableElement
-                    string.Join(" ", proc.StartInfo.Arguments), psi.WorkingDirectory,
-                    stdin != null ? "\r\nStandard input:\r\n" + stdin : "");
-                throw new IOException(sbError.ToString());
-            }
-            // Make to complete the status, if the process succeeded, but never
-            // printed 100% to the console
-            if (percentLast < 100)
+            finally
```
chambm, line 173: I'm not sure why this diff is so ugly but all I did was put the above code inside a try block and added this finally block.

pwiz_tools/Skyline/SettingsUI/EditIsolationSchemeDlg.cs
```diff
@@ -1067,6 +1070,10 @@ private void ImportRangesFromFiles(MsDataFileUri[] dataSources)
             {
                 MessageDlg.ShowWithException(this, TextUtil.LineSeparate(Resources.EditIsolationSchemeDlg_ImportRangesFromFiles_Failed_reading_isolation_scheme_, x.Message), x);
             }
+            finally
+            {
+                ImportingFromFile = false;
+            }
         }
```
chambm, line 1077: @nickshulman Does this seem like a reasonable solution to allow waiting for the isolation scheme to finish importing? I'm not sure whether there's a race condition where the test code can check the sentinel value before it's actually been set to true. I haven't seen it fail yet though.

chambm:
All the build failures are due to the smart build trigger triggering master branch builds instead of release branch. I have fixed the smart build trigger script on Brendan's PR for the release branch but that's not yet merged.

pwiz_tools/Skyline/SettingsUI/EditIsolationSchemeDlg.cs
chambm: @brendanx67 Quick review on this change please. Otherwise this PR is ready to go.

pwiz_tools/Shared/Common/SystemUtil/ProcessRunner.cs
brendanx67: Maybe you changed line endings somehow, or could it be indenting? It looks better if you ignore whitespace changes

pwiz_tools/Skyline/SettingsUI/EditIsolationSchemeDlg.cs
brendanx67: You should remove this hack and instead add a WaitForConditionUI(() => dlg.GetIsolationWindows().Count == expectedCount). Much better to check the UI for what you are expecting to see from this file import than to add a hack like this just to know when it completes.

brendanx67 (changes requested):
Thanks for writing the new test. I suggested a change in approach to your waiting for the isolation scheme to import as expected.

pwiz_tools/Shared/Common/SystemUtil/ProcessRunner.cs
chambm: Ah indeed. I thought I had ignore whitespace enabled but maybe it resets.

pwiz_tools/Skyline/SettingsUI/EditIsolationSchemeDlg.cs
chambm: Unless I'm mistaken, couldn't your proposal keep on adding windows after it reaches the expected count (if the expected count is wrong) and lead to the test continuing while the dialog is still loading? However, you are of course right that my approach doesn't test the results of the loading: that was by design because it wasn't the point of the test, but it certainly won't hurt to check it anyway.

nickshulman: @chambm The method "AddIsolationScheme" will not return until after the dialog has been closed, so one thing you could do is set a boolean at the end of the action that you pass into ShowDialog:
```
            bool finishedReading = false;
            var isolationScheme =
                ShowDialog<EditIsolationSchemeDlg>(()=>
                {
                    importPeptideSearchDlg.FullScanSettingsControl.AddIsolationScheme();
                    finishedReading = true; // <<<<<<<<<<
                });
            RunUI(() =>
            {
                isolationScheme.IsolationSchemeName = _instrumentValues.IsolationSchemeName;
                isolationScheme.UseResults = false;
            });
            RunDlg<OpenDataSourceDialog>(isolationScheme.ImportRanges, importRangesDlg =>
            {
                importRangesDlg.CurrentDirectory = new MsDataFilePath(Path.GetDirectoryName(searchFiles[0]));
                importRangesDlg.SelectFile(Path.GetFileName(DiaFiles[0]));
                importRangesDlg.Open();
            });

            WaitForConditionUI(()=>finishedReading);
```

I have never actually written test code like this, but I think it would work.
(But, Brendan's way works too)

chambm: @nickshulman How/when would the dialog be closed?

nickshulman:
Oops. My mistake. It's actually the call to "ImportRanges" which will return after it has finished reading the raw file.
So, the code I should have suggested is:
```
            bool finishedImporting = false;
            RunDlg<OpenDataSourceDialog>(()=>
            {
                isolationScheme.ImportRanges();
                finishedImporting = true; // <<<<<<<<<<<<
            }, importRangesDlg =>
            {
                importRangesDlg.CurrentDirectory = new MsDataFilePath(Path.GetDirectoryName(searchFiles[0]));
                importRangesDlg.SelectFile(Path.GetFileName(DiaFiles[0]));
                importRangesDlg.Open();
            });

            WaitForConditionUI(() => finishedImporting);
```

(but I might still have made a mistake somewhere)

pwiz_tools/Skyline/SettingsUI/EditIsolationSchemeDlg.cs
brendanx67: @chambm You are mistaken. The wait is on the UI thread, which means that it cannot interrupt the code that fills the grid. For anything that needs to run on the UI thread, that entire function becomes atomic. I still prefer adding a meaningful test that proves the code is functioning in something like the way you expect.

brendanx67: I also think it is more readable and understandable than Nick's "clever" solution.

chambm: Ah, so in order for any further *UI() code to run, the import has to finish? I can see that now. WaitForCondition does some looping to keep testing the predicate every 100ms, which is what threw me for a loop. So in that case can't we just assert once about the expected count with RunUI? And if we don't care about the count[1], simply proceed to `OkDialog(isolationScheme, isolationScheme.OkDialog);`? 

1. Not that it's anything but trivial for me to get the expected count here and add the assert, I just hadn't focused on that.

chambm: Huh. If I just proceed to OkDialog it doesn't wait on import and gives me the "Specify Start and End values for at least one isolation window." error. But OkDialog is doing a RunUI call.

chambm: Just trying to understand how this UI test code works. I added `WaitForConditionUI(() => 1 < isolationScheme.GetIsolationWindows().Count);` before `OkDialog(isolationScheme, isolationScheme.OkDialog)` and I changed `public static bool WaitForConditionUI(int millis, Func<bool> func, Func<string> timeoutMessage = null, bool failOnTimeout = true, bool throwOnProgramException = true)` so that it writes the number of waits needed to console when the predicate goes true. Turns out it waits 4-5 times before it gets the first range, and only needs to wait 5 times to get all 28 ranges. So once it starts importing then it's blocking, but there is time before that are, I guess, UI interruption points or something, where the import hasn't started?

brendanx67: RunDlg<OpenDataSourceDialog>(isolationScheme.ImportRanges, importRangesDlg =>
            {
                importRangesDlg.CurrentDirectory = new MsDataFilePath(Path.GetDirectoryName(searchFiles[0]));
                importRangesDlg.SelectFile(Path.GetFileName(DiaFiles[0]));
                importRangesDlg.Open();
            });

// At this point the UI thread is dropping out of GetDataSourcePaths()
// The LongWaitDlg moves reader.Import() to a background thread and pushes a message pump
//     which allows the UI to be responsive to the user while the import is processed, e.g. allowing the "Cancel" button to be clicked
//     and any other RunUI() call get get through the message queue and get run
//     Once the LongWaitDlg is complete the function resumes once it can get onto the UI thread
//     and it will complete the loop that fills the grid entirely on the UI thread without relinquishing it to other code

            WaitForConditionUI(() => 28 == isolationScheme.GetIsolationWindows().Count);  // So, you need this wait!
            OkDialog(isolationScheme, isolationScheme.OkDialog);

PR #1765, opened 2021-09-22
Fix a problem processing data with multiple TIMS calibrations - it's …

bspratt:
…not clear to me why the existing code seems to have a problem with memory getting swapped out from under some important cache values when the next calibration cache is added, but pre-allocating all caches seems to fix the problem.

### DIFF
--- a/pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.cpp
+++ b/pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.cpp
@@ -214,9 +214,16 @@ TimsDataImpl::TimsDataImpl(const string& rawpath, bool combineIonMobilitySpectra
         " ORDER BY Id"; // we currently depend on indexing the frames_ vector by Id (which so far has always been sorted by time)
     sqlite::query q(db, querySelect.c_str());
 
-    int maxNumScans = 0;
-    vector<TimsFramePtr> representativeFrameByCalibrationIndex; // the first frame for each calibration index
-    vector<map<double, int>> scanNumberByOneOverK0ByCalibrationIndex;
+    string queryCalibrations = "SELECT MAX(TimsCalibration) FROM Frames";
+    size_t calibrationsCount = sqlite::query(db, queryCalibrations.c_str()).begin()->get<sqlite3_int64>(0);
+    vector<map<double, int>> scanNumberByOneOverK0ByCalibrationIndex(calibrationsCount);
+    oneOverK0ByScanNumberByCalibration_.resize(calibrationsCount);
+
+    string queryNumScans = "SELECT MAX(NumScans) FROM Frames";
+    size_t maxNumScans = sqlite::query(db, queryNumScans.c_str()).begin()->get<sqlite3_int64>(0);
+    vector<double> scanNumbers(maxNumScans + 1);
+    for (int i = 0; i <= maxNumScans; ++i)
+        scanNumbers[i] = i;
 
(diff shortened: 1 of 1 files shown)


chambm:
Did you get this to break in a debug build? I'm trying to break when it crashes but so far no luck with unoptimized debug builds. Trying optimized now.

bspratt:
Yes, in a debug build. Where, presumably, it intentionally makes more mischief with memory relocation at reallocation etc.

bspratt:
Though I've just discovered that this change messes up the Skyline test BrukerPrmPasefImportTest so I'm not out of the woods yet,

chambm:
Yeah I'm not sure any of our non-perf tests would actually exercise this change except for the first calibration index.

bspratt:
Poking around briefly I didn't find any examples of multiple calibration in our test data - do you know of any?

chambm:
Hmm, the BrukerPrmPasefImportTest test file doesn't have multiple calibrations? If not then I'm really curious why it would fail but the unit test data wouldn't (which should be checking that the arrays of 1/k0 values are the same with the reference mzMLs).

bspratt:
It doesn't have multiple calibrations, no. It does use the combined ion mobility mode though, which is maybe a difference. This is still a debug build I'm talking about, too.

chambm:
It was crashing because it's keeping a const ref to the vector passed into the frame ctor. Resizing invalidates references and iterators if capacity has to be increased. So I think your approach should work although I refactored it back so it only reserves the capacity ahead of time instead of rearranging the loops. Or were you able to fix the perf test just by adding that intensities.empty() check?

bspratt:
I guess your change is smaller than mine, which is generally a good thing, but I think mine results in simpler code, which is also desirable. Why reserve+resize when you can just set the size? And I don't understand why maxNumScans shouldn't just be set rather than developing it along the way (actually I worry that this will be a problem if we ever encounter a file with different sections have different numbers of scans).

chambm:
Yes I agree just resizing from the get-go makes more sense. But I don't think it's simpler to move the after loop into the frame loop.

bspratt:
To my mind it's simpler (and thus easier to understand/maintain), especially in that it eliminates the vector<TimsFramePtr> representativeFrameByCalibrationIndex(calibrationsCount); business. It's just less code, which is generally a good thing.

chambm:
It is a bit simpler to not have representativeFrameByCalibrationIndex, but putting that extra loop (even if it's only run once per calibration) inside the frame loop looks like it could cause branch prediction issues. That's just a hypothesis though, I don't have any profiling to back it up.

Does this code still fail BrukerPrmPasefImportTest though?

bspratt:
>> branch prediction issues
I don't think performance is a huge concern here. And as we're already asking the question "is this a new calibration?" I can't see any difference in cost.

>> Does this code still fail BrukerPrmPasefImportTest though?
The code as I left it passes everything on my system. It's not clear to me, though, why I had to add that empty check where we didn't seem to need it before, but I guess we were already kind of getting lucky so maybe not a total surprise that other side effects crop up. 

Or are you asking about your revised code passing?

bspratt:
The code as currently checked in also passes the perf tests on my system with no problems. I'm not that bothered if you want to leave it the way you've got it currently set.

chambm (approved):
I do prefer keeping the extra loop outside but I agree it's almost certainly a negligible performance difference. Sounds like this is good to go.

bspratt:
Sounds good. I'll just let it get through all the TC checks then merge to master.

PR #1767, opened 2021-09-24
Prevent the "ThreadAbortException" from being written to the console …

nickshulman:
…when running the Topograph unit tests.

Also, prevent the message about security debugging when launching Topograph.


nickshulman:
This is more tricky to fix than I expected. I will keep trying.

PR #1770, opened 2021-09-28
Make TestExportMethodShimadzu less strict

kaipot:


chambm:
Is there any testing done as to whether the exported binary file is valid?

kaipot:
> Is there any testing done as to whether the exported binary file is valid?

Not at the moment, I can't think of any way to do that. Maybe pass the created method back into the library and use it as a template file? But if the library creates a method without throwing an error and the file is about the right size, that might be good enough. Checking that the file is exactly N bytes doesn't seem like a great way to check (it changed when we updated the libraries to new versions)

chambm:
But it only changed on Linux/Wine, not on Windows? That's what makes me raise an eyebrow here.

kaipot:
> But it only changed on Linux/Wine, not on Windows? That's what makes me raise an eyebrow here.

It changed on both but yeah the new sizes vary (540,672 bytes with old version; 548,864 / 569,344 bytes with new version on Wine/Windows respectively).

PR #1771, opened 2021-09-28
Fix exception when trying to automatically train mProphet model after decoys removed

kaipot:


nickshulman (commented):
I would think that training a model could fail for a lot of different reasons, so it seems unnecessary to write code to specifically handle the "there are no decoys in the document" case. Have we tested what happens if the model training fails for some other reason?

I do not know of a reliable way to make training a model fail, but I wonder what would happen if the user deleted all but one of the decoys.

kaipot:
> I would think that training a model could fail for a lot of different reasons, so it seems unnecessary to write code to specifically handle the "there are no decoys in the document" case. Have we tested what happens if the model training fails for some other reason?
> 
> I do not know of a reliable way to make training a model fail, but I wonder what would happen if the user deleted all but one of the decoys.

Handling the case of zero decoys might be a little more helpful for users. But now I added a more general case also (your idea of deleting all but one of the decoys works).
It looks like I changed a bunch of code in MProphetScoringModel.cs, but I just wrapped the code in a try/finally so that we can complete the progress monitor even if an exception happens (otherwise it gets stuck on the status bar).

brendanx67:
I tend to prefer making the current function a sub-function and putting the try {} finally in the calling function, rather than adding indentation to a large block of code.

PR #1773, opened 2021-09-29
Upgrading our test infrastructure to the current MSTest V2 framework.

bspratt:
Still some issues here (with ABi reader dll in particular) but curious to get this into TeamCity and see what else breaks.


chambm:
Did you forget to add `packages.config`?

chambm:
IDPicker uses NuGet packages.config to install:
```
  <package id="Castle.Core" version="3.3.3" targetFramework="net451" />
  <package id="Excel-DNA.Interop" version="14.0.1" targetFramework="net451" />
  <package id="TestStack.White" version="0.13.3" targetFramework="net40" />
```

There's probably quite a few packages we could replace in Shared/Lib with NuGet references. Would be a bit nice to (slightly) reduce repository size but probabaly no reason to do it other than that.

bspratt:
So those are sufficient for a build to go out and find the necessary package, and install it?

chambm:
Yeah, although I think I had to check in the .nuget directory as well. Maybe that's not necessary anymore, although I recently had to update the nuget.exe because of the TLS version changes.

bspratt:
> Yeah, although I think I had to check in the .nuget directory as well. Maybe that's not necessary anymore, although I recently had to update the nuget.exe because of the TLS version changes.

I see where idPicker does indeed have a ,nuget directory with a nuget.exe in it, but so far I don't see how to use that on the Skyline side. I copied that over, even put a copy in each subproject needing MSTest, but still get the error message advising a manual update. What am I missing?

chambm:
This is what IDPicker.csproj has at the end of the file to do the NuGet restore:
```
<Import Project="$(SolutionDir)\.nuget\NuGet.targets" Condition="Exists('$(SolutionDir)\.nuget\NuGet.targets')" />
  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
    <PropertyGroup>
      <ErrorText>This project references NuGet package(s) that are missing on this computer. Enable NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
    </PropertyGroup>
    <Error Condition="!Exists('$(SolutionDir)\.nuget\NuGet.targets')" Text="$([System.String]::Format('$(ErrorText)', '$(SolutionDir)\.nuget\NuGet.targets'))" />
  </Target>
```

I don't know whether one approach is better or not. :)

bspratt:
I want to keep this work, so don't delete branch

PR #1774, opened 2021-09-29
Support CVID.MS_scanning_quadrupole_lower_bound_m_z_array

chambm:
* added CVID.MS_scanning_quadrupole_lower_bound_m_z_array as the new standard source of SONAR arrays for combined spectra (although Skyline gets bins instead of m/z)
* added WatersSonarPerfTest to tc-perftests.bat


bspratt (approved):
Looks right from here.

PR #1776, opened 2021-10-01
Avoid loading all of the ChromPeak's from the .skyd file at once

nickshulman:
This change makes it so that the ChromPeak's and Scores do not get read from the .skyd file when you open a document.
Instead, the ChromatogramGroupInfo will read them if they get asked for.

TransitionGroupDocNode.UpdateResults now looks very carefully to decide whether the ChromatogramGroupInfo needs to be looked at, or whether the existing TransitionChromInfo's can be used.


brendanx67:
You should test this with the Bruderer and Selevsek data sets and Skyline Batch.

nickshulman:
> You should test this with the Bruderer and Selevsek data sets and Skyline Batch.

Where is the Selevsek dataset? I don't think I've heard of that one.

Do you have a particular Skyline Batch script that I should test this with?

brendanx67:
What is the memory use difference between the old and the new code with the Bruderer document loaded? It should be a big improvement, right? i.e. after processing while you File > Open that file. Sounds like it is 2 seconds faster to load, but what is the memory usage difference? That is what you set out to improve, right?

nickshulman:
> What is the memory use difference between the old and the new code with the Bruderer document loaded? It should be a big improvement, right? i.e. after processing while you File > Open that file. Sounds like it is 2 seconds faster to load, but what is the memory usage difference? That is what you set out to improve, right?

Yes, the memory improvements are exactly what we would expect. Here's what it looks like without these changes:
image

and here's what it looks like with these changes:
image

PR #1778, opened 2021-10-01
Adding support for previously unseen ionization source in Bruker data…

bspratt:
…. This is based on communication with a customer, we should verify with Bruker.


chambm:
From what I can tell HESI is heated ESI + ACPI ("The Vacuum Insulated Probe Heated ESI source is a dual source with an integrated APCI probe."). So I'm not sure it's accurate to just group it with ESI. Not that it matters much to our workflows. To handle this properly though probably requires a new CV term, or confirmation with the mzML group that multi-type sources should just have a CV param for each type (the same as we do for activation types e.g. CID+ETD).

bspratt:
Just to keep things moving, how about I put a TODO in there in addition to the current simple ESI assignment?

chambm (approved):

PR #1781, opened 2021-10-03
Fix intermittent TeamCity failure in ConsoleImportPeptideSearchTest

nickshulman:
Put try/catch around code that tries to set Process.PriorityClass.
Sometimes, the process has exited by the time that line of code is reached.


brendanx67 (approved):
Looks good.

PR #1785, opened 2021-10-06
Skyline: add a code inspection test to catch forgotten debug code ena…

bspratt:
…bling PerfTests regardless of TestRunner settings


bspratt:
> Is this a real code inspection which will not break the required checks, or a test that will break the required checks?

I don't quite understand the question, but it's a real test in that it failed until I commented out that offending line in TestDiaUmpireWiffFile(). 

The funny thing is that I didn't even expect to find that there - I added this inspection just because I was developing a different perf test and thought "huh, this is really something our custom code inspection ought to be looking out for, same way we do PauseTest()"

chambm:
I meant ReSharper code inspection. But you answered it by saying the test failed. ReSharper inspections aren't tests.

chambm (approved):

bspratt:
Yes, this is our own homegrown inspection that looks for things that we can't conveniently look for with ReSharper.

PR #1786, opened 2021-10-06
Skyline: Improvements for all-ions DIA, in particular for Bruker timsTOF bbCID ("broadband CID")

bspratt:
- Skyline was confused by DDA-like "MassTrigger" value in DIA data
- Skyline was also confused by initial run of (presumed to be all-ions DIA, but actually not) MS1 scans before switching to MS1/MS2 alternation. Now using data header information to anticipate that switch.

Reported by user Michael Witting, who also contributed the data for the new associated test.


pwiz_tools/Skyline/Model/Results/SpectrumFilter.cs
```diff
@@ -101,6 +103,8 @@ public sealed class SpectrumFilter : IFilterInstrumentInfo
                 _isWatersFile = instrumentInfo.IsWatersFile;
                 _isWatersSonar = instrumentInfo.IsWatersSonarData;
                 _configInfoList = instrumentInfo.ConfigInfoList;
+                _fileContentList = instrumentInfo.FileContentList;
+                _sourceHasDeclaredMS2Scans = _fileContentList.Contains(@"MSn spectrum");
```
chambm, line 107: I recommend going with CVID so if the term name ever changes in the future we won't need to support multiple strings (the old string for existing mzMLs and the new string for vendor files and new mzMLs).

pwiz_tools/Skyline/TestPerf/PerfImportBrukerPasefBbCIDTest.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// <summary>
+    /// Verify consistent import of Bruker bbCID data.
+    /// </summary>
+    [TestClass]
+    public class PerfImportBrukerPasefBBCIDTest : AbstractFunctionalTestEx
+    {
+        [TestMethod]
+        public void BrukerAllIonsbbCIDImportTest()
+        { 
+RunPerfTests = true; // Uncomment this to force test to run in IDE
```
nickshulman, line 39: Don't forget to comment out this line before you merge.

chambm (changes requested):
How is the header information being used here to anticipate the switch from MS1 only to MS1/MS2 alternation? Is there an actual difference between the MS1s that are "all-ion" and the ones that aren't?

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/Model/Results/SpectrumFilter.cs
bspratt: Makes sense. I was following a pattern I'd seen elsewhere but I think you're right.

bspratt:
> How is the header information being used here to anticipate the switch from MS1 only to MS1/MS2 alternation? Is there an actual difference between the MS1s that are "all-ion" and the ones that aren't?

In some formats, "all ions" is represented as MS1 only with alternating hi/low CE, in others it's alternating MS1/MS2. If the user tells us it's all-ions data, and the first thing we see a run of MS1, we assume it's the former. But in this file, it's the latter - the actual all-ions MS1/MS2 scans start after a bunch of MS1-only scans. So, we need to know that MS2 data will be forthcoming, and that header info is a cheap and convenient way to do that.

chambm:
> 
> In some formats, "all ions" is represented as MS1 only with alternating hi/low CE
You mean other than Bruker? Or Bruker can have both schemes?

pwiz_tools/Skyline/Model/Results/SpectrumFilter.cs
chambm: Whoops I didn't notice this file was outside MsDataFileImpl. Hmm. Maybe a dedicated property for HasDeclaredMSn?

bspratt: Yeah, probably a more robust and succinct approach.

chambm: Nevermind. It's very unlikely this particular term will change name, and we could always switch to a property later if it does.

bspratt: Oops, too late. I like your way better, anyway.

PR #1788, opened 2021-10-07
Fix document schemas

kaipot:


kaipot:
Skyline_21.11.xsd didn't exist, and synchronize integration got added as 21.11 when it was supposed to be 21.12.

PR #1789, opened 2021-10-07
Prompt user to generate decoys on Import Results; autotrain peak scoring model after import

kaipot:


pwiz_tools/Skyline/Model/Results/Scoring/AutoTrainManager.cs
```diff
@@ -0,0 +1,176 @@
...
+
+        protected override bool LoadBackground(IDocumentContainer container, SrmDocument document, SrmDocument docCurrent)
+        {
+            SrmDocument docNew;
+            var loadMonitor = new LoadMonitor(this, container, container.Document);
+
+            bool End(string message)
+            {
+                // Show an error message and set the AutoTrain flag to false.
+                IProgressStatus status = new ProgressStatus();
```
nickshulman, line 76: Usually, you need to keep using the same ProgressStatus object all through the process, and pass it to everyone when you also pass the LoadMonitor.
Otherwise, the status bar ends up thinking that some part of the process has never ended.

I guess the way the code is here, each of the steps of this like "Score Features" and "Change Peaks" will appear as a separate task going from 0 to 100%, and nothing bad will happen, so I guess that's ok.

PR #1796, opened 2021-10-14
Skyline: the unzoomed view of an ion mobility heatmap had a fixed lower value of zero…

bspratt:
…, which made for a lot of useless whitespace. With this change, we now use the actual range of values instead for the default view.

So you'd formerly get this view:
image
when this view makes more sense
image

### DIFF
--- a/pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
+++ b/pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
@@ -53,6 +53,7 @@ namespace pwiz.Skyline.Controls.Graphs
         private HeatMapData _heatMapData;
         private double _maxMz;
         private double _maxIntensity;
+        private double _minIonMobility;
         private double _maxIonMobility;
         private bool _zoomXAxis;
         private bool _zoomYAxis;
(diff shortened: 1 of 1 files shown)


pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -876,11 +877,13 @@ private void GetMaxMzIntensity(out double maxMz, out double maxIntensity)
...
         }
 
-        private void GetMaxMobility(out double maxIonMobility)
+        private void GetIonMobilityRange(out double minIonMobility, out double maxIonMobility)
         {
+            minIonMobility = double.MaxValue;
             maxIonMobility = 0;
             foreach (var spectrum in _msDataFileScanHelper.MsDataSpectra)
             {
+                var spectrumMinMobility = Math.Abs(spectrum.MinIonMobility ?? spectrum.IonMobility.Mobility ?? 0);
```
brendanx67, line 886: Seems inconsistent to use zero here and double.MaxValue above. Maybe you should just use zero in both places?

brendanx67 (approved):
Looks good. Though, I question the use of double.MaxValue.

PR #1803, opened 2021-10-19
Fix problem where blank document created from the Start Page sometimes never loads

nickshulman:
If the default settings has an ion mobility library in it, and you use the "Blank Document" button on the Start Page, then Skyline never finishes loading the document.

This happens because of the following:
1. The "Blank Document" ActionBoxControl causes two "SkylineWindow.NewDocument" calls to happen in rapid succession because of the way its event handler was written.
2. IonMobilityLibraryManager.LoadBackground will refuse to load a library if it notices the document has changed:
```
            if (ionMobilityLibrary == null || !ReferenceEquals(document.Id, container.Document.Id))
            {
                // Loading was cancelled or document changed
                EndProcessing(document);
                return false;
            }
```
3. Because of the way that "StateChanged" was implemented, it would not notice that the new document also needed to get loaded.

IrtDbManager and LibraryManager also had similar problems.

### DIFF
--- a/pwiz_tools/BiblioSpec/src/WatersMseReader.cpp
+++ b/pwiz_tools/BiblioSpec/src/WatersMseReader.cpp
@@ -29,6 +29,8 @@
 #include "WatersMseReader.h"
 #include "boost/algorithm/string.hpp"
 #include "boost/filesystem.hpp"
+#include <boost/range/adaptor/map.hpp>
+#include <boost/range/algorithm/copy.hpp>
 #include "pwiz/utility/chemistry/Ion.hpp"
 
 #ifdef USE_WATERS_READER
@@ -619,8 +621,11 @@ void WatersMseReader::parseModString(LineEntry& entry,
                 break;
         }
         if (j == mods_.rend()) {
-            throw BlibException(false, "The modification '%s' on line %d is not recognized.",
-                                i->c_str(), lineNum_);
+            // We support a very limited hardcoded list of understood modifications, go ahead and list them in the error message.
+            // If we ever expand this we may want to stop listing them all, but for now this is reasonably helpful.
+            string modList = boost::algorithm::join(mods_ | boost::adaptors::map_keys, "\", \"");
+            throw BlibException(false, "The modification '%s' on line %d is not recognized. Supported modifications include: \"%s\".",
+                                i->c_str(), lineNum_, modList.c_str());
         }
         // find the position in the sequence
         size_t openBrace = i->rfind('(');


kaipot (approved):
Maybe some of the checks in IrtDbManager.StateChanged can be merged together, but looks good to me.

pwiz_tools/Skyline/Model/IonMobility/IonMobilityLibraryManager.cs
```diff
@@ -50,7 +50,8 @@ public override void ClearCache()
 
         protected override bool StateChanged(SrmDocument document, SrmDocument previous)
         {
-            return !ReferenceEquals(GetIonMobilityLibrary(document), GetIonMobilityLibrary(previous));
+            return !ReferenceEquals(GetIonMobilityLibrary(document), GetIonMobilityLibrary(previous))
+                   || !ReferenceEquals(document.Id, previous?.Id);
```
brendanx67, line 54: Should this go up a level into the calling logic that checks for a state change? Do we ever want to consider changing the document ID as potentially not a state change?

pwiz_tools/Skyline/Model/Irt/IrtDbManager.cs
```diff
@@ -66,6 +66,10 @@ protected override bool StateChanged(SrmDocument document, SrmDocument previous)
             {
                 return true;
             }
+            if (!ReferenceEquals(document.Id, previous.Id))
```
brendanx67, line 69: I am again not sure we want to force every implementation of BackgroundLoader to check for the document ID changing. This should be a very fast check. So probably not necessary to protect in the state change logic of the implementation. Better to just check it first in the code that calls StateChanged()

pwiz_tools/Skyline/Model/Lib/Library.cs
```diff
@@ -69,8 +69,9 @@ public override void ClearCache()
         protected override bool StateChanged(SrmDocument document, SrmDocument previous)
         {
             return previous == null ||
-                !ReferenceEquals(document.Settings.PeptideSettings.Libraries, previous.Settings.PeptideSettings.Libraries) ||
-                !ReferenceEquals(document.Settings.MeasuredResults, previous.Settings.MeasuredResults);
+                   !ReferenceEquals(document.Settings.PeptideSettings.Libraries, previous.Settings.PeptideSettings.Libraries) ||
+                   !ReferenceEquals(document.Settings.MeasuredResults, previous.Settings.MeasuredResults) ||
+                   !ReferenceEquals(document.Id, previous.Id);
```
brendanx67, line 74: More proof that this check is needed always.

PR #1818, opened 2021-10-28
The import peptide search wizard dialog is taller than we'd like when IMS is in play…

bspratt:
…, so go for a more careful arrangement of the Full Scan Settings tab.

We save some real estate when IMS values are found in the spectral library by no longer asking the user if they want to use them (of course they do) although we also now will accept 0 as a valid resolving power meaning "no filtering" (which actually makes sense - divide by 0 gives an infinitely wide window)

image


brendanx67 (changes requested):
Please make the spacing consistent with the space between MS1 and MS2. This should pack things in even a bit more. All of the spacing between MS1, MS2, high-selectivity, IMS, RT filtering should be consistent.

bspratt:
> Please make the spacing consistent with the space between MS1 and MS2. This should pack things in even a bit more. All of the spacing between MS1, MS2, high-selectivity, IMS, RT filtering should be consistent.

image

bspratt:
> What happened to the checkbox? I think it needs to stay and be part of the consistent spacing.

Oops, thanks for noticing. Accidentally left out of the formatting due to being disabled, not what we meant.

bspratt:
image

brendanx67 (approved):

brendanx67:
Looks right to me.

bspratt:
I suppose there are tutorial screenshots that should be updated?

brendanx67:
I think only the diaPasef tutorial will really need it. The others should not change perceptibly when the IMS group box is hidden.

PR #1820, opened 2021-10-28
per issue #842 "Add a setting for RT annotation display to Chromatogr…

bspratt:
…am graph right-click > properties form", users really want to be able to show more than a single decimal place in the RT annotations on chromatograms. So we've added a control to the chromatogram Properties form for that.

image
image


brendanx67:
Ha! You wrote one of the best examples in TestMs1Tutorial where you query all the chromatogram annotation strings into a single line and test that they do not change. That was a long time ago, though. Review that code now.

bspratt:
Thanks - I knew it was in there somewhere!

brendanx67:
Looks like that test should work. I assume you actually ran it in French?

brendanx67 (changes requested):
Despite how much vertical space it requires, you need to put the label above the control. The way you have it is non-standard. The problem is that any change in label length impacts the control positioning. To get this right you would need to use a flow layout panel, which Nick has done in some cases, but easier just to be standard and avoid localization issues.

bspratt:
> Looks like that test should work. I assume you actually ran it in French?

Mais oui!

bspratt:
> Despite how much vertical space it requires, you need to put the label above the control.

No problem.
image

brendanx67 (approved):

PR #1822, opened 2021-10-29
fix an unhandled exception when attempting to parse an ill-formed adduct description

bspratt:
Came in on Exception web without any data, but I was able to repro by feeding the parser a description like "+M" in a new test step.

### DIFF
--- a/pwiz_tools/Skyline/Util/Adduct.cs
+++ b/pwiz_tools/Skyline/Util/Adduct.cs
@@ -213,7 +213,7 @@ namespace pwiz.Skyline.Util
                 {
                     // No leading + or - : is it because description starts with a label, or because + mode is implied?
                     var limit = input.IndexOfAny(new[] { '+', '-', ']' });
-                    if (limit < 0)
+                    if (limit < posNext)
                     {
                         return null;
                     }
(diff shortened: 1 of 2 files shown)


pwiz_tools/Skyline/Util/Adduct.cs
```diff
@@ -213,7 +213,7 @@ private Adduct(int charge, bool protonated)
                 {
                     // No leading + or - : is it because description starts with a label, or because + mode is implied?
                     var limit = input.IndexOfAny(new[] { '+', '-', ']' });
```
nickshulman, line 215: I think this should maybe be:
var limit = input.IndexOfAny(new[] { '+', '-', ']' }, **posNext**);

You pretty much want to ignore any sign that is before the "M". But, if there is a sign somewhere before the M, and also after the M, you'd want to be able to find the sign after the M, right?

bspratt: Thanks, as usual, for a thoughtful review.
If there's a sign before the M it's not a well formed adduct description, so we'd rather know about that.

nickshulman (approved):

PR #1823, opened 2021-10-29
Fix file associations that were broken by PR#1697.

nickshulman:
Reported by Matt Karasu


brendanx67:
Ouch! How long have these been broken and who broke them?

brendanx67:
Okay. PR created in July, but only merged into the last Skyline-daily. Phew. Good reason to do another Skyline-daily soon.

PR #1828, opened 2021-11-02
Fix ChangePeak parameters for synchronized integration

kaipot:


nickshulman:
It looks like that "syncGeneratedChange" property ends up getting passed in as the parameter for "preserveMissingPeaks"?

Is the bug that you're trying to fix that the user has explicitly removed a peak for a transition, and then when they integrate some other replicate, but they have synchronize integration turned on, those removed transitions get peaks again?

Did anyone complain about this? It's not clear to me why anyone would have a strong opinion on this behavior one way or the other.

nickshulman (approved):
It looks like you are trying to fix the error that people get:
No results found for the precursor XXX in the replicate YYY

This looks like a good enough fix for that exception.
We might need to look into a more complicated fix to handle the scenario where the replicate that the user set the peak in is the one which has fewer chromatograms than the other replicates.

kaipot:
Thanks @nickshulman

PR #1829, opened 2021-11-03
Fix an issue (#841) where a mass-only description of a labeled precursor would come up with the unlabeled mass…

bspratt:
… for the precursor transition


pwiz_tools/Skyline/TestFunctional/PasteMoleculesTest.cs
```diff
@@ -1772,6 +1773,39 @@ void TestFileImportTransitionList(string knownGood)
...
 
+        // Test for an issue where a mass-only description of a labeled precursor would come up with the unlabeled mass for the precursor transition
+        void TestHeavyPrecursorNoFormulas()
+        {
+            var input = 
+                "Molecule List Name,Precursor Name, Precursor m/z,Precursor Charge, Product m/z,Product Charge, Label Type,Explicit Retention Time,Explicit Retention Time Window\n" +
+                "molecules1,2′deoxycitidine,330.1095,-1,330.1095,-1,light,7.7,2\n" +
+                "molecules1,2′deoxycitidine,336.12963,-1,336.12963,-1,heavy,7.7,2\n";
+            var asFile = false;
+            for (var pass = 0; pass < 2; pass++)
```
nickshulman, line 1784: I probably would have written this loop as:
`foreach (bool asFile in new[] {false, true})`

pwiz_tools/Skyline/TestFunctional/PasteMoleculesTest.cs
```diff
@@ -1772,6 +1773,39 @@ void TestFileImportTransitionList(string knownGood)
...
+                    File.WriteAllText(tempFile, input);
+                    RunUI(() => SkylineWindow.ImportMassList(tempFile));
+                }
+                else
+                {
+                    SetClipboardText(input);
+                    RunUI(() => SkylineWindow.Paste());
+                }
+                var pastedDoc = WaitForDocumentChange(docOrig);
+                AssertEx.IsDocumentState(pastedDoc, null, 1, 1, 2, 2);
```
nickshulman, line 1799: It would be good to add the line:
`AssertEx.Serializable(pastedDoc);`
That will make sure that the document can round trip to XML, and could catch any problems if the serialized m/z is incorrect.

nickshulman (approved):

pwiz_tools/Skyline/TestFunctional/PasteMoleculesTest.cs
bspratt: Great stuff, thanks. I really do appreciate your reviews.

PR #1831, opened 2021-11-03
fix for Issue 794 "Unable to deselect (or remove) the Prosit library for library matching in Molecule UI" …

bspratt:
… which refers to https://skyline.ms/announcements/home/support/thread.view?rowId=50798

If we aren't going to show the menu items because we're not in proteomic mode, then make sure they are unchecked and the backing members are set false.


pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -1104,14 +1104,21 @@ void GraphSpectrum.IStateProvider.BuildSpectrumMenu(bool isProteomic, ZedGraphCo
...
+                    prositLibMatchItem.Checked = Settings.Default.Prosit;
+                    menuStrip.Items.Insert(iInsert++, prositLibMatchItem);
+                    mirrorMenuItem.Checked = Settings.Default.LibMatchMirror;
+                    menuStrip.Items.Insert(iInsert++, mirrorMenuItem);
+                    menuStrip.Items.Insert(iInsert++, toolStripSeparator61);
+                }
+            }
+            else
+            {
+                prositLibMatchItem.Checked = _graphSpectrumSettings.Prosit = false;
```
nickshulman, line 1120: It's weird that you're changing the value of "_graphSpectrumSettings.Prosit" (which changes Settings.Default.Prosit).

You would not think that the act of bringing up the right-click menu would change settings. (And also, that if this setting does need to be false, it's not going to be the correct behavior if you wait until the context menu is shown to change it).

Normally, you would make it so that would check whether what it is currently displaying is proteomic, and AND that with Settings.Default.Prosit.

bspratt: Yeah, I take your point. Starting over...

nickshulman (approved):
Looks good to me.

PR #1832, opened 2021-11-04
Fix for issue 777 "Training IMSDB hangs when raw data files are not present

bspratt:
When training an ion mobility database quite a lot of work is done before we even try to access the raw data files. When they turn out to be missing, it's a frustrating user experience. With this change, we check for the raw data before we do anything else.


pwiz_tools/Skyline/TestData/Results/MeasuredDriftValuesTest.cs
```diff
@@ -99,10 +107,27 @@ private void PerformTestMeasuredDriftValues(bool asSmallMolecules)
...
+                            libraryName0, dbPath0));
+                }
+                catch (FileNotFoundException) // If we catch the missing file early, we get this unwrapped exception. If caught deeper in the process it's wrapped in a IOException
+                {
+                    if (provokeError)
+                    {
+                        return; // Expected
+                    }
+                }
+                // ReSharper disable once PossibleNullReferenceException
```
nickshulman, line 130: Since this is a test, you should just do:
Assert.IsNotNull(newIMFiltering);

nickshulman (commented):
The bug description for Issue 777 says that the LongWaitDlg did not go away on its own when the file could not be found.
That makes me think that there is something else going on.
Which file had you downloaded from Panorama where you saw the LongWaitDlg hang?

I think users might already see the "data file could not be found" message coming from "ScanProvider.GetDataFile".

bspratt:
> The bug description for Issue 777 says that the LongWaitDlg did not go away on its own when the file could not be found. That makes me think that there is something else going on. Which file had you downloaded from Panorama where you saw the LongWaitDlg hang?

I wasn't the one reporting the error, so I don't know what file was in use. But my strategy here is to never even put up that LongWaitDlg if we can see that the operation is going to fail ahead of time. It's likely that yes, we'd still have that lingering LongWaitDLg issue if the file was somehow made to go away after the initial confirmation of existence, but I didn't think that worth pursuing.
 
> I think users might already see the "data file could not be found" message coming from "ScanProvider.GetDataFile".

They do with the original code, but only after a fair bit of work has been done. My change performs the file existence check up front so there's less frustration.

pwiz_tools/Skyline/Model/Results/IonMobilityFinder.cs
```diff
@@ -103,6 +103,17 @@ public bool UseHighEnergyOffset
             if (_totalSteps == 0)
                 return measured;
 
+            // Before we do anything else, make sure the raw files are present
+            foreach (var f in fileInfos)
+            {
+                if (!ScanProvider.FileExists(_documentFilePath, f.FilePath))
+                {
+                    throw new FileNotFoundException(TextUtil.LineSeparate(Resources.IonMobilityFinder_ProcessMSLevel_Failed_using_results_to_populate_ion_mobility_library_,
+                    string.Format(Resources.ScanProvider_GetScans_The_data_file__0__could_not_be_found__either_at_its_original_location_or_in_the_document_or_document_parent_folder_,
```
nickshulman, line 112: Something is wrong with the indentation here.

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/Model/Results/IonMobilityFinder.cs
bspratt: fixed, thanks

PR #1835, opened 2021-11-04
Somehow, on Sept 9, Agilent dll's changed from "Include (Auto)" to "P…

nickshulman:
…rerequisite (Auto)"

This changes them to "Include"

### DIFF
--- a/pwiz_tools/Skyline/Skyline.csproj
+++ b/pwiz_tools/Skyline/Skyline.csproj
@@ -5528,16 +5528,6 @@
       <IncludeHash>True</IncludeHash>
       <FileType>File</FileType>
     </PublishFile>
-    <PublishFile Include="OFX.Core.Contracts">
-      <Visible>False</Visible>
-      <Group>
-      </Group>
-      <TargetPath>
-      </TargetPath>
-      <PublishState>Include</PublishState>
-      <IncludeHash>True</IncludeHash>
-      <FileType>Assembly</FileType>
-    </PublishFile>
     <PublishFile Include="ProteomeDb.pdb">
       <Visible>False</Visible>
       <Group>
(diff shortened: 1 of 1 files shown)


bspratt:
Bringing MattC into the conversation, he knows more about this than I do

chambm:
https://docs.microsoft.com/en-us/visualstudio/deployment/how-to-specify-which-files-are-published-by-clickonce?view=vs-2019
> References to assemblies (.dll files) are designated as follows when you add the reference: If Copy Local is False, it is marked by default as a prerequisite assembly (Prerequisite (Auto)) that must be present in the GAC before the application is installed. If Copy Local is True, the assembly is marked by default as an application assembly (Include (Auto)) and will be copied into the application folder at installation. A COM reference will appear in the Application Files dialog box (as an .ocx file) only if its Isolated property is set to True. By default, it will be included.

So I guess something changed with the DLL so it got changed to Copy Local = False. Googling for it I can't find out what that would be though. When I added the test for the administrator installer I didn't add one for the ClickOnce install because it seemed quite complicated. But I can take another look at it.

chambm:
Actually I don't see where BaseDataAccess et al. are referenced at all. Maybe recursively through pwiz_data_cli.

nickshulman:
@chambm There is something suspicious about the references in pwiz_data_cli.dll.
When I open it up in JetBrains dotpeek, a lot of the dll's are repeated with slightly different version numbers:
image

This is what it looked like before that checkin on Sept 9:
image

nickshulman:
I bet MassSpecDataReader.dll has something to with what went wrong.
It is referencing dll's with the 10.0.10305.0 version, but the actual dll's we have are version 10.0.1.10035.
image

chambm:
Is it strongly signed? If not we could probably fix the reference ourselves.

bspratt:
Clearly we need some kind of short term fix like that, but we should really go to Agilent and ask for a proper set of DLLs.

PR #1839, opened 2021-11-09
Fix performance on changing multiple peak bounds

kaipot:


nickshulman (commented):
I think we can speed up this code a lot by using "SrmDocument.BeginDeferSettingsChange" and "SrmDocument.EndDeferSettingsChange" so it won't be necessary to show a LongWaitDlg.

By the way, if you are bringing up a LongWaitDlg (or any UI) inside of a call to "ModifyDocument", you should probably have a lock on "SkylineWindow.GetDocumentChangeLock" which can prevent a background thread from invalidating the work that was done. Otherwise, in this case, the user might see multiple LongWaitDlg's pop up if some background loader changed the document, which might not make sense to them.

nickshulman (approved):

PR #1843, opened 2021-11-12
Fix loading libraries when cache file cannot be written (Issue 844)

nickshulman:
Make it so that libraries can be loaded even if it is not possible to save a cache file in the folder containing the library.

Currently, all I have fixed is BiblioSpec and EncyclopeDIA, but I imagine that I should probably also fix this for Nist, XHunter, SpectraST, and Chromatogram Libraries.


brendanx67 (approved):
Looks good.

PR #1845, opened 2021-11-15
Use ColumnSelectDlg instead of PasteDlg for importing transition lists

henrytsanford:
Edit > Insert > Transition List now brings up a window prompting the user to paste. The user can then assign columns using ColumnSelectDlg.

Also adds the ability to associate peptides with proteins from the background proteome in ColumnSelectDlg. After selecting 'Associate Proteins' the user can view more information about the associated proteins by hovering over the protein name cells.


pwiz_tools/Skyline/TestTutorial/TutorialAuditLogs/en/TestExistingExperimentsTutorial.log
```diff
@@ -17,308 +17,306 @@ Settings > Peptide Settings -- Modifications > Isotope modifications > "heavy" >
...
-YEIISQQPENVSNLSK	928.9790995	1338.686899	YKL035W	UGP1 SGDID:S000001518, Chr XI from 369534-371033, Verified ORF, "UDP-glucose pyrophosphorylase (UGPase), catalyses the reversible formation of UDP-Glc from glucose 1-phosphate and UTP, involved in a wide variety of metabolic pathways, expression modulated by Pho85p through Pho4p"
-YEIISQQPENVSNLSK	924.972	1115.5802	YKL035W	UGP1 SGDID:S000001518, Chr XI from 369534-371033, Verified ORF, "UDP-glucose pyrophosphorylase (UGPase), catalyses the reversible formation of UDP-Glc from glucose 1-phosphate and UTP, involved in a wide variety of metabolic pathways, expression modulated by Pho85p through Pho4p"
-YEIISQQPENVSNLSK	928.9790995	1123.594399	YKL035W	UGP1 SGDID:S000001518, Chr XI from 369534-371033, Verified ORF, "UDP-glucose pyrophosphorylase (UGPase), catalyses the reversible formation of UDP-Glc from glucose 1-phosphate and UTP, involved in a wide variety of metabolic pathways, expression modulated by Pho85p through Pho4p"
-				
+Imported a transition list from the clipboard
+Extra Info: TANDVLTIR	501.778	830.474
```
bspratt, line 24: I feel like we've lost some useful information in these logs, namely the header line. At a minimum we should be including the original if there was one, and our understanding of the columns if there wasn't. Probably both would be good, really - what we saw and what we decided it meant.

pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -1909,14 +1909,14 @@ public void ImportMassList(string fileName)
         /// Process and then add the mass list to the document
         /// </summary>
         public void ImportMassList(MassListInputs inputs, string description, bool assayLibrary, 
-            SrmDocument.DOCUMENT_TYPE inputType = SrmDocument.DOCUMENT_TYPE.none) // "None" means "don't know if it's peptides or small molecules, go figure it out".
+            SrmDocument.DOCUMENT_TYPE inputType = SrmDocument.DOCUMENT_TYPE.none, bool forceDlg = false) // "None" means "don't know if it's peptides or small molecules, go figure it out".
```
bspratt, line 1912: Let's go with a full-on comment header here. I feel like forceDlg needs as much explanation as inputType does.

pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -1950,14 +1950,14 @@ public void ImportMassList(string fileName)
                 // Try importing that list to check for errors
                 docCurrent = docCurrent.ImportMassList(input, importer, null,
```
bspratt, line 1951: Seems like this check could be skipped if forceDlg is true

pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -1950,14 +1950,14 @@ public void ImportMassList(string fileName)
...
-                    out librarySpectra, out testErrorList, out peptideGroups);
+                    out librarySpectra, out testErrorList, out peptideGroups, null, SrmDocument.DOCUMENT_TYPE.none, hasHeaders);
                 if (!testErrorList.Any())
                 {
                     useColSelectDlg = false;
                 }
             }
 
-            if (useColSelectDlg)
+            if (useColSelectDlg || forceDlg)
```
bspratt, line 1960: I think at this point the code could be simplified by ORing useColSelectDlg and forceDlg?

brendanx67:
The goal of the audit logs is to allow a person to reproduce the same
outcome with a new document. So, yes, you need to include the full text of
a paste, and how the columns got interpreted.

PR #1846, opened 2021-11-15
Another synchronized integration exception fix

kaipot:


nickshulman:
I wonder whether, instead of loading the chromatogram, we should instead look at the TransitionGroupDocNode.Results[replicateIndex] in order to figure out which ChromFileInfoId's have results.
So, the code to do that would look like:
```
....
            if (nodeGroup == null)
                throw new IdentityNotFoundException(groupId);
            if (nodeGroup.Results == null)
            {
                yield break;
            }

            foreach (var chromSet in syncTargets)
            {
                var replicateIndex =
                    document.Settings.MeasuredResults.Chromatograms.IndexOf(c => ReferenceEquals(c, chromSet));
                if (replicateIndex < 0 || replicateIndex >= nodeGroup.Results.Count)
                {
                    continue;
                }

                var replicateResults = nodeGroup.Results[replicateIndex];
                foreach (var info in chromSet.MSDataFileInfos)
                {
                    if (ReferenceEquals(thisChromSet, chromSet) && ReferenceEquals(thisFile, info.FileId))
                    {
                        continue;
                    }

                    if (!replicateResults.Any(transitionGroupChromInfo =>
                        ReferenceEquals(transitionGroupChromInfo?.FileId, info.FileId)))
                    {
                        continue;
                    }
                    var start = thisStart;
....
```

I can't think of a strong reason whether it might be better to figure out which files apply by looking in the .skyd file (like you have it in this pull request) versus looking in the Results&lt;TransitionGroupChromInfo> (like I have it above). Let's talk about it tomorrow.

nickshulman (approved):
Your way looks better than what I was proposing because your way is using the same logic as "SrmDocument.ChangePeakBounds" uses.

PR #1847, opened 2021-11-16
Distinguish between 'no decoys' and 'no decoy chromatograms' on auto-train model error

kaipot:


kaipot:
We should also not automatically train model if RT regression fails, but I will make a separate PR for that.

pwiz_tools/Skyline/Model/Results/Scoring/AutoTrainManager.cs
```diff
@@ -101,12 +107,9 @@ bool End(string message)
             // Get scores for target and decoy groups.
             targetDecoyGenerator.GetTransitionGroups(out var targetTransitionGroups, out var decoyTransitionGroups);
             if (!decoyTransitionGroups.Any())
```
nickshulman, line 109: I wonder if you should also give a different message if targetTransitionGroups is also empty.
Like, it's a tiny bit confusing to be saying "none of the decoys have chromatograms" if the real problem is that none of the targets have chromatograms either. (In the scenario we are thinking of, the only peptides with any chromatograms will be the iRT standards, and standards are not included in either targets or decoys).

nickshulman (approved):
Looks good to me.

PR #1849, opened 2021-11-18
Add 'Align to' option in Synchronized Integration dialog

kaipot:


nickshulman:
I am not sure whether it is really necessary to ask the user which result file to align to when changing peak boundaries.
The most logical thing to do is to align to the result file in which the peak boundary was performed.
So, all you need to have is a checkbox that says "Align Times". (But, then, you also need to ask the user whether to align things using iRT).

brendanx67:
At this point it is just a shortcut to doing what we were allowing before,
which is only aligning when the x-axis was set to align, which seemed
pretty hard to find. So, I suggested adding this as a way of advertising
it. It just performs the matching menu action to set the x-axis to the
aligned values.

Your idea sounds cool. I think the all vs all alignment case is less common
that iRT, though.

Let’s not change it right now. Let’s leave it just and advertisement for
what was already implemented.

nickshulman (approved):
Looks good enough to me for code complete.

In the future I would recommend decoupling the synchronizing of peak boundaries from the way that they are displayed in the chromatogram window. The user really should only need to be asked whether they want to align peak boundaries using iRT or MS/MS IDs.

brendanx67:
I am not sure I totally agree with you on that, Nick. I sort of do, but I
also wonder how confusing it could be to people who are less abstract than
you (pretty much 99% of the population). With the way it is, we are
guaranteed that the user is currently visualizing the data in a way that
makes sense for the synchronization, and which provides viewing
compatibility with synchronized zooming.

Your way makes the best sense in the abstract but could be highly confusing
to a user less comfortable with the nuance of the abstraction.

PR #1853, opened 2021-11-22
Dotp cutoff highlight

rita-gwen:
Highlight the dotp points that are below a user-defined cutoff value.


pwiz_tools/Skyline/Controls/Graphs/AreaReplicateGraphPane.cs
```diff
@@ -1,1473 +1,1518 @@
...
+            UpdateAxes(resetAxes, aggregateOp, dataScalingOption, normalizeOption);
+
+            if (Settings.Default.PeakAreaDotpCutoffShow && DotProductDisplayOption.line.IsSet(Settings.Default) && _dotpData != null)
+            {
+                var cutoff = Settings.Default.PeakAreaDotpCutoffValue;
+                var highlightValues = new PointPairList(graphData.DotpData.Select(point => point.Y < cutoff ? point : new PointPair(){X = point.X, Y = float.NaN}).ToList());
+                var cutoffHighlightLine = new LineItem("", highlightValues, Color.DimGray, SymbolType.Circle)
+                {
+                    IsY2Axis = true,
+                    Line = new Line() { Color = Color.Transparent},
```
nickshulman: What is the effect of this line of code? Is it overwriting the "Color.DimGray" that was specified three lines above this?

pwiz_tools/Skyline/Controls/Graphs/SummaryBarGraphPaneBase.cs
```diff
@@ -36,6 +36,41 @@ public abstract class SummaryBarGraphPaneBase : SummaryGraphPane
...
+                }
+                public new void Add(CurveItem curve)
+                {
+                    //all targets must be on the same axis
+                    if (Count > 0)
+                        Assume.AreEqual(base[0].GetYAxis(_parent), curve.GetYAxis(_parent), @"All target curves for a tooltip must be on the same axis.");
+                    base.Add(curve);
+                }
+
+                public CurveItem ClearAndAdd(CurveItem curve)
```
nickshulman, line 55: Usually, you clear the list at the very start of the method, and then add stuff later on.
For this reason, "Clear()" and "Add()" are usually separate methods that you call at different times.
A method "ClearAndAdd" would only make sense if you really wanted the behavior to be "If there end up being no curves, then display the previous tooltip text".

pwiz_tools/Skyline/Controls/Graphs/GraphHelper.cs
```diff
@@ -599,6 +600,13 @@ public static double GetMaxY(CurveList curveList, GraphPane g)
             }
             return maxY;
         }
+        public static Color Blend(Color baseColor, Color blendColor, double blendAmount)
+        {
+            return Color.FromArgb(
+                (int)(baseColor.R * (1 - blendAmount) + blendColor.R * blendAmount),
```
nickshulman, line 606: I guess this code has always been this way, but when I was Googling how to blend colors, Stack Overflow said that you should square the components before averaging them. This is because 254 is four times as bright as 127, so, if your two colors were (254, 0, 0) and (0, 254, 0) and you blended them, you would end up with (127, 127, 0) which is a less bright total color than either of the colors you were taking the average of.

nickshulman (commented):
Here are a couple of things that I noticed in the code.

I would recommend adding some screenshots to the description of the pull request. It's a great way to increase the amount of feedback you get.

pwiz_tools/Skyline/Controls/Graphs/AreaReplicateGraphPane.cs
```diff
@@ -1,1473 +1,1518 @@
...
+        }
+
+        public override ImmutableList<float> GetToolTipDataSeries()
+        {
+            return _dotpData;
+        }
+        public override void PopulateTooltip(int index)
+        {
+            ToolTip.ClearData();
+            ToolTip.AddLine(Resources.SummaryReplicateGraphPane_SummaryReplicateGraphPane_Replicate + @":", XAxis.Scale.TextLabels[index]);
```
brendanx67: Colons need to go into the RESX text. Japanese and Chinese use a different colon symbol.

pwiz_tools/Skyline/Controls/Graphs/AreaReplicateGraphPane.cs
```diff
@@ -1,1473 +1,1518 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Drawing.Drawing2D;
+using System.Globalization;
```
brendanx67: It appears you change the line endings in this file. The diffs are only shown appropriately when you hide whitespace changes.

pwiz_tools/Skyline/Controls/Graphs/AreaReplicateGraphPane.cs
```diff
@@ -1,1473 +1,1518 @@
...
+            else
+            {
+                Y2Axis.IsVisible = false;
+            }
+
+            UpdateAxes(resetAxes, aggregateOp, dataScalingOption, normalizeOption);
+
+            if (Settings.Default.PeakAreaDotpCutoffShow && DotProductDisplayOption.line.IsSet(Settings.Default) && _dotpData != null)
+            {
+                var cutoff = Settings.Default.PeakAreaDotpCutoffValue;
```
brendanx67: Realistically, the user may want a different cutoff for each of the 3 different types: dotp, idotp, rdotp. Their reliability of the target values are different in the different cases.

brendanx67 (changes requested):
Some minor comments, and the layout of the properties form needs to be fixed.

brendanx67:
Here is a screenshot of the Properties form which needs to be fixed.

image

Let's talk about the control UI for this feature on Zoom.

pwiz_tools/Skyline/Controls/Graphs/AreaReplicateGraphPane.cs
rita-gwen: In my experience the color specified in the constructor arguments is ignored, but that argument is required, so I had to redefine it in the attribute definitions.

pwiz_tools/Skyline/Controls/Graphs/SummaryBarGraphPaneBase.cs
rita-gwen: Well, if there are no curves then there is no tooltips shown. The intention was to make the client code a bit more concise, since the typical pattern is to make sure no old curve are stuck in the collection and then add a new one, one right after the other.

rita-gwen:
The new layout of the properties dialog:
image

pwiz_tools/Skyline/Controls/Graphs/AreaReplicateGraphPane.cs
rita-gwen: Fixed

pwiz_tools/Skyline/Controls/Graphs/AreaReplicateGraphPane.cs
rita-gwen: Looks like it happened somehow during this merge from master: https://github.com/ProteoWizard/pwiz/pull/1853/commits/c0155f6ea9999f4080bf0938c37f5d6008ac8502
Current version of the file has consistent Windows style line endings (CR/LF)

rita-gwen:
Fixes as discussed at the last UI review:
- removed unnecessary "dotp" inside the group box
- fixed the property assignment
- disable the checkbox and the grid if "none" or "labels" are selected
- disable the grid label to show that the grid is disabled.
- added the Alt-shortcuts
- added the dotp and idotp to the test

PR #1855, opened 2021-11-23
Fix bug with synchronized integration not adjusting other precursors

kaipot:


pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -2159,44 +2159,57 @@ public SrmDocument ChangePeakBounds(SrmDocument document, IEnumerable<ChangedPea
                 document = document.BeginDeferSettingsChanges();
             }
 
-            var changedGroupIds = new HashSet<IdentityPath>();
-            var peptideChanges = new Dictionary<IdentityPath, ChangedPeakBoundsEventArgs>();
+            var changedGroupIds = new HashSet<Tuple<IdentityPath, string>>();
```
nickshulman, line 2162: I think this should be HashSet&lt;Tuple&lt;IdentityPath, MsDataFileUri>>
(and the Dictionary on the next line should also use MsDataFileUri instead of string)
You can trust the "Equals" and "GetHashCode" methods on MsDataFileUri to do the right thing-- no reason to convert to string.

nickshulman:
This looks like it would work.

I am not sure I understand why there are two separate "foreach" loops in ChangePeakBounds: one that loops over "changesArr" and which adds things to "peptideChanges", and then another loop which runs afterwards and deals with the things in peptideChanges. I would have thought that it would make more sense to do all of the things associated with the things in "changesArr" at one time, without needing to add to another list.

(but this is the way the code already is in master, so I guess it's fine)

pwiz_tools/Skyline/SkylineGraphs.cs
kaipot: I changed it - originally I did consider that, but MsDataFileUri doesn't have Equals or GetHashCode. So it's not guaranteed that inherited classes will override the default implementation is it?

nickshulman (approved):

PR #1857, opened 2021-11-23
Add shift-click to Synchronized Integration dialog and preserve selections when switching 'Group By'

kaipot:


pwiz_tools/Skyline/EditUI/SynchronizedIntegrationDlg.cs
```diff
@@ -38,6 +38,10 @@ public partial class SynchronizedIntegrationDlg : Form
         private readonly bool _originalAlignRtPrediction;
         private readonly ChromFileInfoId _originalAlignFile;
 
+        private int _idxLastSelected = -1;
+
+        private Tuple<string, HashSet<string>> _memory = Tuple.Create((string)null, new HashSet<string>());
```
nickshulman, line 43: I would recommend making this a
Tuple&lt;ReplicateValue, IColllection&lt;object>>
You only need to convert things to strings if you need to persist them in Settings or something. If they only need to live for the life of dialog, you can keep everything as objects.

You can use "null" for the ReplicateValue for when they have not chosen a property.

pwiz_tools/Skyline/EditUI/SynchronizedIntegrationDlg.cs
```diff
@@ -134,16 +138,64 @@ private void SelectAlignOption()
 
         private void comboGroupBy_SelectedIndexChanged(object sender, EventArgs e)
         {
-            var newItems = SelectedGroupBy.GetItems(Document, new AnnotationCalculator(Document)).ToArray();
-            if (!ArrayUtil.EqualsDeep(listSync.Items.Cast<object>().ToArray(), newItems))
+            var annotationCalc = new AnnotationCalculator(Document);
+            var newItems = SelectedGroupBy.GetItems(Document, annotationCalc).ToArray();
+            if (ArrayUtil.EqualsDeep(listSync.Items.Cast<object>().ToArray(), newItems))
```
nickshulman, line 143: If you use "Enumerable.SequenceEqual" here you can avoid the "ToArray()" call.

nickshulman (commented):
If you are going to make it so that shift-click extends the selection, I wonder whether you should also change "listSync.CheckOnClick" to "false" so that the user can move the selection without affecting what is selected.

I don't understand why "_memory" is a field in the class, instead of local to the function "comboGroupBy_SelectedIndexChanged". I would have thought that the way to implement this would be right before they choose a different ReplicateValue, you figure out what is currently selected, and then update the CheckedListBox after they have chosen the new ReplicateValue to group on.

kaipot:
> I don't understand why "_memory" is a field in the class, instead of local to the function "comboGroupBy_SelectedIndexChanged". I would have thought that the way to implement this would be right before they choose a different ReplicateValue, you figure out what is currently selected, and then update the CheckedListBox after they have chosen the new ReplicateValue to group on.

I did it this way because if the "group by" option is changed to something where the groups don't match up nicely, the selections can still be preserved if the user changes back.

pwiz_tools/Skyline/EditUI/SynchronizedIntegrationDlg.cs
```diff
@@ -134,16 +138,64 @@ private void SelectAlignOption()
...
+                    {
+                        toCheck.Clear();
+                        break;
+                    }
+                }
+            }
+
+            listSync.ItemCheck -= listSync_ItemCheck;
+
+            cbSelectAll.CheckedChanged -= cbSelectAll_CheckedChanged;
```
nickshulman, line 194: I would recommend that "SetCheckedItems" sets a field called "_changingCheckedItems" and then make it so that both "cbSelectAll_CheckedChanged" and "listSync.ItemCheck" check whether "_changingCheckedItems" is true and if so, return immediately.

This code to remove event handlers and then add them back after you are done is difficult to maintain.

nickshulman (commented):
> I did it this way because if the "group by" option is changed to something where the groups don't match up nicely, the selections can still be preserved if the user changes back.

Ohhhh... That makes sense. And the "_memory" only gets updated when they actually change the selection in the CheckedListBox. Sounds good.

PR #1860, opened 2021-11-27
Fix JRE URL and add property for testing original URLs

chambm:
* fixed wrong JRE download URL
* added Skyline.Program.UseOriginalURLs property for testing real URLs


brendanx67 (approved):
Looks good to me. Maybe use original for Perf test mode? Those are expected to run long and get run far less frequently.

PR #1862, opened 2021-11-29
Work in progress: Create a new project "CommonStandard" which targets .Net Standard instead of .Net Framework

nickshulman:
If you want to make a library which can be used by both .Net Core and .Net Framework, then that library needs to target .Net Standard.

Here I have moved many of the classes from "CommonUtil" (a .Net Framework library) into "CommonStandard" a .Net Standard library.

My goal is that we will eventually have a project called "Skyline Library" which will be able to be used by things like LabKey but I haven't figured out exactly how that will work.


chambm:
What is it in the Common and CommonUtil libraries that make them not able to just switch over to Standard already? For that matter, isn't Standard going away with .NET 5? https://visualstudiomagazine.com/articles/2020/09/16/net-standard-future.aspx

nickshulman:
> What is it in the Common and CommonUtil libraries that make them not able to just switch over to Standard already? For that matter, isn't Standard going away with .NET 5? https://visualstudiomagazine.com/articles/2020/09/16/net-standard-future.aspx

There are a few things in CommonUtil that use System.Windows.Forms (e.g. "FormUtil.cs")

Do you think we will manage to move Skyline to .Net 5 next version? Yes, I guess that would make this irrelevant.

Thanks for the feedback. These are good points.

PR #1863, opened 2021-11-30
temporary fix for performance issues around IM filtering in library viewer - just disable for peptides until we address the actual problem

bspratt:
User noticed slow load time for large peptide library


brendanx67 (approved):
Seems quick and safe

PR #1864, opened 2021-11-30
* updated BilbioSpec to support reading mods with any of MaxQuant's elements

chambm:
Reported by Ziyang, Valerie, and Stephan.


nickshulman (approved):

brendanx67 (changes requested):
Could we maybe keep the old order for the masses we had before so that it is clearer that masses are being added but the previously existing ones are not being changed? I checked this for N, but it is more challenging for others. Interesting that ["Ox"] is changing +1 and ["Oy"] gets the prior value. Did we have it wrong before?

chambm:
It would be trivial to do that rearrangement but this is the order they're given in the MaxQuant source file. It makes more sense to me to keep that order. Do you disagree? I don't think any of the masses changed, but if they did then it was to change them to the latest masses that MaxQuant uses.

And yes, Ox was wrong before (I checked the MaxQuant source and didn't see it change there since 2014).

I see I missed 15N as a synonym though.

brendanx67 (approved):

PR #1865, opened 2021-11-30
Use peptide level q-values when building spectral libraries from Proteome Discoverer results

kaipot:


kaipot:
Dataset: brendanx@nexus:data\MacCoss\MProphet\20210118_Gotti
(pdResult file: library\RD139_Eco48F_UPS200f_HCDOT_WO_Deam.pdResult)

chambm:
Sorry, I used your PR as a guinea pig for this squash-on-demand action. Looks like it only squashes your commits, not the merge, which was the whole idea. Because the new cherry-pick-to-release action I'm trying doesn't work with merges. It's always something... I've undone the squashing for now.

kaipot:
> Add new test files? Why didn't this break existing tests? Assuming they don't have any foreign key constraints, it should be fairly straightforward to trim down any MSF file you've been testing on with careful SQL DELETE commands and VACUUM.

I think existing tests didn't break because I deleted a necessary table in the test inputs (in an attempt to shrink the size as you mentioned), so it falls back to the old behavior. But good point, I will add some new tests.

chambm:
Remember not to mess with the inputs tarball. Just commit your files directly in the inputs directory (even if they're the same filename as one inside the tarball).

chambm:
That ThermoFAIMS failure looks related to the PR. Would you expect the counts to go down?

kaipot:
> That ThermoFAIMS failure looks related to the PR. Would you expect the counts to go down?

Thanks for noticing that. The counts should change since we are filtering on a different value, I'll update that.

chambm:
Your Linux result for that test is only getting 0s for the scores.

kaipot:
> Your Linux result for that test is only getting 0s for the scores.

Hmm, I run Linux on my PC at home and it passes there. Though I have gcc 11.2.1 installed and it looks like TC is using 4.8.4, maybe that has something to do with it.

chambm:
Yeah, that's odd. TC is supposed to be using GCC 5 but I do see where it reports 4.8.4 in the build log. But it's also reporting diagnostics with color and GCC didn't have color until 4.9. So it can't be using 4.8.4 because it would be giving an error with `-fdiagnostics-color`. Do you know how to launch a specific gcc version with Docker? Can you build/debug with gcc:5?

kaipot:
> Yeah, that's odd. TC is supposed to be using GCC 5 but I do see where it reports 4.8.4 in the build log. But it's also reporting diagnostics with color and GCC didn't have color until 4.9. So it can't be using 4.8.4 because it would be giving an error with `-fdiagnostics-color`. Do you know how to launch a specific gcc version with Docker? Can you build/debug with gcc:5?

I'm not familiar enough with Docker but looking at this more I think I might know what the issue is. Seems like abs() only works with ints while std::abs() is overloaded for floating points. Waiting for the TC build to see if that fixed it.

chambm (approved):
Did Brendan want this going to release?

kaipot:
> Did Brendan want this going to release?

Yeah, will the cherry-pick tag work?

chambm:
Now it will (I hope).

PR #1870, opened 2021-12-01
This addresses a performance issue in the new column picker when there are many columns…

bspratt:
… - the standard ComboBox control is slow enough in construction to introduce a noticeable delay in opening the column picker dialog.

The new LazyComboBox control is implemented as a button which creates a dropdown list when clicked. This change makes the TestPasteMolecules test run in 40 seconds instead of 160.

Old implementation looks like this:
image

and here's the new one:
image

Looks like there's a little highlighting to be done to get a closer visual match,


pwiz_tools/Skyline/Util/LazyComboBox.cs
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        public int FindStringExact(string str)
+        {
+            return Items.IndexOf(str);
+        }
+
+        protected override void OnClick(EventArgs e)
+        {
+            if (_listbox == null)
```
brendanx67, line 83: I worry a bit about this ListBox use where the listbox is actually parented inside the form, but apparently above all other controls. I would have used a menu, since they are designed for this. If the form is not tall enough, this implementation seems like it would clip the listbox.

pwiz_tools/Skyline/Util/LazyComboBox.cs
```diff
@@ -0,0 +1,141 @@
...
+ * limitations under the License.
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Windows.Forms;
+using pwiz.Skyline.Properties;
+
+namespace pwiz.Skyline.Util
```
brendanx67, line 26: I think this should go in the Controls subfolder.

pwiz_tools/Skyline/Util/UtilUIExtra.cs
```diff
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Diagnostics;
```
brendanx67, line 2: I don't see how the changes in this file are related. So, maybe they are just left over from originally putting the class in this file.

brendanx67 (changes requested):
Looks like a good start. Did you try simply using a menu for the dropdown part? It seems like a better fit than a ListBox parented inside the form. A menu floats above everything and is designed to do so.

On the look itself, you should tone down the borders to be a similar gray color to the original combo boxes. You'll want to use something from SystemColors so the border ports to whatever the user does with their colors. You should also expand the controls to have only a single pixel boarder between 2 of them, which will require slight overlap. I would also like to see something visiually distinct for the "Ignore Column" case, which should now be relatively easy to achieve. Maybe just use a dark gray text and/or italics.

brendanx67:
How does it look when you click the column nearest the right edge. Do you need to scroll to see the list? I would expect so, which is unusual for a combo box.

pwiz_tools/Skyline/Util/LazyComboBox.cs
bspratt: I'll try a menu, thanks.

pwiz_tools/Skyline/Util/UtilUIExtra.cs
bspratt: Yeah, that's cruft, I will revert.

bspratt:
> 

image

brendanx67:
Maybe. I guess it is parented but not contained in the form. Maybe I should try playing with it. I am still a little concerned that this may cause unexpected issues due to a ListBox not really being designed for this purpose. But maybe it works. Shrug.

brendanx67:
Are you sure that is your ListBox version you just added the image for? If so, it looks remarkably like the combo box version in its borders and dropdown arrowhead.

chambm:
So DropDownList is spawning a menu? It sure does look like a ListBox, but they could be overriding the painting of course.

brendanx67:
And the shadowing under the dropdown.

bspratt:
> And the shadowing under the dropdown.

Hah, no, that was the existing implementation. My error. But I take your point, mine gets cut off. Menu it is!

brendanx67:
I am not sure what ComboBox uses. It is one of the more complex original Windows controls with quite a lot of code to support all that it does. I am not surprised it ended up being a performance issue.

brendanx67:
I suspect they could have done better with a brand new DropDownList control that was separate from ComboBox. Remember that the default for ComboBox is that it contains and edit box that supports text editing and selection. That is really what is "combo" about it. It is the combination of an edit box and a drop down.

chambm:
I was curious so I looked at the reference source. Looks like they make their own native window:
https://referencesource.microsoft.com/system.windows.forms/R/fc6709aee133dcd0.html

Hopefully a menu is easier than that because that reference implementation looks pretty tedious. :)

brendanx67:
ComboBox is dragging a lot of history behind it. Trying to preserve its implementation in a new control would be a mistake.

pwiz_tools/Skyline/Util/LazyComboBox.cs
```diff
@@ -0,0 +1,141 @@
...
+{
+    /// <summary>
+    /// Substitute for ComboBox which avoids performance issues by only
+    /// creating its dropdown list when user action demands it
+    /// </summary>
+    public class LazyComboBox : Button
```
nickshulman, line 32: I wonder whether this class should be named "LazyDropDownList" since that's the only combo box style it supports. (Actually. maybe just call it "DropDownList" since "Lazy" is too judgmental)
Also, maybe move this class into "Common\Controls" folder.

nickshulman (commented):
Looks really nice.
It also looks like it will be really easy to plug into EditPepModsDlg:
image

I recommend adding the property "SelectedItem":
```
        public object SelectedItem
        {
            get
            {
                return SelectedIndex >= 0 ? Items[SelectedIndex] : null;
            }
            set
            {
                SelectedIndex = Items.IndexOf(value);
            }
        }
```

brendanx67:
Maybe LiteDropDownList

brendanx67:
Why is the window border so black? Even having it the normal button border color of the OK and Cancel buttons in Nick's form would look much better.

bspratt:
> Why is the window border so black? Even having it the normal button border color of the OK and Cancel buttons in Nick's form would look much better.

Just the default behavior, that will get tidied up.

nickshulman:
When I drop the ListBox down, and I move the mouse, the highlighted item in the ListBox does not move with the mouse. It's not that important, but it is a visual difference in behavior between this and a real combo box.
I bet if you used a ContextMenutStrip instead of a NonClippingListBox the mouse tracking would happen.

The other advantage of a ContextMenuStrip is that the different items in the menu could have different fonts. I imagine we are eventually going to be using this awesome control everywhere, so it would be great if users could visually distinguish between their Enzyme that they maliciously named "<Edit List...>" and the "<Edit List...>" option that the SettingsListComboDriver added.

brendanx67:
Let's just try the menu option. I feel like we are going down the ListBox path because that was your inclination, Brian, and it seems that both Nick and I would have started out using a context menu instead.

bspratt:
No argument from me on that - it was definitely the wrong direction. Menu is just easier. I just have to figure out how to get rid of that dead space on the left of the menu,

image

brendanx67:
I don't necessarily think you need to remove that space (which is reserved for icons). It doesn't look exactly like a combo box dropdown list, but I am not sure that is a big deal. It does look like every other menu in the product, whether they use the icon margin or not.

nickshulman:
> I just have to figure out how to get rid of that dead space on the left of the menu,
> 

I bet it's controlled by "ToolStripDropDownMenu.ShowCheckMargin" and "ToolStripDropDownMenu.ShowImageMargin".

bspratt:
Give that man a cigar!

image

brendanx67:
Works for me. Much more comfortable with this as an implemention. Still run the PasteMolecules test 20-100 times to make sure it is not leaking anything with this new implementation.

chambm:
Should it be made about as compact as the original ListBox implementation? Makes me twitch a bit to see a context menu from a DropDownList. :smiley:

brendanx67:
I think it is fine! I think users are much less likely to "twitch" than Matt. We are not actually making any special promises about how this UI works. It already takes some wild trickery just to get these combo boxes where they appear. Grids don't work this way. Nor do they shoe context menus when you left-click on their column headers as we have made Document Grid UI do.

chambm:
I was thinking of Nick's "I imagine we are eventually going to be using this awesome control everywhere". If that's a sincere statement, then using this menu style in more frequently seen combo boxes will change a lot of UI.

brendanx67:
But that is not part of this PR. So, let's not add requirements to this PR because of what Nick imagines for the future.

brendanx67:
Try upgrading to Windows 11 if you want a queasy feeling about UI change. Or are you still using the Windows 7 UI look in Windows 10?

bspratt:
FWIW I did spend some time looking into trying to compact the menu vertically, messing with item Height, Margin etc, but nothing seemed to have any effect.

chambm:
> Try upgrading to Windows 11 if you want a queasy feeling about UI change. Or are you still using the Windows 7 UI look in Windows 10?

I'm still using Windows 7! :laughing: That'll change soon I guess with the new work PC.

brendanx67:
Even if you are using Office 365, it looks like they must have updated to be more Windows 11-like. The rounded scrollbar thumbs are making me cringe a bit. A dropdown that is more menu-like and less list-like still seems pretty minor to me. I use menus all the time for picking.

pwiz_tools/Skyline/Properties/Resources.resx
```diff
@@ -11624,4 +11624,4 @@ If you do not have the original file, you may build the library with embedded sp
   <data name="SimpleFileDownloaderDlg_Show_Downloading_required_files___" xml:space="preserve">
     <value>Downloading required files...</value>
   </data>
-</root>
+</root>
```
brendanx67, line 11627: Seems like you could just revert this file.

brendanx67 (approved):
One minor change to revert the Resources.resx file in Skyline, since this is now in the Common project. Otherwise, it looks pretty good.

brendanx67:
I can't see us using this control for more than this form and Edit > Modify Peptide. Even for the latter, it may need a bit more work to provide an acceptable experience for someone familiar with a ComboBox, like responding to the down arrow key (open the dropdown) and making choices when the user presses alpha keys on the keyboard. I don't think these features are important in this UI, and only slightly in Edit > Modify Peptide, but yeah. Replacing a standard Windows control globally would be definitely more work to make it seem similar enough to what a user expects.

bspratt:
> like responding to the down arrow key (open the dropdown) 

That's not what I'm seeing in the current release implementation - it just rolls the selection without opening the dropdown. But we can add that, yes.

> and making choices when the user presses alpha keys on the keyboard. 

We can add that too.

What else did I miss?

brendanx67:
So, quick list of keyboard shortcuts:
alt-down-arrow and F4: show/hide the menu (toggle)
down-arrow and right-arrow: advance one in the list
up-arrow and left-arrow: back one in the list
a-z: select first thing in the list with the character
second a-z: sometimes supports type-ahead buffering these characters as you type until you pause long enough, or else it advances to the next item if you retype the same character, i.e. "aaa" would select the 3rd element starting with "a" in the list, and cycle (try Peptide Settings - Digestion - Enzyme and typing "l" - lowercase "L")

That's all I can think of. I like that pressing the spacebar should show the menu in your button implementation. I tried that with a normal combo box and was disappointed that it did not work. I had to Google to remember the keys that actually drop the list on a combo box.

brendanx67:
To be as exhaustive as possible, you can search for "combo" in this page:

https://support.microsoft.com/en-us/office/keyboard-shortcuts-for-access-70a673e4-4f7b-4300-b8e5-3320fa6606e2#bkmk_workwithboxes

brendanx67:
I don't think it is necessary to be perfect for this application, but as Matt notes the more ubiquitous we want to make this dropdown list in our UI, the more we need to make it as close in functionality to the original.

PR #1871, opened 2021-12-01
* added Save() call after resetting settings

chambm:


chambm:
@nickshulman Can you chime in here about whether we should Save() here or not? Brendan and I talked on Zoom and got pretty confused about why his Skyline-daily was upgrading from a previous version's settings (after doing a Tools > Options > Misc > ResetSettings) but mine wasn't. We think it might have something to do with the user.configs he has and I don't have, combined with the `Settings.ReloadAndMerge()` function.

brendanx67:
It definitely seemed to me like, if we were going to add a Settings.Default.Save() that we should also update Settings.Default._originalSerializedValues to be the current values. Otherwise, Skyline keeps all its truly original values it started with and things a hodge podge of things changed based on the reset, and the thing on disk must have been created by another version, when in reality the thing on disk was created by the current instance and is completely in synch with what we have in memory.

nickshulman:
If "_originalSerializedValues" needed to be reset, then you should do "_originalSerializedValues.Clear()". However, I don't think that could be causing the behavior that you are seeing. The values in the "_originalSerializedValues" dictionary are only used to determine whether the user has modified a particular setting. The values from that dictionary would never be written to disk.

Yes, it certainly sounds like you would need to add the "Save" call.

If you want to debug installed Skyline, I would recommend replacing the Skyline-Daily.exe in the install folder with one that you have built which has a call to "Debugger.Launch();" at the beginning of Program.Main.

chambm:
Is there any reason to call Reset() and NOT reset the original values? I couldn't think of one.

PR #1874, opened 2021-12-03
- fixed BiblioSpec's DIA-NN speclib reader to support version 3

chambm:
- fixed BiblioSpec's DIA-NN speclib reader to support version 3, and added support for reading scores and ion mobilities (reported by Floriana and Juan)

The DIA-NN speclib reader now requires the TSV report to be in the same directory as the speclib, and they must share some leading characters in the filenames so BlibBuild can automatically find it. The error message when it can't be found is:
`unable to determine DIA-NN report filename for '%s': the TSV report is required to read speclib files and must be in the same directory as the speclib and share some leading characters (e.g. somedata-tsv.speclib and somedata-report.tsv)`
Does that look good @brendanx67 ?


brendanx67:
Yes. Seems like a good idea to have this in release to keep us up to date as possible with DIA-NN

chambm:
Oh I see I didn't merge this because I found out the RT I'm getting from the speclib file is really an iRT. I was hoping to get that fixed in the same PR.

PR #1875, opened 2021-12-05
Fix bug when changing from synchronizing all replicates to none

kaipot:


kaipot:
reported here: https://skyline.ms/announcements/home/support/thread.view?rowId=53819

nickshulman (approved):

PR #1876, opened 2021-12-06
Remove "Normalize to Heavy" context menu item if "HasHeavyModifications" is false

nickshulman:
Hide the menu item "Normalize To > Heavy" when the document does not have any heavy modifications.
Also, if the currently chosen normalization option is unavailable, then switch it to "Default".

Issue 854: Peak Areas - Replicate Comparison - Normalize to strangely chooses poorly
Also, turn off Logarithm scale on Y axis if normalization option is Ratio to Global Standards or Ratio to Isotope Label.

This change should also be ported to 21.2 branch.

Prevent choosing "Normalize to Heavy" when current normalize choice not available. Also, disallow combining Logarithm plot with ratio normalization. (Reported by Brendan, thanks to Nick)


chambm:
How long after the rebase did you wait before merging?

chambm:
Oh I see. The rebase command goes in a GitHub comment, not a commit. :) That way you don't have to mess with git.

PR #1878, opened 2021-12-06
Fix for the NaN or negative dotp values in the peak area plot.

rita-gwen:


brendanx67 (approved):
Does this fix showing zero dotp values when the peak is removed?

rita-gwen:
Yes, it does. It was actually showing -1, not 0. Now the dotp data list will have NaN instead of -1 if the ratio value for a replicate is missing and the graph won't show it.

nickshulman:
Rita,

There are some other places in AreaReplicateGraphPane.cs where it returns "-1" for the dot product value (on lines 1071, and 1100). Should those also be changed to return Float.NaN?

Other than that, this PR looks good to me.

brendanx67:
Not all of the problem values currently show up as -1.00. Simply removing peaks leaves dotp annotations showing 0.00 like this:
image

rita-gwen:
Yes, I just pushed a fix for that. Sorry, I didn't test the (i)dotp use case before. Somehow when the peak is removed the chromInfo for it still exists, it has null area, but not null ratios. The last commit should correct that.
And -1s are removed ass well.

rita-gwen:
Should I merge this one? Looks like the Docker build failure is unrelated.

PR #1887, opened 2021-12-08
"Associate Proteins" checkbox in the Insert Transition List column picker form should be ticked by default…

bspratt:
… when a background proteome is present.

Updated a couple of tutorial tests that were setting this manually, as that should not be necessary.


brendanx67 (approved):

brendanx67:
Approved contingent on passing all of the tests on TeamCity, of course. Probably not a bad idea to let it run perf tests overnight as well.

PR #1895, opened 2021-12-10
ImportTransitionListColumnSelectDlg lags a bit when associate proteins checked…

bspratt:
…, per https://skyline.ms/issues/home/issues/details.view?issueId=857  implementing a quick fix:

 default to Associate Proteins *not* checked
 some performance improvements by batching up the protdb peptide search (was one query per unique peptide)
 make operation cancelable

to be followed by a more complete fix:
  default to checked, but only operate on the handful of visible peptides in the dialog
  defer the full task to when user hits "check for errors" or "OK" (retain result if done in "check for errors")
  add progress indicator to the cancelation mechanism (probably means breaking the search into chunks)

### DIFF
--- a/pwiz_tools/Shared/ProteomeDb/API/Digestion.cs
+++ b/pwiz_tools/Shared/ProteomeDb/API/Digestion.cs
@@ -19,6 +19,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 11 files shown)


brendanx67 (approved):
Looks okay, assuming it passes all tests.

PR #1897, opened 2021-12-11
Added a new custom code inspection…

bspratt:
… to verify that any english language string that begins with "Error:" is localized such that the L10N version starts with Skyline.Properties.Resources.CommandStatusWriter_WriteLine_Error_ (which is "Error:" in english).

There aren't any current examples of this, but I have verified the test by making local changes to non-english resx files.


pwiz_tools/Skyline/Test/CodeInspectionTest.cs
```diff
@@ -215,6 +219,50 @@ List<string> CheckFormsWithoutTestRunnerLookups()
...
+            {
+                Thread.CurrentThread.CurrentCulture = Thread.CurrentThread.CurrentUICulture = new CultureInfo("en"); // We want to compare against the "en" resources
+                var resourceSet = Skyline.Properties.Resources.ResourceManager.GetResourceSet(Thread.CurrentThread.CurrentUICulture, true, true);
+
+                // Build the string of resources.
+                foreach (var resource in resourceSet)
+                {
+                    var pair = resource as DictionaryEntry? ?? default;
+                    var val = pair.Value as string ?? string.Empty;
+                    if (val.StartsWith("Error:"))
```
brendanx67, line 238: I would try to make this always consistent with what is happening in CommandStatusWriter, so that they actually use the same string. Seems unlikely that the text will ever change, but this is a magic text value that is required to match in two places, and I just try to stay away from those as a rule.

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
```diff
@@ -215,6 +219,50 @@ List<string> CheckFormsWithoutTestRunnerLookups()
...
+                    {
+                        var name = pair.Key.ToString();
+                        foreach (var culture in new[] {@"zh-CHS", @"ja"}) 
+                        {
+                            var tryCulture = new CultureInfo(culture);
+                            Thread.CurrentThread.CurrentCulture = Thread.CurrentThread.CurrentUICulture = tryCulture;
+                            var local = Skyline.Properties.Resources.ResourceManager.GetString(name) ?? String.Empty;
+                            var commandStatusWriterWriteLineError = Skyline.Properties.Resources.CommandStatusWriter_WriteLine_Error_;
+                            if (!local.StartsWith(commandStatusWriterWriteLineError, StringComparison.CurrentCulture) &&
+                                !local.StartsWith(@"Error:", StringComparison.InvariantCulture)) // Maybe not yet localized
```
brendanx67, line 248: Way too many uses of "Error:" typed into text.

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
```diff
@@ -215,6 +219,50 @@ List<string> CheckFormsWithoutTestRunnerLookups()
...
+                            {
+                                // Report mismatch
+                                errors.Add(string.Format("The {0} language version of resource string {1} does not begin with the localized version of \"Error:\" (see Skyline.Properties.Resources.CommandStatusWriter_WriteLine_Error_)",
```
brendanx67, line 251: Even this one should be referring to the same value used in CommandStatusWriter and not just matching text you have typed.

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
brendanx67: I would probably just use the "en" resourceSet CommandStatusWriter_WriteLine_Error_ for this value immediately after you make it the current language, and then programmatically refer to it below.

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
```diff
@@ -215,6 +219,50 @@ List<string> CheckFormsWithoutTestRunnerLookups()
...
+        void InspectConsistentErrorMessages(List<string> errors)
+        {
+            var currentCulture = Thread.CurrentThread.CurrentUICulture;
+            try
+            {
+                Thread.CurrentThread.CurrentCulture = Thread.CurrentThread.CurrentUICulture = new CultureInfo("en"); // We want to compare against the "en" resources
+                var resourceSet = Skyline.Properties.Resources.ResourceManager.GetResourceSet(Thread.CurrentThread.CurrentUICulture, true, true);
+
+                // Build the string of resources.
+                foreach (var resource in resourceSet)
```
brendanx67, line 234: Couldn't you just use First() to get the first value instead of using a foreach loop and increasing the indenting. That also has the benefit of failing if for some reason the code to get the "en" language fails. Otherwise, it looks like the code would just fail to do any checking at all and silently allow anything.

brendanx67 (changes requested):
Thanks for the quick response on this request. I had a few suggestions.

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
bspratt: Sounds good. There's a subtle difference between that resource and that string constant, one has a traling space and one does not. I'll make the string constant public and use it in both places.

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
bspratt: Yup.

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
bspratt: Yup.

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
bspratt: Let me clean up the comments here and see if it makes more sense to you then.

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
```diff
@@ -215,6 +219,50 @@ List<string> CheckFormsWithoutTestRunnerLookups()
...
+                {
+                    var pair = resource as DictionaryEntry? ?? default;
```
nickshulman, line 236: In Visual Studio 2017, ReSharper does not like this "default" here and wants it to be "default(DictionaryEntry)".
Are we still supporting VS2017? Is there a setting I could change to make that error warning go away?

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
```diff
@@ -215,6 +219,50 @@ List<string> CheckFormsWithoutTestRunnerLookups()
...
+        /// <summary>
+        /// We have some code, especially in commandline, that expects certain error messages to start with "Error:" or the L10N equivalent
+        /// </summary>
+        void InspectConsistentErrorMessages(List<string> errors)
+        {
+            var currentCulture = Thread.CurrentThread.CurrentUICulture;
+            try
+            {
+                Thread.CurrentThread.CurrentCulture = Thread.CurrentThread.CurrentUICulture = new CultureInfo("en"); // We want to compare against the "en" resources
+                var resourceSet = Skyline.Properties.Resources.ResourceManager.GetResourceSet(Thread.CurrentThread.CurrentUICulture, true, true);
```
nickshulman, line 231: I wonder if you should enumerate over all of the resources (using "[assembly.GetManifestResourceNames](https://stackoverflow.com/questions/26861078/iterating-through-string-resources-in-another-assembly)").

nickshulman (commented):
This code checks that if the English text starts with "Error:", the localized text does as well.
You should also assert the other way: if the localized text starts with "Error:" (or the localized equivalent), the English text really ought to as well.

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
bspratt: I'll just replace it with "new DictionaryEntry()" for max portability.

bspratt:
> This code checks that if the English text starts with "Error:", the localized text does as well. You should also assert the other way: if the localized text starts with "Error:" (or the localized equivalent), the English text really ought to as well.

I thought about that, but I don't think that's actually what we want, since our development happens in English - we don't work backwards from japanese, only forward from English. If anything, we should flag that scenario as an error.

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
bspratt: That just seems to return a list of embedded file resources?

brendanx67:
It does seem to identify the problems that inspired me to suggest this. Now, we will see if they go away when we merge this with my fixes...

brendanx67 (approved):

brendanx67:
The changes look okay to me assuming they pass with my most recent changes.

PR #1900, opened 2021-12-12
Fix NRE using "Import Transition List" button on Start Page

nickshulman:
PasteDlg no longer supports "PasteFormat.transition_list", so the "Import Transition List" button on the Start Page needed to be updated to use "ShowInsertTranstionListDlg" instead of "ShowPasteDlg".

Reported on Exception Web here:
https://skyline.ms/announcements/home/issues/exceptions/thread.view?rowId=54010


bspratt (approved):
Looks good. This whole thing was a bit rushed, it's a pity we didn't have such a test in place to start with.

PR #1904, opened 2021-12-14
Gracefully handle errors reading from the .skyd file during a settings change.

nickshulman:
Fixed error that can occur if there is an error reading chromatogram peaks from the .skyd file while updating document settings. (Reported anonymously, thanks Nick)

This change adds method "ExceptionUtil.IsProgrammingDefect" which returns false if the exception is IOException, or any of the other exceptions which might happen when reading from disk.


pwiz_tools/Skyline/Skyline.cs
```diff
@@ -2451,37 +2451,46 @@ public bool ChangeSettingsMonitored(Control parent, string message, Func<SrmSett
                     bool success = false;
                     using (var longWaitDlg = new LongWaitDlg(this)
                     {
-                        Text = Text,    // Same as dialog box
+                        Text = Text, // Same as dialog box
```
brendanx67, line 2454: Can you revert these ReSharper autoformatting changes to make the important change clearer?

brendanx67 (approved):
Looks good, but please undo the ReSharper autoformatting changes before committing. And cherry-pick to release, of course.

PR #1908, opened 2021-12-14
Fixed a problem handling transition lists with missing product m/z in…

bspratt:
…formation (ref  https://skyline.ms/announcements/home/issues/exceptions/thread.view?rowId=54002). This would cause an exception during the initial inspection of the data to determine its format, but during the inspection phase no exception should be thrown.

### DIFF
--- a/pwiz_tools/Skyline/TestFunctional/PasteMoleculesTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/PasteMoleculesTest.cs
@@ -121,6 +121,7 @@ namespace pwiz.SkylineTestFunctional
         {
             var docEmpty = NewDocument();
 
+            TestMissingProductMZ();
             TestRecognizeChargeState();
             TestLabelsNoFormulas();
             TestHeavyLightPairs();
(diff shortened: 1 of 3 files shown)


pwiz_tools/Skyline/TestFunctional/PasteMoleculesTest.cs
```diff
@@ -1199,6 +1200,37 @@ private void TestRecognizeChargeState()
...
+            var errDlg = ShowDialog<ImportTransitionListErrorDlg>(columnSelectDlg.OkDialog);
+            RunUI(() =>
+            {
+                var allErrorText = string.Empty;
+                foreach (var err in errDlg.ErrorList)
+                {
+                    allErrorText += err.ErrorMessage;
+                }
+                var errText = string.Format(Resources.PasteDlg_ValidateEntry_Error_on_line__0___Product_needs_values_for_any_two_of__Formula__m_z_or_Charge_, 1);
+                Assert.IsTrue(allErrorText.Contains(errText),
```
nickshulman, line 1221: You might want to use the method "StringAssert.Contains" here.

pwiz_tools/Skyline/TestFunctional/PasteTransitionListTest.cs
```diff
@@ -315,5 +318,33 @@ private void TestPeptideOnlyDoc()
...
+            var importDialog = ShowDialog<InsertTransitionListDlg>(SkylineWindow.ShowPasteTransitionListDlg);
+            var columnSelectDlg = ShowDialog<ImportTransitionListColumnSelectDlg>(() => importDialog.textBox1.Text =
+                text.Replace(".", CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator));
+
+            var errDlg = ShowDialog<MessageDlg>(columnSelectDlg.CheckForErrors);
+            RunUI(() =>
+            {
+                Assert.IsTrue(errDlg.Message.Contains(Resources.ImportTransitionListErrorDlg_ImportTransitionListErrorDlg_This_transition_list_cannot_be_imported_as_it_does_not_provide_values_for_) &&
+                              errDlg.Message.Contains(Resources.ImportTransitionListColumnSelectDlg_PopulateComboBoxes_Product_m_z),
+                    string.Format(
```
nickshulman, line 335: You don't need to use "string.Format" for the message part of Asserts.
All of the Assert methods are designed so that they take "string message, params object[] parameters".
The idea is that "string.Format" is expensive and so you pass the format string and parameters so that "string.Format" only needs to be called if the test fails.
(It's a microscopic performance improvement, but that is why those methods all have that sort of signature, and also why methods like "Trace.TraceInformation" are like that).

pwiz_tools/Skyline/TestFunctional/PasteTransitionListTest.cs
```diff
@@ -315,5 +318,33 @@ private void TestPeptideOnlyDoc()
             RunUI(() => documentGrid.Close());
         }
 
+        private void TestMissingFragmentMz()
+        {
+            // Make sure we properly handle missing fragment mz info
+            var saveColumnOrder = Settings.Default.CustomMoleculeTransitionInsertColumnsList;
+            var text =  "Peptide Modified Sequence\tPrecursor m/z\tFragment Name\tProduct Charge\nPEPTIDER\t478.738\ty1\t1\nPEPTIDER\t478.738\ty2\t1";
```
nickshulman, line 325: I think these numbers are going to need to be formatted with the localized decimal separator.

nickshulman (commented):
Looks good, except I think you need to localize the numbers in "TestMissingFragmentMz".
My other comments can be ignored.

nickshulman (approved):
Looks good to me. I forgot that Insert Transition List works with either localized or invariant numbers.

pwiz_tools/Skyline/TestFunctional/PasteTransitionListTest.cs
bspratt: FWIW that happens on line 328.

pwiz_tools/Skyline/TestFunctional/PasteTransitionListTest.cs
bspratt: Good to know, I'll tidy that up so nobody folllows my bad example.

PR #1909, opened 2021-12-15
Improved handling of imported small molecule transition lists with fragment formula and m/z but no charge declaration. …

bspratt:
Specifically, we were not considering the implicitly charged (e.g. [M+2] rather than [M+2H]) case. Reported by Rich Johnson.


pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
```diff
@@ -604,42 +604,45 @@ private int INDEX_DECLUSTERING_POTENTIAL
                 maxCharge, new int[0],
                 TransitionCalc.MassShiftType.none, out _, out _);
 
-            if (isPrecursor && adductInferred.IsEmpty)
+            if (adductInferred.IsEmpty)
             {
                 // See if this can be explained by the more common adduct types, possibly with water loss
                 var matches = new Dictionary<double, Adduct>();
-                foreach (var text in Adduct.DEFACTO_STANDARD_ADDUCTS)
+                foreach (var adductList in new []{Adduct.DEFACTO_STANDARD_ADDUCTS, Adduct.COMMON_CHARGEONLY_ADDUCTS})
```
nickshulman, line 611: If you do:
`                foreach (var text in Adduct.DEFACTO_STANDARD_ADDUCTS.Concat(Adduct.COMMON_CHARGEONLY_ADDUCTS))
`
then you won't need two nested for loops.

nickshulman (approved):
Looks good to me.
I would recommend splitting "PasteMoleculesTest" into separate classes. Shorter tests are usually better because they are easier to debug when they fail.

pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
bspratt: I considered that (I agree that less code is generally better code), but it seems like a lot of memory manipulation in code that can be called quite frequently - not that I actually tried to measure relative performance or anything.

bspratt:
> Looks good to me. I would recommend splitting "PasteMoleculesTest" into separate classes. Shorter tests are usually better because they are easier to debug when they fail.

It's already mostly broken into functions, so once you identify where it's failing it's easy to move that function to the top for a quick repro. And it's nice to still have all the other ones execute afterward, because it's easy to break a parser's correct behavior when tweaking its incorrect behavior.

pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
nickshulman: By the way, "Enumerable.Concat" does not actually have to create a single list containing all of the elements. Instead, it returns an Iterator which first enumerates all of the items in the first list and then the items in the second list.

PR #1911, opened 2021-12-15
Fixed BiblioSpec DIA-NN reader finding first-pass report TSVs

chambm:
- fixed BiblioSpec DIA-NN reader to ignore the report TSV created by the first NN pass (reported by Andy)


pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
```diff
@@ -605,6 +605,8 @@ bool DiaNNSpecLibReader::parseFile()
         {
             if (!bfs::is_regular_file(tsvFilepath))
                 continue;
+            if (bal::contains(tsvFilepath.string(), "first-pass"))
```
nickshulman, line 608: What if they are trying to make a library from the "first-pass.speclib" file? (Is that a valid thing to do?) In that case we would expect the tsv file to contain "first-pass". Maybe this should be:
if (bal::contains(tsvFilepath.string(), "first-pass") && !bal::contains(impl_->specLibFile_, "first-pass"))

chambm: Hah, good point. Yeah, I think checking the speclib file for it is a good start. Probably also need to only be checking the filename in case of "c:\data\first-pass\blahblah.speclib"

chambm: Now there's theoretically nothing stopping a user from putting first-pass in their report filename but not their speclib filename, but to those users I say: JUST DEAL WITH IT (rename your file)! :)

nickshulman: Maybe it should check that the number of times that "first-pass" occurs in the TSV filename is the same as the number of times that it occurs in the speclib filename.

chambm: I really don't expect users to put first-pass in their filename. At least, not more than once, because the first time they may not know about the automatically generated files that have the `-first-pass` suffix in a 2-pass search.

nickshulman (approved):

PR #1912, opened 2021-12-15
Fixed a problem loading small molecule transition lists that provide m/z and no charge information.…

bspratt:
… Skyline will infer the adduct, but in the case where the implied adduct has synonyms known to Skyline (e.g. [M+FA-H] and [M+HCOO]) Skyline would throw an unhandled "System.ArgumentException: An item with the same key has already been added."

Reported anonymously in https://skyline.ms/announcements/home/issues/exceptions/thread.view?entityId=af7f5164-2fc7-103a-9bdb-22f53556416c&_anchor=53634#row:53634


nickshulman (approved):
Looks great.

PR #1914, opened 2021-12-16
Should not throw when a zlib-compressed array decompresses to 0 length

bspratt:
This is not illegal, and if the array was declared as being non-zero in length that still gets flagged by the caller.


bspratt:
See https://github.com/OpenChrom/openchrom/issues/175 for some background - dealing with unusual/malformed mzML output from another open source converter, but I really want to support that converter as it handles a lot of formats we don't such as ChemStation .D files.

chambm (approved):
Looks good. Nice work with the OpenChrom team to get their converter up to spec.

PR #1920, opened 2021-12-16
Skyline/work/20211216 edit insert transitionlist paste dlg improvements

bspratt:
A couple of users have expressed confusion around the new Edit>Insert>TransitionList behavior where Skyline presents this window:
image
this change tries to make Skyline's intentions a little more obvious by moving the prompt to the place where the paste actually happens:
image
this is done programattically, no change to the resx files so no L10N issues in case we want to use it in the release branch:
image
though I'd rather see it say "Press Ctrl-V to paste your transition list here" to be completely obvious.

Note that there are a great number of changes to tests, as they were unfortunately written to assume way too much about the implementation of the paste dialog box.


brendanx67:
I think we can remove the colons without a great deal of concern. Maybe
even replace each with a period, since it is a sentence. There is an Asian
period, which you can copy from other text in Chinese and Japanese.

bspratt:
Added a paste button as requested by Nick and Brendan in this week's team zoom call:

image

PR #1921, opened 2021-12-17
Disable Unifi tests

nickshulman:
Disable unifi tests since the server is returning 504 Gateway Time-out


chambm:
Disabling these tests should not require a code change. I just cleared out the username on the TC agents.

chambm:
Grumble grumble. I see Skyline needs the password to be cleared out to disable the tests. I had cleared out the username.

chambm:
I just factored out all the UNIFI parameters from the TC configs into the "UNIFI test parameters" template. That's now a single config template to modify to enable/disable UNIFI (by blanking out the password, although in the core tests blanking out the username works too: I'm thinking Skyline shouldn't be hardcoding the username either)

PR #1923, opened 2021-12-18
Skyline/work/20211218 update small mol tutorials for new column picker

bspratt:


pwiz_tools/Skyline/FileUI/ImportTransitionListErrorDlg.cs
```diff
@@ -89,6 +89,11 @@ protected override void DoPaste()
             }
         }
 
+        public void OKDialog()
```
brendanx67, line 92: We usually use OkDialog()

brendanx67 (changes requested):
Need docx files in Chinese and Japanese/outgoing for the retranslation of the documents. You just go through the tutorials again and paste Chinese or Japanese screenshots in place of the English screenshots. Text remains English in the outgoing .docx files. See the files in these folders for examples.

pwiz_tools/Skyline/FileUI/ImportTransitionListErrorDlg.cs
bspratt: Fixed, thanks

bspratt:
> Need docx files in Chinese and Japanese/outgoing for the retranslation of the documents. You just go through the tutorials again and paste Chinese or Japanese screenshots in place of the English screenshots. Text remains English in the outgoing .docx files. See the files in these folders for examples.

It's a little tricky - rather a lot of text has changed (because the new interface is WAY easier to use, with many fewer steps) so I'm not really sure what I'd be replacing Certainly I can drop the needed images in there but I don't know exactly how the English text fits in.

Also BTW these images use the prompt text as label rather than the proposed prompt text in the text box, is that what we want?

brendanx67:
It's pretty routine once you understand what needs to be done. You just
take your current English tutorial and replace all the images in it with
new screenshots in the target language. Then we give the translators:

1. The outgoing .docx (English text with target language images)
2. The prior translated .docx
3. The prior English .docx file

They then use tools to decompose the prior .docx files by sentence into a
translation memory. They then translate only new or change sentences, and
then inject that translation memory into the .docx with the target images,
and we get a new translation for a lower translation cost than if you just
gave them #1 alone and they treated it like an entirely new tutorial.
Presumably, despite the extensive changes, there are still a lot of
sentences that remain unchanged from the previous translation.

bspratt:
That makes sense, no problem.

bspratt:
OK, this should be good to go.

Note that I added a "Copy to clipboard" button to the PauseForScreenshot dialog as an alternative to Alt-PrtScn when capturing images of Skyline windows for creating and updating tutorials. 

The code deals with edge pixel schmutz due to overlapping windows by determining which color is most commonly seen in the captured edge pixels and then setting them all to that color. (I was getting really annoyed by the tedium of positioning each window over a clean background.) 

I also added a "Copy Metafile" button for convenience.

N.B. once in a while these buttons don't grab what you think they're going to, it's rare enough that I've chosen not to pursue it - this isn't user facing and you can always go back to the manual way of doing things if needed. Just doesn't seem like a good use of my further time.

image

chambm:
Have you tried TNT Capture? It's a pretty nifty program with a hotkey for capturing the window the mouse is hovering over. It handles my Windows 7 rounded corners very well. Fortunately the current metro style has square corners so it doesn't have that problem. But it would probably avoid the color issues you saw and the cases where your current approach doesn't work.

bspratt:
Interesting suggestion, thanks. I kinda like the fact that this is self contained (though admittedly it won't do well on rounded corners). 

I actually hadn't noticed the issue with edge pixel schmutz until Brendan pointed it out - it's pretty subtle, just slight variations in color due to some kind of blending with whatever is behind the window. Are you sure TNT Capture deals with this?

bspratt:
Although when I think about it, probably not that hard to deal with rounded corners, presumably those areas have a zero alpha channel andit  should be easy to detect where the true edge begins.

chambm:
What is the schmutz? What is up with the weird title bar in your screenshot above? The edge stops next to the red x and also on the left side of the title bar.

brendanx67:
Does the code keep the right window activated? My concern with a button on
the pause window is that the form being captured may not have the focus.

Otherwise, it sounds great. Hopefully this is also using and improving on
Rita’s code which we use to auto-snapshot the “cover shots”.

bspratt:
> What is the schmutz?

Here's an example, Skyline window on top of something else with red lines on white:
image
There's discoloration along the right hand edge. Easy to miss.

>  What is up with the weird title bar in your screenshot above? The edge stops next to the red x and also on the left side of the title bar.

I don't know, it's just how it is, whether you screen cap it or not.

bspratt:
> Does the code keep the right window activated? My concern with a button on the pause window is that the form being captured may not have the focus. Otherwise, it sounds great. Hopefully this is also using and improving on Rita’s code which we use to auto-snapshot the “cover shots”.

Yes, it sets focus on the control before grabbing the screenshot. It does use our existing code for the actual grab.

chambm:
OK, good to know about the discoloration. I don't see it in Win7 because its borders are so thick with "chrome". Haven't tested TNT on Win10 yet.

Is that red x from Windows 11 style?

bspratt:
I guess? Heavy use of ZedGraph in that window, which I don't really know a lot about.

bspratt:
> Yes, it sets focus on the control before grabbing the screenshot. It does use our existing code for the actual grab.

I meant

bspratt:
> I guess? Heavy use of ZedGraph in that window, which I don't really know a lot about.

Actually I think it's the docking window library we use.

bspratt:
Oh, and no, it's not Win11, this is on a Win10 machine. Didn't parse the question properly in my head.

bspratt:
> Although when I think about it, probably not that hard to deal with rounded corners, presumably those areas have a zero alpha channel andit should be easy to detect where the true edge begins.

Oof, maybe not - this is a close up of an Alt-PrtScn on Win11
image
in context:
image

This just seems lazy, they're not even trying to deal with the rounded corners.

Well, I like my implementation for as long as we're on Win10, I guess.  And I can probably still detect that edge, it's just a bit trickier. Or maybe the smart move is programatically putting a big empty window behind the thing being snapshotted.

brendanx67:
When all I need to do is screenshots, I just don’t find it takes that long.
I do this for Chinese and Japanese just after I complete the English
version. By far, the most time construction part is ensuring the tutorial
text and images stay in sync and provide meaningful instruction on best
practices.

I did fully automate the cover shots, but there I can truly leave it
overnight and wake to screenshots for all 20 tutorials in 3 languages in 2
different sizes and formats 120 images in all.

Just don’t want to keep spending a lot of time on this. Try to make it a
small investment when you are actually working on screenshots as you have
been. Otherwise, I am not convinced you are saving me much time and I have
been the primary consumer.

bspratt:
It's done, as far as I'm concerned. I was updating four tutorials in three languages, 198 screenshots in total, and finding it extremely tedious dealing with the edge schmutz issue. For me, at least, this change makes the tutorial maintenance task considerably faster and far less tedious.

brendanx67 (approved):
I did a review and revision pass. Looks good to go for me now.

brendanx67:
Should page 20 in the Existing Experiments tutorial show the transition list column picker UI?

PR #1924, opened 2021-12-18
Remove direct uses of class "DataGridView".

nickshulman:
Subclasses "CommonDataGridView" or "DataGridViewEx" should always be used instead, since they include an important bug workaround which prevents an exception that can happen if the mouse is over the square at the top left corner of the grid when the form is created.

Added code inspection to prevent direct uses of DataGridView in Skyline code.

There are a couple of DataGridViews in SkylineBatch which are still susceptible to this bug. I did not fix SkylineBatch because it does not reference pwiz_tools\Common, so I'd have to write my own class there to work around this bug (or just change the ColumnHeaderHeightSizeMode on the grids to be something other than "AutoSize", such as "EnableResizing".

(This does not need to be ported back to Release branch since it is unlikely anyone will run into this bug).


brendanx67 (approved):
Looks good to me. Thanks, Nick.

PR #1925, opened 2021-12-20
* added CommandLinePerfImportDiaPasefTest

chambm:
* added CommandLinePerfImportDiaPasefTest as the first commandline perftest that can be run on Wine/Docker; a user has reported that import of large diaPASEF files crashes on Wine/Docker Skyline so it'll be good to have an automated test for that


chambm:
I'm stumped on this for now. Need a way to attach a debugger in Wine.

PR #1928, opened 2021-12-30
BiblioSpec: allow score thresholds per file and add new flag to only output score types (i.e. no library build)

kaipot:


pwiz_tools/Skyline/FileUI/PeptideSearch/BuildPeptideSearchLibraryControl.cs
```diff
@@ -223,28 +225,151 @@ public bool FilterForDocumentPeptides
...
+            lock (_gridFilesLock)
             {
-                // Set new value
-                ImportPeptideSearch.SearchFilenames = value;
+                gridSearchFiles.CellValueChanged -= gridSearchFiles_CellValueChanged;
 
-                // Always show sorted list of files
-                Array.Sort(ImportPeptideSearch.SearchFilenames);
+                var existingFiles = new HashSet<string>();
+                for (var i = gridSearchFiles.RowCount - 1; i >= 0; i--)
```
nickshulman, line 247: I wonder if code like this would be easier to write if the DataSource of gridSearchFiles was set to a BindingList instead of adding and removing rows.

pwiz_tools/Skyline/FileUI/PeptideSearch/BuildPeptideSearchLibraryControl.cs
```diff
@@ -46,7 +47,10 @@ namespace pwiz.Skyline.FileUI.PeptideSearch
     public partial class BuildPeptideSearchLibraryControl : UserControl
     {
         private readonly SettingsListComboDriver<IrtStandard> _driverStandards;
-        private MsDataFileUri[] _ddaSearchDataSources = new MsDataFileUri[0];
+        private MsDataFileUri[] _ddaSearchDataSources = Array.Empty<MsDataFileUri>();
+
+        private object _gridFilesLock = new object();
```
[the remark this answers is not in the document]
kaipot: You're right, I'll fix this.

pwiz_tools/Skyline/FileUI/PeptideSearch/BuildPeptideSearchLibraryControl.cs
kaipot: Yeah, the way the files list works is a little strange but I didn't want to mess with it too much. Maybe changes for a future PR.

pwiz_tools/Skyline/FileUI/PeptideSearch/BuildPeptideSearchLibraryControl.cs
```diff
@@ -223,28 +224,139 @@ public bool FilterForDocumentPeptides
...
+            toAdd.ExceptWith(existingFiles);
 
-                // Always show sorted list of files
-                Array.Sort(ImportPeptideSearch.SearchFilenames);
+            foreach (var file in toAdd)
+            {
+                var i = gridSearchFiles.Rows.Add();
+                gridSearchFiles[columnFile.Index, i].Value = new FileCellValue(gridSearchFiles, columnFile, file);
+                gridSearchFiles[columnScoreType.Index, i].Value = new ScoreTypeCellValue();
+                gridSearchFiles[columnThreshold.Index, i] = new ThresholdCell();
```
nickshulman, line 259: I think if you set the CellTemplate property of this DataGridViewTextBoxColumn to a ThresholdCell (in the BuildPeptideSearchLibraryControl constructor), that cell template would end up getting correctly cloned whenever a new row was added to the grid.

pwiz_tools/Skyline/FileUI/PeptideSearch/BuildPeptideSearchLibraryControl.cs
```diff
@@ -223,28 +224,139 @@ public bool FilterForDocumentPeptides
...
+                                for (var j = 0; j < scoreTypesThis.Length; j++)
+                                {
+                                    var scoreType = scoreTypesThis[j];
+                                    if (j == 0)
+                                    {
+                                        row.Cells[columnScoreType.Index].Value = new ScoreTypeCellValue(scoreType);
+                                    }
+                                    else
+                                    {
+                                        gridSearchFiles.Rows.Insert(i + 1);
```
nickshulman, line 314: You might want to write a separate function which inserts a row into the grid and sets these three cell values. That function could be used both here and in the place above where a row is being appended to gridSearchFiles.

PR #1929, opened 2021-12-30
Skyline/work/20211230 assay lib ion mobility

bspratt:
Improvements to transition list reader to support ion mobility values in peptide transition list and assay library imports:

Ion mobility values are now supported in peptide-oriented transition lists (formerly these could only be set as a separate step, using the Document Grid).

Ion mobility values are now supported in Assay Library imports.

Added new column types to support declaring ion mobility value and units in a single column (e.g "Explicit Ion Mobility (1/K0)").

Column recognition code now treats header names like "Explicit Collision Energy" and "Collision Energy" as synonyms for deducing column type (in this example, column type is Explicit Collision Energy).


nickshulman (approved):
Looks good to me.

PR #1933, opened 2022-01-05
* added support for signing Skyline admin installer

chambm:


brendanx67:
Could we maybe also generate a .bat file with the pfx-password value in it that calls Skyline/SignAfterPublish.bat in order to remove the need to change the Skyline.csproj file we have currently? I currently feel I have to be extra careful not to commit this change to the .csproj file. It would be nice if that were not the case and the password existed only in a file I never commit to the project.

chambm:
Are you saying you want to remove the SignAfterPublish call with the dummy password from the Skyline.csproj entirely? If so, perhaps a simpler solution than generating a bat would be to modify your bso.bat so that it calls SignAfterPublish.bat directly. That might require a PAUSE in the bso.bat which you can continue from after you publish. That would avoid having your password in both the generated bat file and your bso.bat file. Or you could just create a bso-after-publish.bat file with the SignAfterPublish.bat call. Or is that not possible because the target name changes? I know the path would have to be different for 32/64 builds but you're only doing 64-bit releases now.

chambm:
> Could we maybe also generate a .bat file with the pfx-password value in it that calls Skyline/SignAfterPublish.bat in order to remove the need to change the Skyline.csproj file we have currently? I currently feel I have to be extra careful not to commit this change to the .csproj file. It would be nice if that were not the case and the password existed only in a file I never commit to the project.

@brendanx67 How is this latest version?

brendanx67 (approved):
I don't have more time to work on this with you right now. So, let's get this much in because it achieves the goals: 1. signing the .msi file, 2. removing the need to edit the .csproj file and put the password in it.

Then we can come back later for the improvements I know you want: 1. PFX file name in one location, 2. signing done by the toolset defined for the build and not hardcoded to VS2013.

chambm:
Well that is interesting. slash-rebase got the wrong commit message.

PR #1934, opened 2022-01-05
Fix problem where Scan Identifier list in .skyd file is twice as big as it should be if there are iRTs.

nickshulman:
If the chromatograms were extracted in two passes (where the iRT standards are extracted in the first pass, and the rest of the peptides in the second pass), then the ScanIds will contain two copies of every scan identifier string.

Also, slight tweak to "CantReadSkydTest" so that it calls "FileStreamManager.CloseAllStreams" multiple times if it fails the first time.


brendanx67:
Nice catch, Nick.

PR #1935, opened 2022-01-05
Dialog title should be "Import Assay Library: Identify Columns"…

bspratt:
… instead of "Transition List: Identify Columns" when importing an assay library.

### DIFF
--- a/pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
+++ b/pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
@@ -124,7 +124,7 @@ namespace pwiz.Skyline.FileUI
         private List<string> smallMolColPositions;
         private List<string> peptideColPositions;
 
-        public ImportTransitionListColumnSelectDlg(MassListImporter importer, SrmDocument docCurrent, MassListInputs inputs, IdentityPath insertPath)
+        public ImportTransitionListColumnSelectDlg(MassListImporter importer, SrmDocument docCurrent, MassListInputs inputs, IdentityPath insertPath, bool assayLibrary)
         {
             Importer = importer;
             _docCurrent = docCurrent;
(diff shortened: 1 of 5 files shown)


bspratt:
Note I experimented with string replacement but it doesn't seem to be reliable, not in locale ch anyway.

nickshulman (approved):

pwiz_tools/Skyline/Properties/Resources.resx
```diff
@@ -11630,4 +11630,7 @@ If you do not have the original file, you may build the library with embedded sp
   <data name="SmallMoleculeTransitionListReader_ReadPrecursorOrProductColumns_Multiple_ion_mobility_declarations" xml:space="preserve">
     <value>Multiple ion mobility declarations</value>
   </data>
+  <data name="ImportTransitionListColumnSelectDlg_Import_Assay_Library___Identify_Columns" xml:space="preserve">
+    <value>Import Assay Library:  Identify Columns</value>
```
brendanx67, line 11634: I think just one space after the colon here is a better choice. Though, I know I was also taught in typing class in the 1980s to use two spaces after a colon.

pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
```diff
@@ -137,6 +137,11 @@ public ImportTransitionListColumnSelectDlg(MassListImporter importer, SrmDocumen
             previousIndices = new int[Importer.RowReader.Lines[0].ParseDsvFields(Importer.Separator).Length + 1];
             InitializeComponent();
 
+            if (assayLibrary) // Dialog title should be "Import Assay Library: Identify Columns" instead of "Transition List: Identify Columns"
+            {
+                Text = Resources.ImportTransitionListColumnSelectDlg_Import_Assay_Library___Identify_Columns;
```
brendanx67, line 142: Do we do this elsewhere? Are there any other forms like the error list form that still refer to a Transition List and not an Assay Library? We added File > Import > Assay Library fairly late in the game as a way of making more helpful decisions by default for a user that knows they have an "Assay Library" which is an extended form of a transition list with iRT values and relative ion intensities.

brendanx67 (approved):
Very minor comment about spaces and a request to check the entire flow from File > Import > Assay Library for consistency of terminology. Since this path was added on top of File > Import > Transition list and really just supposed to make some decisions by default the user is forced to specify with the latter.

pwiz_tools/Skyline/Properties/Resources.resx
bspratt: Easy fix (I'll do that for the Transition List version too)

pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
bspratt: I'll check.

bspratt: Well, there's this:
image

and very briefly
image

but I'm not really sure there's much to be gained by tweaking the language there.

brendanx67:
Sure. They are still transitions in our world. In the end, an Assay Library
is a specific type of Transition List.

PR #1936, opened 2022-01-06
Candidate Peaks grid

nickshulman:
This change adds a menu item "Candidate Peaks".
image
This brings up the Candidate Peaks grid, which shows you, for the currently selected precursor and replicate, the peaks that Skyline detected, and the scores that they got 
image
When you customize the report there are separate columns for each of the feature scores:
image
Those columns with particular scores don't seem to be as useful as the "Weighted Feature" collection, which is a dictionary of feature name to WeightedFeature and ends up begin displayed as "Score x Weight = Product"

If you have manually integrated your peak, it does not show up in the Candidate Peaks grid (maybe it should).
Instead, you can find it's score by looking at the Precursor Result (perhaps in the Document Grid).

image

image


nickshulman:
I have abandoned this branch and now I am working in this branch:
https://github.com/nickshulman/pwiz/tree/Skyline/work/20220205_CandidatePeaks

PR #1940, opened 2022-01-10
Some library loading errors were being mistakenly passed to the ExceptionWeb handler even after shown in UI

bspratt:
My change in commit f839f124 on 2/24/2020 to "improve reported diagnostics when users encounter difficulties with imported spectral libraries" resulted in some messages going to the unhandled exception notification system when they had already been (or should have already been) shown in the UI.


pwiz_tools/Skyline/Model/Lib/BiblioSpecLite.cs
```diff
@@ -1156,7 +1155,7 @@ private bool Load(ILoadMonitor loader, IProgressStatus status, bool cached, out
             }
             catch (Exception x)
             {
-                failureException = new Exception(FormatErrorMessage(x), x);
+                var failureException = new Exception(FormatErrorMessage(x), x);
```
nickshulman, line 1158: This previous "catch (InvalidDataException x)" and "catch (IOException x)" can be combined with this "catch (Exception x)" since each catch block is doing the same thing.

However, since we want some things in the "catch (Exception)" block to be reported to Exception web, what we now recommend doing is using the new "ExceptionUtil.IsProgrammingDefect" method to check whether the error should be reported to exception web:
```
catch (Exception x) 
{
    if (!cached) 
    {
        if (ExceptionUtil.IsProgrammingDefect(x)) 
        {
            throw;
        }
        if (!cached)
        {
            loader.UpdateProgress(status.ChangeErrorException(new Exception(FormatErrorMessage(x), x)));
        }
    }
}
```

nickshulman (commented):
Looks good to me. It might be a good idea to use the new method "ExceptionUtil.IsProgrammingDefect" to check whether to treat this as an unhandled error that should be reported to the exception web.

pwiz_tools/Skyline/Model/Lib/BiblioSpecLite.cs
brendanx67: I suspect that ReSharper will complain about the second if (!cached) always being true. Otherwise, I agree that there could be catch (Exception e) clause and I like the new IsProgrammingDefect() check.

pwiz_tools/Skyline/Model/Lib/BiblioSpecLite.cs
```diff
@@ -995,7 +992,9 @@ private bool Load(ILoadMonitor loader, IProgressStatus status, bool cached, out
                     }
                     catch (Exception e)
                     {
-                        failureException = new Exception(FormatErrorMessage(e), e);
+                        var failureException = new Exception(FormatErrorMessage(e), e);
```
brendanx67, line 995: I think you could get rid of the "var failureException" and simply put all the new Exception() constructors inside the status.ChangeErrorException() calls.

brendanx67 (approved):
Looks very close. I suggested completely removing the vestigial local variable name, since I don't think it adds much to the understandability.

Since, it was apparently confusing to you, I would also suggest adding a comment like:

// This will show the user the error message after which the operation can be treated as canceled.

pwiz_tools/Skyline/Model/Lib/BiblioSpecLite.cs
bspratt: > This previous "catch (InvalidDataException x)" and "catch (IOException x)" can be combined with this "catch (Exception x)" since each catch block is doing the same thing.

I was assuming that there was value in preserving the exception Type where it was known - you don't think so?

brendanx67: I think it is enough to preserve that, and far more importantly the original stack trace, as you have by making that exception the inner exception to the one you create.

brendanx67: You no longer actually throw the exception, and you are clear that the only thing left to throw to is the unhandled exception handler, which we still want for programming error type exceptions.

Makes me realize that you may still not quite want what Nick suggested? Is there anything to be gained with FormatErrorMessage() when it was a programming error? i.e.

if (ExceptionUtil.IsProgrammingDefect(x)) 
{
    throw new Exception(FormatErrorMessage(x), x);
}

bspratt: >  Is there anything to be gained with FormatErrorMessage() when it was a programming error? 

Yes, it includes potentially useful information about the library filename and contents. I think it should be used in this instance.

bspratt:
OK, this is one of those times I wish we were all in the office.

1) Perils of working on legacy code... why would we not want to show errors when "cached" is true? I feel like this is all getting a bit crufty - especially since we apparently no longer worry about whether or not a library actually gets a cached version written.

2) re ExceptionUtil.IsProgrammingDefect -  Do we really believe that all InvalidDataException,  IOException, and UnauthorizedAccessException are equally (in?)appropriate for sharing with the user instead of ExceptionWeb?

brendanx67:
I think the point is that we are going to try to rebuild the cache if we fail to load it, rather than showing the user an error that we were unable to load the cache.

On 2, yes. Fortunately, we have a team Zoom meeting coming up in 1/2 hour.

PR #1942, opened 2022-01-11
Fixed a minor display issue where the "Cancel" buttons on multi-file …

bspratt:
…Import Results progress display could be partially obscured. Reported by Brendan.

There was actually code in there already that tried to anticipate this, but it was timing dependent, so this is a very simple fix.

The problem can be seen here:
image
The corrected code:
image

### DIFF
--- a/pwiz_tools/Skyline/Controls/Graphs/AllChromatogramsGraph.cs
+++ b/pwiz_tools/Skyline/Controls/Graphs/AllChromatogramsGraph.cs
@@ -499,7 +499,7 @@ namespace pwiz.Skyline.Controls.Graphs
             bool first = true;
             var width = flowFileStatus.Width - 2 - // Avoid clipping the cancel/retry button when we need a vertical scrollbar
                         (flowFileStatus.VerticalScroll.Visible || 
-                         status.ProgressList.Count > 8  // If scrollbar isn't visible already, it's about to be
+                         status.ProgressList.Count > (panelFileList.Height / (new FileProgressControl()).Height)  // If scrollbar isn't visible already, it's about to be
                             ? SystemInformation.VerticalScrollBarWidth
                             : 0);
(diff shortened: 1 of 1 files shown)


brendanx67 (approved):

pwiz_tools/Skyline/Controls/Graphs/AllChromatogramsGraph.cs
```diff
@@ -497,8 +497,9 @@ private void AddProgressControls(MultiProgressStatus status)
 
             // Match each file status with a progress control.
             bool first = true;
-            var width = flowFileStatus.Width - 2 -
-                        (flowFileStatus.VerticalScroll.Visible
+            var width = flowFileStatus.Width - 2 - // Avoid clipping the cancel/retry button when we need a vertical scrollbar
+                        (flowFileStatus.VerticalScroll.Visible || 
+                         status.ProgressList.Count > (panelFileList.Height / (new FileProgressControl()).Height)  // If scrollbar isn't visible already, it's about to be
```
brendanx67, line 502: On further through, I was curious why if status.ProgressList.Count is greater than zero (which it should always be, right?) you need to create a new FileProgessControl?

bspratt: Merely to know the expected height of the ones we'll actually employ. If there's a cleaner way to do that, happy to make the change.

brendanx67: Maybe you are right, if they have not been created yet at the time this check is made.

bspratt: I thought about trying to do something clever with the first one that actually gets created and used, but it just didn't seem worth the additional code complexity.

PR #1945, opened 2022-01-13
Fixed a problem reading small molecule transition lists with isotope labels declared in precursor and fragment formulas with integer-only charge declarations

bspratt:
Fix a problem reading small molecule transition lists where isotope labels are given as part of the formula, and charge is given only as an integer. e.g. C3C'2H5 with declared charge 1, which means formula C5H5 with adduct [M2C13+H].  But as there's no adduct column to update/populate we weren't properly preserving the adduct information, and ended up with a target C5H5[M+H] instead of C5H5[M2C13+H].

Also: Fix a problem reading tab separated files with empty fields at end of a line - Trim() was eating the trailing tab characters.

Also: Guarantee preservation of atom order in StripLabelsFromFormula() - eg. C3C'4H2O3 comes out as C7H2O3 and not something dependent on arbitrary dictionary implementation which might give you H2O3C7 etc


bspratt:
reported by Rich Johnson, see issue #864

pwiz_tools/Skyline/Util/BioMassCalc.cs
```diff
@@ -598,7 +598,8 @@ public string StripLabelsFromFormula(string desc)
             }
             var parse = desc;
             var dictAtomCounts = new Dictionary<string, int>();
```
nickshulman, line 600: I think you should use the class "OrderedDictionary" here.
That class is similar to Java's "LinkedHashMap": It's a dictionary, but it remembers the order that you put things into it. I think that would enable you to get rid of the "atomOrder" argument below.

pwiz_tools/Skyline/Util/BioMassCalc.cs
```diff
@@ -613,9 +614,26 @@ public string StripLabelsFromFormula(string desc)
...
+                        }
+                        else
+                        {
+                            atomOrder[index] = unlabeled; // Formula was all heavy - e.g. C' but no C
+                        }
+                    }
                 }
             }
-            return dictAtomCounts.Aggregate(string.Empty, (current, pair) => current + string.Format(CultureInfo.InvariantCulture, @"{0}{1}", pair.Key, (pair.Value>1) ? pair.Value.ToString() : string.Empty)); 
+            return atomOrder.Aggregate(string.Empty, (current, atom) =>
```
nickshulman, line 632: I think "String.Join" is easier to understand than "Aggregate", so I would do:
```
return string.Join(string.Empty, atomOrder.Select(atom=>atom + (dictAtomCounts[atom] == 1 ? string.Empty : dictAtomCounts[atom].ToString(CultureInfo.InvariantCulture))))
```

pwiz_tools/Skyline/Model/Import.cs
```diff
@@ -344,7 +344,8 @@ private IList<string> ReadLinesFromFile(IProgressMonitor progressMonitor, IProgr
             {
                 var inputLines = new List<string>();
                 string line;
-                while ((line = reader.ReadLine()?.Trim()) != null)
+                char[] whitespace = { ' ', '\r', '\n', '\f' }; //The usual whitespace characters, except tab which may be a separator
+                while ((line = reader.ReadLine()?.Trim(whitespace)) != null)
```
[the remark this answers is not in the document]
bspratt: Sounds good, thanks.

pwiz_tools/Skyline/Util/BioMassCalc.cs
bspratt: Turns out OrderedDictionary is pretty dissimilar from Dictionary or SortedDictionary, to the point where using it is more complicated that it's worth due to code reuse issues. I think the atomOrder stuff needs to stay. It was worth investigating though, thanks.

nickshulman: > Turns out OrderedDictionary is pretty dissimilar from Dictionary or SortedDictionary, to the point where using it is more complicated that it's worth due to code reuse issues. I think the atomOrder stuff needs to stay. It was worth investigating though, thanks.

Ohhhh. Yes, I see that "OrderedDictionary" is not a generic class. That is unfortunate.

pwiz_tools/Skyline/Util/BioMassCalc.cs
bspratt: It seems like Aggregate would be more efficient, though, since it doesn't mess with a separator. I mean, it's there for a reason, right?

bspratt: Hah, no: https://docs.microsoft.com/en-us/dotnet/csharp/how-to/concatenate-multiple-strings ". If optimizing performance is critical, consider the StringBuilder class or the String.Concat or String.Join method to concatenate a collection, instead of Enumerable.Aggregate."

bspratt: I think this is the way:
`            return string.Concat(atomOrder.Select(atom =>
            {
                var atomCount = dictAtomCounts[atom];
                return atomCount > 1 ? $@"{atom}{atomCount.ToString(CultureInfo.InvariantCulture)}" : $@"{atom}";
            })); `

bspratt: or rather
 `            return string.Concat(atomOrder.Select(atom =>
            {
                var atomCount = dictAtomCounts[atom];
                return atomCount > 1 ? $@"{atom}{atomCount.ToString(CultureInfo.InvariantCulture)}" : atom;
            })); `

nickshulman: Oooh. Neat. Looks like "String.Join" was added in .Net 4.0.
I am going to find all the places where I did "String.Join(string.Empty,..." and change them to use "String.Concat(".

bspratt: And I should look for the places where I used Aggregate.

PR #1947, opened 2022-01-16
Checkbox to "Synchronize Peptides" so the same peptide can belong to multiple proteins

nickshulman:
This adds a checkbox to the Document Settings Dialog to enable you to synchronize changes to peptides that are duplicated across your document.
image
In theory, that will make it so that things are still manageable even when the same peptide belongs to multiple proteins

Then, when you have multiple copies of the same peptide in your document like this:
image
You can change a peak boundary, and it updates in all the copies of that peptide (see that the "dotp" value changed from 0.86 to 0.8 in the Targets tree for each K.MLSGFIPLKPTVK.K):
image

In theory this will make it so that people won't feel the need to Remove Repeated Peptides. Peptides will be able to belong to multiple proteins, and they will participate in Protein level fold changes etc. for all the proteins they belong to.

(Still needs a lot of work, but might be functional enough for us to figure out whether this is the direction we want to go in terms of dealing with protein groups in peptide search results).

### DIFF
--- a/pwiz_tools/Skyline/Model/DocNode.cs
+++ b/pwiz_tools/Skyline/Model/DocNode.cs
@@ -949,7 +949,7 @@ namespace pwiz.Skyline.Model
         /// <param name="childReplace">A child node with an Id that is reference equal to the node to be replaced</param>
         /// <returns>A new parent node</returns>
         /// <exception cref="IdentityNotFoundException"/>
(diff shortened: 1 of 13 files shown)


nickshulman:
I think the only open question with this pull request is what should duplicated peptides look like in the Targets tree?
In the screenshots here, there is a "*" at the beginning of duplicated peptide sequences.
Should it look like something else?

PR #1951, opened 2022-01-17
Speed up "Refine > Associate Proteins"

nickshulman:
The AssociateProteinsDlg was really slow when the document contained tens of thousands of peptides, especially when using a FASTA file.
This change speeds up finding which peptides are contained in a protein sequence by using a Trie data structure (in the new class "SubstringFinder"). Also, it now does that string matching on multiple threads using ParallelEx.
Also, moving the peptides to different parents is made faster by just changing the Identity objects (Peptide, etc) instead of calling ChangeSettings on DocNode's.


chambm:
Did you consider building an Aho-Corasick trie for all peptides then running all the proteins (in parallel) through it? I ported a C# Aho-Corasick implementation to C++ and it worked quite well so I expect you could re-use the original C# implementation: http://tomasp.net/blog/ahocorasick.aspx/

pwiz_tools/Skyline/Model/PeptideDocNode.cs
```diff
@@ -120,6 +120,22 @@ protected override IList<DocNode> OrderedChildren(IList<DocNode> children)
 
         public Peptide Peptide { get { return (Peptide)Id; } }
 
+        public PeptideDocNode ChangeFastaSequence(FastaSequence newSequence)
+        {
+            int begin = newSequence.Sequence.IndexOf(Peptide.Target.Sequence, StringComparison.Ordinal);
+            int end = begin + Peptide.Target.Sequence.Length;
```
brendanx67, line 126: Put an Assume here that begin is not -1.

pwiz_tools/Shared/Common/SystemUtil/SubstringFinder.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// <summary>
+        /// Constructs a new SubstringFinder which can look for substrings up to a particular length contained inside of a larger string.
+        /// </summary>
+        public SubstringFinder(string stringToBeSearched, int maxSubstringLength)
+        {
+            _maxLengthInTrie = maxSubstringLength;
+            _stringToBeSearched = stringToBeSearched;
+            if (string.IsNullOrEmpty(stringToBeSearched) || maxSubstringLength == 0)
+            {
+                _root = new TrieNode(0);
```
brendanx67, line 52: TrieNode.Empty?

pwiz_tools/Shared/Common/SystemUtil/SubstringFinder.cs
```diff
@@ -0,0 +1,145 @@
...
+            {
+                // Add the substring which begins at character index "start"
+                var node = _root;
+                int end = Math.Min(substringStart + maxSubstringLength, stringToBeSearched.Length);
+                for (int position = substringStart; position < end; position++)
+                {
+                    bool isLastPosition = position == end - 1;
+                    int indexInChildren = stringToBeSearched[position] - _minCharacter;
+                    var child = node._children[indexInChildren];
+                    if (position == end - 1)
```
brendanx67, line 74: Unused if block?

pwiz_tools/Skyline/EditUI/AssociateProteinsDlg.cs
```diff
@@ -178,36 +190,84 @@ public void UseFastaFile(string file)
...
             var peptidesForMatching = ListPeptidesForMatching();
-            using (var reader = new StreamReader(fastaFile))
+            if (peptidesForMatching.Count == 0)
+            {
+                return new List<KeyValuePair<FastaSequence, List<IdentityPath>>>();
+            }
+
+            int maxLength = peptidesForMatching.Max(peptide => GetPeptideSequence(peptide).Length);
+            long streamLength = fastaFile.Length;
+            var proteinAssociations = new List<Tuple<int, FastaSequence, List<IdentityPath>>>();
```
brendanx67, line 203: I really hate Tuple. Can you just define a struct with named properties for this? It makes the code so much easier to read when you return to it later or for the first read for someone not as immersed in your intent.

pwiz_tools/Skyline/EditUI/AssociateProteinsDlg.cs
```diff
@@ -178,36 +190,84 @@ public void UseFastaFile(string file)
...
+                            // TODO(yuval): does digest matter?
+                            if (!substringFinder.ContainsSubstring(GetPeptideSequence(peptide)))
+                            {
+                                continue;
+                            }
+
+                            matches.Add(peptide);
+                        }
+
+                        lock (proteinAssociations)
```
brendanx67, line 230: I prefer not locking in these and instead preallocating an array the size of your fastaRecords. Inside the loop just assign your matches to that array without locking. When the loop finishes you can use a single-threaded loop through that array to build this list.

pwiz_tools/Skyline/EditUI/AssociateProteinsDlg.cs
```diff
@@ -178,36 +190,84 @@ public void UseFastaFile(string file)
...
+                }
+            }
+
+            // Protein associations may be out of order because of multi-threading, so put them back in order.
+            return proteinAssociations.OrderBy(tuple => tuple.Item1).Select(tuple =>
+                new KeyValuePair<FastaSequence, List<IdentityPath>>(tuple.Item2, tuple.Item3)).ToList();
+        }
+
+        /// <summary>
+        /// Returns tuples of FastaData along with the position in the Stream that the fasta record was found.
```
brendanx67, line 256: A little confusing naming and commenting. The progress is not in this function, but it is important that this function uses "yield return" to parse progressively. Your parallel processing would completely break if someone changed this to return an array. Probably worth making that clearer.

brendanx67 (changes requested):
Looks pretty good. I had a few comments. Thanks for writing a unit test for the new class, and I assume there is an existing test for the overall feature. And that you have profiled all of this against the old code. Thanks for giving it some attention.

PR #1959, opened 2022-01-22
VS2022 build wasn't copying the msparser DLLs

bspratt:
Looks like we missed a place where we needed to check for 14.3 as well as 14.2


chambm:
Why didn't the tests pick that up? Hmm. It was probably a conflict resolution mistake when I had to rearrange commits in order to squash them into the right groups.

bspratt:
I think it's only a problem if you're specifying 14.3, which AFAIK the test machines do not.

chambm:
Oh, right. That makes sense. :facepalm:

PR #1961, opened 2022-01-23
Fix problem where if you add the "TMT-126" to your document the first…

nickshulman:
… time that you start Skyline, it ends up having the incorrect mass 127 instead of the 126 that it is supposed to have.

This only happens if you add the ion the first time you have started Skyline (or right after you have pushed the "Reset" button on the "Edit Special Ions" dialog).
This happens because when Skyline creates the list of special ions, it incorrectly marks them as protonated.
If you exit Skyline, that protonation does not round trip to being written out to user settings.
So, the next time you start Skyline, Skyline will correctly know that these special ions do not have an extra proton on them.
Unfortunately, if you then try to open a document which used the incorrectly protonated special ion, Skyline will refuse to open it because a transition has the incorrect m/z.

This problem seems to have been introduced in Skyline 4.2.


bspratt:
Seems appropriate, assuming these things are inherently charged (but clearly I know nothing about this)

pwiz_tools/Skyline/Test/MeasuredIonTest.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace pwiz.SkylineTest
+{
+    [TestClass]
+    public class MeasuredIonTest : AbstractUnitTest
+    {
+        [TestMethod]
+        public void TestMeasuredIonRoundTrip()
+        {
+            var measuredIonList = new MeasuredIonList();
+            measuredIonList.AddDefaults();
```
bspratt, line 32: Might not hurt to verify the mass and m/z for one or two selected ions?

bspratt (approved):

PR #1962, opened 2022-01-23
Skyline (21.2): Orbitrap PRM tutorial for Skyline Online 2022

brendanx67:
PR with work-in-progress for necessary fixes to 21.2 to polish the tutorial steps


pwiz_tools/Skyline/SettingsUI/StartPageSettingsUI.cs
```diff
@@ -53,22 +53,39 @@ public StartPageSettingsUI(SkylineWindow skylineWindow)
         public bool IsIntegrateAll 
         {
             get { return radioBtnQuant.Checked; }
+            set { radioBtnQuant.Checked = true; }
```
nickshulman, line 56: This should use "value" instead of "true".

nickshulman (changes requested):
Looks good except for this one property setter.

PR #1963, opened 2022-01-24
Issue 866: Provide option in Skyline to upload a minimized library to Panorama

vagisha:
Panorama Partners feature request: https://www.labkey.org/Panorama%20Partners/Feature%20Requests/issues-details.view?issueId=44562
Skyline's default behavior when choosing File > Upload to Panorama is to upload the full library. However, some users would like an option to upload a minimized library that contains the targets only. This minimize libraries option is currently only available when doing the File > Share workflow.
- Reuse ShareTypeDlg to determine the ShareType instead of adding new controls to PublishDocumentDlg
- Populate the "Skyline file format" combo with the Skyline versions available for sharing that are also supported on the selected Panorama server
- Add support for the --share-type command-line argument when uploading document to a Panorama server
- Add a Cancel button to the MultiButtonMsgDlg that asks if the document should be uploaded to the last upload location


nickshulman (approved):
Looks good to me

pwiz_tools/Skyline/Util/PanoramaUtil.cs
```diff
@@ -773,32 +776,83 @@ public CacheFormatVersion GetSupportedSkydVersion(FolderInformation folderInfo)
...
 
             if (!cacheVersion.HasValue)
             {
                 // The document may not have any chromatogram data.
                 return shareType;
             }
 
-            CacheFormatVersion supportedVersion = GetSupportedSkydVersion(folderInfo);
+            var supportedVersions = GetSupportedVersionsForCacheFormat(folderInfo, cacheVersion);
+            CacheFormatVersion supportedVersion = supportedVersions.Item1;
```
brendanx67, line 790: Use a custom struct with descriptive field names. There are very few cases where I think a Tuple is an acceptable replacement for named fields, which provide much better in-code documentation. Definitely not as a return type from a public function.

brendanx67 (approved):
One simple change request, and I still want to see the UI demoed, but otherwise, it looks good to me.

pwiz_tools/Skyline/Util/PanoramaUtil.cs
nickshulman: I think that function should be changed to just return a SkylineVersion. The SkylineVersion has a "CacheFormatVersion" in it, and I think that's the CacheFormatVersion that should end up getting used.

PR #1968, opened 2022-01-25
Skip over the wiff2 tests when tests are being run in TestExplorer.

nickshulman:
Added method "IsRunningInTestExplorer" which looks at the class of TestContext to decide whether the tests were launched by TestRunner.
(not sure if there was a better way I could have implemented this).


bspratt:
This seems good, but "...because not running in TestRunner" seems a little cryptic. 

Could that be "... because of known Sciex DLL load order dependency issues that cause problems in TestExplorer. Use SkylineTester/TestRunner if you want to run this test."

bspratt (approved):
Looks great!

PR #1969, opened 2022-01-25
Skyline/work/20220121 new re sharper warnings

bspratt:
The latest ReSharper kicks out a handful of new warnings, one of which appears to point to an actual bug.


pwiz_tools/Skyline/CommandLine.cs
```diff
@@ -2895,8 +2895,8 @@ public bool ImportTool (string title, string command, string arguments, string i
 //                        _out.WriteLine("         tool {0} was not modified.", tool.Title);
                         return true;
                     }
-                    // Ovewrite conflicts
-                    if (resolveToolConflictsBySkipping == false)
+                    // Overwrite conflicts
```
brendanx67, line 2898: The ifs above are a bit strange as well. I would use:

if (r.IsNull)
else if (r.Value)
else

Though, you could also drop the elses, since the blocks all have a return. But it should be consistent.

pwiz_tools/Skyline/Model/DdaSearch/MsFraggerSearchEngine.cs
```diff
@@ -342,7 +342,7 @@ private void EnsureFastaHasDecoys()
 
             // ReSharper disable LocalizableElement 
             IEnumerable<string> commonAffixes = new [] { "decoy", "dec", "reverse", "rev", "__id_decoy", "xxx", "shuffled", "shuffle", "pseudo", "random" };
-            commonAffixes = commonAffixes.Concat(commonAffixes.Select(o => o.ToUpperInvariant()));
+            commonAffixes = commonAffixes.Concat(commonAffixes.Select(o => o.ToUpperInvariant())).ToArray();
```
brendanx67, line 345: Strange that it should need ToArray() for an IEnumberable<string>, but I am sure there are no perf issues in this case.

pwiz_tools/Skyline/CommandLine.cs
```diff
@@ -2548,7 +2548,7 @@ private class LibraryReference : IDisposable
 
             public LibraryReference(Library reference)
             {
-                Reference = Reference;
+                Reference = reference;
```
brendanx67, line 2551: I guess this was only causing a leak, since the only function of LibraryReference was to dispose the library, making it not so surprising that no test or use case ever caught this.

brendanx67 (approved):
Looks good. Thanks. One minor style suggestion.

pwiz_tools/Skyline/Model/DdaSearch/MsFraggerSearchEngine.cs
bspratt: Yeah, I guess "possible" is the keyword in "warning: possible multiple enumeration" but no harm in playing it safe here, I think.

PR #1972, opened 2022-01-26
While working on something else I noticed that the transition list column picker dialog isn't centered…

bspratt:
…, this tidies that up. Purely cosmetic change.

was
image

is
image


brendanx67 (approved):
Noticed that myself while retaking the screenshots for the tutorials that added this form. Thanks for fixing it.

PR #1974, opened 2022-01-28
Added support for heavy copper (Cu' in Skyline formula syntax, Cu65 in adduct syntax) …

bspratt:
… because it appears as an adduct in a list of common mass spec contaminants (MaConDa Contaminants DB https://www.maconda.bham.ac.uk/downloads/MaConDa__v1_0__csv.zip )

Reported by user Sangram


bspratt:
It occurs to me that this could have been handled automatically if we used the information we already have in pwiz.Common.Chemistry,IsotopeAbundances.

For any given element X that we don't already have hard-coded, it's just a matter of identifying the most common stable isotopes and noting them as X', Xmm, X", and Xnn respectively, where mm and nn are the isotope masses rounded to the nearest integer.

I've opened an issue in case we want to revisit this idea.

chambm:
Cu' and Cu65?

bspratt:
> Cu' and Cu65? Ah I see the code has it right. The PR description had me raise an eyebrow. :)

D'oh! Fixed it, thanks

nickshulman (approved):

nickshulman (changes requested):
I guess you also need to update "Skyline_Current.xsd" in order to pass unit tests.

Specifically, copy what is currently "Skyline_Current.xsd" to "Skyline_21.2.xsd", and then modify 
"Skyline_Current.xsd".
Also, add a new version number (21.21) to "DocumentFormat.cs" and a change "SkylineVersion.CURRENT" so that it uses that new version number.

I recommend removing the "&lt;xs:restriction>" from "atomic_symbol_type".
The restrictions that we put in the .xsd should only be for things that would actually cause a problem for Skyline if they were encountered. Basically, things that Skyline would not be able to represent in the document tree, and where Skyline should refuse to open the file.
"IsotopeEnrichments.ReadXml" will happily deal with any unknown elements that might be in that list.

bspratt:
I think we still need the restriction for the moment - if somebody tried to introduce another heavy isotope (e.g. Fe') that we don't explicitly handle, isn't that a problem?

But yeah, one way or another I need to roll the XSD version number. Shouldn't we be moving on to 22.1 rather than 21.2, though?

bspratt:
Of course we *could* drop that restriction if we implemented my aforementioned idea of generating comprehensive heavy isotope support based on pwiz.Common.Chemistry,IsotopeAbundances.

brendanx67:
Let's wait for our users to tell us what they need. This seems to me like a case where adding more flexibility than is interesting to our users could impact ease-of-use. Or at least you would want to also implement "common" versus "all".

bspratt:
>> Let's wait for our users to tell us what they need.

I feel like they have - they need more elements than we though they did, and they need more isotopes than we thought they did. With the elements we just went all in, I think we're now seeing that we should have done so with stable isotopes as well.

>> could impact ease-of-use

I don't think so - I don't propose to add these to our various pick lists. That is to say yes, we'd want to distinguish the common ones we already support from the generated ones.

I don't think it's a change users would notice. It wouldn't get added to any picklists etc.

bspratt:
Are we good with this as it stands? The debate about comprehensive handling of less common isotopes can be tabled - I have entered an issue for that as https://skyline.ms/issues/home/issues/details.view?issueId=868

nickshulman (approved):
Looks good to me.
I would recommend that you change the definition of "atomic_symbol_type" in the .xsd so that it allows all strings. It will make it easier the next time we want to add support for a new element.

bspratt:
> I would recommend that you change the definition of "atomic_symbol_type" in the .xsd so that it allows all strings. It will make it easier the next time we want to add support for a new element.

My first thought was "might as well just remove it then", and then I noticed that it doesn't actually seem to be used anywhere, so I guess I will. (If it was actually in use, I'd probably replace it with a regex though)

bspratt:
BTW, Nick, on Monday's lab meeting call I think I understood you to be against adding any more ' and " notations - but I'm not sure what you see as the alternative? 

When I think about it, it's quite surprising that there doesn't seem to be any standard way to represent this in plaintext formulas.

nickshulman:
> BTW, Nick, on Monday's lab meeting call I think I understood you to be against adding any more ' and " notations - but I'm not sure what you see as the alternative?
> 
> When I think about it, it's quite surprising that there doesn't seem to be any standard way to represent this in plaintext formulas.

If we allowed spaces in the chemical formula, then you could write (for very heavy glucose):
C6 2H12 18O6
but, without spaces there's no way to tell that there are not 62 carbons.

bspratt:
Yes, or perhaps + signs as seen in adduct descriptions (though there it's all turned around, e.g. [M+6C+12H2+6O18]).

But I think we'd want to carry on with the ' and " scheme, for backward compatibility, and at that point it would be quite confusing if it worked for some elements but not others.

In any event, we could programmatically add support for the various representations using the isotope abundance table already in our code.

bspratt:
All that aside, are you good with the changes around schema contents and version numbers? If so I'll merge.

PR #1975, opened 2022-01-29
Skyline/work/20220127 missing text in transition list error dlg

bspratt:
Fixed a few problems with the error dialog for small molecule transition lists:

1) Only the first error was shown. This was irritating when dealing with long, messy transition lists that needed repeated editing.
2) The "Show / hide line text" checkbox had no effect (i.e. the original line data did not appear even though the column was shown)
3) Line numbers were incorrect.
4) The placement of the "OK" button was not making good use of screen real estate.

was:
image

is now:
image

### DIFF
--- a/pwiz_tools/Skyline/Model/Import.cs
+++ b/pwiz_tools/Skyline/Model/Import.cs
@@ -3361,7 +3361,7 @@ namespace pwiz.Skyline.Model
         }
     }
 
-    public class TransitionImportErrorInfo
+    public class TransitionImportErrorInfo : IEquatable<TransitionImportErrorInfo>
     {
         public long? LineNum { get; private set; }
         public int? Column { get; private set; }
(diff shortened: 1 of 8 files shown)


brendanx67:
Just looking at the "now" image, I am not clear why rows 2 and 3 have the prefix text "Error on line n:". Seems like these also should have this redundant text removed. Just as in row 1, the line and column number of the error appear in devoted columns in the grid.

bspratt:
Yes, that's historical accident. I can clean that up too.

brendanx67:
> Yes, that's historical accident. I can clean that up too.

It would be good to be consistent since you fixed it for row 1.

bspratt:
I didn't actually do anything special for row 1, it was already like that. I think we just didn't notice the inconsistency since they weren't ever side by side.

bspratt:
OK, de-crufted:

image

bspratt:
If I can get an amen I'll go ahead and merge this.

pwiz_tools/Skyline/Model/Import.cs
```diff
@@ -3375,6 +3375,43 @@ public TransitionImportErrorInfo(string errorMessage, int? columnIndex, long? li
             Column = columnIndex + 1;   // 1 based column number for reporting to a user
             LineNum = lineNum;
         }
+
+        public bool Equals(TransitionImportErrorInfo other)
```
brendanx67, line 3379: Are the equals functions used? We try to stay away from the operator overrides in general. I imagine ReSharper just added them automatically for you now. Hopefully they can be turned off in the future. I have always used if (Equals(a, b)) on Skyline.

brendanx67 (approved):
The screenshot looks good now. A few minor comments on the code. If you are not actually using TransitionImportErrorInfo equality, it may be best to just remove that again. Otherwise, just remove the operator overrides.

pwiz_tools/Skyline/TestFunctional/PasteMoleculesTest.cs
```diff
@@ -81,11 +81,11 @@ public void TestPasteMolecules()
                     {
                         foreach (var err in errDlg.ErrorList)
                         {
-                            allErrorText += err.ErrorMessage;
+                            allErrorText += err.ErrorMessage + "\n";
```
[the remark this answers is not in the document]
bspratt: Agreed.

pwiz_tools/Skyline/Model/Import.cs
bspratt: I added these purely for use in the test, and yes, ReSharper generated that code.

bspratt: I'll remove the == and != overrides.

PR #1977, opened 2022-01-30
Skyline/work/20220129 nicksh fix ratios in targets tree

nickshulman:
1. Ratio to global standards was never being displayed
2. Ratio to heavy would only be displayed if the user managed to choose it by clicking on the Area Graph pane

The original behavior of earlier versions of Skyline has been restored, where the Targets tree will always be displaying either the ratio to a label or ratio to global standards.

I am planning on adding to "TargetRatiosTest" so that it tests things like choosing different normalization methods in the Area graph pane, and that only changes what is displayed in the Targets tree if the new choice is either a ratio to label or ratio to global standards.


brendanx67 (approved):
Looks reasonable. Much easier for me to test once it is merged into the release branch, which is where I am working on a tutorial for Skyline Online 2022.

brendanx67:
Note that I have updated PR#1962 to include the steps where I was hoping to see Ratio to Heavy values in the Targets view.

nickshulman:
This change has now been cherry picked into PR#1962

PR #1984, opened 2022-02-03
Fix a few issues around VS2022 upgrade (presumably)

bspratt:


chambm:
Crap I think my GitHub edit changed all the line endings. Should we revert that or is it worth the diff to get consistent line endings?

bspratt:
Ugh! Yeah, let's fix that so the squash merge to master shows just the critical line changes. I'll take care of it.

bspratt:
Actually it looks like the line endings got dinged here:
SHA-1: 7d38e60f3ab4f506845727269a6850c32a8af575

* * fixed GCC compile with msparser 2.8 (needed for new Linux build agent)
* refactored .NET version requirements (that use MSVC version as a proxy) into a single rule in Jamroot.jam
* updated logic in smartBuildTrigger.py to not assume username on TeamCity build agent is 'teamcity'
* fixed subset tarball creation to avoid meaningless -C parameter at the end

I'll change them back.

chambm:
Yes, removing the secure-scl-throws rule would require removing anywhere that references it. Not really worth it for this PR I think.

PR #1987, opened 2022-02-05
Performance improvements to "Associate Proteins" option for importing peptide transition lists when a background proteome is in use

bspratt:
- Automatically enable this option (i.e. tick the checkbox on form creation) when there is a background proteome available. This was previously not made automatic for performance reasons.
  -  Initially only perform the operation on the (at most 100) transitions visible in the form, without any check for multiple matches etc. This prevents lag, and avoids confusing interactions due to an only partial evaluation of the transitions.
  -  Do the complete operation (including potential user interaction about multiple matches etc) only when the "Check for errors" or "OK" button is clicked.
  - Retain the results of the completed operation from "Check for errors" rather than redoing it when OK is clicked (unless column selections have changed, of course)


pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
```diff
@@ -195,58 +202,68 @@ private string[] ResolveMatchedProteins(int numWithDuplicates, int numUnmatched,
...
                 {
-                    // If there are headers, add one describing the protein name column we will add
-                    if (Importer.RowReader.Indices.Headers != null)
+                    // If there are headers, add one describing the protein name column we will add - if we haven't already
+                    var hasHeaders = Importer.RowReader.Indices.Headers != null;
+                    if (hasHeaders && Equals(_associateProteinsMode, AssociateProteinsMode.preview))
                     {
                         // Add a header we should recognize
-                        var newHeaders = new string[_originalHeaders.Length + 1];
+                        var newHeaders = new string[_originalColumnIDs.Length + 1];
```
nickshulman, line 256:
```
Importer.RowReader.Indices.Headers = _originalColumnIDs
    .Prepend(Resources.ImportTransitionListColumnSelectDlg_PopulateComboBoxes_Protein_Name)
    .ToArray();
```

pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
```diff
@@ -195,58 +202,68 @@ private string[] ResolveMatchedProteins(int numWithDuplicates, int numUnmatched,
...
+                            newHeaders[i + 1] = _originalColumnIDs[i];
                         }
                         Importer.RowReader.Indices.Headers = newHeaders;
                     }
 
                     // Get a dictionary of peptides under consideration and the proteins they're associated with
-                    var peptides = new List<string>(Importer.RowReader.Lines.Count);
-                    peptides.AddRange(Importer.RowReader.Lines.Select(line => line.ParseDsvFields(Importer.Separator)).Select(fields => fields[Importer.RowReader.Indices.PeptideColumn]));
-                    var stripped = new Dictionary<string, string>();
+                    peptides.AddRange(_originalLines.Skip(hasHeaders ? 1 : 0).Take(count).Select(line => line.ParseDsvFields(Importer.Separator)).Select(fields => fields[_originalPeptideIndex]));
```
nickshulman, line 266: Whenever you have two Select's in a row like this, you can always combine them:
```
peptides.AddRange(_originalLines.Skip(hasHeaders ? 1 : 0).Take(count)
    .Select(line => line.ParseDsvFields(Importer.Separator)[_originalPeptideIndex]));
```

pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
```diff
@@ -1343,59 +1392,82 @@ private void UpdateForm()
         private bool isAssociated; // True if the current transition list contains associated proteins
 
         private void checkBoxAssociateProteins_CheckedChanged(object sender, EventArgs e)
+        {
+            UpdateProteinAssociationState(AssociateProteinsMode.preview, out _);
+        }
+
+        private void UpdateProteinAssociationState(AssociateProteinsMode mode, out bool canceled)
```
nickshulman, line 1399: It's unusual for a "void" function to have an out parameter.
Normally, you would just change return type of this function to be "bool" and get rid of the out parameter.

nickshulman (commented):
Looks good to me.
I found a couple of things that could be simplified.

PR #1990, opened 2022-02-08
Fix the "Install" class so that it can be used in unit tests.

nickshulman:
I wanted to use the class "Install" in pull request #1981, but the class did not work during unit tests.

The class was looking at "Assembly.GetExecutingAssembly()" instead of the Skyline executable, which ends up returning something not useful when running as a unit test.
Also, moved the initializing code into a static initializer. It used to be that "Install.Type" would not return the correct value until after the property "Install.Version" was first queried.

### DIFF
--- a/pwiz_tools/Skyline/Util/Install.cs
+++ b/pwiz_tools/Skyline/Util/Install.cs
@@ -18,6 +18,7 @@
  */
 using System;
 using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


brendanx67 (approved):

chambm:
This class has a confusing name. :) But the changes are good! 👍

PR #1991, opened 2022-02-08
Improve error handling for exporting SCIEX method

kaipot:


nickshulman:
Kaipo, Is there a reason that you aren't using the Exception constructor which takes both a string message and an Exception argument for the innerException? Wouldn't it be useful to preserve the callstack of the original exception?

kaipot:
> Kaipo, Is there a reason that you aren't using the Exception constructor which takes both a string message and an Exception argument for the innerException? Wouldn't it be useful to preserve the callstack of the original exception?



> Kaipo, Is there a reason that you aren't using the Exception constructor which takes both a string message and an Exception argument for the innerException? Wouldn't it be useful to preserve the callstack of the original exception?

The exceptions getting thrown in these cases aren't really helpful for debugging, but I guess it couldn't hurt to include them? I've made that change.

nickshulman (approved):

PR #1993, opened 2022-02-09
Hide molecule-specific reports in proteomics mode

nickshulman:
, and hide peptide-specific reports when in small molecule mode.

Add new default reports "Molecule Quantification" and "Molecule Ratio Results".
Here's what the Reports dropdown at the top of the Document Grid looks like when you are in Small Molecules mode:
image

and here's what the Manage Reports dialog looks like:
image

The peptide-specific reports are greyed out. You can actually bring up the report editor on them (in case you want to change their UI mode). This screenshot also shows a report from the Results Grid ("My Peptide Results Report") which is greyed out and which you would not be able to edit, since that report is designed to be used with the Results Grid.

This pull request also includes some changes where I changed it to use "GetRowSourceInfo" instead of "GetViewInfo", because GetViewInfo does a lot more work than is necessary when all we want to know is whether the report should be shown in the "Reports" dropdown.

It would be nice to add a "Small Molecule Peak Boundaries" report. Currently, in order to import peak boundaries you really do need to have a column called "Peptide Modified Sequence", even if it is for small molecules. We should probably change "ImportPeakBoundaries.cs" so that it allows a small-molecule-appropriate column name, and then design a Small Molecule Peak Boundaries report which uses that name.

### DIFF
--- a/pwiz_tools/Shared/Common/DataBinding/AbstractViewContext.cs
+++ b/pwiz_tools/Shared/Common/DataBinding/AbstractViewContext.cs
@@ -141,11 +141,6 @@ namespace pwiz.Common.DataBinding
 
         protected abstract void SaveViewSpecList(ViewGroupId viewGroupId, ViewSpecList viewSpecList);
 
-        protected RowSourceInfo FindRowSourceInfo(ViewInfo viewInfo)
-        {
-            return FindRowSourceInfo(viewInfo.ParentColumn.PropertyType);
-        }
-
         public ViewInfo GetViewInfo(ViewGroup viewGroup, ViewSpec viewSpec)
         {
             var rowSourceInfo = GetRowSourceInfo(viewSpec);
@@ -170,7 +165,7 @@ namespace pwiz.Common.DataBinding
             return GetViewInfo(FindGroup(viewName.Value.GroupId), viewSpec);
         }
 
-        protected RowSourceInfo GetRowSourceInfo(ViewSpec viewSpec)
+        public RowSourceInfo GetRowSourceInfo(ViewSpec viewSpec)
         {
             return RowSources.FirstOrDefault(rowSource => rowSource.Name == viewSpec.RowSource);
         }
@@ -706,5 +701,10 @@ namespace pwiz.Common.DataBinding
         // SkylineViewContext overrides and uses this event 
 #pragma warning disable 67
         public virtual event Action ViewsChanged;
+
+        public virtual bool CanDisplayView(ViewSpec viewSpec)
+        {
+            return null != GetRowSourceInfo(viewSpec);
+        }
     }
 }
--- a/pwiz_tools/Shared/Common/DataBinding/Controls/Editor/ChooseViewsControl.cs
+++ b/pwiz_tools/Shared/Common/DataBinding/Controls/Editor/ChooseViewsControl.cs
@@ -86,7 +86,7 @@ namespace pwiz.Common.DataBinding.Controls.Editor
                         newGroups.Add(listViewGroup);
                         foreach (var viewSpec in ViewContext.GetViewSpecList(group.Id).ViewSpecs)
                         {
-                            bool validRowSource = null != ViewContext.GetViewInfo(new ViewName(group.Id, viewSpec.Name));
+                            bool validRowSource = ViewContext.CanDisplayView(viewSpec);
                             if (FilterRowSources && !validRowSource)
                             {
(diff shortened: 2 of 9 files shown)


bspratt:
>> It would be nice to add a "Small Molecule Peak Boundaries" report. Currently, in order to import peak boundaries you really do need to have a column called "Peptide Modified Sequence", even if it is for small molecules. We should probably change "ImportPeakBoundaries.cs" so that it allows a small-molecule-appropriate column name, and then design a Small Molecule Peak Boundaries report which uses that name.

There are probably quite a number of supposedly peptide specific reports that are actually applicable to generalized molecules. Duplicating all of them seems like a bad idea, in the same way that duplicating a bunch of UI also seemed like a bad idea when we first added small molecules to Skyline. Is there no way the reports UI could gracefully adapt to show "molecule" instead of "peptide" while still using the same reports where appropriate?

brendanx67:
Many of the peptide reports use "Modified Sequence" which doesn't have a small molecule equivalent.

Can we possibly put "Mixed Transition List" at the bottom? Have it show up below other transition list options? Should it show up at all when you are fully in "molecule mode"? Or only when you are in mixed mode?

nickshulman:
> Can we possibly put "Mixed Transition List" at the bottom?

The reports are in alphabetical order.

> Should it show up at all when you are fully in "molecule mode"? Or only when you are in mixed mode?

I think all reports should show up when you are in mixed mode. Is "Mixed Transition List" a useful report? Maybe we should remove it from the list of default reports. There isn't really such a thing as a Mixed Transition List that you can paste into Skyline. It always makes you choose "Peptides" or "Molecules"
image


> There are probably quite a number of supposedly peptide specific reports that are actually applicable to generalized molecules. Duplicating all of them seems like a bad idea, in the same way that duplicating a bunch of UI also seemed like a bad idea when we first added small molecules to Skyline. Is there no way the reports UI could gracefully adapt to show "molecule" instead of "peptide" while still using the same reports where appropriate?

We certainly could implement something like that.
Currently, the mode is specified in the report definition, so that, no matter what mode you happen to be in in Skyline, the names of the columns will not change.
I could imagine that users should be able to choose a mode of "dynamic" in which case the Peptide column would be called "Peptide" when you are in proteomics mode, or "Molecule" when you are in Mixed or Small Molecules mode.

bspratt:
>> Many of the peptide reports use "Modified Sequence" which doesn't have a small molecule equivalent.

But I wonder how many actually use it as anything more than an identifier that couldn't just be a molecule's display name instead?

brendanx67:
Well, Nick just did the translation for 2 reports from peptides to molecules. Nick is there a way that we can mark our default templates as "auto-translatable" between peptides and molecules to avoid needing to duplicate them?

nickshulman:
> Nick is there a way that we can mark our default templates as "auto-translatable" between peptides and molecules to avoid needing to duplicate them?

There is currently no way to do that, but I could implement such a feature.

bspratt:
>> Currently, the mode is specified in the report definition, so that, no matter what mode you happen to be in in Skyline, the names of the columns will not change.

In the adaptive UI code, UI elements are assumed to be candidates for the "peptide"=>"molecule" treatment unless tagged as being explicitly one or the other. We'd want to do that in reports as well, I think. So a report tagged proteomic wouldn't show at all in Molecules mode, but one not tagged at all would show up and would get the "peptide"=>"molecule" treatment.

bspratt:
>>  Is "Mixed Transition List" a useful report? 
Not for reimport to Skyline, no. But possibly useful for other things downstream from us as a general info dump.

brendanx67:
Probably we should do some auto-translation. I know we need to add a new default reports which expose intensity values pivoted by replicate, which can then easily be fed into heatmap display. Not totally sure which values we want in this format of precursor peak area, peptide peak area, protein peak area, normalized peak area, or quantitative value. But, we got dinged for not having this report in the most recent software comparison paper.

brendanx67:
> > > Is "Mixed Transition List" a useful report?
> 
> Not for reimport to Skyline, no. But possibly useful for other things downstream from us as a general info dump.

Sounds pretty speculative. Did you ever have a request for this or have any reason to believe anyone has used it?

bspratt:
Yes, speculative. I personally find that report useful for looking at various Skyline documents, but I'm a developer. I doubt it would be missed by the general public.

bspratt (approved):
Looks good,

PR #1994, opened 2022-02-09
Fixed some inconsistent handling of attempts to add an empty transiton list to Skyline

bspratt:
ExceptionWeb was reporting "Failed to find peptide column., line 1." exceptions from Edit>Insert>TransitionList that would not occur when the same text  was pasted into the Targets window.

(e.g. just pasting in "Precursor Name, Precursor Formula, Precursor Charge, Precursor Adduct, Precursor m/z" with no lines of actual data following)


nickshulman (approved):
Looks good to me.

PR #1998, opened 2022-02-12
Skyline: Fix chromatogram display when optimization data is asymmetri…

brendanx67:
…c as when CE values go to zero and below


nickshulman (approved):
Looks good to me.

PR #2003, opened 2022-02-15
Fixed a problem reading a transition list where, for example, C2H6N[M+H] and C2H6N[M2C13-H] were incorrectly treated as an m/z ambiguity…

bspratt:
They are different polarities so the m/z match is not actually ambiguous.

Reported by Lisa.


bspratt:
Easy backport to Release if we want it.

bspratt:
Yeah, that's tidier. I'll do that. Damn, you're good!

PR #2009, opened 2022-02-23
Fixed an annoying problem with the "Modify..." right-click menu item …

bspratt:
…in the Targets tree for small molecules. When adding isotope information to an adduct description, the cursor tended to jump back to the start of the text while you were typing.

Not reported by any user, but observed in more than one Skyline training.

Developers note: also improved error checking for mass multipliers (e.g. the "2" in "[2M+H]") which must be postive integers. Things like "[xM+H]" or "[-2M+H]" were not being properly flagged.


pwiz_tools/Skyline/SettingsUI/FormulaBox.cs
```diff
@@ -420,6 +422,15 @@ private void AddFormulaSymbol(string symbol)
             textFormula.SelectionStart = insertAt + symbol.Length;
         }
 
+        private void SetFormulaText(string text)
+        {
+            // Preserve cursor location
```
nickshulman, line 427: I am not sure if it would make a difference here, but you might want to add:
`if (text == textFormula.Text) return;`
so that nothing happens if the text is not actually changing.

nickshulman (approved):

pwiz_tools/Skyline/SettingsUI/FormulaBox.cs
bspratt: Certainly can't hurt.

PR #2012, opened 2022-02-23
Skyline: New diagnostics for ExistingExperimentsTutorialTest failures…

brendanx67:
… with TODOs for Brian Pratt

Race condition causes OkDialog(colDlg, colDlg.OkDialog); to produce a document with a single peptide list:

image


brendanx67:
Can you also add code to cause the test to fail immediately in the right place with an explanatory message if OkDialog() is called when AssociateProteins is required but not complete? Something that would do a much better job of catching this case, if someone writes a new test but forgets to add the WaitForConditionUI() that you have now added to the 3 tests with this issue.

PR #2013, opened 2022-02-24
Use maximum chromatogram intensity as Area for MS2 transitions when acquisition method is DDA

nickshulman:
In order to make it so that the Library Dot Product is a useful number, we have decided that when the acquisition method is DDA, the Area of the MS2 transitions will be based on the intensity at the best point on the chromatogram between the peak boundaries. The best point is determined by finding the point at which the sum of the intensities across the MS2 transitions is highest.


pwiz_tools/Skyline/Model/Results/PeakIntegrator.cs
```diff
@@ -104,6 +118,59 @@ private ChromPeak IntegratePeakWithoutBackground(float startTime, float endTime,
             return new ChromPeak(RawTimeIntensities ?? InterpolatedTimeIntensities, startTime, endTime, flags);
         }
 
+        /// <summary>
+        /// If the Acquisition Method is DDA, then change the Area of the peak to
+        /// be the intensity at the time point with the highest total MS1 intensity
```
brendanx67, line 123: The code looks like you are using highest total MS2 intensity. So, I think this comment is incorrect.

pwiz_tools/Skyline/Model/Results/ChromHeaderInfo.cs
```diff
@@ -1029,7 +1029,7 @@ public struct ChromPeak : ISummaryPeakData
         private readonly float _retentionTime;
         private readonly float _startTime;
         private readonly float _endTime;
-        private readonly float _area;
+        private float _area;
```
brendanx67, line 1032: I think more than just the area needs to change. Background needs to become zero. Height and area should become equal. FWHM also needs to become zero. Mass error needs to become the mass error at the single point, points across peak should be 1. And the RT needs to become the RT of the single spectrum for all transitions, not the time at which each has its maximum intensity within start-end time.

brendanx67 (changes requested):
Looks like a good start. I don't think you quite have all of the peak statistics right yet. It looks like this just corrects Area. Also, I think you are going to need more complete testing of this change.

brendanx67:
I think we can abandon this one in favor of using area without background subtraction in PR#2034

nickshulman:
Closing this pull request since PR #2034 ended up doing much better.

PR #2018, opened 2022-03-03
Added a new standard report "Small Molecule Peak Boundaries" …

bspratt:
… for use in export of peak boundaries for small molecule documents, and updated the peak boundary import code to handle small molecule information.

### DIFF
--- a/pwiz_tools/Skyline/Model/ImportPeakBoundaries.cs
+++ b/pwiz_tools/Skyline/Model/ImportPeakBoundaries.cs
@@ -99,6 +99,16 @@ namespace pwiz.Skyline.Model
         public static int[] REQUIRED_NO_CHROM { get { return REQUIRED_FIELDS.Take(2).ToArray(); }}
 
         // ReSharper disable LocalizableElement
+
+        public static readonly string[] PEPTIDE_SYNONYMS = new[]
+        {
+            "PeptideModifiedSequence", "ModifiedSequence", "FullPeptideName", "EG.ModifiedSequence", ColumnCaptions.PeptideModifiedSequence, ColumnCaptions.ModifiedSequence
+        };
+        public static readonly string[] MOLECULE_SYNONYMS = new[]
+        {
+            "Molecule", "MoleculeName", ColumnCaptions.Molecule, ColumnCaptions.MoleculeName
+        };
+
         // NOTE: The first name is what appears in error messages about missing required fields
         public static string[][] FIELD_NAMES
         {
@@ -108,12 +118,13 @@ namespace pwiz.Skyline.Model
                 // a whole file dwarfs newing up this array.
                 return new[]
                 {
-                    new[] {"PeptideModifiedSequence", "ModifiedSequence", "FullPeptideName", "EG.ModifiedSequence", ColumnCaptions.PeptideModifiedSequence, ColumnCaptions.ModifiedSequence},
+                    PEPTIDE_SYNONYMS.Concat(MOLECULE_SYNONYMS).ToArray(),
                     new[] {"FileName", "filename", "align_origfilename", "R.FileName", ColumnCaptions.FileName},
                     new[] {"Apex", "RetentionTime", "BestRetentionTime", "RT", ColumnCaptions.RetentionTime, ColumnCaptions.BestRetentionTime},
                     new[] {"MinStartTime", "leftWidth", ColumnCaptions.MinStartTime},
                     new[] {"MaxEndTime", "rightWidth", ColumnCaptions.MaxEndTime},
-                    new[] {"PrecursorCharge", "Charge", "FG.Charge", ColumnCaptions.PrecursorCharge},
+                    new[] {"PrecursorCharge", "Charge", "FG.Charge", ColumnCaptions.PrecursorCharge,
+                        "Adduct", "PrecursorAdduct", ColumnCaptions.PrecursorAdduct, ColumnCaptions.PrecursorAdduct},
                     new[] {"PrecursorIsDecoy", "Precursor Is Decoy", "IsDecoy", "decoy", ColumnCaptions.IsDecoy},
                     new[] {"SampleName", ColumnCaptions.SampleName},
                     new[] {"DetectionQValue", "m_score", ColumnCaptions.DetectionQValue},
@@ -221,6 +232,11 @@ namespace pwiz.Skyline.Model
             var requiredFields = changePeaks ? REQUIRED_FIELDS : REQUIRED_NO_CHROM;
             char correctSeparator = ReadFirstLine(line, allFieldNames, requiredFields, out fieldIndices, out fieldsTotal);
 
+            // Determine whether the input list is proteomic or small molecule
+            var header = new DataFields(fieldIndices, line.ParseDsvFields(correctSeparator), allFieldNames);
+            var peptideColumnName = header.GetField(Field.modified_peptide);
+            var listIsProteomic = !MOLECULE_SYNONYMS.Any(s => string.Equals(s, peptideColumnName, StringComparison.CurrentCultureIgnoreCase));
+
             while ((line = reader.ReadLine()) != null)
             {
                 linesRead++;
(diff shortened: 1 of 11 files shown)


pwiz_tools/Skyline/Model/PersistedViews.cs
```diff
@@ -468,6 +492,17 @@ public override void WriteXml(XmlWriter writer)
   </view>
 </views>";
 
+        private const string REPORTS_V12 = @"<views>
+  <view name='Molecule Peak Boundaries' rowsource='pwiz.Skyline.Model.Databinding.Entities.Precursor' sublist='Results!*' uimode='small_molecules'>
+    <column name='Results!*.Value.PeptideResult.ResultFile.FileName' />
+    <column name='Peptide.MoleculeName' />
```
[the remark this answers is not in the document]
bspratt: Yeah, that's nicer UI for sure
image

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/Model/ImportPeakBoundaries.cs
```diff
@@ -99,6 +99,16 @@ public enum Field { modified_peptide, filename, apex_time, start_time, end_time,
         public static int[] REQUIRED_NO_CHROM { get { return REQUIRED_FIELDS.Take(2).ToArray(); }}
 
         // ReSharper disable LocalizableElement
+
+        public static readonly string[] PEPTIDE_SYNONYMS = new[]
+        {
+            "PeptideModifiedSequence", "ModifiedSequence", "FullPeptideName", "EG.ModifiedSequence", ColumnCaptions.PeptideModifiedSequence, ColumnCaptions.ModifiedSequence
+        };
+        public static readonly string[] MOLECULE_SYNONYMS = new[]
```
nickshulman, line 107: This should not be a statically initialized array like this, since it contains localized strings (ColumnCaptions.Molecule and ColumnCaptions.MoleculeName).
Maybe make it into a static property that always returns a new array.

nickshulman (changes requested):

pwiz_tools/Skyline/Model/ImportPeakBoundaries.cs
bspratt: I'm not sure that's a problem - don't we require a restart when user changes the language setting? Here is is running japanese, in testrunner.exe

image

brendanx67:
It is a problem for testing which switches languages without restarting the
process, and so leaves static values unchanged. It is a classic mistake we
have had to fix often to maintain our ability to run multiple tests under
translated languages in the same test run.

bspratt:
Noted, thanks.

bspratt:
FWIW where I've bothered to think about L10N column headers, I've gone all in and made it recognize all supported languages so a en-US process can handle CSV output from a ja (etc) process. Arguably overkill here.

brendanx67:
I think that is a good direction for us to be heading. Thanks, Brian. Last year, I also improved our transition list / assay library import to successfully recognize and import CSV files from any locale (comma separated or semi-colon separated) regardless of the current system locale. More agnosticism in recognizing the file formats users may provide us is a good thing.

pwiz_tools/Skyline/Model/ImportPeakBoundaries.cs
```diff
@@ -100,26 +101,57 @@ public enum Field { modified_peptide, filename, apex_time, start_time, end_time,
...
             {
-                return new[] 
+                if (_peptide_synonyms == null)
                 {
-                    "PeptideModifiedSequence", "ModifiedSequence", "FullPeptideName", "EG.ModifiedSequence",
-                    ColumnCaptions.PeptideModifiedSequence, ColumnCaptions.ModifiedSequence
-                };
+                    var headers = new List<string>()
+                    {
+                        "PeptideModifiedSequence", "ModifiedSequence", "FullPeptideName", "EG.ModifiedSequence",
```
brendanx67, line 113: Are the first 2 entries duplicated by the code below? The old code would have duplicated them also, but only when running with English settings. Otherwise, it allowed us to support both English and the current translation, but not any translated header from any language. With the new code, I think you only want the 3rd and 4th values in this initializer.

pwiz_tools/Skyline/Model/ImportPeakBoundaries.cs
```diff
@@ -100,26 +101,57 @@ public enum Field { modified_peptide, filename, apex_time, start_time, end_time,
...
             get
             {
-                return new[] 
+                if (_molecule_synonyms == null) 
                 {
-                    "Molecule", "MoleculeName", ColumnCaptions.Molecule, ColumnCaptions.MoleculeName
-                };
+                    var headers = new List<string>()
+                    {
+                        "Molecule", "MoleculeName"
```
brendanx67, line 140: Again, I don't see the need for these values which seem to match the English values for the resources used below.

pwiz_tools/Skyline/Model/ImportPeakBoundaries.cs
```diff
@@ -100,26 +101,57 @@ public enum Field { modified_peptide, filename, apex_time, start_time, end_time,
...
-                };
+                    var headers = new List<string>()
+                    {
+                        "PeptideModifiedSequence", "ModifiedSequence", "FullPeptideName", "EG.ModifiedSequence",
+                    };
+                    // Also recognize column captions in all supported display languages
+                    var currentUICulture = Thread.CurrentThread.CurrentUICulture;
+                    foreach (var culture in CultureUtil.AvailableDisplayLanguages())
+                    {
+                        Thread.CurrentThread.CurrentUICulture = culture;
```
brendanx67, line 119: There is disposable CurrentCultureSetter that does this.

pwiz_tools/Skyline/Model/ImportPeakBoundaries.cs
```diff
@@ -100,26 +101,57 @@ public enum Field { modified_peptide, filename, apex_time, start_time, end_time,
...
-                };
+                    var headers = new List<string>()
+                    {
+                        "Molecule", "MoleculeName"
+                    };
+                    // Also recognize column captions in all supported display languages
+                    var currentUICulture = Thread.CurrentThread.CurrentUICulture;
+                    foreach (var culture in CultureUtil.AvailableDisplayLanguages())
+                    {
+                        Thread.CurrentThread.CurrentUICulture = culture;
```
brendanx67, line 146: You can use CurrentCultureSetter here instead.

brendanx67 (approved):
Minor suggestions. Looks good.

pwiz_tools/Skyline/Model/ImportPeakBoundaries.cs
bspratt: I don't believe so, no. "PeptideModifiedSequence" vs "Peptide Modified Sequence".

pwiz_tools/Skyline/Model/ImportPeakBoundaries.cs
bspratt: Same deal -  "MoleculeName" vs  "MoleculeName".  "Molecule" might show up twice though, I guess I can take that out.

pwiz_tools/Skyline/Model/ImportPeakBoundaries.cs
bspratt: Thanks. It seemed like a lot of code duplication.

pwiz_tools/Skyline/Model/ImportPeakBoundaries.cs
bspratt: Presumably from others, yes, and also probably a good guess. I do a similar thing in transition list reading - accept versions of our own known headers with spaces stripped out.  

Also I just realized that there are other columns that need the all languages treatment.

nickshulman:
Maybe it would be good to write a function "GetLocalizedColumnNames" which returns all of the localized forms of a particular column name:
```
        private static IEnumerable<string> GetLocalizedColumnNames(string invariantName)
        {
            return CultureUtil.AvailableDisplayLanguages()
                .Select(culture => ColumnCaptions.ResourceManager.GetString(invariantName, culture))
                .Prepend(invariantName);
        }
```

Then you could do:
```
                    var headers = new List<string>(
                        GetLocalizedColumnNames(nameof(ColumnCaptions.PeptideModifiedSequence))
                            .Concat(GetLocalizedColumnNames(ColumnCaptions.ModifiedSequence))
                            .Concat(new[] {"FullPeptideName", "EG.ModifiedSequence"})
                            .Distinct()
                    );
```

pwiz_tools/Skyline/Model/ImportPeakBoundaries.cs
bspratt: Hmm, that doesn't set the UI culture. Is that a bug, or should I make a CurrentCultureUISetter class as well?

nickshulman: I always use the method "LocalizationHelper.CallWithCulture" instead of "CurrentCultureSetter".

brendanx67:
Probably is a bug. Looks like only MSAmanadaWrapper uses the disposable.
CallWithCulture looks cool. I just went looking because I figured we must
have implemented something for this. Apparently we have implemented 2 ways.
I prefer the one Nick suggests for your case. You probably ought to fix
Viki's as well.

bspratt:
>> You probably ought to fix Viki's as well.

OK, will do.

bspratt:
> Maybe it would be good to write a function "GetLocalizedColumnNames" which returns all of the localized forms of a particular column name...

That's less code, which is usually a good thing, but I don't think it's nearly as readable as the current implementation.

bspratt:
> I prefer the one Nick suggests for your case.

I dunno, the using thing is nice here because I'm not calling a function with a return value. It reads pretty cleanly.

brendanx67:
Fine with me. Just trying to make sure we reuse code where appropriate.

PR #2019, opened 2022-03-04
* added FAIMS support to SeeMS

chambm:
* added FAIMS support to SeeMS (CV is put in the ion mobility column but ion mobility heatmaps are not generated)


brendanx67 (approved):
Looks good. Worked for me. Very helpful for my current work with Thermo.

PR #2020, opened 2022-03-04
Changed small molecule transition list import to create precursors with "auto manage children" turned off.…

bspratt:
… Formerly they were created with this turned on, which could lead to some surprising behavior.

Background: For small molecules, Skyline initially constructs a document with precisely the contents of the transition list. But it was possible for a settings change to alter the document contents, for example creating isotope peaks for molecules with provided chemical formulas - even if the settings change had nothing to do with isotope peak counts.

Now if, for example, you do want the isotope peak count setting to be used to automatically create isotopic peak precursors, you must use the Refine > Advanced menu item and check the "Auto-select all: Precursors" box to enable "Auto manage children" for your document.

Note that small molecule fragments are unaffected by the "Auto manage children" setting since, unlike peptides, Skyline does not have fragment prediction for generalized molecules.


pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
```diff
@@ -1742,7 +1742,7 @@ private PeptideGroupDocNode GetMoleculePeptideGroup(SrmDocument document, Row ro
             if (String.IsNullOrEmpty(name))
                 name = document.GetSmallMoleculeGroupId();
             var metadata = new ProteinMetadata(name, String.Empty).SetWebSearchCompleted();  // FUTURE: some kind of lookup for small molecules
-            return new PeptideGroupDocNode(pepGroup, metadata, new[] { pep });
+            return new PeptideGroupDocNode(pepGroup, metadata, new[] { pep }, false);
```
nickshulman, line 1745: I wonder whether we should just leave this as true, since having automanage turned on for a Molecule List never affects anything.

pwiz_tools/Skyline/TestFunctional/PasteMoleculesTest.cs
```diff
@@ -1587,6 +1591,77 @@ private void TestUnsortedMzPrecursors()
...
+            // Because we created nodes with auto manage off, changing these settings should not change the nodes (1 precursor, two fragments)
+            RunUI(() => SkylineWindow.ModifyDocument("Change isotope peaks count and ion types",
+                doc => doc.ChangeSettings(doc.Settings.ChangeTransitionFullScan(fs =>
+                        fs.ChangePrecursorIsotopes(FullScanPrecursorIsotopes.Count, 3, IsotopeEnrichmentsList.GetDefault())).
+                    ChangeTransitionFilter(f => f.ChangeSmallMoleculeIonTypes(new[] { IonType.custom, IonType.precursor })))));
+            pastedDoc = WaitForDocumentChange(pastedDoc);
+            AssertEx.IsDocumentState(SkylineWindow.Document, null, 1, 1, 1, 3);
+
+            var managed = false;
+            for (var loop = 0; loop < 2; loop++)
```
nickshulman, line 1618: You could get rid of the variable "loop" if you did:
`for (var managed in new[] { false, true })`

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
bspratt: Not for the foreseeable future, no. But I feel like having it off kind of reinforces that expectation.

pwiz_tools/Skyline/TestFunctional/PasteMoleculesTest.cs
bspratt: Yes, tidier, thanks.

bspratt:
I need to update a few tutorials before this is properly considered complete.

bspratt:
Turns out there's just one tutorial affected by this change - if I can get signoff on the English version I'll update the outgoing L10N versions,  See pwiz_tools/Skyline/Documentation/Tutorials/Skyline Hi-Res Metabolomics.docx

brendanx67:
I added a change to the issue for this PR:

https://skyline.ms/issues/home/issues/details.view?issueId=871

I think the change would be good enough to save a fair bit of confusion on small molecule tutorials that use precursor ions. Let's really try to get this done now.

bspratt:
> I added a change to the issue for this PR:
> 
> https://skyline.ms/issues/home/issues/details.view?issueId=871
> 
> I think the change would be good enough to save a fair bit of confusion on small molecule tutorials that use precursor ions. Let's really try to get this done now.

OK, now if the document settings are such thatAuto-select would have an effect, the user sees a dialog like this before the new transitions are created:
image

Tutorial has been update, but I'll hold off updating the Chinese and Japanese until this is approved.

bspratt:
@brendanx67 Hi_Res Metabolomics tutorial (including the never-translated Chinese and Japanese versions) updated, this is good to go if you're happy with the revised text.

PR #2024, opened 2022-03-07
try to add some information to understand the fairly frequently reported exception…

bspratt:
… "Value of '∞' is not valid for 'emSize'. 'emSize' should be greater than 0 and less than or equal to System.Single.MaxValue." in AxisLabelScaler.ScaleAxisLabels

### DIFF
--- a/pwiz_tools/Shared/zedgraph/ZedGraph/FontSpec.cs
+++ b/pwiz_tools/Shared/zedgraph/ZedGraph/FontSpec.cs
@@ -358,18 +358,26 @@ namespace ZedGraph
 			{
 				if ( value != _size )
 				{
-					if (_size == 0)
+					try
 					{
-						Font newFont = null;
-						Remake(value, 1, ref _scaledSize, ref newFont);
-						_font = newFont; 
+						if (_size == 0)
+						{
+							Font newFont = null;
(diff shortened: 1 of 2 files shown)


pwiz_tools/Shared/zedgraph/ZedGraph/FontSpec.cs
```diff
@@ -358,18 +358,26 @@ public float Size
...
+						}
+						_size = value;
 					}
-					else
+					catch (Exception e)
 					{
-						Remake(_scaledSize / _size * value, _size, 
-							ref _scaledSize, ref _font);
+						throw new Exception(string.Format(@"Unable to set FontSpec.Size to {0} for _size {1} _scaledSize {2}", value, _size, _scaledSize), e);
+						throw;
```
nickshulman, line 379: You have two "throw" statements here.

nickshulman (changes requested):

pwiz_tools/Shared/zedgraph/ZedGraph/FontSpec.cs
bspratt: Oops, yeah, thanks

nickshulman (approved):
Looks good.

PR #2029, opened 2022-03-14
Future-proof SkylineTester's code for detecting minimum Visual Studio installation:…

bspratt:
- Changed the minimum supported version from VS2013 to VS2017
- Look for and prefer versions as high as VS2040 (likely to break at some point, of course, but hopefully good for another few releases anyway)

### DIFF
--- a/pwiz_tools/Skyline/SkylineTester/Main.cs
+++ b/pwiz_tools/Skyline/SkylineTester/Main.cs
@@ -594,7 +594,7 @@ namespace SkylineTester
 
                 if (Devenv == null)
                 {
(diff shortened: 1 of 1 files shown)


bspratt:
There are probably more canoncical ways to do this, I'm just following our established pattern.

Big question is what minimum VS version do we want to insist on? For our own purposes we're talking about 2022 but do we want to impose that on anyone else if it isn't strictly necessary?

chambm (approved):
Looks fine to me, and 2040 seems a reasonable choice. Civilization's predicted to collapse by then, right? 👍 I'll have my 99 luftballoons ready!

PR #2033, opened 2022-03-15
Parallel mode in TestRunner for distributing tests between clients

chambm:
* added parallelmode (off/server/client) in TestRunner for distributing tests between clients
* added NetMQ dependency from NuGet

Normally, Windows Docker containers can't run GUI stuff because they run in session 0. So the trick to running this new mode for functional tests (the set that could benefit most from parallelism, but maybe tutorials as well) is a tool I found called [AlwaysUp ](https://www.coretechnologies.com/products/AlwaysUp/) (it's primarily targeted at GUI users but it has a command-line version, AlwaysUpCLT). It runs pretty much any program as a Windows service, so it can run TestRunner's functional tests on session 0. The only issue I've found is that crux fails pretty much 100% of the time when running under a Windows Docker container (so not related to AlwaysUp), but it fails non-deterministically, and I haven't been able to figure out why. So I added a new attribute to mark TestMethods that should run on the host instead of in a container.

I've also added a way for TestRunner in server mode to launch the Docker workers itself. That is certainly the most convenient way to run this new capability. 

However, AlwaysUp isn't free software. Although it doesn't appear to have any license enforcement mechanism in the CLT tools, we should license it. According to an email from them, they don't usually individually license the CLT tools (maybe due to the lack of an enforcement mechanism!), but they said they'd make an exception for us, and that it would be the same price as the GUI: $50 for one "user". ~~I haven't asked them how "users" would work with TeamCity build agents although I did tell them from the outset that I was trying to get it to work for CI.~~ I have bought 2 licenses now. They said they don't have a concept of a "floating" license and that the license is tied to the machine it is installed on. I'm going to interpret that as it being tied to the container image it is installed in. The AlwaysUp devs actually advised me to try a free competitor of theirs, NSSM, which I did, but it apparently does not have the same magic as AlwaysUp because it can't run TestFunctional in session 0.

From the above, it should be apparent that using this PR depends on having a Docker container with AlwaysUpCLT in it. I can't upload that to a public repository since it's not free software. But I think we could get some kind of private repo going either on Docker Hub or our own servers. And I can of course share that with anybody who wants to test this locally to speed up their tests. But we need to figure the licensing thing out.

The result collection is rudimentary (just capturing the output from each client as it runs its assigned test). I don't currently expect this to be used for nightlies because of how variable it would make things like the memory and handle statistics, but it could certainly be discussed.

The bulk of the time I spent on this was finding an easy network/queue approach. I tried Microsoft Message Queue, but that gave me permission errors in Docker and had to be turned on as a Windows feature. I considered RabbitMQ but didn't want to have to a run a separate server process and I didn't find a way to host it in-process. With NetMQ I was able to keep it all in-process with no external configuration. I'm happy to replace it if anybody's got a simpler idea for a solution - perhaps something that doesn't add any new dependencies.

~~I have only been able to test Docker's Hyper-V isolation because of my version of Windows 10. We may get even better performance by using the process isolation available with Windows 11 (the Windows Server version on our build agents should work as well). Are you interested in testing that @bspratt (you'll have to install Docker but that should not be a big deal and you can run it only when you need it)?~~

~~I'll probably upgrade to Windows 11 if there's a significant boost to performance. From my testing on AWS I'm not sure there would be though.~~
I did upgrade to Windows 11 and the process isolation mode does work, but there's no significant difference in performance and it has a tendency to make the workers hang.


bspratt:
I don't actually have a development environment running on Windows11, sorry for the misunderstanding.

chambm:
Does anybody remember why ExportMethod used standard input instead of files?

brendanx67:
I assume you mean why the transition list text is sent to the method export EXE files through StdIn. Just seemed simpler than dealing with temp files you have to delete and communicate names of to the separate processes.

chambm:
Yeesh, finally. This PR now cuts the functional testing time (`TestFunctional.dll,TestConnected.dll` for `language=en,fr`) in half on TC (from ~20m to ~10). That's on only a 4 core machine (8 AWS vCPUs, which are really hyperthreads I suppose). On my new 16-core machine I get much more benefit. I can run `TestFunctional.dll,TestConnected.dll,TestData.dll,TestTutorial.dll` for `language=en,fr` in 7 minutes, vs. 1 hour without parallelmode. We could use a beefier AWS instance type to get more TC benefit from this PR, but that might start to become too expensive when considering all the idle time: it takes 10-15 minutes just to start one our Windows build agents (for some reason I've never been able to figure out), and there's a 20 minute time out to terminate the instance after it goes idle.

chambm:
I've updated the initial description to put the more important information first.

bspratt:
FWIW here's what I wound up with for setting Tools directory. It seems that a static string won't necessarily get nulled between tests, or at least not soon enough. Here I just cache the part that won't change, which is also the expensive part I expect.

`        private static string _skylineDirPath;
        public static string GetToolsDirectory()
        {
            if (_skylineDirPath == null)
            {
                var skylinePath = Assembly.GetExecutingAssembly().Location;
                Assume.IsFalse(string.IsNullOrEmpty(skylinePath), @"Could not determine Skyline installation location");
                _skylineDirPath = Path.GetDirectoryName(skylinePath);
                Assume.IsFalse(string.IsNullOrEmpty(_skylineDirPath), @"Could not determine Skyline installation directory name");
            }

            // Use a unique tools path when running tests to allow tests to run in parallel
            // ReSharper disable once AssignNullToNotNullAttribute
            return Path.Combine(_skylineDirPath, Program.UnitTest ?
                $@"Tools_{Program.TestName}_{Thread.CurrentThread.CurrentCulture.Name}" :
                @"Tools");

        }`

chambm:
My design was not intended to have the path change between tests. Only between processes. So as I mentioned on your PR (after seeing it was failing for your tests and thinking about why), my design is unsuitable if the same instance of TestRunner (or VSTest) is running tests in parallel. I would be surprised if Assembly.GetExecutingAssembly() is actually slow enough to warrant being cached, but I don't think it'll hurt.

chambm:
Somehow broke the Jamfile running tests. Need to fix before merge.

PR #2034, opened 2022-03-16
Integrate MS2 peaks without background when acquisition method is DDA

nickshulman:
Changed so that MS2 peaks are integrated without background subtraction when acquisition method is DDA.

Also includes Pull Request #2032.


brendanx67:
This still doesn't quite deliver the performance I was hoping for. Though it does fix the problem case for the current implementation in master, it adds new problems, which I suspect are also due to the change from peak areas based interpolated data versus raw chromatogram intensities.

master - the problem we are attempting to solve for a peak with 2 spectra straddling the integration boundaries
image

this branch - with the problem case now fixed but two other replicates made noticeably worse
image

brendanx67:
The problem cases look like this (with MS/MS sampled at low intensity early and then much higher intensity outside the integration boundaries):
worst:
image
second worst:
image

brendanx67:
Looks much better:
image

PR #2039, opened 2022-03-22
Fix for the double painting of the PeakAreaReplicateComparison graph.

rita-gwen:


rita-gwen:
I looked a bit further in the GitHub history and it looks like I added it during initial dotp graph implementation. 
There seems to be just two places where UpdateGraph is called (I used VS Call Hierarchy tool): from the GraphSummary.UpdateUI which always calls Invalidate at the end, and SummaryReplicateGraphPane.ChangeSelection. There the control is invalidated when the sequence tree selection is updated.
Let me know about other possible scenarios that should be tested.

brendanx67:
Okay, if that line was added by you recently, then I am much more comfortable seeing it removed again, especially after your further research. Thanks for the closer look.

PR #2047, opened 2022-03-28
Insert a 60 second pause between retries of failed commands in SkylineTester's…

bspratt:
… build and checkout system


brendanx67 (changes requested):
This is not exactly what I was asking for. This will cause the issue we were seeing to take 10 minutes to complete its first set of failures based on Git failing to checkout into a folder with files in it. But, I don't think it will do anything for the subsequent failures where SkylineTester is failing to delete a folder that is in use. That problem will still iterate on SkylineTester run every 4 seconds. I was asking for changes that would be entirely localized within SkylineNightly and would not change SkylineTester at all.

bspratt:
I guess we need another conversation - I can't see where this problem can be addressed other than in SkylineTester, where the git clone occurs.  Agreed that the 60 second retry delay doesn't solve the "failure to clone to a non-empty directory" issue - for that I I'll have to take a closer look at the git failure itself - but the retry delay is still a generally good idea, I think.

brendanx67 (approved):
Okay. Funny that there is already a number of minutes to wait, and we just weren't waiting in this case.

PR #2053, opened 2022-03-30
Fix Issue #863, "Problem loading a small mol document with isotope labels...

nickshulman:
Fixed error that can happen if small molecule documents if you remove an isotope label type that is still being used by some of the precursors in the document. (Reported by Johannes)

### DIFF
--- a/pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
+++ b/pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
@@ -245,6 +245,19 @@ namespace pwiz.Skyline.Model.DocSettings
             return TryGetPrecursorCalc(labelType, mods) != null;
         }
 
+        public bool AllowPrecursor(TransitionGroupDocNode transitionGroup, ExplicitMods mods)
+        {
+            if (transitionGroup.IsLight)
+            {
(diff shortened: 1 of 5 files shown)


bspratt:
This leads to a consistent document, but does the user expect nodes to disappear, or rather expect missing label types to be created? It's not clear to me how the document gets into this state in the first place.

nickshulman:
> This leads to a consistent document, but does the user expect nodes to disappear, or rather expect missing label types to be created? It's not clear to me how the document gets into this state in the first place.

The way that the document gets in this state is that you go to:
Settings > Molecule Settings > Labels
and press "Edit List" and remove one of the labels which the document is using.

Brendan, BrianP and I all discussed this bug in email a few years ago. The issue was raised that maybe we should not silently delete precursors from the document, but Brendan replied on Sept 28, 2017:
> Yes, we need to just remove the label types that are removed from the document. I don't think we need to warn with a message box. We don't do that with any other settings change. e.g. If you change your instrument m/z range, we rip out transitions that can no longer be measured without warning about it. This is what Undo is for.

bspratt:
OK, seems good then. Looks like there are currently build issues but I'm guessing you inherited those from master.

PR #2054, opened 2022-03-30
Add explicit support for VS2022, seems to be needed on machines where VS2022 is the only option

bspratt:
It's a minimal change and seems to resolve the issue building the release branch on KAIPOT-PC1

### DIFF
--- a/Jamroot.jam
+++ b/Jamroot.jam
@@ -324,6 +324,7 @@ project pwiz
         <toolset>msvc,<toolset-msvc:version>14.0:<define>_HAS_TR1
         <toolset>msvc,<toolset-msvc:version>14.1:<define>_HAS_TR1
         <toolset>msvc,<toolset-msvc:version>14.2:<define>_HAS_TR1
+        <toolset>msvc,<toolset-msvc:version>14.3:<define>_HAS_TR1
 
         # link-time optimization for release builds
         #<variant>release:<lto>on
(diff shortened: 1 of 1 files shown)


brendanx67 (approved):
Makes sense. Thanks for finding and fixing it.

bspratt:
> Hmm. I think the release branch on TC is set to build with VS 2017. So that's probably why this issue didn't show up there. Does it actually make sense to do nightlies on the release branch on a VS version it wasn't tested with?

That's a fair question, but we have a number of other machines still building it with VS2017 so I think we're OK with this.

chambm (approved):

chambm:
That Linux failure was real. There was that other Jamroot update we did on master to add the -fno-pie flag. I guess it doesn't matter much on this target branch though.

bspratt:
Well, I hate to see errors. I'll clean that up (it hadn't occurred to me that a VS related change would ding linux?)

bspratt:
Which config are we talking about here?

chambm:
Jamroot.jam changes builds everything (the smart build trigger script can't really tell what configs a Jamfile change could affect so it assumes all of them). Your change didn't break the build, it just triggered a build that was already broken by the AWS agent rebuild I had to do when the old images got erased. I don't understand why the core Linux build worked and the Bumbershoot one failed though. :/

bspratt:
No harm in me patching that too, I guess - like I say, I don't like having errors hanging around on active branches.

PR #2059, opened 2022-03-31
Warn and offer to fix iRT database if target is a standard and also in the library

kaipot:


pwiz_tools/Skyline/Model/Irt/IrtDb.cs
```diff
@@ -591,5 +591,23 @@ private static void EnsureDocumentXmlTable(ISession session)
...
+        public void RemoveLibraryPeptides(IEnumerable<Target> peptides)
+        {
+            using (var session = OpenWriteSession())
+            using (var transaction = session.BeginTransaction())
+            {
+                foreach (var pep in peptides)
+                {
+                    using (var cmd = session.Connection.CreateCommand())
+                    {
+                        cmd.CommandText = @"DELETE FROM IrtLibrary WHERE Standard = 0 AND PeptideModSeq = ?";
```
nickshulman, line 604: If you want to remove all non-standard peptides which have the same sequence as a standard peptide you could do:
DELETE FROM IrtLibrary L1 WHERE L1.Standard = 0 and L1.PeptideModSeq IN (SELECT L2.PeptideModSeq FROM IrtLibrary L2 WHERE L2.Standard = 1)
Afterwards, if you wanted to make sure that there were no duplicates you could do:
DELETE FROM IrtLibrary L1 WHERE L1.Id <> (SELECT MIN(L2.Id) FROM IrtLibrary L2 WHERE L2.PeptideModeSeq = L1.PeptideModSeq)

kaipot: You're right, that's a better way to do it.

nickshulman (commented):
Since there isn't any user interface involved, it might be simpler if you move the checking for duplicates and deleting them to "IrtDbManager.LoadBackground".

That way, all of the database reading/writing is happening on a background thread and you don't need to worry about what might happen if the Document changes in between calls to "SetDocument".

kaipot:
> Since there isn't any user interface involved, it might be simpler if you move the checking for duplicates and deleting them to "IrtDbManager.LoadBackground".
> 
> That way, all of the database reading/writing is happening on a background thread and you don't need to worry about what might happen if the Document changes in between calls to "SetDocument".

Thanks, those are all good suggestions. I made the changes.

nickshulman (approved):
Looks good to me.

PR #2060, opened 2022-04-01
Calculate Skewness and Kurtosis on transition chromatogram peaks

nickshulman:
Add "Peak StdDev", "Skewness", "Kurtosis" and "ShapeCorrelation" (correlation of peak shape to the median chromatogram for the precursor).
These numbers are currently stored as half-precision floating point numbers, so they add 6 bytes to the size of the ChromPeak data structure.

image
image


nickshulman:
This branch has been merged into Skyline/work/20220511_nicksh_SpectrumGrid.
Both of these branches need to use the new class "PeakGroupIntegrator", so they will be worked on together. Also, both branches need to increment the skyd file format version, so they might as well be combined.

PR #2061, opened 2022-04-04
Yet another new instrument source type for which we need to consult B…

bspratt:
…ruker

### DIFF
--- a/pwiz/data/vendor_readers/Bruker/Reader_Bruker_Detail.cpp
+++ b/pwiz/data/vendor_readers/Bruker/Reader_Bruker_Detail.cpp
@@ -175,6 +175,7 @@ std::vector<InstrumentConfiguration> createInstrumentConfigurations(CompassDataP
 
         case InstrumentSource_APCI:
         case InstrumentSource_GC_APCI:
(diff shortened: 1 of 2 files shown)


pwiz/data/vendor_readers/Bruker/Reader_Bruker_Detail.cpp
```diff
@@ -175,6 +175,7 @@ std::vector<InstrumentConfiguration> createInstrumentConfigurations(CompassDataP
 
         case InstrumentSource_APCI:
         case InstrumentSource_GC_APCI:
+        case InstrumentSource_VIP_APCI:  // TODO: yet another new source, this is a guess along with the VIP_HESI question above
```
chambm, line 178: I bet what's happening here is there is only one new source type, VIP-HESI. But it has 2 modes, ESI and APCI. VIP_HESI means it's operating in ESI mode, and VIP_APCI means it's operating in APCI mode. But yeah, we should confirm this with Bruker. Sorry for this dropping off my radar.

chambm (approved):

PR #2063, opened 2022-04-04
Extended retries when ReSharper code coverage is in use, and guard against parallel tests fighting over Tools directory

bspratt:
Trying to get the TeamCity code coverage build going again, in particular hoping to clear up the failure to remove the crux.exe file after a test run - possibly we just need to wait a bit longer for some process to complete


bspratt:
> It's curious the DDA functional tests take so long in coverage mode. Most of the time in those is spent in child processes that aren't .NET.

Maybe it's really just that they're built in debug mode? But I guess we don't build those child process images, do we?

pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs
```diff
@@ -109,6 +109,13 @@ public void TestDdaSearch()
         [TestMethod]
         public void TestDdaSearchMsgfPlus()
         {
+            if (Helpers.RunningResharperAnalysis)
+            {
+                // This test just takes much too long under code coverage analysis
+                Console.Write(MSG_SKIPPING_SLOW_RESHARPER_ANALYSIS_TEST);
```
brendanx67, line 115: We really have nothing better than writing to "Console"? I feel like we should have something more formal, like a method on TestContext, to distinguish between code that we really want to keep as important testing output and code that accidentally got left behind during debugging, which we see plenty of.

chambm:
I noticed a previous run took 2 hours but the latest one took 3h20m (after you fixed a string of timeouts):
image

That looked promising, but was it a fluke?

pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs
```diff
@@ -109,6 +109,13 @@ public void TestDdaSearch()
         [TestMethod]
         public void TestDdaSearchMsgfPlus()
         {
+            if (Helpers.RunningResharperAnalysis)
```
brendanx67, line 112: Let's have a function, so that it looks like this:

if (SkipForResharperAnalysis())
    return;

chambm:
We could (I should) adjust this config (and probably the perf test one too) to use an AWS instance type that has fewer processors. I think 2 processors would be enough? Then we could let it run longer if that would mean we get better coverage results.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -551,12 +551,18 @@ private static int GetWaitCycles(int millis = WAIT_TIME)
...
+            // Wait a little longer for debug build.
             if (ExtensionTestContext.IsDebugMode)
             {
                 waitCycles = waitCycles * 4;
             }
 
+            // Code coverage testing may be slower yet.
+            if (Helpers.RunningResharperAnalysis)
+            {
+                waitCycles *= System.Diagnostics.Debugger.IsAttached ? 2 : ExtensionTestContext.IsDebugMode ? 4 : 16;
```
brendanx67, line 563: I don't totally understand all of these options. These multipliers are additive and IsDebugMode and Debugger.IsAttached get added above. Let's also make the code above consistently use

waitCycles *= multiplier;

instead of sometimes:

waitCycles = waitCycles * multiplier;

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -551,12 +551,18 @@ private static int GetWaitCycles(int millis = WAIT_TIME)
                 waitCycles *= Program.UnitTestTimeoutMultiplier;
             }
 
-            // Wait a little longer for debug build. (This may also imply code coverage testing, slower yet)
+            // Wait a little longer for debug build.
             if (ExtensionTestContext.IsDebugMode)
             {
                 waitCycles = waitCycles * 4;
```
brendanx67, line 557: Make waitCycles *= 4;

brendanx67 (approved):
Minor comments. Otherwise, it looks okay, assuming you have tested it all yourself under the expected conditions.

pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs
bspratt: Ultimately we do want Console output, but I take your point about putting that in just one place where it's clear that it's intentional.

bspratt:
> We could (I should) adjust this config (and probably the perf test one too) to use an AWS instance type that has fewer processors. I think 2 processors would be enough? Then we could let it run longer if that would mean we get better coverage results.

That would be great - I defer to your experience selecting the instance type.

bspratt:
> I noticed a previous run took 2 hours but the latest one took 3h20m (after you fixed a string of timeouts): image
> 
> That looked promising, but was it a fluke?

Really the important change here is that I realized that tests were wrestling over the Tools directory (that was the crux.exe noise). I think with that fixed, and disabling some long-duration tests, there's just a different constellation of tests that find time to run?

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
bspratt: No problem. I just habitually don't change code that I don't have to, but agreed that consistency is good.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
bspratt: Added a comment to clarify:
// Code coverage testing may be slower yet. But don't apply multipliers that have already been applied above.

brendanx67: Maybe this could be made clearer by having a "waitMultiplier" variable that accumulates above and just gets set to 16 in this case? And then the final line could be "return waitCycles * waitMultiplier;

bspratt:
> > We could (I should) adjust this config (and probably the perf test one too) to use an AWS instance type that has fewer processors. I think 2 processors would be enough? Then we could let it run longer if that would mean we get better coverage results.
> 
> That would be great - I defer to your experience selecting the instance type.

Although the perf tests often benefit from more cores since many are doing chromatogram extraction on multiple replicates.

chambm:
> > > We could (I should) adjust this config (and probably the perf test one too) to use an AWS instance type that has fewer processors. I think 2 processors would be enough? Then we could let it run longer if that would mean we get better coverage results.
> > 
> > 
> > That would be great - I defer to your experience selecting the instance type.
> 
> Although the perf tests often benefit from more cores since many are doing chromatogram extraction on multiple replicates.

Perf tests would continue to run on the same instance type as the regular tests. Only this code coverage build would be a skimpy instance type.

bspratt:
> This is what I went with for the tools dir:
> 
> ```
> private static string _toolsDirectory;
>         public static string GetToolsDirectory()
>         {
>             if (_toolsDirectory != null)
>                 return _toolsDirectory;
> 
>             string skylinePath = Assembly.GetExecutingAssembly().Location;
>             if (string.IsNullOrEmpty(skylinePath))
>                 return null;
>             string skylineDirPath = Path.GetDirectoryName(skylinePath);
>             if (string.IsNullOrEmpty(skylineDirPath))
>                 return null;
> 
>             // use a random tools path when running tests to allow tests to run in parallel
>             if (Program.FunctionalTest)
>                 _toolsDirectory = Path.Combine(skylineDirPath, "Tools-" + Path.GetRandomFileName());
>             else
>                 _toolsDirectory = Path.Combine(skylineDirPath, @"Tools");
>             return _toolsDirectory;
>         }
> ```

Excellent, I'll use that too. Though I think Program.UnitTest is probably the check we want for best generality.

chambm:
Hmm, my approach is only process-unique. If everything is running under one process, it won't be parallel-safe. My PR's parallel testing code runs tests in separate processes (Docker containers, actually). Your test-name-hash-based approach may be superior then. It should work for both cases. The only drawback is it won't re-use the tools directory for non-coverage testing, which will continue to be serial. There are definitely other cases where tests aren't safe to run in parallel, usually because they two or more want a filesystem exclusive lock.

Of course, both of our make-the-tools-directories-unique approaches change the behavior from what a real Skyline instance (or instances) would be doing.

bspratt:
Yeah, I'm finding that the current approach doesn't seem to guarantee a unique directory name. I'll go back to the test name + language hash.

chambm:
Where are you finding the command-line might be too long? The command-line limit is 1024 characters, so if we're hitting that with your fairly short tool directory names, it's probably something we need to fix a different way.

bspratt:
There's one remaining test that's failing, TestDdaSearchMsFragger. But the commandline iit fails on (invoking java / msfragger) is only 738 characters long so perhaps that's not the issue.

chambm:
The limit for any given path (without \\?\ suffix) is 260 characters though, so that does seem more likely as an issue. I'll look at the test.

chambm:
Yeah those paths aren't approaching 260. Not even half that. Is that test not failing for you locally?

bspratt:
Unfortunately I only get that failure on TC, in the CodeCoverage config. But yeah, I agree that path or command length does not appear to be the issue.

chambm:
What about these Skyline x64 test failures?

bspratt:
You're right, I was overly focused on that one config. So there's something else going on there, I'll figure it out.

bspratt:
Thanks for pointing that out - there's some logic in those tests that assumes a constant Tools directory name for finding Java. How evil is it to download Java each time? Does that add up in AWS bandwith charges etc?

chambm:
My parallel PR has some local caching to alleviate that (kind of like the persistent perf tests downloads except in my case the cached files get copied into their final location instead of downloaded again). I got tired of having to redo the downloads 😛

bspratt:
Sounds good. I'll call my too-much-downloading implementation good for now, and we'll see who gets into master first then merge carefully,

bspratt:
OK, this is ready for a final review. In the end the bulk of these changes are just about improved debug output, but they are handy and they are unobtrusive when not in use, so I think we should retain them.

The meat of it is dealing with interactions between tests which only seem to crop up under code coverage on TeamCity, the most obvious symptom being struggles over the contents of the Tools directory. It looks like the tests may be running in a combination of multiple processes, and reusing those processes for multiple tests (as evidenced by static assignments sometimes persisting from test to test.)

Note that with these changes I can *nearly* get unit tests to run with Test Explorer's "Run Tests In Parallel" enabled on my local machine, but the Wiff reader gets unlucky once in a while - I believe it's known to have issues doing more than one thing at a time. It's also the roadblock to moving to the current MSTest version, which is eventually going to be a problem - we're using a legacy version that ReSharper no longer officially supports.

All those changes around "run small molecule test versions" are a result of a similarly implemented "skip test for code coverage" routine that in the end I haven't used (yet).

This will require a careful merge with Matt's work to parallelize TestRunner.

pwiz_tools/Skyline/Model/DdaSearch/MsFraggerSearchEngine.cs
```diff
@@ -451,67 +451,67 @@ private void EnsureFastaHasDecoys()
...
-use_topN_peaks = 50			# Pre-process experimental spectrum to only use top N peaks.
-min_fragments_modelling = 2			# Minimum number of matched peaks in PSM for inclusion in statistical modeling.
-min_matched_fragments = 4			# Minimum number of matched peaks for PSM to be reported.
-minimum_ratio = 0.01			# Filters out all peaks in experimental spectrum less intense than this multiple of the base peak intensity.
-clear_mz_range = 0.0 0.0			# Removes peaks in this m/z range prior to matching.
-
-add_Cterm_peptide = 0.000000
-add_Nterm_peptide = 0.000000
-add_Cterm_protein = 0.000000
+data_type = 0			# Data type (0 for DDA, 1 for DIA, 2 for DIA-narrow-window).
```
chambm, line 454: Are these line ending changes? Consider excluding those from the PR to reduce noise. (they might also be important to MSFragger, I'm not sure).

pwiz_tools/Skyline/Model/Tools/ToolDescription.cs
```diff
@@ -887,15 +887,30 @@ public static void GetReport(SrmDocument doc, string reportTitle, string toolTit
             }
         }
 
+
```
chambm, line 890: Did the abbreviation end up solving a problem? If not then I personally would rather have the full test names (possibly without the redundant "Test" prefix or suffix). It also seems like it could fairly easy collide where the full versions would be unique: TestMyCoolFunc and TestMyCoolFeat would both abbreviate the same, right?

pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs
```diff
@@ -1,50 +1,50 @@
...
-    [TestClass]
-    public class DdaSearchTest : AbstractFunctionalTest
-    {
-        public struct ExpectedResults
-        {
-            public int proteinCount;
-            public int peptideCount;
-            public int precursorCount;
+﻿/*
+ * Original author: Matt Chambers <matt.chambers42 .at. gmail.com >
```
chambm, line 2: It's hard to tell what changed here. Line endings again?

pwiz_tools/Skyline/Model/DdaSearch/MsFraggerSearchEngine.cs
bspratt: Yeah, those happened without me noticing. I'll change them back, and add a comment
        // N.B. leave these line endings alone, they may matter to MsFragger
 but if that's really the case, this seems like an accident waiting to happen (again). This would be better off using \n or \r or whatever.

pwiz_tools/Skyline/Model/Tools/ToolDescription.cs
bspratt: It did, yes. An unzip operation failed because it was trying to unzip to a deeply nested directory under the java install, which already had quite a long path due to the length of the test name.

TestMyCoolFunc and TestMyCoolFeat would indeed abbreviate the same, but that's a bridge I'm willing to jump off when I come to it.

Fundamentally I'd rather just fix Skyline so it doesn't care about path lengths - I believe it's possible but I haven't look into it.

pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs
bspratt: Yeah, that file was a mixed bag of line endings, and I got tired of being asked about it.

bspratt: (You can tell the github diff to ignore line endings, BTW)

pwiz_tools/Skyline/Model/DdaSearch/MsFraggerSearchEngine.cs
bspratt: Actually I think it's OK to leave it as is - it gets run through a StringBuilder so line endings are going to be whatever they're going to be. And if it was a problem for MsFragger wouldn't the tests fail?

chambm: True, it must be fine if the tests are passing. I like to leave line endings alone because it makes it hard to see the semantic differences, but I just rediscovered GitHub's "Hide whitespace changes" option. They hide it so well...

bspratt: I feel you, but if folks wouldn't commit janky line endings we wouldn't have to worry about non-semantic differences (and we wouldn't be pestered by the idea offering to fix them). Maybe I should make that a new custom code inspection...

bspratt:
@chambm I think you've been over this closely enough that you can give it the formal thumbs up. If you'll do that I'll go ahead and merge to master.

chambm (approved):

PR #2065, opened 2022-04-05
Fix small molecule library error when no chemical formulas.

nickshulman:
Fixed error trying to use small molecule spectral library which did not contain chemical formulas. (Reported by Allison)


pwiz_tools/Skyline/Model/Lib/BiblioSpecLite.cs
```diff
@@ -723,7 +724,17 @@ private byte[] CreateCache(ILoadMonitor loader, IProgressStatus status, int perc
...
                             var inChiKey = iInChiKey >= 0 && !reader.IsDBNull(iInChiKey) ? reader.GetString(iInChiKey) : null;
                             var otherKeys = iOtherKeys >= 0 && !reader.IsDBNull(iOtherKeys) ? reader.GetString(iOtherKeys) : null;
-                            smallMoleculeLibraryAttributes = SmallMoleculeLibraryAttributes.Create(moleculeName, chemicalFormula, inChiKey, otherKeys);
+                            if (string.IsNullOrEmpty(chemicalFormula))
+                            {
+                                var precursorMz = reader.GetDouble(iPrecursorMZ);
+                                smallMoleculeLibraryAttributes = SmallMoleculeLibraryAttributes.Create(moleculeName,
+                                    null, new TypedMass(precursorMz, MassType.Monoisotopic),
+                                    new TypedMass(precursorMz, MassType.Average), inChiKey, otherKeys);
+                            }
```
bspratt, line 733: We're looking for a mass here, rather than an m/z, so that will need to involve a calculation with the adduct.

bspratt: Though I don't see where we're pulling an adduct here, just a charge. Charge is not generally sufficient to get from mz to mass since different adducts can have the same charge.

nickshulman: Oh, good point. I had forgotten that m/z and mass were different.
I have coded up something which tries to use the adduct and/or charge columns to calculate the mass from the precursor m/z.
I should probably write a unit test which makes sure that when you export a spectral library from Skyline the masses are the same when Skyline reads the library again.

bspratt:
(and yeah, a test would be ideal)

bspratt:
Nick - I think we're good to go here. Perhaps you can eyeball my changes then do the commit to master?

PR #2066, opened 2022-04-06
Support for show/hide neutral losses spectrum annotation

rita-gwen:
Replaced the context menu items for ion type and charge selection with specialized menu-hosted controls.


pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -506,7 +506,7 @@ private void LoadLayoutLocked(Stream layoutStream)
             _listGraphDetections.ToList().ForEach(DestroyGraphDetections);
             DetectionsGraphController.GraphType = type;
 
-            FormUtil.OpenForms.OfType<FoldChangeForm>().ForEach(f => f.Close());
+            CollectionUtil.ForEach(FormUtil.OpenForms.OfType<FoldChangeForm>(), f => f.Close());
```
rita-gwen, line 509: This is a strange change, I don't remember doing it. Might has come from a master merge.

rita-gwen:
This PR is ready for the code review.

rita-gwen:
This is how it looks with enough losses defined:

brendanx67 (changes requested):
Found my review. Mostly pretty minor stuff, but needs a bit more testing.

rita-gwen:
I was thinking how to address that extra border problem when a button on the control has focus. Do you think this style will work? It doesn't have borders at all.

PR #2069, opened 2022-04-08
Some bug fixes

nickshulman:
Prevented clipboard error pressing "Copy" button on message dialog (reported on Exception Web)

Also:
Change a few other places that were using "Clipboard.SetText" to use ClipboardHelper.SetClipboardText
Remove unused setting "StackTraceListVersion"
Always include product version in body of exception posts (used to leave it off if not ClickOnce install)
Add code inspection to prevent using Clipboard.SetText outside of test code.
Change CodeInspectionTest to use compiled regular expression because it's faster


bspratt:
Looks solid.

PR #2070, opened 2022-04-09
Fixed a couple of issues building spectral libraries from Proteome Discoverer files

kaipot:


nickshulman (approved):
Looks good to me.

pwiz_tools/BiblioSpec/src/MSFReader.cpp
```diff
@@ -561,7 +569,7 @@ namespace BiblioSpec
             "   AND psm_spec.TargetPsmsWorkflowID = psms.WorkflowID";
         if (peptideGroups) {
             string joins =
-                " JOIN TargetPsmsTargetPeptideGroups psm_pep ON psms.PeptideID = psm_pep.TargetPsmsPeptideID"
+                " JOIN " + pepPsmTable + " psm_pep ON psms.PeptideID = psm_pep.TargetPsmsPeptideID"
```
nickshulman, line 572: It looks like "pepPsmTable" might be an empty string if neither TargetPeptideGroupsTargetPsms nor TargetPsmsTargetPeptideGroups exists. Do you need to check for that here?

PR #2075, opened 2022-04-18
Dotp line graph points should be clickable for changing the active replicate. 

rita-gwen:
Fix for the issue https://skyline.ms/issues/home/issues/details.view?issueId=885.


brendanx67 (changes requested):
This seems to work as I would expect now. The painting is quite flashy over Remote Desktop. Which I would really like to see fixed. It seems related to the graph painting once without the red cut-off lines and then again with them, which also causes everything in the legend to shift because the red lines are the first elements in the legends. I can give you a demo on Zoom.

rita-gwen:
The fix for double painting was a separate PR and it was not merged into this PR yet at the time when you checked it out. The problem should go away after current master is merged into this branch.

brendanx67:
Can you do the merge? It says it requires manual attention.

rita-gwen:
It is merged now.

brendanx67 (approved):

PR #2076, opened 2022-04-19
Updated library match to show spectra when multiple precursors are selected

kaipot:


pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
```diff
@@ -598,14 +603,44 @@ public bool ShouldShowMirrorPlot
...
+            }
+
+            public void Add(IEnumerable<SpectrumDisplayInfo> spectra)
+            {
+                Spectra = (Spectra != null ? Spectra.Concat(spectra) : spectra).ToArray();
+            }
+
+            private TreeNodeMS SelectedTreeNode { get; }
+            public TransitionGroupDocNode Precursor { get; }
+            public SpectrumDisplayInfo[] Spectra { get; private set; }
```
nickshulman, line 625: This should probably be a List<SpectrumDisplayInfo> because it's easier to add to a list than an array.

nickshulman (commented):
I don't think this works correctly if I have the "Decoys" peptide list selected. The Library Match window is trying to show me the spectrum of the decoy peptide, whereas if I had an individual decoy peptide selected, it would show me a spectrum.
image
image

brendanx67:
Updates seem to be synchronously attached to selection change. So, if I give the combo box the focus and press down arrow a lot quickly, the UI gets stuck advancing through each spectrum. You should be using a Timer like the SkylineWindow._timerGraphs. When the selection change happens, you start the timer and only respond to it when you get the timer event, where you check the selection at that time. Timer events are lower priority, so you can make a lot of selection changes before they make it through the message queue.

brendanx67:
Especially heinous if you do it for a large set of decoys with Prosit mirroring turned on. Though Prosit mirroring is also quite broken for decoys. It just increases the amount of work being done for each down-arrow press and so takes longer to release the UI once you stop pressing down-arrow.

brendanx67:
Actually, the Prosit CE combo box does this correctly. So, maybe you can just use the same method.

pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
```diff
@@ -911,6 +987,240 @@ private void ClearGraphPane()
...
+                _precursorStringCache = Cache.Create(outStrings, spectra);
+                return false;
+            }
+
+            public void Dispose()
+            {
+                _timer.Dispose();
+            }
+
+            private class Cache
```
nickshulman, line 1181: The "Cache" class here seems to make things more confusing, since it ends up obscuring the type of the keys and values that are used here.

pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
```diff
@@ -911,6 +987,240 @@ private void ClearGraphPane()
...
+                    catch (Exception)
+                    {
+                        // ignored
+                    }
+                }
+
+                _spectraLoadCache = Cache.Create(null, precursorSpectra, settings);
+            }
+
+            public bool GetPrecursorStrings(IReadOnlyList<PrecursorSpectra> spectra, out string[] outStrings)
```
nickshulman, line 1164: I would recommend using "ImmutableList&lt;string>" instead of "string[]" since it would prevent the elements in the list from being changed.

pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
```diff
@@ -911,6 +987,240 @@ private void ClearGraphPane()
...
+                    {
+                        return _timer.Tag != null;
+                    }
+                }
+            }
+
+            public void QueueUpdate(bool isUserAction)
+            {
+                // Restart the timer at 100ms, giving the UI time to interrupt.
+                lock (_timer)
```
nickshulman, line 1024: It looks to me like everything involving this timer is happening on the UI thread. Is anything here happening on a background thread that would make this "lock" necessary?

kaipot: GraphSpectrum implements IUpdatable, I think it's safer to use a lock since UpdateUI could potentially be called from anywhere?

pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
```diff
@@ -598,14 +610,72 @@ public bool ShouldShowMirrorPlot
...
+                Peptide = peptide;
+                LookupTarget = lookupTarget;
+                LookupMods = lookupMods;
+                Precursor = precursor;
+                Spectra = new List<SpectrumDisplayInfo>();
+                _prositCacheSettings = null;
+                _prositSpectra = new Dictionary<Tuple<IsotopeLabelType, int>, SpectrumDisplayInfo>();
+            }
+
+            private TreeNodeMS SelectedTreeNode { get; }
```
nickshulman, line 631: It looks like SelectedTreeNode is only getting used for checking whether the selected item is a PeptideGroupTreeNode.
It probably isn't a good idea to hang onto a TreeNodeMS, since that's a control which belongs to the SequenceTree. I would recommend just having an IdentityPath here instead.

PR #2077, opened 2022-04-20
Fix TIC for wiff files

nickshulman:
Fixed extracting the "Total Ion Current Area" from wiff files. (Reported by Karin)


chambm:
This should fail tests. Since you already made the PR, here's the updated file.
[PressureTrace1-6500SysSuit1269.zip](https://github.com/ProteoWizard/pwiz/files/8524440/PressureTrace1-6500SysSuit1269.zip)

PR #2078, opened 2022-04-20
extend our custom code inspection test to raise an error on any source code files with mixed line endings

bspratt:
After yet another code review where line ending changes caused confusion, I decided to extend our custom code inspection test to raise an error on any of our source code files with mixed line endings, so that we won't waste time on that in any future code reviews. Note that it doesn't insist on Windows line endings, it just insists on consistent line endings. 

All but one of the files changed here are just line endings cleanup, the only one with new logic is Skyline\Test\CodeInspectionTest.cs

I have also updated https://skyline.ms/wiki/home/development/page.view?name=Skyline%20Custom%20Code%20Inspections as follows:

> If you see this:
> Inconsistent line endings in (filename)
> 
> the problem is that your git installation, or code editor, or both, isn't configured to properly handle line endings. See https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings for information on fixing that. The reason this matters is that Visual Studio will complain about inconsistent line endings, or even just fix them silently, and when they are changed that creates code thrash and muddies code reviews. So we just don't allow ourselves to get into that state in the first place.


nickshulman (approved):
Looks good to me. Impressively simple implementation.

brendanx67 (approved):
Looks good to me.

PR #2079, opened 2022-04-20
Per Brendan's request, make the use of tabs instead of spaces a ReSharper code inspection Warning.

bspratt:
Only real change here is to the DotSettings file, everything else is cosmetic changes from tab to 4 spaces.

Note that ReSharper does not complain about tab characters inside strings, which is good since a lot of our tests have that (in FASTA data especially)


bspratt:
It appears that there was a handful of files that were not originally UTF8 encoded (e.g. pwiz_tools/Skyline/Controls/TreeViewMS.cs), they are now thanks to my last commit.

brendanx67 (approved):
Nice that ReSharper had this.

PR #2081, opened 2022-04-25
Deal with transition lists with entries which describe the same molecule but with a different (non-conflicting) set of accessions…

bspratt:
This seems to happen when users cobble together transition lists from multiple sources to flesh out the set of measured fragments.


nickshulman:
It might be easier to understand the code if you move "MoleculeNameAndAccessions" into a top-level class in its own file.

Then, make it so that SmallMoleculeTransitionListReader makes an array of MoleculeNameAndAccessions with one entry per row, and fills it in with the information from the grid.

And, then, the logic of merging those MoleculeNameAndAccession values could be done by a static method in the class "MoleculeNameAndAccessions" which modifies an array in place:
`public static void GetConsensusAccessionNumbers(MoleculeAccessionNumbers[] moleculeAccessionNumbersArray)
`
In that way, you can move the logic of merging accession numbers into a class which is separate from "SmallMoleculeTransitionListReader" and would be easier to test separately.

I am thinking that it's probably cleaner to use array MoleculeAccessionNumbers[] instead of a Dictionary<int, MoleculeAccessionNumbers> since you know how many rows there are and most of the rows will probably have a value.

bspratt:
Thanks for the review, I'll look at these refactorings.

The dictionary thing is needed because the input rows may be sorted by mass before final parsing, so we need a mapping from the original index value.

bspratt:
@nickshulman if you have a moment to give this (with your suggested refactoring) the blessing I'd be grateful, thanks

nickshulman (approved):
Looks good to me.

PR #2086, opened 2022-04-28
Add msconvert.exe to Administrator Installer

nickshulman:
Fixed missing msconvert.exe from Admin Installer (reported by Dilip)
Also, change capitalization of a couple of dll's and "zh-TW" folder to be consistent between admin installer and the regular installer.


chambm:
Hmm. Maybe we should switch the installer test to run TestFunctional instead of (or in addition to) TestData. That would have caught this (I actually explicitly added TestDdaSearch to TestData.dll, but that would only be testing MSAmanda without needing msconvert).

chambm (approved):
Nevermind, this is fine.

I'm going to do parallel functional testing on TC, and I don't want the test progress tied up with Boost.Build's wait-until-target-is-finished-to-print-everything (which Boost.Build does, ironically, to better support parallel building of targets). But it's still a good idea to run those tests (at least sometimes) from the installed location. So I might move that installer test action script into something that can run standalone (i.e. install Skyline, run TestRunner from installed location, uninstall Skyline). But that's a lot more involved than this PR so I'll do it separately.

PR #2089, opened 2022-04-30
Catch SQLite exception when creating cache

nickshulman:
Changed "unhandled error" when .blib missing into an ordinary error message. (reported anonymously on exception web).

Here is an example of what the error message will look like:
image


pwiz_tools/Skyline/Model/Lib/BiblioSpecLite.cs
```diff
@@ -931,6 +931,18 @@ private byte[] CreateCache(ILoadMonitor loader, IProgressStatus status, int perc
...
 
+        private SQLiteDataReader ExecuteReader(SQLiteCommand command)
+        {
+            try
+            {
+                return command.ExecuteReader();
+            }
+            catch (SQLiteException ex)
+            {
+                throw new IOException(string.Format(Resources.BiblioSpecLiteLibrary_CreateCache_Failed_reading_library_header_for__0__, FilePath), ex);
```
brendanx67, line 942: Could we add a file existence check so that we can more clearly report the file is missing? I think that could be a more helpful message to the user than that we failed to read a header, whatever that means to a non-programmer.

brendanx67 (changes requested):
I would like to see a more helpful message for the case when the file is completely missing. That could also allow us to improve the "Failed reading library header for {0}" message to suggest that the file may be corrupted.

pwiz_tools/Skyline/Model/Lib/BiblioSpecLite.cs
```diff
@@ -1168,13 +1156,19 @@ private bool Load(ILoadMonitor loader, IProgressStatus status, bool cached)
                 // should be suppressed because we know Skyline is going to try again.
                 if (!cached)
                 {
-                    var failureException = new Exception(FormatErrorMessage(x), x);
-                    if (ExceptionUtil.IsProgrammingDefect(x))
+                    // SQLiteExceptions are not considered programming defects and should be shown to the user
+                    // as an ordinary error message
+                    if (x is SQLiteException || !ExceptionUtil.IsProgrammingDefect(x))
+                    {
+                        var message = string.Format(Resources.BiblioSpecLiteLibrary_Load_Failed_loading_library__0__, FilePath);
```
brendanx67, line 1163: I keep asking for a separate message for when the FilePath does not exist. Then it would be helpful, I think, to report that it is missing instead of just the we "Failed loading library". It would be really easy to do that test and have 2 messages.

brendanx67 (changes requested):
Really would prefer to include a File.Exists(FilePath) test, and have a separate message explaining that the file does not exist.

PR #2090, opened 2022-05-02
Tweaks for code coverage speed and stability

bspratt:
These changes should fix the problem with InstallToolsTest taking an hour to complete (apparently just noise around trying to delete directories that don't exist), and the occasional failure that seems to be from an interaction of TestGroupedStudiesTutorialDraft and TestGroupedStudies1Tutorial

### DIFF
--- a/pwiz_tools/Skyline/TestTutorial/GroupedStudiesTutorialTest.cs
+++ b/pwiz_tools/Skyline/TestTutorial/GroupedStudiesTutorialTest.cs
@@ -47,8 +47,6 @@ namespace pwiz.SkylineTestTutorial
         {
             // Set true to look at tutorial screenshots.
             //IsPauseForScreenShots = true;
-ForceMzml = (DateTime.Now.DayOfYear % 2) == 0;   // TODO(bspratt) remove this once we're convinced that WIFF isn't the source of leaks and hangs
-
 
             TestFilesZipPaths = new[]
                 {
@@ -127,7 +125,7 @@ ForceMzml = (DateTime.Now.DayOfYear % 2) == 0;   // TODO(bspratt) remove this on
                 RunUI(() =>
                 {
                         editIrtCalcDlg.CalcName = irtCalcName;
-                        editIrtCalcDlg.CreateDatabase(GetOcRawTestPath("iRT-OC-Study.irtdb")); // Not L10N
+                        editIrtCalcDlg.CreateDatabase(GetTestPath("iRT-OC-Study.irtdb")); // Not L10N
                         SetTransitionClipboardText(new[] {15, 17}, c =>
                             {
                                 if (string.Equals(c[8], "protein_name"))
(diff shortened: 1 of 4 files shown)


pwiz_tools/Skyline/TestTutorial/GroupedStudiesTutorialTest.cs
```diff
@@ -202,19 +200,23 @@ protected override void DoTest()
...
+            WaitForDocumentLoaded();
+
             ImportResultsFiles(GetOcRawTestPath(), ExtAbWiff,
                 IsFullData ? "R201203" : "R201203_plasma_revision_F", true);
             WaitForCondition(2*60*1000, () =>
                 SkylineWindow.Document.Settings.HasResults && SkylineWindow.Document.Settings.MeasuredResults.IsLoaded);
+            WaitForDocumentLoaded();
 
             PauseForScreenShot();
+            LoadNewDocument(true); // Clean up
```
brendanx67, line 219: Are you sure this is necessary? AbstractFunctionalTest.EndTest() does essentially this, I think with:

                // Release all resources by setting the document to something that
                // holds no file handles.
                var docNew = new SrmDocument(SrmSettingsList.GetDefault());
                // Try twice, because this operation can fail due to active background processing
                RunUI(() => Helpers.TryTwice(() => SkylineWindow.SwitchDocument(docNew, null)));

pwiz_tools/Skyline/TestTutorial/GroupedStudiesTutorialTest.cs
```diff
@@ -127,7 +125,7 @@ protected override void DoTest()
                 RunUI(() =>
                 {
                         editIrtCalcDlg.CalcName = irtCalcName;
-                        editIrtCalcDlg.CreateDatabase(GetOcRawTestPath("iRT-OC-Study.irtdb")); // Not L10N
+                        editIrtCalcDlg.CreateDatabase(GetTestPath("iRT-OC-Study.irtdb")); // Not L10N
```
brendanx67, line 128: Okay. Though, in the true sense of the original we would want the .sky file to at least go into the "Ovarian Cancer" folder, because it was meant to share space with the "Heart Failure" study in GroupedStudies1TutorialTest. This probably should have been GroupedStudies2TutorialTest.

pwiz_tools/Skyline/TestTutorial/GroupedStudiesTutorialTest.cs
```diff
@@ -47,8 +47,6 @@ public void TestGroupedStudiesTutorialDraft()
         {
             // Set true to look at tutorial screenshots.
             //IsPauseForScreenShots = true;
-ForceMzml = (DateTime.Now.DayOfYear % 2) == 0;   // TODO(bspratt) remove this once we're convinced that WIFF isn't the source of leaks and hangs
```
brendanx67, line 50: I guess we feel confident enough about this, given the lack of failures.

pwiz_tools/Skyline/Util/UtilIO.cs
```diff
@@ -954,7 +954,10 @@ public static void SafeDelete(string path)
         {
             try
             {
-                Helpers.TryTwice(() => Directory.Delete(path, true), $@"Directory.Delete({path})");
+                if (path != null && Directory.Exists(path)) // Don't waste time trying to delete something that's already deleted
```
brendanx67, line 957: Surprising that this is meaningful, but it sounds like you have testing proof.

pwiz_tools/Skyline/TestTutorial/GroupedStudiesTutorialTest.cs
```diff
@@ -202,19 +200,23 @@ protected override void DoTest()
...
 
-            RunUI(() => SkylineWindow.SaveDocument(GetOcRawTestPath("OC-study.sky")));
+            RunUI(() => SkylineWindow.SaveDocument(GetTestPath("OC-study.sky")));
 
             // Importing Data
             ImportResultsFiles(GetOcRawTestPath(), ExtAbWiff,
                 IsFullData ? "R201217" : "R201217_plasma_revision_A", true);
             WaitForCondition(5*60*1000, () =>
                 SkylineWindow.Document.Settings.HasResults && SkylineWindow.Document.Settings.MeasuredResults.IsLoaded);
+            WaitForDocumentLoaded();
```
brendanx67, line 210: These WaitForDocumentLoaded() changes look like the only meaningful changes here to me. Though, mildly surprising that the results loading is not the slowest background loader.

brendanx67 (approved):
Okay. I don't see anything problematic. Just a little surprised that these are such meaningful changes.

pwiz_tools/Skyline/TestTutorial/GroupedStudiesTutorialTest.cs
bspratt: Yeah, likely not required. I'll delete it.

pwiz_tools/Skyline/Util/UtilIO.cs
bspratt: Those "directory does not exist" exceptions slow things down, especially under code coverage where we allot extra retry time. A similar check already exists for FileEx.SafeDelete.

brendanx67: Good point. Throwing exceptions is slow, especially when we are going to try multiple times. Makes you wonder if we should have a way to exclude certain exception times from the try again loop.

PR #2091, opened 2022-05-02
Add "app.manifest" to Skyline Nightly

nickshulman:
This manifest file contains <supportedOS id=> tags which indicate that this app is compatible with all currently known versions of Windows.
Having these tags in the manifest make it so that "Environment.OSVersion" will return the actual Windows version number, instead of the version number that Windows was pretending to be (which was "6.2.9200.0" which meant Windows version 8).
The version number for the current version of Windows 11 is "10.0.22000.194".

The app.manifest in this pull request is the one that Visual Studio gives you when you add a manifest file to the project, which just has a lot of commented placeholders in it. I have uncommented the supportedOS tags.

### DIFF
--- a/pwiz_tools/Skyline/SkylineNightly/app.manifest
+++ b/pwiz_tools/Skyline/SkylineNightly/app.manifest
@@ -1,0 +1,79 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+  <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
+  <trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
+    <security>
+      <requestedPrivileges xmlns="urn:schemas-microsoft-com:asm.v3">
+        <!-- UAC Manifest Options
+             If you want to change the Windows User Account Control level replace the 
+             requestedExecutionLevel node with one of the following.
+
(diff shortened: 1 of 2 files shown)


pwiz_tools/Skyline/SkylineNightly/app.manifest
```diff
@@ -0,0 +1,79 @@
...
+      <supportedOS Id="{35138b9a-5d96-4fbd-8e2d-a2440225f93a}" />
+
+      <!-- Windows 8 -->
+      <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}" />
+
+      <!-- Windows 8.1 -->
+      <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}" />
+
+      <!-- Windows 10 (also Windows 11, Windows Server 2016, Windows Server 2019 and Windows Server 2022) -->
+      <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}" />
```
bspratt, line 43: Is that going to be ambiguous in nightly reports?

nickshulman: The build number part of the version number can be used to tell the difference between Windows 10 and Windows 11.
With this change, my Windows 11 computer will report itself as being "Microsoft Windows NT 10.0.22000.0" and my Windows 10 computer will say "Microsoft Windows NT 10.0.19042.0".

bspratt (approved):
Seems good. It will be nice to have that OS version info in the reports.

PR #2094, opened 2022-05-05
Add (or modify) app.manifest files to Skyline, SkylineTester, TestRun…

nickshulman:
…ner, SkylineCmd and SkylineBatch to indicate that Skyline is compatible with the latest versions of Windows.

Questions:
1. Are there any other projects that should have an app.manifest added at this time?
2. Should we try to figure out some way to share these app.manifest files between projects so that when we want to change them (and also be sure that they are consistent with each other now)? Maybe that &lt;ApplicationManifest> element in the .csproj could point to a file in a shared location.

### DIFF
--- a/pwiz_tools/Skyline/Skyline.csproj
+++ b/pwiz_tools/Skyline/Skyline.csproj
@@ -19,6 +19,7 @@
     </ManifestKeyFile>
     <GenerateManifests>true</GenerateManifests>
     <TargetZone>LocalIntranet</TargetZone>
+    <ApplicationManifest>Properties\app.manifest</ApplicationManifest>
     <SignManifests>false</SignManifests>
     <SignAssembly>false</SignAssembly>
     <FileUpgradeFlags>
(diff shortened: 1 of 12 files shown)


pwiz_tools/Skyline/SkylineTester/SkylineTester.csproj
```diff
@@ -100,6 +100,10 @@
     </StartupObject>
   </PropertyGroup>
   <PropertyGroup />
+  <PropertyGroup />
```
brendanx67, line 103: Why the empty property group tags?

pwiz_tools/Skyline/Skyline.csproj
```diff
@@ -19,7 +19,6 @@
     </ManifestKeyFile>
     <GenerateManifests>true</GenerateManifests>
     <TargetZone>LocalIntranet</TargetZone>
-    <ApplicationManifest>Properties\app.manifest</ApplicationManifest>
```
brendanx67, line 22: Why does this get removed when it is getting added to other .csproj files? Is it redundant?

brendanx67 (approved):
A couple minor things I find confusing, but if they make sense to you, then you can merge this.

PR #2098, opened 2022-05-09
Fix an apparent race condition in MethodEditTutorialTest…

bspratt:
… commonly seen in TC "code coverage with debug" build


brendanx67 (approved):
Okay. Though, it would be nice to understand why so that we don't have to add this kind of thing everywhere. Is it because the graph somehow updates based on the selection change, but then has some kind of background processing that updates it again later? If so, we may need to rethink how "WaitForGraphs()" works to wait for all background processing to complete.

bspratt:
Agreed 100%.  I just want to see first if this fixes the immediate problem. I suspect it's an issue with the graphs update timer - test thread getting ahead of it and checking before it's even been triggered, or something like that - so WaitForGraphs probably needs some work.

brendanx67:
The change in selection should immediately put all graphs on an update queue and WaitForGraphs() should be waiting until that queue empty. So, what you describe shouldn't really be possible. What could happen is that an individual graph might implement its on secondary timer where it updates a first time, and then updates itself again without being put on the queue. I know that Tobi added some background updating for Prosit and Kaipo is expanding on that for his new multi-peptide combo box. This fix makes me wonder what more we may need to do about graphs that add this type of background updating so that being removed from the update queue doesn't necessarily mean the plot has been finalized.

bspratt:
Let's be sure to mention it at today's team meeting.

PR #2099, opened 2022-05-10
Fix FormatException reading .bcfg file in international.

nickshulman:
Fixed "Input string was not in correct format" error reading .bcfg file on systems with different decimal separator (reported by Lutho).


brendanx67 (approved):
Looks good. Have you tried any other testing with European language settings? e.g. round-tripping the .bcfg file to make sure that it is also written correctly?

nickshulman:
> Looks good. Have you tried any other testing with European language settings? e.g. round-tripping the .bcfg file to make sure that it is also written correctly?

Good point. We were also using the wrong CultureInfo when writing Decimal values to XML. Fixed.

PR #2100, opened 2022-05-12
SkylineNightly: when a nightly run hangs before it even runs any test…

bspratt:
…s, provide the full log in the alert email instead of just the log tail


bspratt:
One potential concern is that the entire log might be too big for the LabKey email relay to handle - @bconn-proteinms do you know if there's a limit? Typically a log of this variety will be less than 1.5MB.

PR #2102, opened 2022-05-14
Fix TestHugeAssociateProteins

nickshulman:


nickshulman:
There seems to be another problem where the test always fails the second time it is run. I'll close this for now, since I think the real fix is going to be something more complicated.

pwiz_tools/Skyline/EditUI/AssociateProteinsDlg.cs
```diff
@@ -456,16 +473,5 @@ private void AssociateProteinsDlg_HelpButtonClicked(object sender, EventArgs e)
             else
                 WebHelpers.OpenSkylineLink(this, proteinAssociationWikiPath);
         }
-
```
chambm, line 459: Why remove these and the Accept/CancelButton properties?

chambm:
Yes, that 2nd-time-failure is what I've been banging my head on for several days. Somehow the ApplyChanges action itself is changing _newDocument, but only on the second (or later) iteration?! But stepping through it in the debugger on the second iteration, I haven't been able to figure out where/why the change is happening.

chambm:
OK I think it has to do with automatically repopulating the last used FASTA in the OnShown handler. The functional test code is running before OnShown; when OnShown runs, it is resetting the dialog settings to the defaults (I'm not sure why).

PR #2105, opened 2022-05-17
Protein association fixes

chambm:
* refactored LastFastaFileName to a proper UserSetting
* added IsBusy property to allow test code to wait properly
* fixed test code not waiting for correct condition (FinalResults != null before _newDocument is finished)
* fixed changing group proteins checkbox causing two UpdateParsimonyResults() calls
* fixed MinPeptidesPerProtein not being persisted correctly to AssociateProteinsSettings
* added explicit testing for reusing the last used FASTA filepath


pwiz_tools/Skyline/TestFunctional/AssociateProteinsDlgTest.cs
```diff
@@ -83,7 +88,8 @@ private void TestDialog(ImportType type)
             var dlg2 = ShowDialog<AssociateProteinsDlg>(SkylineWindow.ShowAssociateProteinsDlg);
             if (type == ImportType.FASTA)
             {
-                RunUI(() => dlg2.FastaFileName = _fastaFile);
+                if (Settings.Default.LastProteinAssociationFastaFilepath.IsNullOrEmpty())
```
brendanx67, line 91: Is this ever not the case anymore? Settings.Default gets reset between test runs.

pwiz_tools/Skyline/EditUI/AssociateProteinsDlg.cs
```diff
@@ -275,7 +278,10 @@ private void rbCheckedChanged(object sender, EventArgs e)
         private void tbxFastaTargets_TextChanged(object sender, EventArgs e)
         {
             if (File.Exists(tbxFastaTargets.Text))
+            {
+                Settings.Default.LastProteinAssociationFastaFilepath = tbxFastaTargets.Text;
```
brendanx67, line 282: Normally, we would save a value in OkDialog() after we are sure everything is valid, to avoid saving a value the user didn't really end up using, e.g. I typed "foobar.fasta" realized I was wrong and canceled out of the form.

pwiz_tools/Skyline/EditUI/AssociateProteinsDlg.Designer.cs
```diff
@@ -80,7 +80,6 @@ private void InitializeComponent()
             // btnCancel
             // 
             resources.ApplyResources(this.btnCancel, "btnCancel");
-            this.btnCancel.DialogResult = System.Windows.Forms.DialogResult.Cancel;
```
brendanx67, line 83: We usually leave the DialogResult for the cancel button. If you remove it, I think you need a click action handler. What happens when you click the Cancel button?

pwiz_tools/Skyline/EditUI/AssociateProteinsDlg.cs
```diff
@@ -44,20 +44,19 @@ public partial class AssociateProteinsDlg : ModeUIInvariantFormEx,  // This dial
...
         private static string[] _sharedPeptideOptionNames = Enum.GetNames(typeof(ProteinAssociation.SharedPeptides));
 
         public string FastaFileName
         {
             get { return tbxFastaTargets.Text; }
-            set { LastFastaFileName = tbxFastaTargets.Text = value; }
+            set { tbxFastaTargets.Text = value; }
         }
 
+        public bool IsBusy => _newDocument == null;
```
brendanx67, line 57: I don't see this getting used.

brendanx67 (changes requested):
Unsure about removing the DialogResult on the Cancel button.

pwiz_tools/Skyline/EditUI/AssociateProteinsDlg.cs
chambm: OK, makes sense.

pwiz_tools/Skyline/TestFunctional/AssociateProteinsDlgTest.cs
chambm: Yes, by doubling up the calls to TestDialog above I am explicitly testing the re-use-last-FASTA-filepath feature.

pwiz_tools/Skyline/EditUI/AssociateProteinsDlg.cs
chambm: Conflict resolution mistake.

pwiz_tools/Skyline/EditUI/AssociateProteinsDlg.Designer.cs
chambm: With the CancelButton property set but no DialogResult set on the button, clicking the button does close the dialog and return DialogResult.Cancel. But TBH I didn't even think about it, I just removed the DialogResult from both buttons thinking they'd both be redundant. But I can see why you wouldn't expect so since DialogResult.Cancel is never returned explicitly. And when I tested removing the DialogResult = DialogResult.OK from the OkDialog() function, the OK button no longer closes the form. So maybe those Accept/CancelButton properties only affect the DialogResult if there's no other click handler.

brendanx67: Maybe CancelButton is special in that way. I had always thought they really only determined what button got a clicked message when you press Enter = Accept, Esc = Cancel. Clearly, I had never worked out whether those settings might also set DialogResult in the absence of an OnClicked handler.

brendanx67 (approved):

PR #2108, opened 2022-05-24
Sanity check FASTA for ASCII text

chambm:
- added sanity checking code that FASTA files contain only ASCII text (the FASTA format predates Unicode)

Fixes [EncoderFallbackException](https://skyline.ms/announcements/home/issues/exceptions/thread.view?rowId=55746) to give a more helpful error message

This was something I did a couple weeks ago but just now took the time to give it its own PR.


nickshulman (commented):
Currently, we seem to use UTF8 whenever we parse FASTA files. It wasn't necessarily a conscious choice, but it's what StreamReader gives us by default.

Even if we are allowing UTF8, we can still be pretty effective at detecting binary files, just by looking for character 0. Most binary files have lots of zeroes in them, and text files never do.

pwiz_tools/Skyline/Model/Import.cs
```diff
@@ -3463,9 +3463,14 @@ public static IEnumerable<FastaData> ParseFastaFile(TextReader reader, bool read
             string line;
             string name = string.Empty;
             StringBuilder sequence = new StringBuilder();
+            int lineNum = 0;
 
             while ((line = reader.ReadLine()) != null)
             {
+                for (int i=0; i < line.Length; ++i)
+                    if (line[i] < 32 || line[i] > 126)
```
nickshulman, line 3471: You need to allow character number 10 here. The FASTA spec says that description lines should contain character number 10 to separate the protein name from the description.

chambm: You mean linefeed/newline? I don't think we would ever get that here because we're using ReadLine, right? Did you mean character 9, horizontal tab? Because that one is quite probable so it needs to be added.

chambm:
> Currently, we seem to use UTF8 whenever we parse FASTA files. It wasn't necessarily a conscious choice, but it's what StreamReader gives us by default.
> 
> Even if we are allowing UTF8, we can still be pretty effective at detecting binary files, just by looking for character 0. Most binary files have lots of zeroes in them, and text files never do.

The most common case may be someone picking a .fasta.gz file and in that case yeah, there's tons of 0s. I'm trying to also catch rarer cases where there might be a small amount of garbage somewhere in the file and it won't necessarily have 0s but if it's random garbage each byte only has a 96/256 chance to be valid.

pwiz_tools/Skyline/Model/Import.cs
nickshulman: I think maybe the character I was thinking of is actually ASCII code 1, not 10.

If you look at "WebEnabledFastaImporter.cs", there's this code:
`            String[] alternatives = line.Substring(1).Split((char) 1);`
Meanwhile, the [Wikipedia page](https://en.wikipedia.org/wiki/FASTA_format) says:

> In a deprecated practice, the header line sometimes contained more than one header, separated by a ^A (Control-A) character.

I think the Wikipedia page confused me, and the special character I was thinking of was 1 not 10.

chambm: Yeah, I just had to read up on the caret notation. I thought it was a Windows thing but it's older than that. I have never seen a FASTA with ^A in it, have you? The original spec also says the header line (really a comment) could start with `;` instead of '>' and there can be any number of comments before and between sequence lines. But I think very few modern parsers would support that. :/

chambm: Updated code supports ^A and adds a test with it.

chambm:
Ping @nickshulman - any reason this shouldn't go in?

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
```diff
@@ -94,6 +94,10 @@ public void UseFastaFile(string file, Func<FastaSequence, IEnumerable<Peptide>>
                     }
                 }
             }
+            catch (InvalidDataException)
```
nickshulman, line 97: I would be inclined to get rid of this whole try/catch block, and just let all exceptions bubble up. The wrapping with "InvalidDataException" a few lines later doesn't appear to add much information to the error message.

nickshulman (approved):
Looks good to me.

PR #2111, opened 2022-05-26
Optimize finalization in MsDataFileImpl

chambm:
* switched C# version to 8.0 in ProteowizardWrapper.csproj and Skyline.csproj
* added using declarations for most of the calls into the pwiz_data_binding.dll in MsDataFileImpl, so the binding code disposes the underlying resources immediately instead of gumming up the GC thread with millions of binding objects (which was the cause of the slowdown with ReSharper's Dynamic Program Analysis feature)

There is a significant improvement in my timings for TestDiaTtofFullSearchTutorial, which was the test case Brendan discovered the DPA slowdown on.  My times go from average ~430 seconds to ~380 seconds. However, when I tested it on zzzNativeVsMz5_ThermoDiaChromatogramPerformanceTest I got no noticeable difference. I'd like to see how it performs on nightlies.

If everybody has their VS2022 up and running now, C# 8.0 should not be a problem. The csproj file(s) has to be manually edited (or it can be put in the Directory.Build.props file) because the IDE pegs the C# version to the .NET version used, and the .NET Framework is now AFAIK frozen at 7.3. The [default interface members feature won't work](https://stackoverflow.com/questions/56651472/does-c-sharp-8-support-the-net-framework) but everything else from 8.0 does (or can be added with NuGet packages).

Nick, I think we can in the long run switch to .NET Core and still be able to use C++/CLI interop, but I haven't proven that yet and I think it's a longer term change.


nickshulman (commented):
My impression is that we should stick with the old C# way of doing things where the using (x) statement has to be followed by curly braces.
We can turn on "Hide Whitespace" in this code review, and it won't look like that much of a code change.

chambm:
> My impression is that we should stick with the old C# way of doing things where the using (x) statement has to be followed by curly braces. We can turn on "Hide Whitespace" in this code review, and it won't look like that much of a code change.

Look again. :) I actually chose not to convert the existing using statements to using declarations because it would make the diff harder to understand. All the differences are *new* using declarations. I could have done them with using statements, but in some of these cases it would be pretty ugly, e.g.
```
            using var scans = spectrum.scanList.scans;
            if (scans.Count == 0)
                return null;
            using CVParam param = scans[0].cvParam(CVID.MS_scan_start_time);
            if (param.empty())
                return null;
            return param.timeInSeconds() / 60;
```
vs
```
        using (var scans = spectrum.scanList.scans)
        {
            if (scans.Count == 0)
                return null;
            using (CVParam param = scans[0].cvParam(CVID.MS_scan_start_time))
            {
                if (param.empty())
                    return null;
                return param.timeInSeconds() / 60;
            }
        }
```

And in every case I think the using declaration is easier to parse.

chambm:
New syntax-only features in C# 8.0 (that will work without switching to .NET Core):
[Static local functions](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#static-local-functions)
[Using declarations](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#using-declarations)
[Null-coalescing assignment](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#null-coalescing-assignment)
[Readonly members](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#readonly-members)
[Disposable ref structs](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#disposable-ref-structs)
[Positional patterns](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#positional-patterns)
[Tuple patterns](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#tuple-patterns)
[Switch expressions](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#switch-expressions)
[Nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#nullable-reference-types) are also supported, but the new [nullable attributes](http://docs.microsoft.com/en-us/dotnet/csharp/nullable-attributes) required to design the more complex nullable use cases are not.

Reading about those features I learned the null-conditional `.?` operator has the related version, `?[` version for conditional index access. Seems it's used in Skyline already in a few places.

nickshulman (approved):

PR #2112, opened 2022-05-28
Attempt to fix clickonce installer error "The element 'assembly' has invalid child element 'SignedInfo'".

nickshulman:
This error is the same as mentioned on this StackOverflow page:
https://stackoverflow.com/questions/27106385/clickonce-manifest-has-invalid-children-but-xml-matches-old-published-manifest
where they said the solution is to remove the Vista line from the application compatibility section.
From testing on my Windows 11 machine, it appeared to be necessary to remove both the Windows 7 and Windows Vista GUIDs.

### DIFF
--- a/pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/Properties/app.manifest
+++ b/pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/Properties/app.manifest
@@ -27,9 +27,6 @@
            and is designed to work with. Uncomment the appropriate elements
            and Windows will automatically select the most compatible environment. -->
 
-      <!-- Windows Vista -->
-      <supportedOS Id="{e2011457-1546-43c5-a5fe-008deee3d3f0}" />
-
       <!-- Windows 7 -->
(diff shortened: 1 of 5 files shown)


brendanx67:
Hmmm... Obviously we need to fix this. Are we ready to remove Windows 7 compatibility? That seems bigger than something we should do in a rush like this. Can you rollback the change that put us in this situation? Wouldn't that just leave us not reporting the Windows system version correctly?

Please clarify the cost of reverting versus accepting these changes. At first glance it would seem to be:
1. These changes - dropping Windows 7 support
2. Reverting - going back to poor OS version reporting

nickshulman:
I am not sure whether removing a <supportedOS> element would actually mean that we would not be able to run on that OS.
I will test this out with my Windows 7 virtual machine.

nickshulman:
Skyline will still install fine on my Windows 7 virtual machine even if I remove the &lt;supportOS> tag for Windows 7.

brendanx67:
Okay. Go ahead, then. I will give list a final test on my Windows 7 machine
before releasing a patch to Skyline-daily

PR #2120, opened 2022-06-02
Replace PeptidesPerProteinDlg with AssociateProteinsDlg in Import Peptide Search wizard

chambm:
image

I commented out the `NewTargetsAll` sections of tests, but I could instead just delete them.

I haven't changed the default options for parsimony as you want Brendan, because I'm not sure what you want the new defaults to be and when I do change them it'll mean overriding the previous defaults in all the relevant tests.

The wizard-launched version of the form does conditionally add the "Existing protein associations will be discarded" sentence based on whether the input document to the dialog has any protein nodes. Unfortunately that is the case with the current implementation. The ImportFastaControl creates preliminary protein nodes before running the AssociateProteinsDlg (behavior kept from the PeptidesPerProteinDlg approach). Should I eliminate that step?


pwiz_tools/Skyline/FileUI/PeptideSearch/ImportFastaControl.cs
```diff
@@ -420,72 +420,92 @@ public bool ImportFasta(IrtStandard irtStandard)
...
                 var nodeInsert = _sequenceTree.SelectedNode as SrmTreeNode;
                 IdentityPath selectedPath = nodeInsert != null ? nodeInsert.Path : null;
                 var newPeptideGroups = new List<PeptideGroupDocNode>();
 
-                if (!_fastaFile)
+                // if user provided FASTA content, write it to a temporary file
+                string fastaFilepath = _fastaFile ? tbxFasta.Text : Path.GetTempFileName();
+                try
                 {
-                    FastaText = tbxFasta.Text;
```
chambm, line 430: @nickshulman Here's a case where a using declaration would come in pretty handy to avoid this Path.GetTempFileName()/try/finally construct which incidentally makes the diff hard to read. If I made a TemporaryFile class like the existing TemporaryDirectory class, I could use that with a using declaration instead of this try/finally. (The using statement version would probably still make a hard-to-read diff like this one due to indentation changes.)

nickshulman: I am not sure I understand why a temporary file needs to be created here. Do we need to pass that file path to some other code that really requires a real file on disk and could not take a stream?

brendanx67:
How much more time does this add to the functional tests?

chambm:
10 seconds. I'm thinking about taking 10 seconds off the 2 other tests by testing one file instead of two but I can do that in a separate PR. (I actually only did the 2nd loop thing in this PR to try to make sure coverage is increasing, because I finally got ReSharper coverage working in my local VS)

brendanx67:
Sounds like a relatively expensive way to get increased coverage to me.
Maybe write some new instant unit tests.

chambm:
@brendanx67 I think this is ready for review if you want.

PR #2123, opened 2022-06-04
Remove spurious "ratio to" on the Targets tree if no ratio is selected.

nickshulman:
Fixed problem where if the Targets tree was not supposed to be showing ratios, it would display raw peak areas followed by the word "ratio". (Reported by Brendan)


pwiz_tools/Skyline/Controls/SeqNode/TransitionTreeNode.cs
```diff
@@ -21,6 +21,7 @@
 using System.Drawing;
 using pwiz.Common.SystemUtil;
 using pwiz.Skyline.Model;
+using pwiz.Skyline.Model.GroupComparison;
```
brendanx67, line 24: Seems a bit odd that this is suddenly required now. I guess NormalizationMethod must be in here?

brendanx67 (approved):

PR #2128, opened 2022-06-14
Fix problem where opening a document with a background proteome and a peptide uniqueness constraint sometimes results in all peptides being removed from document.

nickshulman:
I have run into a problem a few times where someone has sent me a .sky.zip file, and, when I open it, all of the peptides have been deleted from the document.
This happens if the document has a Background Proteome and a PeptideUniquenessConstraint.

This does not typically happen if you open a document which you already have on your computer, because it only happens if the path in the BackgroundProteomeSpec changes in "SkylineWindow.ConnectBackgroundProteome".

The change in this pull request fixes the problem by making it so that we only pay attention to a background proteome change if they libraries have been loaded.

I feel like there is a more fundamental problem that should be addressed here which is that if "SrmSettingsDiff.DiffPeptides" happens to be true, and the libraries have not been loaded, then all peptides will be deleted from the document (if the PeptidePick setting indicates that the peptides need to be in the library).

Does anyone have any ideas for a more robust way to fix this bug?

If you want to observe this behavior on your own, just open the .sky file inside of the test .zip that I have added to this pull request. When you open that Skyline document you will see that the two peptides that are supposed to be in there have been deleted.
This bug has been around for many years.


pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
```diff
@@ -2578,7 +2578,7 @@ public SrmSettingsDiff(SrmSettings settingsOld, SrmSettings settingsNew)
             // changes, or if constraint is non-None and background proteome or digestion enzyme changes
             // Background proteome uniqueness constraints - only considered a change if constraint type
             // changes, or if constraint is non-None and background proteome or digestion enzyme changes
-            bool uniquenessConstraintChange = !Equals(newPep.Filter.PeptideUniqueness, oldPep.Filter.PeptideUniqueness);
+            bool uniquenessConstraintChange = uniquenessConstraintChange = !Equals(newPep.Filter.PeptideUniqueness, oldPep.Filter.PeptideUniqueness);
```
bspratt, line 2581: this seems weird
uniquenessConstraintChange = uniquenessConstraintChange =

pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
```diff
@@ -2594,6 +2594,8 @@ public SrmSettingsDiff(SrmSettings settingsOld, SrmSettings settingsNew)
                 }
                 uniquenessConstraintChange = uniquenessConstraintChange || !newPep.DigestSettings.Equals(oldPep.DigestSettings);
             }
+            // Don't try to update peptide uniqueness if the libraries have not yet been loaded
+            uniquenessConstraintChange = uniquenessConstraintChange && newPep.Libraries.IsLoaded;
```
bspratt, line 2598: maybe (or am I showing my C++ roots?)
uniquenessConstraintChange &= newPep.Libraries.IsLoaded;

pwiz_tools/Skyline/TestFunctional/OpenDocWithBackgroundProteomeTest.cs
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using pwiz.SkylineTestUtil;
+
+namespace pwiz.SkylineTestFunctional
+{
```
bspratt, line 24: A bit of commenting explaining what this test is checking for would be good - that is, explaining what the problem was

bspratt (commented):
Seems good but I did have a few notes

nickshulman:
I think there is a more general bug here that needs to be fixed.
If a SettingsChange comes through where SrmSettingsDiff.DiffPeptides is true, and the libraries have not yet been loaded, all peptides get deleted from the document.
Another way that I could make that happen is to bring up the Peptide Settings dialog and change the enzyme, and also add a new library.
I think the real fix is going to be something that causes the PeptideGroupDocNode to preserve its children inside of ChangeSettings if it see that the libraries have not yet been loaded.

bspratt:
It's admirable the way Skyline works to remain responsive to the user when certain long-running processes like library load are completing, but I wonder if it isn't acceptable to just lock out the settings change UI when already in a half-baked state that is likely to result in a change to the document on completion. I think users would understand and tolerate "Please wait for current operation to complete" if it came with a "Cancel current operation" option.

brendanx67:
In the spirit in which this was implemented, you could certainly turn off DiffPeptides if:
1. The document is not fully loaded.
2. The diff between the current document and the loaded document will cause DiffPeptides.
Being not fully loaded is essentially an IOU for the processing that will happen when loading is completed.

I think, you could add this to SrmSettingsDiff by simply checking whether the libraries are fully loaded, and turning off DiffPeptides if they are not, right?

pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
```diff
@@ -2785,6 +2785,14 @@ public SrmSettingsDiff(SrmSettings settingsOld, SrmSettings settingsNew)
             // Force update if the bulk import has just completed
             else if (settingsOld.HasResults && settingsOld.MeasuredResults.IsResultsUpdateRequired)
                 DiffResults = true;
+
+            if (settingsNew.HasLibraries && !settingsNew.PeptideSettings.Libraries.IsLoaded &&
+                settingsNew.PeptideSettings.Libraries.Pick != PeptidePick.filter)
+            {
+                // If the libraries have not been loaded then we will not be able to determine which peptides
+                // should be chosen
+                DiffPeptides = false;
```
brendanx67, line 2794: I think you want to turn off all of DiffPeptides, DiffTransitionGroups, and DiffTransitions. Any of these could have disastrously bad results if applied before the libraries are loaded.

brendanx67 (changes requested):
Glad turning off DiffPeptides works for your test case. I think we still need to turn off the other element diffs to avoid erroneous large-scale changes when this type of diff is applied before spectral libraries are loaded.

brendanx67 (approved):

nickshulman:
This fix is causing a failure in "ConsoleChangeFilterSettingsTest". I am thinking that the fix is that CommandLine.ProcessDocument needs to wait for the document to be loaded before doing its stuff.

pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
nickshulman: Oops. That's a weird line of code for me to have written. There was supposed to be a "&&" in there instead of the "=".

pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
brendanx67: Yeah. Bad C++ habits die hard, like using bitwise & on booleans. ReSharper would be well within its rights to warn you about this.

bspratt: Seems like more concise syntax - why is it problematic?

brendanx67: The trend from C++ to Java/C# is to make operations more type-explicit, and & is a bitwise operator, not a boolean operator. Many of the "cleaver" conveniences of C++ have died in Java/C# and thankfully even some of the stylistically  cleaver syntax to avoid C++ pitfalls have also become unnecessary, e.g. if (null == myVal) to avoid assignment within an if statement. I applaud the Java developers who initiated this change, which makes me not very appreciative of C++ cleverness.

Interesting discussion here:

https://stackoverflow.com/questions/6346001/why-are-there-no-or-operators-in-c

Even the person with the top response who says you can use &= (instead of &&=) is not a big fan.

nickshulman: I believe the line of code that Brian originally commented on is very different than the one we are discussing now.

I was shocked to see that my initial checkin had this completely nonsensical line of code in it:
`bool uniquenessConstraintChange = uniquenessConstraintChange = !Equals(newPep.Filter.PeptideUniqueness, oldPep.Filter.PeptideUniqueness);`

And I was like: "[Good catch, Brian] that's a weird line of code for me to have written..."

There's no coding style that would decide between "&=" or "= ... && "; both would be acceptable.

brendanx67: So, on we go. Yes, I saw that change, and it was clearly a good one. Surprised that ReSharper wouldn't warn about the redundant assignment in the original.

chambm: I think C++ is IMO even more type specific than C# and Java. Especially early Java and C#. Both latter languages had a lot of revisions to make them more type explicit. For example, all the reference types derive from object, there's ArrayList, and `*Collection`s in a bunch of WinForms controls that return objects instead of base types. But C++ definitely takes the cake for confusing paradigms (template metaprogramming, anyone?) and baffling and absurdly verbose compile errors. But hey, _if you do it right_, you get [zero-cost abstraction](https://joshpeterson.github.io/a-zero-cost-abstraction) (unless you need to turn off inlining and optimizations...then it's a big cost :) ).

brendanx67: The stackoverflow poster has a point, though:
a = a && b can also be if (a) a = b
a = a || b can also be if (!a) a = b
And you get the ability to set a breakpoint in the deal!

chambm: One reason to use bitwise operators for bools is if performance is a concern (e.g. inside a deep deep computational loop) and short-circuiting isn't needed. With bitwise you can avoid branches.

PR #2129, opened 2022-06-14
SkylineTester: add a couple of UI hints encouraging the use of the Tests tab for demos instead of the Tutorials tab…

bspratt:
… , which has fewer options. 

Nick and I both were confused by this when trying to get a rolling multi-language demo going. Arguably we should just rip that control out but it's presumably there for a reason.

### DIFF
--- a/pwiz_tools/Skyline/SkylineTester/SkylineTesterWindow.Designer.cs
+++ b/pwiz_tools/Skyline/SkylineTester/SkylineTesterWindow.Designer.cs
@@ -871,7 +871,8 @@ namespace SkylineTester
             this.tutorialsDemoMode.Name = "tutorialsDemoMode";
             this.tutorialsDemoMode.Size = new System.Drawing.Size(82, 17);
             this.tutorialsDemoMode.TabIndex = 6;
-            this.tutorialsDemoMode.Text = "Demo mode";
+            this.tutorialsDemoMode.Text = "Demo mode (deprecated, use Tests tab)";
+            this.toolTip1.SetToolTip(this.tutorialsDemoMode, "Use the Tests tab's 'Tutorials only', 'Mode', and 'Run indefinitely' settings for more fine-grained control of demos");
             this.tutorialsDemoMode.UseVisualStyleBackColor = true;
             // 
             // label5


nickshulman:
Do we know why anyone would want to use the "Demo mode" option on the tutorials tab?
If we cannot think of reason it would be used, we should remove the radio button.
If there is a good reason for it, maybe we could change the text on the button to reflect what you might want to use it for. Is it "Single language tutorial only demo mode"? "Quick Review Mode"?

brendanx67:
I guess we could do a full clean-up of this UI to remove redundancy and stick with only the things we really use... It is not user facing, and we only use this particular feature once a year (when ASMS is not canceled). What you used this year was how I designed and used it originally at ASMS. The radio buttons on the Tutorials page were added by Don Marsh as perhaps an easy way to get to the feature for testing? This past year, I added even more flexibility on the Test tab, so that I almost only us the Tutorials page for screenshots now. In theory, all of the radio buttons could be removed from the Tutorials tab without impacting me, since I made it possible to run the Tests tab in both Demo mode and Covershots mode. Having control of the number of seconds we pause for Demo mode was never that interesting.

Or, we could leave it as it is and hope that no one will ever lose as much time to this UI as we have this year, after 3 years off.

bspratt:
>> Or, we could leave it as it is and hope that no one will ever lose as much time to this UI as we have this year, after 3 years off.

My proposed change should accomplish this. Not a big deal to just tear it out, though. Your call.

pwiz_tools/Skyline/SkylineTester/SkylineTesterWindow.Designer.cs
```diff
@@ -871,7 +871,8 @@ private void InitializeComponent()
             this.tutorialsDemoMode.Name = "tutorialsDemoMode";
             this.tutorialsDemoMode.Size = new System.Drawing.Size(82, 17);
             this.tutorialsDemoMode.TabIndex = 6;
-            this.tutorialsDemoMode.Text = "Demo mode";
+            this.tutorialsDemoMode.Text = "Demo mode (deprecated, use Tests tab)";
```
brendanx67, line 874: Let's just "Demo mode (deprecated)". Presumably the curious can get the other information from the tip. I know this is not user-facing, but the full text for me comes close to the line where I am tempted to consider a better looking solution.

brendanx67 (changes requested):

PR #2134, opened 2022-06-23
Switch from hybrid runsettings/testsettings to pure runsettings

chambm:
* turned off deployment, legacy mode, and testsettings inclusion in runsettings; ReSharper seems to run much better in VS 2022 without it (coverage mode won't work with testsettings and legacy mode), but need to test on TeamCity as well


chambm:
I started a code coverage run on TC. I think that's the only config that uses runsettings, am I right?

brendanx67:
Note that Nick and I still use Run All in Visual Studio and expect that all to pass, which requires the runsettings/testsettings configuration.

chambm:
> Note that Nick and I still use Run All in Visual Studio and expect that all to pass, which requires the runsettings/testsettings configuration.

That is not true as far as I can tell. I've been able to run all the tests in both ReSharper and native Test Explorer with just runsettings.

chambm:
I actually had switched us to RunSettings only a long time ago (with deployment turned off) but IIRC Nick reverted because it made the test directories get created in the release/binaries directory instead of the TestResults directory. So this time around I've addressed that issue as well.

chambm:
And run all takes so long without parallel testing that even though I haven't got it to work on TeamCity reliably yet I'm tempted to go ahead and merge #2033 for us to use personally (I would just disable it from running in parallel on TeamCity until I can get that figured out).

chambm:
The TeamCIty issue has been resolved. @brendanx67 @nickshulman do you want to try out this PR to check whether Test Explorer and ReSharper coverage are working for you in VS 2022? @bspratt and I have both tested ReSharper coverage, and I have tested Test Explorer (not sure about Brian).

nickshulman:
Running the tests in Test Explorer seems to work for me.
I imagine you should update "TestSettings_x86.runsettings" even though we don't care about the 32 bit build anymore.
Also, you should probably delete the "*.testsettings" files so that nobody gets confused by them.

pwiz_tools/Skyline/TestData/TestData.csproj
```diff
@@ -224,10 +224,6 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
-    <None Include="..\app.config">
-      <Link>app.config</Link>
```
chambm, line 228: This interesting change was needed to get the code coverage build working on TeamCity: otherwise I got a userSettings configuration error in some of the TestFunctional tests (not TestData, which is the odd part). Removing it seems to have had no other impact as far as I can tell.

chambm:
Like the C# 8.0 upgrade, this change could be pretty annoying when switching between release branch and master (you'd have to switch which settings file to use in the IDE). So I think we can delay merging until we're pretty sure we don't need to switch to 21.2 (or cherry pick it to there as well? @brendanx67 I can't think of a reason this wouldn't work fine with 21.2 built with VS 2022; built with VS 2017 is a different story).

PR #2137, opened 2022-06-27
Added SendGa4AnalyticsHit() to send an event to GA4

chambm:
* added SendGa4AnalyticsHit() to send an instance event to the new GA4 property

We continue to send events to UA. I modeled the GA4 event after the UA event, but I'm not sure if there isn't some better way to do it. There's a lot of builtin events in GA4, but none of them seem to fit what we're doing with the custom "Instance" event. The "login" event seemed closest but I wasn't clear whether you're supposed to add custom properties to builtin event types. And it's not really a login anyway.

I am considering whether to make the validation variable dependent on whether it's running as a test. But we'd only want to test it once (could be its own test). Do I need to make it public to call it from a unit test?

### DIFF
--- a/pwiz_tools/Skyline/Program.cs
+++ b/pwiz_tools/Skyline/Program.cs
@@ -390,15 +390,16 @@ namespace pwiz.Skyline
             // ReSharper restore LocalizableElement
         }
 
-        internal static void SendGa4AnalyticsHit(bool validateEvent = false)
+        internal static string SendGa4AnalyticsHit(bool useDebugUrl = false)
         {
             // ReSharper disable LocalizableElement
             const string measurementId = "G-CQG6T54XQR";
             const string apiSecret = "8_Ci004BQSKdL1bEazPK3A"; // does this need to be kept secret somehow?
-            string analyticsUrl = $"https://www.google-analytics.com/{(validateEvent ? "debug/" : "")}mp/collect?measurement_id={measurementId}&api_secret={apiSecret}";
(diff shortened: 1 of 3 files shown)


nickshulman (approved):
Looks good to me.

chambm:
How bad is my unit-test-an-internal hack? Better than just making `SendAnalyticsHitAsync()` public? I'm not sure if UA has an equivalent debug mode which doesn't actually send an event.

pwiz_tools/Skyline/Program.cs
```diff
@@ -386,6 +390,52 @@ private static void SendAnalyticsHit()
             // ReSharper restore LocalizableElement
         }
 
+        internal static void SendGa4AnalyticsHit(bool validateEvent = false)
```
nickshulman, line 393: I think this method should return "responseString". Skyline will ignore the return value, but your unit test will assert that the responseString is what it is supposed to be.
Also, I think this "validateEvent" parameter should be named "debug" or "useDebugUrl".

chambm:
@brendanx67 What's your position on cherry picking this to release and making a patch release so we start getting GA4 hits for the current release?

brendanx67:
> @brendanx67 What's your position on cherry picking this to release and making a patch release so we start getting GA4 hits for the current release?

I am okay with that. Brian Pratt also has a fix that missed the last patch. We do not track how quickly patch updates are adopted. It may not be as immediate as we might hope.

PR #2139, opened 2022-06-28
The UniprotKB API has shifted again, here are a couple of tweaks to resume working with that

bspratt:
MattC noticed this because one of the perf tests that accesses Uniprot slowed down dramatically

### DIFF
--- a/pwiz_tools/Shared/ProteomeDb/API/ProteinMetadata.cs
+++ b/pwiz_tools/Shared/ProteomeDb/API/ProteinMetadata.cs
@@ -306,13 +306,14 @@ namespace pwiz.ProteomeDatabase.API
         public WebSearchTerm(char service, string query)
         {
             Service = service;
-            if (service == WebEnabledFastaImporter.UNIPROTKB_TAG)
+            if (service == WebEnabledFastaImporter.UNIPROTKB_TAG && !string.IsNullOrEmpty(query))
             {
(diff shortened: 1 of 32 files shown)


pwiz_tools/Shared/ProteomeDb/API/ProteinMetadata.cs
```diff
@@ -306,7 +306,26 @@ public class WebSearchTerm: IEquatable<WebSearchTerm>
...
+                }
+                else
+                {
+                    Query = query;
+                }
+            }
+            else
+            {
+                Query = query;
+            }
```
chambm, line 328: You could flatten/simplify this a bit like:
```
Query = query;
if (service == WebEnabledFastaImporter.UNIPROTKB_TAG)
{
                // UniprotKB has gotten tweaky about SGD yeast entries, as of 6/28/2022 only wants to see "S000000001" of "SGD:S000000001"
                if (query?.StartsWith(@"SGD:")) 
                {
                    Query = query.Substring(4);
                }
                else if (query?.StartsWith(@"SGDID:"))
                {
                    Query = query.Substring(6);
                }
}
Can query actually be null? What would that mean?

bspratt: >> Can query actually be null? 

It only happens during unit tests when we're not actually accessing the web.

PR #2142, opened 2022-06-30
Improved handling of transition list imports when the number of columns varies from line to line.

bspratt:
Reported anonymously via exception web.


nickshulman (approved):
Looks good to me.

PR #2144, opened 2022-07-01
Added natural sort ordering preset for selection menus and added tests to verify accuracy

ClarkUPS:
Added natural sort ordering preset for selection menus and added tests to verify accuracy


pwiz_tools/Shared/Common/Collections/NaturalComparer.cs
```diff
@@ -0,0 +1,102 @@
...
+                // If both values are numbers
+                if (int.TryParse(left, out var l) && int.TryParse(right, out var r))
+                {
+                    var comp  =l.CompareTo(r);
+                    if (comp != 0)
+                    {
+                        return comp;
+                    }
+                }
+                else if (!int.TryParse(left, out _) || !int.TryParse(right, out _)) // If both are letters
```
bspratt, line 69: Let's try cacheing these TryParse results - that's a fairly expensive call, this could start to be a performance issue in a long list

pwiz_tools/Shared/Common/Collections/NaturalComparer.cs
```diff
@@ -0,0 +1,102 @@
...
+                {
+                    var comp = string.Compare(left, right, StringComparison.CurrentCulture);
```
nickshulman, line 71: This should probably be "StringComparison.CurrentCultureIgnoreCase".

pwiz_tools/Shared/Common/Collections/NaturalComparer.cs
```diff
@@ -0,0 +1,102 @@
...
+                {
+                    return 1;
+                }
+                else if (int.TryParse(right, out _)) //If left is a number, left wins
+                {
+                    return -1; 
+                }
+            }
+            //If all else equal go by length. If length is the same they are identical
+            if (y1.Length > x1.Length)
```
nickshulman, line 87: This whole set of if/else blocks should be replaced by:
return x1.Length.CompareTo(y.Length);

pwiz_tools/Shared/Common/Collections/NaturalComparer.cs
bspratt: Actually Nick is of the opinion that we should just replace all this with a call to the Windows implementation, that's probably not a bad idea, and an interesting thing to learn how to do. Look for "DLLImport" for examples of that in our code.

bspratt: https://www.pinvoke.net/default.aspx/shlwapi.strcmplogicalw

nickshulman (commented):
This looks great, but it would probably be best if you used the Windows API "StrCmpLogicalW".
There is some example code for how to do that here:
https://www.pinvoke.net/default.aspx/shlwapi.strcmplogicalw

By the way, here is someone else's C# implementation of natural sort:
https://gist.github.com/mstum/63a6e3e8cf54e8ae55b6aa28ca6f20c5
The implementation is interesting to look at because they implemented it without using "Regex.Split", so it's probably faster.
One tricky thing to deal with is things which have too many digits in them to be successfully parsed as an integer, which is why this other C# implementation had to add that class "ParsedNumber".

Someone has made a page listing all of the quirks in the way that the Windows API "StrCmpLogicalW" behaves including the treatment of leading zeroes and decimal points:
https://www.geoffchappell.com/studies/windows/shell/shlwapi/api/strings/strcmplogicalw.htm

pwiz_tools/Shared/Common/Collections/CollectionUtil.cs
```diff
@@ -202,7 +202,7 @@ public static int CompareColumnValues(object o1, object o2)
             {
                 return 1;
             }
-            return Comparer.Default.Compare(o1.ToString(), o2.ToString());
+            return NaturalComparer.Compare(o1.ToString(), o2.ToString()); // Natural Compare
```
bspratt, line 205: Not as useful a comment as it could be - basically just repeating the name of the function call. We want to let the casual reader easily understand what that function does, so something more like 

// Sort in the same manner as Windows Explorer, where for example [z11ab, z2ab] sorts as ]z2ab, z11ab] (because 11 is more than 2)

pwiz_tools/Skyline/FileUI/OpenDataSourceDialog.cs
```diff
@@ -457,15 +459,26 @@ private void populateListViewFromDirectory(MsDataFileUri directory)
...
 
+            listSourceInfo = listSourceInfo.Where(l => l != null).ToList(); // Ignore null entries
+
+            listSourceInfo.Sort((x, y) =>
+            {
+                if (x.isFolder != y.isFolder)
+                {
+                    return x.isFolder ? -1 : 1;
+                }
+                return NaturalComparer.Compare(x.name, y.name);
```
bspratt, line 470: Again, worthy of a comment

pwiz_tools/Skyline/Test/TextUtilTest.cs
```diff
@@ -96,5 +97,155 @@ private IEnumerable<string> AddSuffix(string[] baseStrings, string fixText, int
             foreach (string baseString in baseStrings)
                 yield return baseString + fixText.Substring(fixText.Length - i, i++);
         }
+
+        [TestMethod]
```
bspratt, line 101: Needs a comment explaining what the test is testing for

pwiz_tools/Skyline/Test/TextUtilTest.cs
```diff
@@ -96,5 +97,155 @@ private IEnumerable<string> AddSuffix(string[] baseStrings, string fixText, int
...
+                "XfWhr","XfY9w","xgThP","-XI_N","XiSQ3","XJ0Ij","xJl17","XjWWj","XlGMw","xlt5k","XMgjD","XPCHC",
+                "XPxAQ","XQkWA","xRiK2","XSkvk","XUO_k","XWPlc","Xx0hO","XXA97","XxgDy","XZ_VJ","xZfiP","Y6Egb",
+                "Y6KYN","y7C8P","y7VP8","Y8P0V","Y8tqk","YBhcq","YbyUR","yE59F","yewLE","YFavT","yFJtl","YFnAx",
+                "yIaqR","YJ_Bw","ykUfT","Ylp6y","ymA2t","yPNFs","-yqQU","YS_68","YsfqU","YtEmi","ytYCY","YtZZ8",
+                "yUzgJ","YZ09K","yZeZY","Yzlgb","Z-0NM","z1hKB","z7DVy","Z8FV9","Z8qKD","Z96lA","zAu-r","ZBE3X",
+                "ZbPja","zbUz-","ZDjTb","Zdrnb","zFjSv","zFunl","zIW8u","zjUA6","zL3Mn","ZLImV","ZLpfb","zMVI0",
+                "zOpUQ","Zuzaq","ZvFIb","zXdQ1","zxWHd","zz0Pq","ZzIxa","zzlPp","ZZPVw"
+            };
+
+
```
bspratt, line 222: Not a single comment anywhere in this code. It's possible to over-comment, of course, but a quick sketch of the logic of the test is a good idea

// Take a known-sorted list, randomize it, make sure that our sort algorithm puts it back in order again

bspratt (commented):
Looking solid, just lacking comments. On a project like this you're coding for the ages - somebody is going to walk up to this 5-10 years from now and will expect a brief english language explanation of what's going on

brendanx67:
> Looking solid, just lacking comments. On a project like this you're coding for the ages - somebody is going to walk up to this 5-10 years from now and will expect a brief english language explanation of what's going on

And also just for a few months from now after you leave and someone reports a bug in the code.

bspratt:
I'm fine with just moving that test to TestFunctional. Clark, let me know if you need any help with that refactoring.

bspratt:
Though when I think about it, maybe the smart thing is just to figure out where they disagree, and edit the test to avoid that. I'll bet it's all weird corner cases that we don't much care about, and that could break in future Windows versions as well (we know that such things have happened in the past). All we're really after is "File2" comes before "File11", after all.

bspratt:
(So, Clark, change the test's failure behavior to not only call out the first found disagreement, but to also emit the entire sorted list)

brendanx67:
> Though when I think about it, maybe the smart thing is just to figure out where they disagree, and edit the test to avoid that. I'll bet it's all weird corner cases that we don't much care about, and that could break in future Windows versions as well (we know that such things have happened in the past). All we're really after is "File2" comes before "File11", after all.

I wouldn't say it is quite that simple. People put a lot of numerical information in their mass spec data file names and they don't always get the digits right. Our own Skyline installers have this problem in a more complex way than File2 and File11.

Skyline-64_1_4_0_4421.zip
Skyline-64_20_1_0_155.zip
Skyline-64_2_5_0_5675.zip

Windows sorts these files well, but we would not.

pwiz_tools/Shared/Common/Collections/CollectionUtil.cs
```diff
@@ -202,7 +202,7 @@ public static int CompareColumnValues(object o1, object o2)
             {
                 return 1;
             }
-            return Comparer.Default.Compare(o1.ToString(), o2.ToString());
+            return NaturalComparer.Compare(o1.ToString(), o2.ToString()); // Compare using natural non-lexicographical order for easier human readability (1,20,3 --> 1,3,20)
```
bspratt, line 205: Closer - but the point is that strings of mixed letters and numbers sort a certain way. Just 
(a1b,a20b,a3b --> a1b,a3b,a20b)
would be an improvement in clarity

pwiz_tools/Shared/Common/Collections/NaturalComparer.cs
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+using System.Runtime.InteropServices;
+
+namespace pwiz.Common.Collections
+{
+    /// <summary>
+    /// Added by Clark Brace (cbrace3)
+    /// https://www.pinvoke.net/default.aspx/shlwapi.strcmplogicalw
+    /// </summary>
```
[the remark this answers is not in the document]
bspratt: Better, I'd add an example like (a1b,a20b,a3b --> a1b,a3b,a20b). It's better to err on the side of verbosity.

pwiz_tools/Skyline/FileUI/OpenDataSourceDialog.cs
bspratt: Use that example here too

pwiz_tools/Skyline/Test/TextUtilTest.cs
bspratt: Closer - but it's not shuffling random strings, it's shuffling sorted strings then checking to see if we can reestablish that oder

pwiz_tools/Shared/Common/Collections/NaturalComparer.cs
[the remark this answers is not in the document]
bspratt: I really would like to see an example in the summary.

bspratt:
Huh, so Wine and Windows also disagree on the proper sort for "4tuzE","04ZaO". Once again I don't think we actually care enough to insist that it's one or the other. I'd just pull the "04ZaO" element from the test. 

Really, we could remove everything but "Skyline-64_1_4_0_4421.zip","Skyline-64_1_4_0_4422.zip","Skyline-64_2_5_0_5675.zip","Skyline-64_20_1_0_155.zip" and achieve what we're after, but I'm not actually proposing that.

pwiz_tools/Shared/Common/Collections/NaturalComparer.cs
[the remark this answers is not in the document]
bspratt: "e.g." is more usual than "ex:" in our (and most, I'd say) codebase

also "https://www.pinvoke.net/default.aspx/shlwapi.strcmplogicalw" should probably read "See https://www.pinvoke.net/default.aspx/shlwapi.strcmplogicalw"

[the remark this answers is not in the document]
bspratt: Still says "ex:" here. The IDE is pretty quick at searching for strings in the entire project, you should use that. (But don't mess with any files you aren't already involved with)

PR #2149, opened 2022-07-06
Fixed concurrent transition count in RT scheduling graph

kaipot:


brendanx67 (changes requested):
Looks good. Just a bit more testing to make sure the View > Retention Times > Scheduling graph works as expected also.

PR #2150, opened 2022-07-06
Add "Show commandline" button to MSConvertGUI

chambm:
- added "Show commandline" button to MSConvertGUI to make it easy to get the equivalent msconvert.exe command

### DIFF
--- a/pwiz_tools/MSConvertGUI/MainForm.Designer.cs
+++ b/pwiz_tools/MSConvertGUI/MainForm.Designer.cs
@@ -63,7 +63,36 @@ namespace MSConvertGUI
             this.RemoveFileButton = new System.Windows.Forms.Button();
             this.StartButton = new System.Windows.Forms.Button();
             this.FilterGB = new System.Windows.Forms.GroupBox();
+            this.DiaUmpirePanel = new System.Windows.Forms.Panel();
+            this.DiaUmpireParamsFileBrowseButton = new System.Windows.Forms.Button();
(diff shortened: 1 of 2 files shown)


brendanx67:
I like the feature. Unfortunately, when I start MSConvertGUI, immediately remove the "titleMaker" filter and then press the button, I still get a command like with the titleMaker filter in it. It does seem to respond to my adding other filters like peak picking and demultiplexing, but it always seems to add a titleMaker filter at the end, even when...

Oh, strange. I can only get rid of the titleMaker filter by unchecking the "TPP compatibility" checkbox. So, I think really the bug is that the "titleMaker" filter shows up at all in the "Filters" list. Once you delete it, there is no way I can figure out to get it back using the Filters dropdown list, or unchecking and rechecking the "TPP compatibility" checkbox.

Either make the two linked so that deleting the filter unchecks the checkbox and checking the checkbox puts it back, or just make the checkbox cause a hidden filter that doesn't show up in the list.

brendanx67 (approved):
Other than the confusion over the titleMaker filter, this looks good.

chambm:
Yeah that was some annoying inconsistent behavior with the TPP checkbox. It wasn't causing a real bug, just confusion. I fixed a couple other issues that were also confusing and actually more problematic (multiple subset filters for the same category even though they're not intended to be used that way, and msLevel/scanNumber filters always being pushed to the end where they are least efficient)..

PR #2153, opened 2022-07-08
Fix Qe/Ttof DIA perf tests

chambm:
* removed attempt to calculate and test PeptidesPerProteinDlg-style pre-filtering target counts
* updated audit logs


brendanx67 (approved):
Okay. Hate to see intermediate testing lost, but I guess the audit logs will enforce the right results in the end here.

chambm:
What do you mean? We're still testing the FinalTargetCounts. It's just that the AssociateProteinsDlg doesn't really have a concept of the TargetCounts, because it redoes the mapping from scratch.

PR #2155, opened 2022-07-09
Normalization bug fixes

nickshulman:
Fix a couple of bugs related to normalization:
1. When normalizing, Skyline would never divide by a number which was less than 1 (there was a "Math.Max(1, divisor)". When doing something like Global Standard normalization, this would not be a problem, since global standard areas are typically much larger than 1 already. However, when doing median normalization or tic normalization, typically half of the replicate have a divisor which is less than one. Brendan noticed that half of the protein abundance values did not end up changing when switching between no normalization, and TIC normalization.
2. The peak area graph was incorrectly multiplying by the normalization divisor instead of dividing by it.

This change adds new Document Grid columns on the Result File object:
image
image

Skyline used to calculate the medians separately for all of the light transitions and the heavy transitions. I have now changed it so that only one median value gets calculated per Result File. This makes it much simpler to have the "Median Peak Area" column in the Document Grid. Since median normalization was already significantly broken because of the Math.Max(1, ...) bug, I don't think we need to worry about this small change in the median behavior if you happen to have light and heavy transitions in the document.

Fix some bugs related to normalization.
Fixed values calculated using TIC normalization and "Equalize Medians" normalization methods. For half of the replicates, the numbers were being divided by one, instead of by the divisor they were supposed to be divided by (reported by Brendan)
Fixed displaying median-normalized areas in the peak area graph: the numbers getting multiplied by the normalization divisor instead of divided (reported by Mike)
Fixed problem displaying Protein Abundance value in the document grid if the report did not include any columns from the peptide level or lower (reported by Brendan)


brendanx67:
@nickshulman Could you possibly have misunderstood the MSstats code that you said your median normalization was based on? I have seen Meena's slides on this a lot, I would have said her intent was to use the median of the "Internal standard type" ILT for normalization. I suppose that could require multiple normalization values just as we produce when creating ratios to standards when there are multiple "Internal standard types".

But, in the 99% simple case with just light and heavy with one being the internal standard type, then you would calculate the median of each sample based on its internal standards. This is the only way that median normalization makes any sense in a targeted experiment with labeled reference peptides. You can't at all depend on the median values of the targeted endogenous peptides, but you should be able to depend on the medians of your internal standard reference peptides.

After seeing these slides a few times and implementing median normalization in R, I came to view this as a cleaver statistical method for normalizing against reference peptides while keeping the resulting values in normal intensity units (ions per second) instead of producing ratios... Now that I say that, I would have said they were doing this "median normalization" for each individual peptide based on its reference standard. i.e. transform the endogenous intensity by taking the median intensity of the reference and then exp(log(endogenous intensity) + log(median standard intensity) - log(standard intensity))

Anwyay, when reference standards are involved, I still think median normalization is more complicated than you are probably accounting for yet. Though, you may not have gotten it right in the past either.

brendanx67:
Here is the figure from Meena's slides:
image

brendanx67:
If you want a value per sample, then I think you really, really want the median of the "internal standard type" precursors, if there is an internal standard type, since those are expected to stay constant, and you really have no idea about the endogenous peptides, and when you have an internal standard type it is pretty unusual that you can expect over 50% of the endogenous peptides to be unchanging.

brendanx67:
This seems to be failing a lot right now.

PR #2156, opened 2022-07-11
Add ion match tolerance unit transition setting (Issue 830)

rita-gwen:


rita-gwen:
Thank you, Nick! Should I also add Skyline_22_14.xsd at this point? Or it is done at the release time?

nickshulman:
Brendan will create the file "Skyline_22_14.xsd" when he releases the next Skyline-Daily.
The reason that we delay doing that is so that we don't increment the version number twice if two developers add features during the same Daily release cycle. The only .xsd file that you ever edit is "Skyline_Current.xsd". All of the numbered .xsd files should be considered immutable.
When Skyline-Daily is being released, the "IsOfficialBuild" part of "TestDocumentFormatSchemaFiles" reminds the person if the numbered .xsd file needs to be created.

pwiz_tools/Shared/Common/Chemistry/MzTolerance.cs
```diff
@@ -60,16 +60,30 @@ public MzTolerance(double value = 0, Units units = Units.mz)
...
+                    return Math.Abs(mz) * Value * 1e-6;
+            }
+
+            return 0;
+
+        }
+
         /// <summary>returns true iff a is in (b-tolerance, b+tolerance)</summary>
-        public static bool IsWithinTolerance(double a, double b, MzTolerance tolerance)
+        public bool IsWithinTolerance(double a, double b)
```
rita-gwen, line 78: I have a question. Why the inequality is non-inclusive here while it is inclusive for the method match tolerance?

nickshulman: Good question.
@chambm, is there a reason that this method is using an exclusive range?
I see that this C# method is not used anywhere but there is a C++ method in "MZTolerance.cpp" which has the same implementation.
Skyline always (as far as I know) interprets tolerances using an inclusive range so that if someone sets the tolerance to zero there will be one number which would still fall in the tolerance range.

pwiz_tools/Skyline/TestFunctional/FullScanGraphTest.cs
```diff
@@ -298,6 +298,39 @@ private void TestSpecialIonsAnnotations()
...
+        {
+            RunUI(() => {
+                SkylineWindow.ShowBIons(true);
+                SkylineWindow.ShowCIons(true);
+            });
+
+            SetZoom(false);
+            ClickChromatogram(33.11, 15.055, PaneKey.PRODUCTS);
+            WaitForGraphs();
+            Assert.AreEqual(Skyline.Program.SkylineOffscreen? 70 : 20, SkylineWindow.GraphFullScan.IonLabels.Count());
```
brendanx67, line 314: Why are these numbers different on-screen vs. off-screen? This at least needs a comment explaining the issue.

pwiz_tools/Skyline/TestFunctional/FullScanGraphTest.cs
```diff
@@ -298,6 +298,39 @@ private void TestSpecialIonsAnnotations()
             //check that the special ion annotation shows in both library and full scan viewers
             TestAnnotations(new [] {"Reporter_Test+"});
             TestLibraryMatchAnnotations(new[] { "Reporter_Test+" });
+            Settings.Default.ShowSpecialIons = false;
+        }
+
+        private void TestIonMatchToleranceUnitSetting()
```
brendanx67, line 304: To test this more thoroughly will require using the TransitionSettingsUI form and also a case of using PPM in the Import Peptide Search wizard to exercise the TransitionSettingsControl.

brendanx67 (changes requested):
I mostly applied myself the necessary changes I saw. However, it does not look as well tested as I would like to see, and there is one case were a label count is different on-screen vs. off-screen which should be fixed or explained.

pwiz_tools/Skyline/TestFunctional/FullScanGraphTest.cs
rita-gwen: The labels are not drawn in off-screen mode because it happens during Paint event. In the off-screen mode this tests total number of peaks matching the ion type filter. I'll add a comment in the code.

pwiz_tools/Skyline/TestFunctional/FullScanGraphTest.cs
rita-gwen: I replaced the changing the settings  with invocation of the Settings dialogue and added tolerance unit verification to the ImportPeptideSearchTest.

pwiz_tools/Skyline/TestFunctional/FullScanGraphTest.cs
```diff
@@ -334,7 +338,7 @@ private void TestIonMatchToleranceUnitSetting()
             });
             WaitForGraphs();
             var graphLabels = SkylineWindow.GraphFullScan.IonLabels;
-            Assert.AreEqual(Skyline.Program.SkylineOffscreen ? 49 : 1, graphLabels.Count());
+            Assert.AreEqual(Skyline.Program.SkylineOffscreen ? 48 : 1, graphLabels.Count());
```
rita-gwen, line 341: This number has changed because calling the settings dialogue resets the special ion list and there was one reporter ion in this spectrum.

pwiz_tools/Skyline/TestFunctional/FullScanGraphTest.cs
brendanx67: Another approach can be to implement a method that gets the ion labels that would be added if the graph were painted. But, looking for a good example, I just ran into another case where we simply look at the CurveList for TextObj items. And I can see some benefit to testing the text that actually ends up on the graph. So, I can live with this.

pwiz_tools/Skyline/TestFunctional/ImportPeptideSearchTest.cs
```diff
@@ -621,6 +622,8 @@ private void TestIrts()
                 Assert.IsTrue(importPeptideSearchDlg.CurrentPage == ImportPeptideSearchDlg.Pages.match_modifications_page);
                 Assert.IsTrue(importPeptideSearchDlg.ClickNextButton());
                 Assert.IsTrue(importPeptideSearchDlg.CurrentPage == ImportPeptideSearchDlg.Pages.transition_settings_page);
+                importPeptideSearchDlg.TransitionSettingsControl.IonMatchToleranceUnits = MzTolerance.Units.ppm;
```
brendanx67, line 625: Test the prior value gets multiplied by 1000.

pwiz_tools/Skyline/TestFunctional/FullScanGraphTest.cs
```diff
@@ -311,17 +312,20 @@ private void TestIonMatchToleranceUnitSetting()
...
+
             RunUI(() =>
             {
-                var settings = SkylineWindow.DocumentUI.Settings;
-                var newLibs = settings.TransitionSettings.Libraries.ChangeIonMatchMzTolerance(new MzTolerance(10.0, MzTolerance.Units.ppm));
-                var newSettings = settings.ChangeTransitionSettings(settings.TransitionSettings.ChangeLibraries(newLibs));
-                SkylineWindow.ModifyDocument("Set test settings",
-                    doc => doc.ChangeSettings(newSettings));
+                transitionSettingsUI.IonMatchToleranceUnits = MzTolerance.Units.ppm;
+                transitionSettingsUI.IonMatchTolerance = 10.0;
```
brendanx67, line 324: Test the prior value gets multiplied by 1000 before changing it.

brendanx67 (approved):
Super close. Apply minor changes and then you can merge.

PR #2158, opened 2022-07-12
Serialize protein association settings and protein groups

chambm:
- Move the ParsimonySettings from PeptideFilter to PeptideSettings as Brendan suggested, so that PeptideFilter continues to represent only things that are configurable in the settings UI.
- Add serialization for ParsimonySettings (as protein_association)
- Add protein_group type to schema
- Add subclass of FastaSequence, FastaSequenceGroup, which acts like a FastaSequence most of the time but serializes all of its proteins in XML (keeping just a single peptide_list though)


pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
```diff
@@ -190,7 +193,8 @@ public void UseBackgroundProteome(BackgroundProteome backgroundProteome, Func<Fa
             return proteinAssociations;
         }
 
-        public class ParsimonySettings
+        [XmlRoot("protein_association")]
+        public class ParsimonySettings : IXmlSerializable
         {
             [Track(ignoreDefaultParent:true)]
             public bool GroupProteins { get; set; }
```
brendanx67, line 200: Ack! This is an immutable object. It cannot have public setters and it must use Change<property>() methods if it allows any direct changes to the object. Otherwise, it should have a constructor that takes its value, and a validation routine that is called during construction and XML serialization, if the values are at all limited in any way, like MinPeptidesPerProtein being at least non-negative.

brendanx67 (changes requested):
This looks pretty good, but the fact that no tests and no audit logs changed indicates that this is undertested.

Also noticed that your implementation of an immutable object is not immutable. That needs to get fixed.

chambm:
The audit logs didn't change (which surprised me) because the settings were already being stored in the audit log, and their representation didn't change from my refactoring here. That's up to AssociateProteinsSettings.CreateEntry(). Can you can check the audit log format so you can say whether or not it's not good the way it is now?

I did (minimally) change the unit test for serialization to make sure protein_group works. And I did some manual testing and checking of the document as well. Is there a test pattern for saving and reloading a document after it's been run through a test just to check the serialization works?

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
```diff
@@ -190,7 +193,8 @@ public void UseBackgroundProteome(BackgroundProteome backgroundProteome, Func<Fa
             return proteinAssociations;
         }
 
-        public class ParsimonySettings
+        [XmlRoot("protein_association")]
+        public class ParsimonySettings : IXmlSerializable
```
nickshulman, line 197: As Brendan mentioned, this class needs to be Immutable.
Also, I would recommend that you add a member:
public static readonly ParsimonySettings EMPTY = ...;
that you use everywhere, instead of allowing PeptideSettings.ProteinAssociationSettings to ever be null.

When you are writing out the XML, you can, in PeptideSettings.WriteXml, check "if (!ParsimonySettings.EMPTY.Equals(peptideSettings.ProteinAssociationSettings))" when you are deciding whether any XML needs to be written to the .sky file.

nickshulman (commented):
You are probably going to need to add something to "SrmSettings.RemoveUnsupportedFeatures" which removes the parsimony settings from the PeptideSettings, since we don't want that element written out when saving to an older version of Skyline. You should try doing "File > Share" and saving as version 21.2 and make sure that Skyline 21.2 can read a file which used to have protein groups in it.

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
chambm: Partially done. I assume private set is ok?

chambm:
Just committed the ProteinGroupMetadata changes. That part warrants a re-review @nickshulman @brendanx67 . I'm touching a lot of code I'm not familiar with.

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
```diff
@@ -0,0 +1,216 @@
...
+            {
+                return new ProteinGroupMetadata(String.IsNullOrEmpty(Name) ? sourceGroup.Name : Name,
+                    String.IsNullOrEmpty(Description) ? sourceGroup.Description : Description,
+                    sourceGroup.ProteinMetadata);
+            }
+
+            return new ProteinGroupMetadata(String.IsNullOrEmpty(Name) ? source.Name : Name,
+                String.IsNullOrEmpty(Description) ? source.Description : Description,
+                ProteinMetadata);
+            //throw new InvalidOperationException("cannot merge ProteinMetadata into ProteinGroupMetadata");
```
brendanx67, line 154: Seems unnecessary.

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
```diff
@@ -0,0 +1,216 @@
...
+        /// returns a copy of this, with any blanks filled in
+        /// using the members of source
+        /// </summary>
+        /// <param name="source">the source to merge from</param>
+        /// <returns>a copy of this merged with source, with this winning when fields conflict</returns>
+        public override ProteinMetadata Merge(ProteinMetadata source)
+        {
+            if (source == null)
+                return this;
+            if (source is ProteinGroupMetadata sourceGroup)
```
brendanx67, line 144: A comment explaining would be good. Because the functions Name and Description are virtual, these really appear the same except for how ProteinMetadata gets handled. So you might make that clearer by having just one "return new ProteinGroupMetadata()" call and the if determining only what ProteinMetadata gets used with an explanation why the sourceGroup overrides this when it has ProteinMetadata. Seems obvious whey this.ProteinMetadata gets used when source has none.

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
```diff
@@ -0,0 +1,216 @@
...
+
+        private ProteinGroupMetadata() : base(null, null)
+        {
+        }
+
+        private ProteinGroupMetadata(ProteinGroupMetadata other, WebSearchInfo webSearchInfo = null) : base(other.Name, other.Description)
+        {
+            webSearchInfo ??= other.ProteinMetadata.First().WebSearchInfo;
+            ProteinMetadata = ImmutableList<ProteinMetadata>.ValueOf(other.ProteinMetadata);
+            //ProteinMetadata[0] = ProteinMetadata[0].ChangeWebSearchInfo(webSearchInfo);
```
brendanx67, line 165: Necessary to keep the commented code?

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
```diff
@@ -0,0 +1,216 @@
...
+            /*    Name = ProteinMetadata.All(p => p.Name == null)
+                ? null
+                : string.Join(GROUP_SEPARATOR, ProteinMetadata.OrderBy(p => p.).Select(p => p.Name));*/
+        }
+
+        public ProteinGroupMetadata(string name, string description, IList<ProteinMetadata> proteinMetadata) : base(name, description)
```
brendanx67, line 171: Prefer to see the constructors closer to the top just after privates and statics.

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
```diff
@@ -0,0 +1,216 @@
...
+        {
+            ProteinMetadata = ImmutableList<ProteinMetadata>.ValueOf(proteinMetadata);
+        }
+
+        public new object GetDefaultObject(ObjectInfo<object> info)
+        {
+            return EMPTY;
+        }
+
+        public bool Equals(ProteinGroupMetadata other)
```
brendanx67, line 181: How does this thing get serialized? I am surprised not to see serialization code in here.

pwiz_tools/Skyline/Model/Proteome/ProteinMetadataManager.cs
```diff
@@ -200,35 +200,45 @@ private SrmDocument LookupProteinMetadata(SrmDocument docOrig, IProgressMonitor
...
                                     {
-                                        proteinMetadata = proteinMetadata.Merge(parsedProteinMetaData);  // Fill in any gaps with parsed info
+                                        // We note the sequence length because it's useful in disambiguating search results
+                                        proteinsToSearch.Add(new ProteinSearchInfo(new DbProteinName(null, proteinMetadata), node.PeptideGroup.Sequence == null ? 0 : node.PeptideGroup.Sequence.Length));
+                                        docNodesWithUnresolvedProteinMetadata.Add(proteinsToSearch.Last(), node);
                                     }
                                 }
-                                if (proteinMetadata != null)
+
+                                if (node.ProteinMetadata is ProteinGroupMetadata proteinGroupMetadata)
```
brendanx67, line 231: Not a fan of the heavy use of "is" to get to the list of metadata objects. Use a virtual instead and just have the single object return itself. You have derivation and virtuals.

pwiz_tools/Skyline/Model/Proteome/ProteinMetadataManager.cs
```diff
@@ -237,7 +247,20 @@ private SrmDocument LookupProteinMetadata(SrmDocument docOrig, IProgressMonitor
                             foreach (var result in FastaImporter.DoWebserviceLookup(proteinsToSearch, progressMonitor, false)) // Resolve them all, now
                             {
                                 Assume.IsTrue(!result.GetProteinMetadata().NeedsSearch());
-                                _processedNodes.Add(docNodesWithUnresolvedProteinMetadata[result].Id.GlobalIndex, result.GetProteinMetadata());
+                                var node = docNodesWithUnresolvedProteinMetadata[result];
+                                var resultMetadata = result.GetProteinMetadata();
+                                if (node.ProteinMetadata is ProteinGroupMetadata proteinGroupMetadata)
```
brendanx67, line 252: Not so easy to see how this can be written better, but wish it didn't require so much "is" and "as" type checking and casting.

pwiz_tools/Skyline/Test/SrmSettingsTest.cs
```diff
@@ -1140,6 +1141,24 @@ TransitionIonMobilityFiltering CheckIonMobilitySettingsBackwardCompatibility(str
...
+                                                        "remove_subset_proteins=\"true\" " +
+                                                        "shared_peptides=\"AssignedToBestProtein\" />";
+            AssertEx.DeserializeNoError<ProteinAssociation.ParsimonySettings>(proteinAssociationSerialized, DocumentFormat.VERSION_22_13);
+
+            var parsimonySettings = AssertEx.Deserialize<ProteinAssociation.ParsimonySettings>(proteinAssociationSerialized);
+            Assert.AreEqual(3, parsimonySettings.MinPeptidesPerProtein);
+            Assert.AreEqual(true, parsimonySettings.GroupProteins);
+            Assert.AreEqual(true, parsimonySettings.FindMinimalProteinList);
+            Assert.AreEqual(true, parsimonySettings.RemoveSubsetProteins);
+            Assert.AreEqual(ProteinAssociation.SharedPeptides.AssignedToBestProtein, parsimonySettings.SharedPeptides);
```
brendanx67, line 1159: I would add a bit more here. Does the class do any validation at all? If so, you should check DeserializeError() for out of bounds value, and maybe at least test min_peptides_per_protein="0".

brendanx67 (changes requested):
Mostly looks good. A few minor suggestions.

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
chambm: The commented out exception?

pwiz_tools/Skyline/Test/SrmSettingsTest.cs
chambm: No validation yet but I'll add it. Why 0 specifically?

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
chambm: I might not understand the question. ProteinGroupMetadata is serialized in the same place as ProteinMetadata, in DocumentReader/Writer. Quite a few changes there to handle the subclass.

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        public ProteinMetadata ChangeSingleProteinMetadata(ProteinMetadata singleProteinMetadata)
+        {
+            Assume.IsTrue(singleProteinMetadata != null && !(singleProteinMetadata is ProteinGroupMetadata));
+            var proteinMetadataList = new List<ProteinMetadata>(ProteinMetadata);
+            var matchingProteinIndex = proteinMetadataList.IndexOf(m => m.Name == singleProteinMetadata?.Name);
+            Assume.IsTrue(matchingProteinIndex >= 0, $@"no matching protein in group {this} for protein {singleProteinMetadata}");
+            proteinMetadataList[matchingProteinIndex] = singleProteinMetadata;
+            return new ProteinGroupMetadata(Name, Description, proteinMetadataList);
```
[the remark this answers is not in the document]
brendanx67: Good point. You should probably move much of this code to the ImmutableList implementation in a ChangeElementAt<T>(T item, index i) function. You could also add ChangeElement<T>(T item, Func<bool, T> isElement)

Anyway, enhancing ImmutableList would be better than putting the code here.

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
brendanx67: Right. Didn't realize this wouldn't be clearly attached to the line I added the comment on.

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
brendanx67: Oh, right. The difference between classes inside SrmSettings and those associated with targets.

pwiz_tools/Skyline/Test/SrmSettingsTest.cs
brendanx67: Often testers will test the acceptable edges of allowed values. Zero is the lowest this should be allowed to go, but off-by-ones are common and if zero became unallowed that would be an important bug. A professional tester should be able to tick off: 1) allowed values, 2) edge case allowed values, 3) error cases.

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
```diff
@@ -0,0 +1,216 @@
...
+        public override string Species => ProteinMetadata.All(p => p.Species == null)
+            ? null
+            : string.Join(GROUP_SEPARATOR,
+                ProteinMetadata.Where(p => !string.IsNullOrWhiteSpace(p.Species)).Select(p => p.Species));
+        [Track]
+        public override WebSearchInfo WebSearchInfo => ProteinMetadata.Select(p => p.WebSearchInfo).First();
+
+        public override ProteinMetadata ChangeName(string name)
+        {
+            return new ProteinGroupMetadata(this);
```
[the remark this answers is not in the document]
chambm: I think I put these in as placeholders planning to come back to it. Should non-protein-group-aware Skyline code be able to change protein groups' metadata properties? Or should they always be calculated from the constituent protein metadata (e.g. Protein1/Protein2/Protein3)?

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
brendanx67: You could use Assume.Fail() to make it clearer (and then return null - which will never happen but ReSharper doesn't know that). Programming error type messages which are expected to end up as an unhandled exception and reported to us through the unhandled exception form do not need to be localized. So, you can just make them constant strings.

brendanx67:
Do you understand why TestAuditLogLocalization is failing? It seems the existing code has decided that suddenly ProteinMetadata.WebSearch is a tracked property that might end up in the audit log. Though, I can't see why.

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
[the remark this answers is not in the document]
brendanx67: I think the latter. Certainly would be strange for a user to change WebSearch on ProteinMetadata of any kind, since that is a hidden field. This maybe explains why the code fails TestAuditLogLocalization complaining about WebSearch not being in PropertyNames.resx

chambm:
> Do you understand why TestAuditLogLocalization is failing? It seems the existing code has decided that suddenly ProteinMetadata.WebSearch is a tracked property that might end up in the audit log. Though, I can't see why.

It was an errant track attribute in the subclass' copy-pasta that nobody noticed. Including me, until I commented out the derived property (going back to a non-virtual in the base class to see if that was the issue) and got a compile error.

pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
```diff
@@ -2051,6 +2051,12 @@ public SrmSettings RemoveUnsupportedFeatures(DocumentFormat documentFormat)
                             Array.Empty<string>())));
             }
 
+            if (documentFormat < DocumentFormat.VERSION_22_13)
```
bspratt, line 2054: I'd rather see something more descriptive, in the style of DocumentFormat.FLAT_CROSSLINKS instead of a hardcoded version number

pwiz_tools/Skyline/Model/PeptideGroup.cs
```diff
@@ -332,11 +332,12 @@ public bool Equals(FastaSequence obj)
         {
             if (ReferenceEquals(null, obj)) return false;
             if (ReferenceEquals(this, obj)) return true;
-            return Equals(obj._name, _name) &&
-                Equals(obj._description, _description) &&
-                Equals(obj._sequence, _sequence) &&
-                ArrayUtil.EqualsDeep(obj.Alternatives, Alternatives) &&
-                obj.IsDecoy == IsDecoy;
+            var equal = Equals(obj._name, _name) &&
```
bspratt, line 335: if this is just a formatting change, it should be backed out (along with any similar changes)

pwiz_tools/Shared/ProteomeDb/API/ProteinMetadata.cs
```diff
@@ -39,72 +39,90 @@ namespace pwiz.ProteomeDatabase.API
...
+        public virtual WebSearchInfo WebSearchInfo { get; private set; }
 
-        public ProteinMetadata ChangeName(string name)
+        /// <summary>
+        /// For a single protein, returns an ImmutableList with just that protein's ProteinMetadata.
+        /// For a protein group, returns all of the group's proteins' ProteinMetadata
+        /// </summary>
+        public virtual ImmutableList<ProteinMetadata> ProteinMetadataList => ImmutableList<ProteinMetadata>.Singleton(this);
+        
+        public virtual ProteinMetadata ChangeSingleProteinMetadata(ProteinMetadata other)
```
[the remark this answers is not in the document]
chambm: You're right. This needs to set the properties on a new object.

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
```diff
@@ -0,0 +1,219 @@
...
+
+        public override ImmutableList<ProteinMetadata> ProteinMetadataList { get; }
+
+        [Track]
+        public override string Name => ProteinMetadataList.All(p => p.Name == null)
+            ? null
+            : string.Join(GROUP_SEPARATOR, ProteinMetadataList.Select(p => p.Name));
+
+        [Track]
+        public override string Description => ProteinMetadataList.All(p => p.Description == null)
```
bspratt, line 61: These seem like they ought to include "none" or somesuch instead of just collapsing over nulls. Hard to tell what goes with what otherwise, I'd think

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
```diff
@@ -0,0 +1,219 @@
...
+
+        /// <summary>
+        /// The Change* functions do nothing for a protein group. Protein groups' properties are always computed from their constituents' properties.
+        /// </summary>
+        public override ProteinMetadata ChangeName(string name)
+        {
+            return new ProteinGroupMetadata(this);
+        }
+
+        public override ProteinMetadata ChangeDescription(string descr)
```
bspratt, line 100: Piling on, maybe, but yes these seem incomplete.

pwiz_tools/Skyline/Model/Serialization/DocumentWriter.cs
```diff
@@ -114,16 +118,16 @@ private void WritePeptideGroupXml(XmlWriter writer, PeptideGroupDocNode node)
             {
                 writer.WriteAttributeString(ATTR.label_description, node.ProteinMetadataOverrides.Description);
             }
-            WriteProteinMetadataXML(writer, node.ProteinMetadataOverrides, true); // write the protein metadata, skipping the name and description we already wrote
+            if (!(node.PeptideGroup is FastaSequenceGroup) || SkylineVersion.SrmDocumentVersion < DocumentFormat.VERSION_22_13)
```
bspratt, line 121: Again I'd rather not see a hardcoded version number here

bspratt (commented):
I don't completely understand what's going on with protein group metadata - seems like only name and description matter? If so it seems odd that it inherits from the richer protein metadata class that also tracks gene, organism, etc

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
chambm: Yeah, good point.

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
```diff
@@ -0,0 +1,219 @@
...
+            : string.Join(GROUP_SEPARATOR, ProteinMetadataList.Select(p => p.Name));
+
+        [Track]
+        public override string Description => ProteinMetadataList.All(p => p.Description == null)
+            ? null
+            : string.Join(GROUP_SEPARATOR,
+                ProteinMetadataList.Where(p => !string.IsNullOrWhiteSpace(p.Description)).Select(p => p.Description));
+
+        [Track]
+        public override string PreferredName => ProteinMetadataList.All(p => p.PreferredName == null)
```
bspratt, line 67: Should probably apply the same test to see if there are any items of interest as we do for individual items - 
ll(p => string.IsNullOrWhitespace(p.PreferredName)
rather than
ll(p => p.PreferredName == null)

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
```diff
@@ -0,0 +1,219 @@
...
+        {
+            webSearchInfo ??= other.ProteinMetadataList.First().WebSearchInfo;
+            ProteinMetadataList = ImmutableList<ProteinMetadata>.ValueOf(other.ProteinMetadataList);
+            //ProteinMetadataList[0] = ProteinMetadataList[0].ChangeWebSearchInfo(webSearchInfo);
+            /*    Name = ProteinMetadataList.All(p => p.Name == null)
+                ? null
+                : string.Join(GROUP_SEPARATOR, ProteinMetadataList.OrderBy(p => p.).Select(p => p.Name));*/
+        }
+
+        public ProteinGroupMetadata(string name, string description, IList<ProteinMetadata> proteinMetadata) : base(name, description)
```
bspratt, line 48: I still don't understand why this class has a ctor that takes name and description when it doesn't take all the others. I get that "all the others" are aggregates from the proteins in the group, but so are Name and Description if I'm reading the code correctly. Seems like the ctor name and description are unrecoverable?

chambm: Yeah, that is pretty pointless. I'll make it just take the IList and pass null to the base class.

pwiz_tools/Skyline/Model/Proteome/ProteinGroupMetadata.cs
chambm: @brendanx67 @nickshulman What should I put in as the "(missing)" placeholder value in cases like this:
Protein1/Protein2/Protein3
Gene1/(missing)/Gene3 (it would be weird if it was Gene1/Gene3)

Something like a localized "None" or "Missing"?

pwiz_tools/Skyline/TestUtil/Schemas/Skyline_Current.xsd
```diff
@@ -845,6 +855,22 @@
       </xs:extension>
     </xs:complexContent>
   </xs:complexType>
+  <xs:complexType name="protein_group_type">
+    <xs:complexContent>
+      <xs:extension base="doc_parent_node_type">
+        <xs:sequence>
+          <xs:element name="protein" type="protein_type" minOccurs="1" maxOccurs="unbounded" />
```
labkey-jeckels, line 862: Brendan mentioned in a call yesterday that Skyline was going to continue writing out a <protein> element when a group only has a single member. If that's correct, should this be minOccurs="2"?

chambm: Yes, I think you're correct. Semantically, it should be 2.

brendanx67:
I think space around the slashes would be good. Monospaced for the sequence also would be interesting to see. The way these tips get shown is a two-step interaction, where first the node is asked how big the tip needs to be and then it is asked to paint the tip on a resulting window. It seems like this shouldn't be that hard to expand to support scrolling. Maybe returning (optionally) two sizes: tip window size, and content size. Then the tip could have scrollbars added based on the content size (if specified).

chambm:
I didn't intend for the tooltip/scrolling discussion to delay this PR (if that's what you mean by 'we have gone beyond my initial review'). Only the (missing) placeholder Brian mentioned should be addressed IMO.

Adding the spaces between slashes helped for moderate groups:
image

But large groups still look unwieldy:
image

I'm thinking the descriptions should be separated by linefeeds instead of slashes. And maybe for Species, if all the values are the same, collapse it down to that value (e.g. just "Homo sapiens" if all the proteins are in the group are Homo sapiens). Could also do that for Gene, although it's not as common that a group will have just one gene.

chambm:
@bspratt Is it normal to get multiple gene names for a single protein (e.g. aliases)?

bspratt:
My understanding is that while it's not common, multiple genes can indeed produce the same protein.

chambm:
No I mean aliases: https://www.genecards.org/cgi-bin/carddisp.pl?gene=DLD
image

chambm:
Multiple genes would not produce the same protein accession, but could produce two different accessions with the same protein sequence. That would not be uncommon between species or subspecies, but would be fairly uncommon within a species.

brendanx67:
You can also implement wrapping for some of the fields that appear very
wide right now. You can see that Description and sequence already have
wrapping implemented.

bspratt:
@chambm I guess I don't understand the question, sorry

chambm:
The value for "Gene" in the last screen capture has 4 different space separated aliases for the same gene. That's not my code doing that (I don't think). Is it supposed to be that way?

bspratt:
Yeah, that's normal: we ask

https://rest.uniprot.org/uniprotkb/stream?query=P09622&format=tsv&columns=id,genes,organism,length,entry%20name,protein%20names,reviewed

and they respond with

Entry	Entry Name	Reviewed	Protein names	Gene Names	Organism	Length
P09622	DLDH_HUMAN	reviewed	Dihydrolipoyl dehydrogenase, mitochondrial, EC 1.8.1.4 (Dihydrolipoamide dehydrogenase) (Glycine cleavage system L protein)	DLD GCSL LAD PHE3	Homo sapiens (Human)	509

chambm:
OK after fixing a bug with the first sequence length of the protein group always being used for web lookups (leading to lots of unknowns), and adding wrapping for all fields, I get this:
image

PR #2159, opened 2022-07-13
Fix EncyclopeDIA spectrum display

nickshulman:
Fixed bug where spectrum from incorrect file from EncyclopeDIA library would be displayed in Library Match window (reported by Deanna, thanks Nick)

Also, remove many Tuple's from EncyclopeDiaLibrary.cs and replace them with named types.


brendanx67 (commented):
Any way to add a test that fails with the old code and passes with the new to prove the point?

PR #2164, opened 2022-07-16
Fix TestMs1Tutorial

nickshulman:
Here's what I think the fix should be:
1. If showRedundantSpectra is true, then labelFilename should say "Fil&amp;e:"
2. If showRedundantSpectra is false, then labelFilename should say "File: " followed by the filename
3. We should never try to parse the text in labelFilename, and instead. the file name should be stored in "_sourceFile".


brendanx67 (approved):
Looks good now, I think.

PR #2165, opened 2022-07-19
Fixed case of CiRT peptides not covering RT range

kaipot:


brendanx67:
Kaipo, can you get this ready for merge again, and we can push it through to completion this time?

brendanx67:
I am working on this now and it is getting closer. Still not entirely sure it achieves the original intent of storing custom iRT standards with ranked transitions and relative abundance information in all cases.

pwiz_tools/Skyline/Model/Irt/IrtDb.cs
```diff
@@ -618,66 +619,108 @@ public static string GenerateDocumentXml(IEnumerable<Target> standards, SrmDocum
...
-                    addPeptides.Add(nodePep.Merge(nodePepOld));
-                    oldPeptides.Remove(nodePep.Target);
-                }
-                else
-                {
-                    addPeptides.Add(nodePep);
-                }
+                var addPep = nodePep;
+                if (addPeptides.TryGetValue(nodePep.ModifiedTarget, out var existing))
+                    addPep = addPep.Merge(existing);
```
nickshulman, line 626: This "Merge" call did not seem to be doing the correct thing at all.
I had a Skyline document which had multiple copies of the peptides that I was using for the iRT standard, and the peptides in the .irtdb ended up having duplicate transitions as a result of this "Merge" call.
In my pull request #3222 I made it just choose the first PeptideDocNode in the document instead of trying to merge multiple ones.

brendanx67: I think I have been adding tests. Keep adding tests that cover the cases you are concerned about, and fixes.

nickshulman: Actually, nothing bad happens in the current code as a result of this PeptideDocNode.Merge call.
The merged PeptideDocNode does have duplicate TransitionGroupDocNode's in it, but there is some code later on which only uses the first TransitionGroupDocNode.
If we wanted to allow iRT standards to have more than one charge state then we would need to fix this Merge call.
The reason that this Merge call is problematic is that Merge uses the "Equals" method on the TransitionGroup to decide if two transition groups are the same. However, because TransitionGroup includes the Peptide which includes the FastaSequence, so merging to PeptideDocNode's with different protein FastaSequence's does not really work right.
It might be that we somehow want to detach the Peptide from the FastaSequence before calling "Merge".

PR #2167, opened 2022-07-23
Issue 497: Ignore full scan settings when doing peak picking in SRM result files

nickshulman:
Fixed so Skyline ignores Transition Full Scan settings when doing peak picking in SRM result files (reported by Brendan)

This required adding a flag to ChromFileInfo and ChromCachedFile so that we can tell which files we extracted by ChromatogramDataProvider, and which came from SpectraChromDataProvider.

Added a unit test which makes sure that library retention times are never used for peak scoring in SRM runs regardless of the Transition Full Scan settings.


pwiz_tools/Skyline/Model/Results/ChromHeaderInfo.cs
```diff
@@ -1638,6 +1642,29 @@ public bool UsedMs2Centroids
             get { return UsedMs2CentroidsFlags(Flags); }
         }
 
+        public bool IsSrm
```
brendanx67, line 1645: I prefer having a single way of doing things to avoid confusion. Can you either make the other getters that call private static functions do their flag manipulation in-line or give this a parallel private static function? I am fine with either way. Maybe this way is better since it keeps others from wondering why the extra functions were defined, as I wonder, unless it is simply to be ready in case they need to be made public like IsSingleMatchMzFlags() and IonMobilityUnitsFromFlags().

brendanx67 (approved):
Mostly looks good. A few minor stylistic comments.

brendanx67:
Obviously need to fix the warnings in your new test to pass code inspection.

PR #2168, opened 2022-07-23
Undo/Redo from document grid

nickshulman:
Issue 897: Make Ctrl-Z (Undo) and Ctrl-Y (Redo) work in Document Grid
Issue 898: Record the Document Grid selection in the Undo/Redo buffer during Document Grid editing events

### DIFF
--- a/pwiz_tools/Skyline/Controls/Databinding/DataboundGridForm.cs
+++ b/pwiz_tools/Skyline/Controls/Databinding/DataboundGridForm.cs
@@ -18,9 +18,13 @@
  */
 
 using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
 using System.Windows.Forms;
 using pwiz.Common.DataBinding;
 using pwiz.Common.DataBinding.Controls;
+using pwiz.Common.SystemUtil;
 using pwiz.Skyline.Model.Databinding;
 using pwiz.Skyline.Util;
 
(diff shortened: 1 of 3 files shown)


brendanx67:
Probably would be good to hop on Zoom together and I could do manual testing with you, and you could just track the things I try that don't work as expected. In maybe 3-4 scenarios I have tried, I am getting about a 50% success rate. Here is what hasn't worked as I might have hoped:
- Paste several rows into Protein Note (for a document with multiple proteins), Ctrl-Z to undo - it just bings, despite having an undo record that can be undone with the Undo button on the Skyline Window.
- Select several rows with the row selectors, click the red X delete button, okay the deletion, Ctrl-Z to undo - this works but you don't get the selection back to what it was before the action, i.e. the rows come back but they are not selected.

nickshulman:
It would actually be fairly tricky to make the Document Grid restore the selection when an undo happens.
When the current document changes, values inside individual cells in the document grid are redrawn immediately to reflect the new Document. However, the number of rows and columns in the Document Grid never changes immediately on the event thread.

Whenever the document changes, a background thread figures out what new columns and rows need to be displayed. If the columns and the number of rows are exactly the same, then the DataGridView can be invalidated, and the new values are displayed without affecting the selection at all. But, if the columns or row count are different, then the DataGridView gets sent a ListChangedEvent.Reset, which moves the focus to the first displayed column, but  typically keeps the row number that you are on the same.

When the DataGridView has a lot of rows, it becomes dangerous to ask the DataGridView which rows are selected. The only way I have found to ask the DataGridView which rows are selected is with the "SelectedRows" property. If the number of selected rows happens to be a large number, then retrieving that property takes a really long time, since a DataGridViewRow object needs to be constructed for each row in the selection.

Potentially, we could make it so that when the Document Grid has fewer than, for instance, 1000 rows in it, then the selection is preserved when the Document changes, and once the Document Grid has more than 1000 rows, selection preservation stops, but that would be a testing nightmare.

brendanx67:
The most recent changes help. When I paste now, though, I end up with an undo record that leaves the selection where it was at the end of the paste and not before the paste. So, paste -> undo -> paste does not cause the same action to be repeated. Rather, the second paste will start at the last cell of the undone paste.

I guess I am okay with simply preserving the selected cell and not the entire selection, but it isn't quite there yet. Type -> enter -> undo -> type -> enter works correctly, which seems pretty important.

brendanx67:
It is definitely a big improvement already, even if it never reaches the ideal. Undoing a big Fill Down is pretty cool. Maybe you could make Ctrl-D (the Excel shortcut for Fill Down) work. Since you already have the ProcessCmdKey handler. :-)

brendanx67 (approved):
Looks good. Huge improvement over keyboard support before this.

PR #2169, opened 2022-07-23
Issue 888: File > Import > Peptide Search with existing library repor…

nickshulman:
…ts the build settings in the Audit Log

Fixed audit log entries when choosing to use existing library during Import Peptide Search (reported by Brendan)

If you choose to use an existing library file, an entry gets added after the first page indicating which library was chosen. Here's what the audit log might end up looking like:
image


brendanx67 (approved):
Looks good. Thanks.

PR #2176, opened 2022-07-26
Use classes from Google.Protobuf.WellKnownTypes instead of "OptionalDouble", "OptionalFloat"

nickshulman:


bspratt (approved):
Looks great, including tests. This will be very helpful for serializing classes that have nullable members.

PR #2178, opened 2022-07-26
Improved version of ValueCache which should use less memory

bspratt:
As of .Net 4.7.2, you can now recover the actual stored value from a HashSet  - they added 
    public bool TryGetValue (T equalValue, out T actualValue);
so we no longer need to use a dictionary whose values are the same as its keys to retrieve the originally stored value.

Internal issue only, no user report


nickshulman (approved):
Looks great.

PR #2179, opened 2022-07-29
When TestDocumentFormatSchemaFiles() finds that Skyline_Current.xsd has been changed but DocumentFormat.CURRENT has not…

bspratt:
…, just show the error message, rather than dumping the full text of the two xsd files being compared followed by the error message.

### DIFF
--- a/pwiz_tools/Skyline/Test/SchemaDocumentsTest.cs
+++ b/pwiz_tools/Skyline/Test/SchemaDocumentsTest.cs
@@ -73,10 +73,11 @@ namespace pwiz.SkylineTest
             var xsdCurrentResourceName = SchemaDocuments.GetSkylineSchemaResourceName(DocumentFormat.CURRENT.ToString()); // Special case, returns "Skyline_Current.xsd"
             var currentXsdContents = GetResourceText(xsdCurrentResourceName); 
             Assert.IsNotNull(currentXsdContents);
-            if (explicitCurrentXsdContents != null && !explicitCurrentXsdContents.Equals(currentXsdContents))
+            if (explicitCurrentXsdContents != null)
             {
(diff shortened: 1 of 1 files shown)


nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/Test/SchemaDocumentsTest.cs
```diff
@@ -73,11 +73,10 @@ public void TestDocumentFormatSchemaFiles()
             var xsdCurrentResourceName = SchemaDocuments.GetSkylineSchemaResourceName(DocumentFormat.CURRENT.ToString()); // Special case, returns "Skyline_Current.xsd"
             var currentXsdContents = GetResourceText(xsdCurrentResourceName); 
             Assert.IsNotNull(currentXsdContents);
-            if (explicitCurrentXsdContents != null)
+            if (explicitCurrentXsdContents != null && !explicitCurrentXsdContents.Equals(currentXsdContents))
             {
-                Assert.AreEqual(currentXsdContents, explicitCurrentXsdContents, 
-                    string.Format("Expected resource files {0} and {1} to be identical - did you change the contents of Skyline_Current.xsd without incrementing the value of DocumentFormat.CURRENT?",
-                        explicitCurrentResourceName, xsdCurrentResourceName));
+                AssertEx.Fail(string.Format("Expected resource files {0} and {1} to be identical - did you change the contents of Skyline_Current.xsd without incrementing the value of DocumentFormat.CURRENT?",
```
brendanx67, line 78: We could also use AssertEx.NoDiff() which outputs the first lines that are different, and may offer more information, but still not blast the entire file contents into the log.

brendanx67 (approved):
You could do it this way. Sure. You could also use AssertEx.NoDiff() to include the first differing line that might remind the user what has changed.

pwiz_tools/Skyline/Test/SchemaDocumentsTest.cs
bspratt: Yeah, that would be better - will do.

PR #2180, opened 2022-07-29
Prevent blank graph when Legend is too big

nickshulman:
If the size of the Legend does not leave any room to draw the graph, then draw the graph and the legend on top of each other.

It looks like this:
image
It's not perfect, but I think the user has a better chance of figuring out what is going on than when they just get a blank graph.

The reason the graph is blank is because of the following lines in GraphPane.cs:
```
			// do a sanity check on the ChartRect
			if ( _chart._rect.Width < 1 || _chart._rect.Height < 1 )
				return;

```
That code has been there forever, and it prevents any part of the graph being drawn if the chart area is empty. Often there are other things outside of the chart area, such as the axes and Legend, which could be drawn even though the user would not be able to see any data on the chart itself.


pwiz_tools/Shared/zedgraph/ZedGraph/Legend.cs
```diff
@@ -1020,8 +1021,15 @@ private PaneList GetPaneList( PaneBase pane )
 				}
 			}
 
-			_rect = newRect;
-		}
+            if (tChartRect.IsEmpty && !originalChartRect.IsEmpty)
```
brendanx67, line 1024: Indenting looks wierd. Maybe a mixture of tabs and spaces?

brendanx67 (approved):
Something's up with your indenting, but I definitely approve of changing the behavior. There has been a lot of confusion around why graph panes go completely blank.

PR #2183, opened 2022-08-02
* fixed AssociateProteinsDlg's OK button never being enabled

chambm:
* fixed OK button never being enabled when AssociateProteinsDlg's PeptideCount == 0
* fixed TestImportPeptideSearch not waiting for OK button to be enabled

I don't know if this will fix the actual intermittent failure but it definitely fixes a usability bug when the FASTA mapping results in no peptide matches.


brendanx67 (approved):
Okay for this release, since we are feature complete, but I think we should show an error here telling the user that the result will not include any targets. It does not make sense to continue this wizard through importing results and failing on all of the chosen files in the case where there are no targets. This is just letting the user shoot themselves in the foot. They would be much better off being able to OK the error and consider how they might change the settings to include some targets.

PR #2189, opened 2022-08-04
Added a static code inspection for uses of "Form" where we really should use "FormEx"…

bspratt:
…. This currently throws two inspection errors, so this should not be merged to master until those are corrected:

Found prohibited use of
"class .*\:.* Form(?!Ex)\s*(,|$)"
(use FormEx instead of Form - this ensures proper interaction with automated tests, small molecule interface operation, and other enhancements. If this really is a legitimate use add this comment to the offending line: '// Purposely using Form here instead of FormEx') at
C:\Dev\master_clean\pwiz_tools\Skyline\EditUI\EditLinkedPeptidesDlg.cs(36)
    public partial class EditLinkedPeptidesDlg : Form

Found prohibited use of
"class .*\:.* Form(?!Ex)\s*(,|$)"
(use FormEx instead of Form - this ensures proper interaction with automated tests, small molecule interface operation, and other enhancements. If this really is a legitimate use add this comment to the offending line: '// Purposely using Form here instead of FormEx') at
C:\Dev\master_clean\pwiz_tools\Skyline\EditUI\SynchronizedIntegrationDlg.cs(34)
    public partial class SynchronizedIntegrationDlg : Form

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -387,6 +387,8 @@ Reader*Test.data
 /pwiz_aux/msrc/utility/vendor_api/ABI/LicenseKey.h
 /pwiz_aux/msrc/utility/vendor_api/Agilent/x64/
 /pwiz_aux/msrc/utility/vendor_api/Agilent/x86/
+/pwiz_aux/msrc/utility/vendor_api/Bruker/BAF2SQL-README.txt
+/pwiz_aux/msrc/utility/vendor_api/Bruker/BAF2SQL-THIRD-PARTY-LICENSE-README.txt
 /pwiz_aux/msrc/utility/vendor_api/Bruker/x64/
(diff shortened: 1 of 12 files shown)


nickshulman:
I wonder whether it would be better to write this as an ordinary unit test which uses "Assembly.GetTypes" to find all of the classes in Skyline.exe and pwiz.Common.dll that extend "Form" and make sure that the ones in Skyline.exe inherit from either "FormEx" or "DockableFormEx" and the form classes in Common inherit from "CommonFormEx".

bspratt:
That would be more robust in the face of things like a class declaration split across lines, e.g.

public class foo \:
Form

bspratt:
OK, using reflection now.  Also promoted failure to inherit from AbstractUnitTest from warning to error.

PR #2190, opened 2022-08-04
Skyline Batch: Download template file, open remote folder, template file bug fix

rayhankhanna:


spallanck:
Downloading a template file changes:
- If a user tries to open a template file before the configuration is run, attempt to download the associated panorama file (using the new downloadDlg.cs which uses FileSaver.cs) if there is one and then open the valid template file in Skyline. 
- If there is no associated panorama file, tell the user the template file has not been downloaded.
- The downloadDlg.cs uses async and await for the download process

Open remote folder changes:
- Added a new button that shows available remote folders and allows a user to click on one to open

brendanx67:
Did you also remove the minimize box? It is okay to leave the close box and even the control box when the form is sizable, since the control box will include Close, Move, and Size.

spallanck:
I added back in the control box so the close box is still available, but removed the maximize and minimize options.

PR #2192, opened 2022-08-08
Extending precursor sense to include things like CE, CCS, instrument configuration etc

bspratt:
This allows for ion mobility "multiple conformers", awareness of CE values in spectral libraries, handling mass spec data with multiple fragmentation modes etc


brendanx67:
Multiple conformers are really a special case, since most other things I can think of are about deciding whether or not to extract a point from a specific spectrum, where today we would frequently extract from too many spectra and get spikey peaks. Because we so often use 3D spectra for IMS, the multiple conformer case requires applying a filter after the decision of whether the 3D spectrum should be considered a candidate for extraction. Minor point, but in other cases I can think of it is really about deciding whether a spectrum should be considered for extraction where there is a certain place in the code for making that decision.

This doesn't mean we can go back to 2D spectra for IMS, since that would cause a major slowdown where many 2D spectra are now batched into a single 3D spectrum.

pwiz_tools/Skyline/Model/TransitionGroupDocNode.cs
```diff
@@ -50,14 +50,16 @@ public class TransitionGroupDocNode : DocNodeParent
...
+        /// <param name="explicitPrecursorFilter">Optional values like explicit ion mobility, CE etc</param>
+        /// <param name="libraryPrecursorFilter">Library (imsdb or blib) ion mobility value, if known</param>
+        public TransitionGroupDocNode(TransitionGroup id, TransitionDocNode[] children, PrecursorFilter explicitPrecursorFilter = null, PrecursorFilter libraryPrecursorFilter = null)
             : this(id,
                    Annotations.EMPTY,
                    null,
                    null,
                    null,
-                   explicitTransitionGroupValues ?? ExplicitTransitionGroupValues.EMPTY,
+                   explicitPrecursorFilter,
```
nickshulman, line 61: There should probably be a class which has two properties: "ExplIcitPrecursorFilter" and "LibraryPrecursorFilter" and that one object is what gets passed to this constructor. Anything that we can do to reduce the number of parameters that this constructor takes is a good thing. I also have a feeling that there might be other things that we would want to add to this new class when it exists.

pwiz_tools/Skyline/Model/TransitionGroupDocNode.cs
```diff
@@ -137,12 +159,28 @@ protected override IList<DocNode> OrderedChildren(IList<DocNode> children)
         /// Deep copy for use in small molecule user editing - we have to deep copy entire branches for changes to IDs
         /// </summary>
         /// <returns> Copy of transition group and its transitions, all with new IDs and backpointers</returns>
-        public TransitionGroupDocNode UpdateSmallMoleculeTransitionGroup(Peptide parentNew, TransitionGroup groupNew, SrmSettings settings)
+        public TransitionGroupDocNode DeepCopyTransitionGroup(Peptide parentNew, TransitionGroup groupNew,
```
nickshulman, line 162: There is already a method "ChangeTransitionGroupId" which I think does the same thing as this.

nickshulman (commented):
This looks exciting!

I think it's a mistake to add "PrecursorFilter" to "LibraryKey" (and LibKeyIndex). In general, we will always need to get a list of spectra that match the peptide sequence, and then do a linear search on that small list because the rules for what constitutes a "match" will depend on exactly what we are looking at. Sometimes we will want the spectrum with a particular CE, sometimes with the closest CE, and sometimes the highest scoring spectrum with a CE which falls in a particular range.

PR #2194, opened 2022-08-09
Fixed exception on iRT standards changed

kaipot:


pwiz_tools/Skyline/Skyline.cs
```diff
@@ -2285,11 +2285,12 @@ public void HandleStandardsChanged(ICollection<Target> oldStandard, RCalcIrt cal
                     }
                 }
             }
-            if (newDoc != null)
+
+            var standardPepGroup = newDoc?.PeptideGroups.FirstOrDefault(nodePepGroup => new IdentityPath(nodePepGroup.Id).Equals(firstAdded));
```
nickshulman, line 2289: Usually, if you have an IdentityPath, you would use "FindNode" to find it.
(but I guess you also have to check whether the identity path is null):
`            var standardPepGroup = firstAdded == null ? null : (PeptideGroupDocNode) newDoc?.FindNode(firstAdded);
`
FindNode is better than the linear search because it's able to take advantage of the Dictionary in "DocNodeChildren".

nickshulman (approved):

PR #2196, opened 2022-08-10
Fix ga4 hit geolocation

chambm:
OK, so after fiddling with Puppeteer for a while, looking at what it was sending, and doing a bit more research about what the cryptically named query parameters meant, I decided that maybe we can just fudge it and make direct requests to google-analytics. It seems to be working. The fact it uses a POST method but actually sends the parameters in the URL is odd, but that's the way it does it in the browser too. When useDebugUrl is true, the page_view hit can be seen in [DebugView](https://analytics.google.com/analytics/web/#/m/p318139389/debugview) and when it's false you can see it in [Realtime view](https://analytics.google.com/analytics/web/#/p318139389/realtime/overview) (and this is also the only way you can see the geolocation is working; DebugView doesn't do the geolocation apparently).

The one thing I'm not happy about here is we aren't really testing that the hit is showing up in analytics, just that the status code comes back as 204 (No Content). It gives that status code even if I mess up the parameters in a way that the hit won't be shown. So I'm a bit concerned that if they change gtag.js in the future such that the parameter names change, our analytics might stop working. @bconn-proteinms do you think we can somehow monitor analytics DebugView to make sure we're getting hits there every day (during nightly tests)?

### DIFF
--- a/pwiz_tools/Skyline/Program.cs
+++ b/pwiz_tools/Skyline/Program.cs
@@ -30,7 +30,7 @@ using System.Text;
 using System.Threading;
 using System.Windows.Forms;
 using Microsoft.Win32;
-using Newtonsoft.Json;
+using pwiz.Common.Collections;
 using pwiz.Common.Controls;
 using pwiz.Common.SystemUtil;
 using pwiz.Skyline.Alerts;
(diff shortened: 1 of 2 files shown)


bconn-proteinms:
@chambm you asked "do you think we can somehow monitor analytics DebugView to make sure we're getting hits there every day (during nightly tests)?"

It is possible to use something like DebugView, but we will need a different page for this as the documentation says you need to include the debug_mode parameter (ie `gtag(.... { ..., 'debug_mode':true });`). Also if you were going to do this, as part of the nightly tests you will need to add an additional parameter to the tag to distinguish the test "event" from an "event" produced by any other application startup .

chambm:
> @chambm you asked "do you think we can somehow monitor analytics DebugView to make sure we're getting hits there every day (during nightly tests)?"
> 
> It is possible to use something like DebugView, but we will need a different page for this as the documentation says you need to include the debug_mode parameter (ie `gtag(.... { ..., 'debug_mode':true });`). Also if you were going to do this, as part of the nightly tests you will need to add an additional parameter to the tag to distinguish the test "event" from an "event" produced by any other application startup .

Gtag.js sets _dbg=1 when the query actually gets sent, so that's what I'm already doing. I have manually confirmed that the hit does show up in the DebugView. I'm wondering how we can automate testing that it's showing up in DebugView though.

chambm:
Do you want to review/test this Brendan or are we good to go?

chambm:
/squash
* changed GA4 analytics hit to use page_view type instead of measurement protocol in order to preserve geographic information

PR #2203, opened 2022-08-11
Speed up EditPepModsDlg by using LiteDropDownList

nickshulman:
Change font and padding in LiteDropDownList to look more like a combo box.
image

Reduced time to display Edit Modifications dialog.
image


brendanx67 (approved):

bspratt:
Looks good. Still one "TODO" in there, I note.

brendanx67:
Just a thought, but maybe we should make these fast combo boxes delay load their values by allowing them to accept an interface for getting the values that gets called when they are needed, since we are now retrofitting several places in our code to load the value list when it is shown to the user.

nickshulman:
> Just a thought, but maybe we should make these fast combo boxes delay load their values by allowing them to accept an interface for getting the values that gets called when they are needed, since we are now retrofitting several places in our code to load the value list when it is shown to the user.

Good idea. 
We have been trying to make it so that this class has properties and events which closely match the original ComboBox control, so that you can usually start using this new control just by changing the classes in the .designer.cs file. That means the way that we would accomplish delay loading of items is by adding a "DropDown" event which fires before the drop down menu is shown. As soon as we have a scenario that wants to use that, I am sure that it will be straightforward to add that event.

brendanx67:
So, you think in this case it is irrelevant that we are producing a lot of identical lists of modifications for all of the combo boxes? Probably you are right if they are now just in memory lists. Just saw your image with the very long list of heavy modifications, and likely a set of blank combo boxes to start, and thought "Eewww!"

PR #2204, opened 2022-08-15
* added support for parallelmode to SkylineTester

chambm:
* added support for parallelmode to SkylineTester including building chambm/always_up_runner for first time users and killing Docker workers when pressing "Stop"

* added simple password hiding to KeyValueGridDlg (if key contains "password" so not really localized)
* moved some TestRunner parallelization functions to TestRunnerLib so they can be used in SkylineTester
* added CloseMainWindow() call for ProcessUtilites.Kill() to allow managed process to clean up gracefully (this was intended to let TestRunner clean up its own Docker workers but catching the "close window" event did not seem to work very well)


brendanx67 (approved):
Looks okay and it worked for me. Definitely makes it more usable.

chambm:
Wow that did not do what it was supposed to do.

PR #2212, opened 2022-08-17
Restore and improve my previous attempt at finding forms that aren't …

bspratt:
…using FormEx, this time avoiding assemblies that aren't needed (and which can cause trouble due to brittle load order issues etc).

Also change the "WARNING: {test name} does not derive from AbstractUnitTest!" message in TestRunner to say "ERROR" = even though it doesn't actually show as an error in our test stats, it should be harder to ignore.


bspratt:
Passes perf tests on TC

PR #2214, opened 2022-08-17
Remove UI references from Model namespaces

nickshulman:
I was thinking of adding an interface "IUserInterface" which could be passed to model code which does things, and which potentially needs to display information to users, and ask them questions.
Then, there is the "TextUserInterface" implementation which just writes messages to an output stream, and if a user gets asked a question, it assumes that the user gave the default response.
And there is "GraphicalUserInterface" implemented in Skyline which displays message boxes using AlertDlg, etc.

I am not sure whether this is a good approach in terms of removing UI code from SkylineDataSchema and SkylineViewContext.


bspratt:
This seems like its headed in the right direction. I wonder if we could minimize code changes by requesting an interface from the document container at point of use, rather than passing it in?

brendanx67:
Typically, I prefer polymorphism where the model simply supplies a message
it wants to show the user, and the interface decides how to show it. Often
we achieve this by throwing an exception out of the model and letting the
caller decide how to show the message in the exception. Also, we should
never be using MessageBox because our test framework can't handle it. Any
use of MessageBox implies error handling code that is never tested.

bspratt:
>> Any use of MessageBox implies error handling code that is never tested.

We do have a code inspection test for that, FWIW

bspratt:
I think you may have forgotten to add something to git:

Z:\pwiz\pwiz_tools\Skyline\Alerts\GraphicalUserInterface.cs(10,43): error CS0246: The type or namespace name 'AbstractGui' could not be found (are you missing a using directive or an assembly reference?) [Z:\pwiz\pwiz_tools\Skyline\Skyline.csproj]

pwiz_tools/Shared/Common/UserInterfaces/IUserInterface.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Threading;
+using System.Windows.Forms;
+using JetBrains.Annotations;
+using pwiz.Common.SystemUtil;
+
+namespace pwiz.Common.UserInterfaces
+{
+    public interface IUserInterface
+    {
+        void DisplayMessage(Control owner, string message);
```
brendanx67, line 11: You need to get rid of the Windows.Forms use in the interface at least. That means no Control and now MessageBoxButtons in the interface. I would expect the owner control to be passed in at construction time of the object. It just knows the owner control and correctly displays the message box when asked to show a message.

DisplayMessageWithButtons() would need to become DisplayMessageWithOptions() and the Windows.Forms implementation would need to translate between the options and buttons, while the text version would ignore them? That worries me a little, since they seem likely to be questions expecting an answer.

brendanx67 (changes requested):
Just one brief explanation of a direction to take this, as we discussed over Zoom.

nickshulman:
Replaced by PR#2344

PR #2215, opened 2022-08-17
FAIMS was not working for MS2 data

bspratt:
SureQuant + FAIMS seems to have been the first seen use of MS2 data with FAIMS, and we weren't handling it properly.

Also noticed a bug in Share Minimized which would leave .imsdb file empty.

### DIFF
--- a/pwiz_tools/Skyline/Model/Results/SpectrumFilterPair.cs
+++ b/pwiz_tools/Skyline/Model/Results/SpectrumFilterPair.cs
@@ -166,6 +166,12 @@ namespace pwiz.Skyline.Model.Results
         private ExtractedSpectrum FilterSpectrumList(MsDataSpectrum[] spectra,
             SpectrumProductFilter[] productFilters, bool highAcc, bool useIonMobilityHighEnergyOffset)
         {
+
+            if (HasIonMobilityFAIMS() && spectra.All(s => !Equals(IonMobilityInfo.IonMobility, s.IonMobility)))
+            {
+                return null; // No compensation voltage match
+            }
+
             int targetCount = 1;
             if (Q1 == 0)
                 highAcc = false;    // No mass error for all-ions extraction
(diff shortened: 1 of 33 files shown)


pwiz_tools/Skyline/Model/Results/SpectrumFilterPair.cs
```diff
@@ -626,6 +632,11 @@ public override int GetHashCode()
             }
         }
 
+        public override string ToString() // For debug convenience
+        {
+            return $@"mz={TargetMz} w={FilterWidth} id={FilterId} heo={HighEnergyIonMobilityValueOffset}";
```
nickshulman, line 637: I wonder whether we should only include "heo" if it's not null.

bspratt: Doesn't seem worth the bother for debug code

nickshulman (commented):
Looks good.
I wonder whether part of the problem with getting an IonMobilityLibrary when you are minimizing is that "PersistMinimized" converts the LibKeyMap to a dictionary, and so is looking for entries with an exact match. There might be entries in the .imsdb which are being used by the document, but they are not a perfect match because the number of digits of precision in a modification mass is different. It might also be necessary to remove the "ToDictionary" call on line 185 of IonMobilityLibrary.cs and do something else.

bspratt:
> not a perfect match because the number of digits of precision in a modification mass is different

That actually was the issue, now it uses a LibraryKey so the comparison should be tolerant, yes?

nickshulman:
> That actually was the issue, now it uses a LibraryKey so the comparison should be tolerant, yes?

Unfortunately, no. The fuzzy comparison happens in "LibKeyIndex.ItemsMatching".
If you put all the entries into a Dictionary, then they get looked up using LibraryKey.Equals, which is an exact match.
If you want to find the entries that are being used by the document, you should call "_database.DictLibrary.KeyPairsMatching" for each of the TransitionGroupDocNodes. I will make this change and push it to this pull request.

bspratt:
Excellent, thanks!

nickshulman (approved):
Looks good to me.

chambm:
NB: automatic cherry pick won't work if there are any of those "Merge branch 'master' into PR" commits. You have to slash-squash or slash-rebase first. (I can't remember whether putting that in the middle of a comment will trigger it or not). But it's too late after it's been merged. You'll have to CP this one manually.

bspratt:
No worries, thanks for the heads up.

PR #2218, opened 2022-08-18
Added fix for exporting Waters methods containing compounds with multiple modifications

kaipot:


pwiz_tools/Skyline/Model/Export.cs
```diff
@@ -4125,6 +4125,33 @@ protected override void WriteHeaders(TextWriter writer)
             }
             writer.WriteLine();
         }
+
+        // Hack to replace modification brackets with parentheses.
```
brendanx67, line 4129: Make this a block comment.

pwiz_tools/Skyline/Model/Export.cs
```diff
@@ -4125,6 +4125,33 @@ protected override void WriteHeaders(TextWriter writer)
...
+
+        // Hack to replace modification brackets with parentheses.
+        // MassLynx or VerifyESkylineLibrary.dll has some problem with multiple sets of brackets that causes
+        // an incomplete method to be generated.
+        public static string FormatMods(string modSeq)
+        {
+            if (!modSeq.Contains('['))
+                return modSeq;
+
+            var newSeq = new StringBuilder(modSeq.Length);
```
brendanx67, line 4137: I wonder if this is really necessary. It seems like replacing this entire function with:

return modeSeq.Replace('[', '(').Replace(']', ')');

Would be just fine given all the other work that is happening. I can't imagine the double allocation ever showing up in a profiler as important.

brendanx67 (changes requested):
Thanks for fixing this so quickly.

pwiz_tools/Skyline/Model/Export.cs
```diff
@@ -4125,6 +4125,15 @@ protected override void WriteHeaders(TextWriter writer)
             }
             writer.WriteLine();
         }
+
+        /* Hack to replace modification brackets with parentheses.
```
brendanx67, line 4129: Sorry, I meant our standard C# function header comment style:
    /// <summary>
    /// text
    /// </summary>

brendanx67 (approved):

chambm:
OK I think this is good for cherry picking.

PR #2220, opened 2022-08-19
Add support for reading ProxlXML files from Byonic in BiblioSpec

chambm:
- added support for reading ProxlXML files from Byonic (converted from mzIdentML) in BiblioSpec


kaipot (approved):
Looks good to me.

PR #2227, opened 2022-08-19
Skyline/work/20220713 share replicate files

ClarkUPS:
Added the ability to include raw files when sharing a skyline document. Users can now select which files to include in the zip file. Should files be misplaced users can now navigate though their file system and either select the individual missing files to include or select a folder that contains missing files. All discovered missing files are added to the selection where users can decide weather or not to include them in the zip.


pwiz_tools/Skyline/Alerts/ShareResultsFilesDlg.cs
```diff
@@ -0,0 +1,452 @@
...
+
+        /// <summary>
+        ///  Close form after updating files
+        /// </summary>
+        public void OkDialog()
+        {
+            // Add auxiliary file data
+            UpdateAuxFiles();
+            this.DialogResult = DialogResult.OK;
+            this.Close();
```
nickshulman, line 161: Setting the DialogResult of a modal dialog causes it to clause. You probably should not call "Close" here.

pwiz_tools/Skyline/Alerts/ShareResultsFilesDlg.cs
```diff
@@ -0,0 +1,452 @@
...
+            {
+                _checkBoxFiles = checkedFiles;
+                _missingCheckBoxFiles = missingFiles;
+            }
+        }
+
+        /// <summary>
+        /// Checked list box information
+        /// </summary>
+        public class CheckedListBoxItems
```
nickshulman, line 440: I think this class should not be plural and should be named "CheckedListBoxItem".

pwiz_tools/Skyline/FileUI/OpenDataSourceDialog.cs
```diff
@@ -241,8 +245,49 @@ public string SourceTypeName
...
+            bool validInclude = false;
+            var subFiles = Directory.GetFiles(dirInfo.FullName, "*.*", SearchOption.AllDirectories);
+            var subDirectories = Directory.GetDirectories(dirInfo.FullName, "*.*", SearchOption.AllDirectories);
+
+
+            // Convert file paths to file names
+            var subFileNames = subFiles.Select(Path.GetFileName).ToList();
+            var subDirectoryNames = subDirectories.Select(Path.GetFileName).ToList();
+
+            foreach (var missingFiles in _specificFileFilter)
```
nickshulman, line 260: This code needs to check whether "_specificFileFilter" is null.

nickshulman (commented):
This is nice looking code.
I get a NullReferenceException whenever I try to import results into Skyline because "_specificFileFilter" in OpenDataSourceDialog is null.

The specific file filter feature looks interesting. When doing an Import Peptide Search there is a part where the user is asked to find the raw files that go with the peptide search results. This happens in "ImportResultsControl.browseToResultsFileButton_Click". Then, after the user has chosen their raw files, everything that does not match what they were supposed to be looking for gets thrown away. It would be better if import peptide search was using this new feature.
The Import Peptide Search case is a little different than what you have, because in that case, only the filename without the extension is known, and the user is allowed to choose any .raw, .mzML etc. which matches the base name.

bspratt:
Thanks for the review. Clark is headed back to school so I'll handle the last bits of polishing on this PR.

bspratt:
OK, I have this pretty well buttoned up now. It would be great if you guys can find some time to try it out.

nickshulman:
If I try to include some Bruker .d folders in my .sky.zip, I get the error "The zip file is not a shared file".
I think we need to make SrmDocumentSharing.FindSharedSkylineFile tolerant, so it allows folders whose name ends in ".d".
Also, I think we should probably disable the "Include results files" checkbox if the target Skyline file format is less than or equal to 22.2.

bspratt:
Yes, your first point rather leads to the second. Thanks!

brendanx67:
Remember that Waters uses .raw folders. We have a class that checks if a
folder is a mass spec “file”.

bspratt:
> Remember that Waters uses .raw folders. We have a class that checks if a folder is a mass spec “file”.

Note that what's being inspected here isn't a directory on disk but rather the contents of a .zip file, so the class you mention isn't directly applicable here, but I expect I can abstract it to deal with both cases.

brendanx67:
Moved these changes to a new pull request.

PR #2236, opened 2022-08-22
Fixed path separator in built BiblioSpec libraries

kaipot:


chambm:
I was surprised this didn't trigger needing to record new reference checks for everything, but I see we always trim the path to the BiblioSpec root. Otherwise tests would fail between Linux/Windows or just a different source tree location. That actually brings up the point that in rare cases, a BiblioSpec library may have been created on Linux, so its path will use forward slashes with `/` as the root. I suppose that's not a common enough case to worry about though, eh @brendanx67 ?

chambm (approved):

brendanx67:
I don’t have a problem with that case, nor where the user used full network
paths on Windows and has paths like:

\\server\share\path\to\file.mgf

I was only objecting to the mixed path separators obviously created by
BlibBuild and not something a user would otherwise ever expect to see.

PR #2239, opened 2022-08-23
Button to sort Report Editor alphabetically

nickshulman:
Add button to toolbar at top of Report Editor to toggle whether columns are sorted alphabetically.

Here's what it looks like without the button pressed (there are a lot of columns because the UI mode is "Mixed"):
image

And here is what it looks like after pressing the button to sort alphabetically. The expanded state of the tree and the selected column are maintained.
image


brendanx67 (approved):
Looks good to me.

PR #2245, opened 2022-08-24
Load Library Match data in background

kaipot:
@nickshulman This change is to fix the issue you mentioned in an email last week (freezing due to blib queries happening on the UI thread). Brendan mentioned that we should also not populate the toolbar comboboxes until the user clicks/focuses them but I wanted to get your thoughts on this change first.

For some reason a lot of the indentation was weird in GraphSpectrum.cs so I fixed those too.


nickshulman (commented):
It might not have been necessary to cache the result of "RetentionTimesPsmCount()".
That method gets called by "BiblioSpecLitLibrary.GetRedundantSpectra", but, really all it wants to know is whether the RetentionTimes table exists, so the performance would be almost as good if there was a separate method "RetentionTimesTableExists()" which GetRedundantSpectra could call instead of the slow method "RetentionTimesPsmCount()".

I wonder whether you should get rid of the "Timer" in UpdateManager. Now that you are querying for spectra on a background thread, is there anything that is happening on the foreground thread which is so time consuming you need to wait before doing it?

kaipot:
> I wonder whether you should get rid of the "Timer" in UpdateManager. Now that you are querying for spectra on a background thread, is there anything that is happening on the foreground thread which is so time consuming you need to wait before doing it?

That's a good point. I'll try removing the timer and see how it goes.

kaipot:
> I wonder whether you should get rid of the "Timer" in UpdateManager. Now that you are querying for spectra on a background thread, is there anything that is happening on the foreground thread which is so time consuming you need to wait before doing it?

@nickshulman I think the timer is still useful. Drawing the graph is slow enough to be noticeable (the original reason for the timer was selecting the combobox and holding the down arrow, so that it cycles through spectra very quickly).

kaipot:
@nickshulman I updated the PR based on your suggestion to use the known cached retention times, rather than doing anything on a background thread. The spectra only get loaded when needed.

I'm not sure what's wrong with the code inspection check.

kaipot:
scrapping this in favor of #2367

PR #2248, opened 2022-08-25
- added support for Bruker TSF format

chambm:
- added support for Bruker TSF format
- fixed timsTOF instrument type not being recorded in instrumentConfiguration
* updated PSI-MS CV to 4.1.99


chambm:
@brendanx67 What do you think about including this in 22.2?

PR #2251, opened 2022-08-29
Skyline: Attempt to fix DLL locking caused by adding TestInfo.IsAuditLogTest

brendanx67:


brendanx67:
It turned out the reflection used to determine which tests do audit log recording was not actually deferred until it was needed. Instead, in the constructor for TestInfo, a new instance of each test class was created and that instance was created:

var testObj = Activator.CreateInstance(TestClassType);

It seems that this could cause longer-lasting DLL locking than other reflection operations we do. This PR attempts to keep this code from being called unless the test type in the Tests tab is set to "Audit log tests".

PR #2260, opened 2022-09-06
Old KillChildren() behavior and TestRunner logging fixes

chambm:
* restored old KillChildren() behavior (combining it with the new behavior)
* fixed some logging output inconsistencies in TestRunner parallel mode


brendanx67:
I don’t think this is the problem. I am still working on the true problem
which seems to be two sets of commands starting at once and interfering
with each other.

brendanx67:
Not yet, but there was certainly a lot of accumulated cruft and poorly
tested, buggy code to contribute to the problem.

This last issue is some kind of race condition in SkylineTester I don’t yet
understand and which doesn’t happen every time. It also is not really
possible to debug after the fact.

I am now emitting stack traces into the log to try to understand, but it
doesn’t look like that helped. So, I need to think of something else…

chambm:
Won't hurt to merge this back in though, right?

brendanx67:
Let's not while I am still working on the problem. After that, I will have
a closer look.

chambm:
> The second call fails immediately trying to run Git because the pwiz folder already exists, after that, it now correctly waits for 1 minute, which is not enough for the first call to finish its Git processing. So, the second call starts deleting files while the first Git clone is still running, then fails and waits again. When the first call's Git ends it fails because the repository is not valid after all of the deletions, and then SkylineTester crashes.

Wouldn't the old KillChildren haev avoided the symptoms of this race condition by killing the git process from the first execution (presumably that aborts the test run?) so that the second execution can go ahead and delete the files and start the checkout again?

pwiz_tools/Skyline/TestRunner/Program.cs
```diff
@@ -543,7 +550,7 @@ private static int ListenToTestQueue(StreamWriter log, CommandLineArgs commandLi
 
                 while (!cts.IsCancellationRequested)
                 {
-                    Console.WriteLine("Waiting for message");
+                    //TeeLog("Waiting for message");
```
brendanx67, line 553: Is this getting commented out for a reason? Maybe add a comment explaining why?

pwiz_tools/Skyline/TestRunner/Program.cs
```diff
@@ -634,11 +641,16 @@ private static string LaunchDockerWorker(int i, CommandLineArgs commandLineArgs,
             var pwizRoot = Path.GetDirectoryName(Path.GetDirectoryName(GetSkylineDirectory().FullName));
             string workerName = bigWorker ? $"docker_big_worker_{i}" : $"docker_worker_{i}";
             string dockerRunRedirect = string.Empty;
+            string testRunnerLog = @$"c:\AlwaysUpCLT\TestRunner-{workerName}.log";
             if (commandLineArgs.ArgAsBool("keepworkerlogs"))
-                dockerRunRedirect = $"> c:\\pwiz\\TestRunner-{workerName}-docker.log 2>1";
+            {
+                //dockerRunRedirect = $"> c:\\pwiz\\TestRunner-{workerName}-docker.log 2>1";
```
brendanx67, line 647: Doesn't seem to add anything leaving this commented out.

brendanx67 (approved):
Looks good. I did decide when I wrote the original code that it was preferable to kill any of the named processes that might be running on the system when a SkylineNightly run starts than to continue seeing issues we were seeing at the time. It seemed unlikely really that these commands should be running on the system at that time anyway. I never heard anyone complain in the many years it worked this way.

PR #2265, opened 2022-09-08
Discard unused scan id's when minimizing chromatograms.

nickshulman:
Fixed so that when minimizing results, spectrum identifiers which are no longer referenced by any extracted chromatogram are discarded, resulting in a smaller .skyd file.

The one thing that concerns me about this fix is that it makes it so that when you minimize a .skyd file, the scan identifiers in the MsDataFileScanIds are reordered to be the order in which they were found in ChromGroupHeaderInfo's, instead of what they started at, which was the order in which they were encountered during chromatogram extraction.


brendanx67:
I guess I don't understand why the order would need to change, if you just store the original order and then sort before writing. In my experience, it is always possible to preserve an original order, if you want to.

nickshulman:
I don't think this fix is worth doing since it's a weird scenario where it matters and it is a little weird that the order of the scanid's in the file have to change unless we want to make two passes through the .skyd file.

PR #2267, opened 2022-09-08
Skyline: Fix peptide losses with multiple charge states

brendanx67:
for the same formula / m/z
- Fix charged losses in peptides to preserve peptide adducts
- Added testing to cover the fixed cases


pwiz_tools/Skyline/Model/TransitionGroup.cs
```diff
@@ -726,14 +726,20 @@ protected override IList<ExplicitLoss> CreateState(ExplicitLoss[] mods)
...
                 TransitionLosses firstLosses = null;
                 List<TransitionLosses> allLosses = null;
-                HashSet<double> allLossMasses = null;
+                HashSet<Tuple<double, int>> allLossMassCharges = null;
                 foreach (var losses in potentialLosses)
                 {
-                    double lossMass = CalcTransitionLossesMass(type, cleavageOffset, massType, losses);
+                    double lossMass = CalcTransitionLossesMass(type, cleavageOffset, massType, losses, out var lossCharge);
+                    if (type != IonType.precursor)
+                        lossCharge = 0; // Only precursors care about the loss charge
```
nickshulman, line 734: I don't understand this comment. Fragments do sort of care about the loss charge in that the fragment charge has to be less than or equal to the precursor's charge minus the loss charge.

pwiz_tools/Skyline/Model/TransitionGroup.cs
```diff
@@ -726,14 +726,20 @@ protected override IList<ExplicitLoss> CreateState(ExplicitLoss[] mods)
                 // there should be only one loss
                 TransitionLosses firstLosses = null;
                 List<TransitionLosses> allLosses = null;
-                HashSet<double> allLossMasses = null;
+                HashSet<Tuple<double, int>> allLossMassCharges = null;
                 foreach (var losses in potentialLosses)
                 {
-                    double lossMass = CalcTransitionLossesMass(type, cleavageOffset, massType, losses);
+                    double lossMass = CalcTransitionLossesMass(type, cleavageOffset, massType, losses, out var lossCharge);
```
nickshulman, line 732: Would it be possible to get rid of this function call, and just call "CalcTransitionLosses" on line 744?
Then you would have a TransitionLosses object and you could take advantage of its ".Equals" implementation, and put it in the hashset, instead of having a HashSet&lt;Tuple&lt;double, int>>.

Then, you could also make "allLosses" into a "HashSet&lt;TransitionLosses>" and you would not need "allLossMassCharges".

pwiz_tools/Skyline/Model/DocSettings/Loss.cs
```diff
@@ -363,7 +363,8 @@ public bool Equals(TransitionLosses other)
         {
             if (ReferenceEquals(null, other)) return false;
             if (ReferenceEquals(this, other)) return true;
-            return other.Mass == Mass;
+            return other.Mass == Mass &&
```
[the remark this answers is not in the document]
brendanx67: I am going to note that I am not changing this comparison, and if you want to address this issue, it should be done systematically in a separate PR, not to mention the exploring and addressing possibility that ReSharper generates Equals() implementations like this, possibly adding something to our own static analysis to catch these, etc.

pwiz_tools/Skyline/Model/TransitionGroup.cs
brendanx67: Until I added this the case of interest was adding multiple equivalent losses to y- and b-ions.

e.g. H20, H20[M-1], H20[M-2] on a triply charged precursor could all lead to:

y7-16+ (neutral loss from the y7+)
y7-16+ (-1 loss from the y7++)
y7-16+ (-2 loss from the y7+++)

Without this condition all 3 got added to the list of possible transitions as if they were different transitions. The case of whether a loss can be applied without reducing the charge to zero is handled elsewhere.

brendanx67: I mean -18 of course for water loss.

pwiz_tools/Skyline/Model/TransitionGroup.cs
brendanx67: Unfortunately, we can't rely on the loss equality for the reason I explained above. Losses with unequal charges can end up colliding with differently charged losses in fragments where the fragmentation event causes a charge loss from the precursor that was not related to the loss.

brendanx67: I improved the comment about why this only changes for precursor ions. It was a surprise to me too, and another confirmation of the importance of writing tests.

pwiz_tools/Skyline/Model/TransitionGroup.cs
```diff
@@ -726,14 +726,25 @@ protected override IList<ExplicitLoss> CreateState(ExplicitLoss[] mods)
...
+                    // ions can lose charge during fragmentation no matter what the charge of the loss is,
+                    // and all charges down to charge 1 are allowed for fragments. So, it really is only
+                    // the mass of the loss that matters for fragment ions.
+                    if (type != IonType.precursor)
+                        lossCharge = 0;
                     if (lossMass == 0 ||
-                            (firstLosses != null && firstLosses.Mass == lossMass) ||
-                            (allLossMasses != null && allLossMasses.Contains(lossMass)))
+                        (firstLosses != null && firstLosses.Mass == lossMass &&
+                         firstLosses.TotalCharge == lossCharge) ||
```
nickshulman, line 742: This "firstLosses.TotalCharge == lossCharge" looks suspicious.

This function is behaving correctly, but I don't think it's doing so for the reasons that you intended.
The only thing that is necessary is that, for fragment ions, if there are multiple losses which have the same mass, this function needs to make sure that the loss with the lowest charge is included in the list that gets returned.

It would make more sense if the logic was:
Does the current loss have the same mass as a loss that we are already planning on returning? If so, compare the losses. If it's a precursor, then keep the current loss if its charge is different than all the ones we already have. If it's not a precursor, then keep the loss with the lowest charge state.

The way the code here is written, there is this other variable "lossCharge" which gets set to zero, but that zero never gets added into the hashSet allLossMassCharges.

pwiz_tools/Skyline/Model/TransitionGroup.cs
```diff
@@ -743,15 +754,15 @@ protected override IList<ExplicitLoss> CreateState(ExplicitLoss[] mods)
...
+                            allLossMassCharges.Add(new Tuple<double, int>(firstLosses.Mass, firstLosses.TotalCharge));
                             firstLosses = null;
                         }
                     }
                     if (allLosses != null)
                         allLosses.Add(tranLosses);
-                    if (allLossMasses != null)
-                        allLossMasses.Add(tranLosses.Mass);
+                    if (allLossMassCharges != null)
+                        allLossMassCharges.Add(new Tuple<double, int>(tranLosses.Mass, tranLosses.TotalCharge));
```
nickshulman, line 765: If your goal is to prevent this function from returning losses which have the same mass, then it would make more sense if you used "lossCharge" instead of "tranLosses.TotalCharge" (but, then, the behavior of this function would be incorrect).

brendanx67: This is a very good point, Nick. I cleaned up the code to use a new struct called LossId to make the intention clearer and more correct. And I added a new test with just charged losses which failed before, because it turned out that the old code relied on the neutral loss being present two exclude the other two loss states.

brendanx67:
Note that it looks like the code is trying very hard to avoid allocating a lot of lists in the simple single loss case, which is highly likely, and I assume resulted from profiling massive DIA target lists where losses were present.

pwiz_tools/Skyline/Model/TransitionGroup.cs
brendanx67: I see what you mean about the lowest charge. Reversing the order of the losses in the list breaks the test. Okay. Back to work...

brendanx67: Okay, added changes to use lowest charge, and added a test the used to fail because of this issue. Hooray for testing!

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/Model/TransitionGroup.cs
```diff
@@ -767,6 +780,52 @@ protected override IList<ExplicitLoss> CreateState(ExplicitLoss[] mods)
...
+            // The loss of the charge is only interesting to preserve separate from the mass in
+            // the case of the precursor, because it has a fixed starting charge. Whereas, fragment
+            // ions can lose charge during fragmentation no matter what the charge of the loss is,
+            // and all charges down to charge 1 are allowed for fragments. So, it really is only
+            // the mass of the loss that matters for fragment ions.
+
+            public bool Equals(LossId other)
+            {
+                return Mass.Equals(other.Mass) && 
+                       (IonType != IonType.precursor || Charge == other.Charge);
```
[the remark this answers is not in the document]
brendanx67: Okay, put more effort into the struct, which definitely grew from what I originally expected.

pwiz_tools/Skyline/Model/TransitionGroup.cs
```diff
@@ -726,14 +726,27 @@ protected override IList<ExplicitLoss> CreateState(ExplicitLoss[] mods)
                 // there should be only one loss
                 TransitionLosses firstLosses = null;
                 List<TransitionLosses> allLosses = null;
-                HashSet<double> allLossMasses = null;
+                HashSet<LossId> allLossMassCharges = null;
```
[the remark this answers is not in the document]
brendanx67: Almost went that direction on the last iteration, but backed off to something less invasive. Gave it a try and it does yield nicer code. Hope none of this impacts the performance case of File > Import > Assay Library with millions of lines, which is how it ended up as complicated as it is.

PR #2274, opened 2022-09-09
Prevent duplicate entries in imsdb for crosslinked peptides

nickshulman:
Fixed bug where ion mobility library entries for crosslinked peptides would be duplicated each time you pressed the "Use Results" button (reported by Juan)

### DIFF
--- a/pwiz_tools/Skyline/TestUtil/TestFunctional.cs
+++ b/pwiz_tools/Skyline/TestUtil/TestFunctional.cs
@@ -263,7 +263,7 @@ namespace pwiz.SkylineTestUtil
             }
         }
 
-        protected static void RunDlg<TDlg>(Action show, [InstantHandle] Action<TDlg> act = null, bool pause = false, int millis = -1) where TDlg : Form
+        protected static void RunDlg<TDlg>([InstantHandle] Action show, [InstantHandle] Action<TDlg> act = null, bool pause = false, int millis = -1) where TDlg : Form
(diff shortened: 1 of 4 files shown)


pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -263,7 +263,7 @@ private static void SkylineBeginInvoke(Action act)
             }
         }
 
-        protected static void RunDlg<TDlg>(Action show, [InstantHandle] Action<TDlg> act = null, bool pause = false, int millis = -1) where TDlg : Form
+        protected static void RunDlg<TDlg>([InstantHandle] Action show, [InstantHandle] Action<TDlg> act = null, bool pause = false, int millis = -1) where TDlg : Form
```
bspratt, line 266: Just curious - what's this about?

bspratt (approved):
Looks good. Thanks for jumping on this.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
nickshulman: It lets ReSharper know that the Action "show" is not going to be invoked after RunDlg returns.
Without it, ReSharper would have given me the "access to modified closure" warning because the "transitionSettingsUi" variable that I was using inside of my lambda got modified later on in the DoTest method.

PR #2281, opened 2022-09-13
Fix for a problem with Agilent IMS data where fringe CCS and/or DT values cannot roundtrip through the CCS/DT conversion…

bspratt:
… (throws exception "Cannot solve cubic fit"). When we hit these, just stick with any measure DT value since the CCS conversion isn't useful.

Reported by user Alan

### DIFF
--- a/pwiz_aux/msrc/utility/vendor_api/Agilent/MidacData.cpp
+++ b/pwiz_aux/msrc/utility/vendor_api/Agilent/MidacData.cpp
@@ -29,6 +29,8 @@
 
 
 #pragma managed
+#using <System.dll> // Provides System::IO::InvalidDataException
+
 namespace pwiz {
 namespace vendor_api {
 namespace Agilent {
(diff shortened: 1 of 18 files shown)


chambm (commented):
A new perftest to test fringe values seems a bit extreme (unless it's very fast to run?). Was there no practical way to just test those values standalone?

bspratt:
> A new perftest to test fringe values seems a bit extreme (unless it's very fast to run?). Was there no practical way to just test those values standalone?

Fair question.

It takes 50 seconds exclusive of the time to download the data set. It's only in perf because of the data download issue, but the data is needed to reproduce the problem.  Possibly I could write some custom code to get that runtime down but for something that runs just once in a test suite it didn't seem worth the bother. I could be convinced otherwise, I suppose.

chambm:
I'm not aware of a way to subset Agilent MIDAC data files. Can you get the fringe value failures from the other MIDAC perf data sets? (I.e. explicitly testing the problematic values on other datasets where those values might not occur in any of the spectra)

bspratt:
> I'm not aware of a way to subset Agilent MIDAC data files. Can you get the fringe value failures from the other MIDAC perf data sets? (I.e. explicitly testing the problematic values on other datasets where those values might not occur in any of the spectra)

Conceivably, but it just doesn't seem like a good use of my time - but possibly I am underestimating the evilness of what I've done here?

chambm:
How big is the new zip file?

bspratt:
> How big is the new zip file?

image

chambm (approved):
OK, no big deal then.

PR #2282, opened 2022-09-13
Add method ion match tolerance unit instrument setting

rita-gwen:


pwiz_tools/Skyline/Model/DocSettings/TransitionSettings.cs
```diff
@@ -1922,6 +1930,32 @@ public bool IsMeasurable(double mz, double precursorMz)
         [Track]
         public double MzMatchTolerance { get; private set; }
```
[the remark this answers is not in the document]
brendanx67: Does the audit logging impact this? I guess it doesn't need to be seen as two properties in the audit log.

MzMatchTolerance change from 0.05 m/z to 20 ppm

Would actually be preferable to:

MzMatchTolerance changed from 0.05 to 20
MzMatchToleranceUnits changed from m/z to ppm

How are these changes getting logged with this code?

pwiz_tools/Skyline/Model/TransitionCalc.cs
```diff
@@ -60,8 +61,7 @@ public static class TransitionCalc
                     double delta = mz - calculatedMz;
                     double deltaAbs = Math.Abs(delta);
                     int potentialShift = (int) Math.Round(deltaAbs);
-                    double fractionalDelta = deltaAbs - potentialShift;
-                    if (MatchMz(fractionalDelta, tolerance) && MatchMassShift(potentialShift, massShifts, massShiftType))
+                    if (tolerance.IsWithinTolerance(deltaAbs, potentialShift) && MatchMassShift(potentialShift, massShifts, massShiftType))
```
rita-gwen, line 64: Need to verify if this is correct.

pwiz_tools/Skyline/Model/Results/SpectrumFilter.cs
```diff
@@ -1147,11 +1147,11 @@ public override string ToString()
                 // if more than one match at this m/z value return a list
 
                 double minMzDelta = double.MaxValue;
-                double mzDeltaEpsilon = Math.Min(_instrument.MzMatchTolerance, .0001);
+                double mzDeltaEpsilon = Math.Min(_instrument.IonMatchMzTolerance.GetAbsoluteValue(isoWin.IsolationMz.Value), .0001);
 
                 // Isolation width for single is based on the instrument m/z match tolerance
                 var isoTargMz = isoWin.IsolationMz.Value;
-                var isoWinSingle = new IsolationWindowFilter(isoTargMz, _instrument.MzMatchTolerance * 2);
+                var isoWinSingle = new IsolationWindowFilter(isoTargMz, _instrument.IonMatchMzTolerance.GetAbsoluteValue(isoTargMz) * 2);
```
rita-gwen, line 1154: Need to verify if this is correct.

pwiz_tools/Shared/Common/Chemistry/MzTolerance.cs
```diff
@@ -60,16 +63,32 @@ public MzTolerance(double value = 0, Units units = Units.mz)
             return 0;
         }
 
+        public static implicit operator MzTolerance(double tolerance)
```
nickshulman, line 66: I don't think this implicit conversion should exist, because it's inherently ambiguous to anyone looking at the code whether it's going to be mz or ppm.
How about two static methods:
`public static MzTolerance FromPpm(double ppm)
`and
`public static MzTolerance FromMz(double mz)`

pwiz_tools/Shared/Common/Chemistry/MzTolerance.cs
```diff
@@ -60,16 +63,32 @@ public MzTolerance(double value = 0, Units units = Units.mz)
...
+                case Units.ppm:
+                    return ionMz * Value * 1e-6;
+                default:
+                    return Value;
+            }
         }
 
         /// <summary>returns true iff b - a is greater than the value in tolerance (useful for matching sorted mass lists)</summary>
-        public static bool LessThanTolerance(double a, double b, MzTolerance tolerance)
+        public bool LessThanTolerance(double a, double b)
```
nickshulman, line 89: This method does not seem to be used anywhere. I think this method should be removed, since its name says "Less Than" but the implementation uses "<=".

pwiz_tools/Shared/Common/Collections/Range.cs
```diff
@@ -29,4 +32,32 @@ public Range(int start, int end) : this()
         public int End { get; private set; }
         public int Length { get { return End - Start; } }
     }
+
+    public struct Interval<T> where T:IComparable
```
nickshulman, line 36: Given that one of the fields in this struct is a Delegate, it kind of defeats the (memory overhead saving) purpose of it being a struct.

pwiz_tools/Skyline/Model/AbstractModificationMatcher.cs
```diff
@@ -580,16 +580,15 @@ private string NoteIfAnnotationMzDisagrees(LibKey key, SpectrumPeaksInfo.MI peak
                 var averageMass = charge.MassFromMz(peak.Mz, MassType.Average);
 
                 if (!(peakAnnotation.Ion.MonoisotopicMass.Equals(monoisotopicMass,
-                          Settings.TransitionSettings.Instrument.MzMatchTolerance) ||
-                      peakAnnotation.Ion.AverageMass.Equals(averageMass,
-                          Settings.TransitionSettings.Instrument.MzMatchTolerance)))
+                          Settings.TransitionSettings.Instrument.IonMatchMzTolerance) ||
+                      Settings.TransitionSettings.Instrument.IonMatchMzTolerance.IsWithinTolerance(peakAnnotation.Ion.AverageMass, averageMass)))
```
nickshulman, line 584: It's confusing that the code for the mono mass uses the ".Equals(double, MzMatchTolerance)" function, but the average mass uses "IsWithinTolerance".
Could this expression be rewritten so that the same function is used for comparing mono mass and average mass?

pwiz_tools/Skyline/Model/Results/IonMobilityFinder.cs
```diff
@@ -258,8 +258,8 @@ private bool ProcessFile(ChromFileInfo fileInfo)
             // Only use the transitions currently enabled
             var transitionPointSets = chromInfo.TransitionPointSets.Where(
                 tp => nodeGroup.Transitions.Any(
-                    t => (t.Mz - (tp.ExtractionWidth ?? tolerance)/2) <= tp.ProductMz &&
-                         (t.Mz + (tp.ExtractionWidth ?? tolerance)/2) >= tp.ProductMz))
+                    t => (t.Mz - (tp.ExtractionWidth ?? tolerance.Interval) / 2) <= tp.ProductMz &&
```
nickshulman, line 261: This "Tolerance.Interval" does not seem to be the right thing to be using if the tolerance is ppm.

pwiz_tools/Shared/Common/Chemistry/MzTolerance.cs
rita-gwen: It is used to reduce the amount of code changes. There is a lot of places in the code, especially in tests, where tolerance is passed to a method as a single number. With this operator in place it is implicitly converted to mz tolerance without need for explicit conversion.

pwiz_tools/Shared/Common/Collections/Range.cs
rita-gwen: Well, I think struct copy constructor performs a shallow copy, so a new heap allocation happens only during initial construction, but not at assignment or parameter passing.

pwiz_tools/Skyline/Model/AbstractModificationMatcher.cs
rita-gwen: Agree.

pwiz_tools/Skyline/Model/Results/IonMobilityFinder.cs
rita-gwen: Agree.

brendanx67:
Let's just close this PR. Nick and I were unable to think of a case where it would be appropriate, given that he is already not using this property when the MS/MS full-scan settings are set to DDA, which was my motivating case because then the precursor m/z is actually instrument supplied based on the MS1 spectra, which means they are not really appropriate for the "method match tolerance" which is supposed to match only m/z values provided in the instrument method by an instrument operator, which makes them prone to error in the m/z range and not really in PPM units.

Fortunately, it seems like this PR at least positively impacted PR#2156

brendanx67:
We are going to abandon this PR, since it no longer seems necessary.

PR #2283, opened 2022-09-13
Detect Waters lockmass function by noting if any function lacks an accompanying chromatogram where the others do not

bspratt:
Possibly should also be checking that the supposed lockmass function is the right kind of data?


pwiz_aux/msrc/utility/vendor_api/Waters/WatersRawFile.hpp
```diff
@@ -135,12 +136,28 @@ struct PWIZ_API_DECL RawData
...
         {
             string fileName = BFS_STRING(functionFilepaths[i].filename());
+            if (useChromatogramsToDetectLockmassFunction)
+            {
+                string chromatogramFileName = fileName;
+                boost::algorithm::replace_all(chromatogramFileName, "_FUNC", "_CHRO");
+                if (boost::range::find_if(chromatogramFilepaths, 
+                    [chromatogramFileName](const bfs::path& cfp) { return BFS_STRING(cfp.filename()) == chromatogramFileName; }) == chromatogramFilepaths.end())
+                {
+                    continue; // No matching chromatogram - presumably the lockmass function, don't add to list
```
chambm, line 158: This is going to make the functions entirely disappear from pwiz? I'm not sure that's a good idea unless you want to make it a Reader::Config option. It's entirely likely that users will want to see their lockmass spectra. Instead I think we could annotate those spectra with a userParam (at this point in the code it would need to be a new API function like `IsFunctionLockmass(int functionNumber)` which SpectrumList_Waters would call) and clients could exclude the lockmass spectra that way. But the Reader::Config route is valid too.

bspratt: I'm open to better ways to use this information. Though I thought msconvert already was trying to not pass lockmass scans through (Skyline is certainly trying to ignore them).

Is there any likelihood that a function lacks a matching chromatogram for reasons other than being lockmass data?

chambm: It's only Skyline that currently has the heuristic to exclude lockmass functions. I have no idea about the CHRO files. How did you come up with this heuristic? Have you checked PDA functions? If not I'll have to sleuth around to remember which of my test files has PDA.

chambm: Photo Diode Array. Records wavelength vs. intensity spectra (which can have wavelength vs intensity chromatograms, but I don't know if that maps to CHRO files). Skyline ignores such spectra of course but I've definitely had users that use them.

bspratt: Yeah, I'll need to make sure that it only considers functions that are mass spec data.

bspratt: Matt, can you take another look? I think I've addressed your concerns.

chambm (approved):
Looks good. From a Skyline POV I was wondering how important it is that this heuristic should work on already converted mzMLs. It would be do-able I think since you can just look at the sourceFileList to check for the CHRO* files instead of the actual filesystem. And in that case the entire heuristic could be moved to Reader_Waters and SpectrumList_mzML (to have access to the sourceFileList). But that would be a new kind of feature for our mzML reading: currently none of the Reader::Config options affect it.

PR #2290, opened 2022-09-20
Add a test to verify that tutorial download links in the Start page are valid,…

bspratt:
… and fix one that wasn't

### DIFF
--- a/pwiz_tools/Skyline/Controls/Startup/StartPage.cs
+++ b/pwiz_tools/Skyline/Controls/Startup/StartPage.cs
@@ -22,7 +22,6 @@ using System.Collections.Generic;
 using System.Drawing;
 using System.IO;
 using System.Linq;
-using System.Net.Http;
 using System.Windows.Forms;
 using pwiz.Common.Controls;
 using pwiz.Common.SystemUtil;
@@ -264,6 +263,8 @@ namespace pwiz.Skyline.Controls.Startup
 
         private void PopulateTutorialPanel()
         {
+            Tutorial = TutorialAction;
+
             var labelFont = new Font(@"Arial", 12F, FontStyle.Regular, GraphicsUnit.Point, 0);
             var labelAnchor = AnchorStyles.Left | AnchorStyles.Right;
             var labelWidth = flowLayoutPanelTutorials.ClientRectangle.Width - SystemInformation.VerticalScrollBarWidth;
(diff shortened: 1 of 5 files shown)


pwiz_tools/Skyline/Controls/Startup/StartPage.cs
```diff
@@ -637,6 +638,23 @@ private void Tutorial(string skyFileLocation, string pdfFileLocation, string zip
         {
             Assume.IsNotNull(skyFileLocation);
 
+            // Test support
+            if (TutorialActionsTestDataUrisOnly)
```
brendanx67, line 642: I'd rather see the test support limited to an accessor to an IEnumerable<Uri> and let the test itself do the HTTP access with an Assert.IsTrue(). I think the fact that you are using Assume.IsTrue() as a substitute for Assert.IsTrue() is a dead give-away that you are doing too much in the main project. Assume is not supposed to be for testing.

pwiz_tools/Skyline/Controls/Startup/StartPage.cs
```diff
@@ -728,6 +746,8 @@ public IFormView ShowingFormView
             }
         }
 
+        public bool TutorialActionsTestDataUrisOnly { get; set; }
```
brendanx67, line 749: Use an accessor to an IEnumerable<Uri> instead of a switch to tell the StartPage to perform a test.

brendanx67 (changes requested):
Make the StartPage accessor less of a switch to do a test inside the main code base to serving up information in the data model for the test to act on.

pwiz_tools/Skyline/Controls/Startup/StartPage.cs
bspratt: The issue is that enumerating the URIs isn't trivial. Those values are encoded as part of an EventAction and I haven't found any way to get at them at runtime other than actually running the action.

A more extensive refactor would be needed to make those URIs trivially enumerable. I can do that if you feel strongly about it.

brendanx67: Good point. It was trickier than I could see from my review. I played around with an Action member that could be set by the test, and felt it achieved my goal of keeping more of the test code in the test, feeling a bit more like when we insert test interfaces in other places to override code like querying UniProt.

Nice test. I tried it with the old URL and it nicely captures the issue and running it indefinitely mostly keeps the full test around 1 second.

brendanx67 (approved):

brendanx67:
I expect my changes to fix the extraneous using clause caught by the code inspection team city agent.

bspratt:
Nice, much tidier. I'll merge and backport once TC is done.

PR #2291, opened 2022-09-20
Text PSM -> mzIdentML transformation tool used by the CPTAC CDAP pipeline

edwardsnj:
I can maintain this tool on a fork indefinitely, but it would be nice to bring it into the fold. This program reads a (very) verbose, but easy to generate, text stream representing the PSM data from the CPTAC CDAP DDA peptide identification analysis pipeline, and represents it as mzIdentML. 

The Python based parsers live in the [cptac3-cdap/cptac-mzid](https://github.com/cptac3-cdap/cptac-mzid) repository. 

This code has worked, essentially unmodified, for many years, only requiring re-compilation to pick up new terms from the PSI-MS and PSI-MOD ontologies. 

Let me know if you think this is suitable for inclusion in the main repository.

I will attach a small example of the text-based PSM stream this tool consumes and the resulting mzIdentML file.

[example.txt.gz](https://github.com/ProteoWizard/pwiz/files/9608873/example.txt.gz)
[example.mzid.gz](https://github.com/ProteoWizard/pwiz/files/9608872/example.mzid.gz)

Cheers!


chambm (changes requested):
Hi Nathan,

Thanks for the contribution. Just a few changes and I think it's good to go as an example. (It would also be nice to have a Serializer_CPTAC or something in IdentData so it could work with idconvert, but it's not a really common format so this is fine.)

edwardsnj:
Great, thanks for the careful review. I'll work on the various things you've identified. I haven't used C++ actively for many years, probably haven't kept up with current best practices. Thanks for the pointers.

edwardsnj:
@chambm OK, used explicit std:: throughout, got rid of using namespace std; declared namspace pwiz::examples for my classes in TextPSMReader.hpp and used in TextPSMReader.cpp and textpsm2mzid.cpp; replaced all runtime lookups of hardcoded MS CV terms with CV enums; replaced irrelevant/misleading filename string with the empty string. 

I think this addresses all of your concerns. Let me know if you find more...

Cheers!

pwiz_tools/examples/TextPSMReader.hpp
```diff
@@ -0,0 +1,171 @@
...
+#include <map>
+#include <list>
+
+#include "pwiz/data/identdata/IdentData.hpp"
+#include "pwiz/data/common/CVTranslator.hpp"
+
+using namespace pwiz::data;
+using namespace pwiz::identdata;
+
+namespace pwiz::examples {
```
chambm, line 16: Hah, apparently this syntax was added in C++17. I didn't know about it!

pwiz_tools/examples/TextPSMReader.cpp
```diff
@@ -0,0 +1,1451 @@
+//
+// Copied from examples.cpp
+//
+//
+// Original author: Robert Burke <robert.burke@proteowizard.org>
```
chambm, line 5: Update name and copyright.

chambm (changes requested):
Update the header author/copyrights and it'll be ready (sorry I didn't notice them on first review).

pwiz_tools/examples/TextPSMReader.cpp
edwardsnj: Caught! OK, I've removed the old verbiage. I don't know if you want me to retain some of the previous boilerplate, such as license. I've put my email at the top of all three files, at least.

pwiz_tools/examples/TextPSMReader.hpp
edwardsnj: Well, in that case I should dumb it down, right? should I nest instead?

namespace pwiz {
  namespace examples {
     ....
   }
}
???????

chambm: Ah actually I was thinking about a global search and replace to switch all the files to the new format. I think we have C++17 as our minimum version. I could be wrong and it's at C++14 though, in which case this change would be for the best.

edwardsnj: OK, that's what I did. Don't want to be that guy that forces the need for a recent language feature. Seems to work.

pwiz_tools/examples/TextPSMReader.cpp
chambm: Best to keep the Apache license verbiage for consistency with the rest of the codebase.

PR #2308, opened 2022-09-27
Update Library Viewer context menu

rita-gwen:


brendanx67 (changes requested):
There are still consistency issues with these menus across all of the places they are shown:
1. The View menu on the main menu
2. Right-click in the Library Match view
3. Right-click in the Spectral Library Explorer
4. Right-click in the Full-Scan graph (when spectrum annotation is on)

The Spectral Library Explorer has an extra splitter above "Charges" that should be removed. The View menu doesn't use the new menu pushbutton UI.

Also, the right-click menus end with a splitter instead of having a splitter above Undo Zoom.

rita-gwen:
All menus that set ion types and charges are synchronized. The screenshots are here:
image

brendanx67 (approved):

pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.resx
```diff
@@ -721,7 +721,7 @@
     <value>Magenta</value>
   </data>
   <data name="btnAIons.Size" type="System.Drawing.Size, System.Drawing">
-    <value>21, 20</value>
+    <value>31, 28</value>
```
brendanx67, line 724: All these size changes make me a little concerned.

brendanx67 (changes requested):
The RESX file changes in ViewLibraryDlg make this change incompatible with a cherry-pick back to 22.2. You will need to implement a much smaller change there to fix the menus that got broken between 21.2 and 22.2, but not go this far with fully normalizing all menu cases.

rita-gwen:
I added missing Special Ions item to the main window View menu and fixed broken logic of the Score menu item in the Library Match viewer.

PR #2309, opened 2022-09-28
Skyline: Fixes for large transition list / assay library import

brendanx67:
Bruker reported an OutOfMemoryException trying to import a large assay library from PaSER. Testing with 6,000,000 line file (5x the Pan Human library) exposed a number of areas broken in this handling by changes in ImportTransitionListColumnSelectDlg with the introduction of small molecule support and a fix to handle non-rectangular data provided by a user.

These fixes now only support showing columns for long lines in the first 100 rows that we show in the form. Otherwise, the first unexpectedly long line now produces a normal error giving the line number with the line containing extra fields.

Attempting to restore save column selections also read the entire file and information derived from clicking "Check For Errors" was always discarded.


bspratt (approved):
Looks good. I do wish we had a proper performance test for very large input lists - one that actually fails when performance degrades. Those are tricky, of course, as performance can be highly system dependent.

PR #2310, opened 2022-09-28
Skyline: Simplify use of ImportTransitionListColumnSelectDlg in tests

brendanx67:
- Add ManualResetEvent and remove WindowShown boolean to reduce need for tests to add waiting
- Hopefully fix intermittent failures in TestInsert

### DIFF
--- a/pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
+++ b/pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
@@ -33,6 +33,7 @@ using System.Data;
 using System.Drawing;
 using System.Globalization;
 using System.Linq;
+using System.Threading;
 using System.Windows.Forms;
 using pwiz.Common.Controls;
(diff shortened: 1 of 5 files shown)


pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
```diff
@@ -989,6 +1013,9 @@ private void ButtonOk_Click(object sender, EventArgs e)
 
         public void OkDialog()
         {
+            // Wait until the window is fully shown to do processing below
+            WaitForShown();
```
nickshulman, line 1017: Isn't this line of code guaranteed to cause a hang if we happen to get here and we haven't been shown yet? 
There is no way for event thread code to run while we are waiting, so I don't see how _windowShownEvent.Set() will ever manage to be called.

brendanx67:
Yeah. I realized that. Disregard this PR.

brendanx67:
Realized this is not what I was hoping for.

PR #2317, opened 2022-10-04
Add support for pressure chromatograms from Thermo

chambm:
- removed empty BP chromatograms for WIFF2 files
- added pressure trace support for Thermo files
- Skyline: improved message displayed by TIC chromatogram graph if the file has no MS1 spectra

Review requested about the TIC chromatogram message.


brendanx67 (approved):
Looks reasonable.

PR #2318, opened 2022-10-06
Prevent NaN's in fold change values in PRM tutorial

nickshulman:
Fix NaN's that appear in fold change values in PRM tutorial (reported by Brendan)

The bug was originally introduced by the change to GroupComparer.cs in PR #2155.

This is my best guess as to how to fix this bug, but I need to do some more investigation including comparing the results that we get with this fix to what the results were before the bug was introduced.


brendanx67 (approved):
Thanks for adding the extra testing and the record mode.

PR #2319, opened 2022-10-06
Fix TestPerfTicArea

nickshulman:


brendanx67 (approved):
Okay. Makes sense.

nickshulman:
This fixes the test, but it might have the effect of undoing the behavior that Matt was trying to achieve with his checkin.

chambm:
Yes I'm pretty sure this will break that. Oh well, it's a pretty minor improvement, I may look at it again later in concern with some simAsSpectra fixes.

PR #2322, opened 2022-10-10
Fix location of drop targets when multiple screens have different dpi

nickshulman:
Fixed display of dockable window drop targets when second monitor has different "Scale" (dots per inch) (reported most recently by Juan)

The source code for this change is in this pull request:
https://github.com/uw-maccosslab/developers/pull/2


brendanx67:
Nice! This one drives me crazy, as I usually have different scales on my two monitors at home. I just switched to 100% scaling on both monitors for Skyline-Online today. Otherwise, when not demoing, I just "feel around" for the drop targets. It will be nice to have a fix.

brendanx67:
Can you merge this back to 22.2?

nickshulman:
There is still a bit of a problem where, if the Skyline window is spread across two monitors with different resolutions, it's not possible to drop on a drop target which is on the other screen.

image

brendanx67:
Worked for my tests with monitors at 1920x1080@100% and 2160x3840@150% starting the main window on both screens for the main window.

It still does not work, however, for docking multiple windows into a floating window. There I am still not seeing docking indicators for the floating windows. Though, the cursor hit-testing is working correctly (as it always did for the main window), you just need to hunt around for what you are looking for, knowing when you have it because of the blue shaded rectangle which also shows up correctly. It is just the docking indicators that are entirely missing on my screens.

brendanx67:
image

brendanx67:
image

brendanx67:
It is an improvement. So, maybe we should merge this fix but leave an issue active for the floating window case.

nickshulman:
It would be better to comment on the code changes in the docking windows source code repository:
https://github.com/uw-maccosslab/developers/pull/2

I think the fix I have there happens to fix the bug, but the reasoning behind it is completely flawed.
Also, I did not realize there were two different states of DockIndicator objects: sometimes their DockPane is null, and sometimes not null, and in the latter case, the DockPanel rectangle is irrelevant.

I think I'm close to figuring out what the actual right fix might be.

brendanx67:
Definitely want this in the 22.2 patch for Sunday night.

PR #2323, opened 2022-10-10
Enable exporting and importing Standard Type

nickshulman:
Enabled setting the Standard Type of a peptide using "--import-annotations" command line argument (requested by Jianqiao)


brendanx67 (changes requested):
Is this possible to test?

PR #2324, opened 2022-10-12
PRM optimization

kaipot:


nickshulman:
Currently, PeptideChromDataSets.IsSaveRawTimes looks like this:
```
        public bool IsSaveRawTimes
        {
            get
            {
                // We do not save raw times if there is an optimization function because it is too hard.
                return OptimizableRegression == null;
            }
        }
```
I believe the thing that was hard was getting the old CE optimization unit tests to pass.

If it's PRM optimization, I think we should try to save the raw times. For one thing, the class "RawTimeIntensities" has a lot more places to store extra information.

pwiz_tools/Skyline/Model/Results/ChromHeaderInfo.cs
```diff
@@ -174,7 +174,8 @@ public enum FlagValues
             raw_chromatograms = 0x80,
             ion_mobility_type_bitmask = 0x700, // 3 bits for ion mobility type none, drift, inverse_mobility, spares
             dda_acquisition_method = 0x800,
-            extracted_qc_trace = 0x1000
+            extracted_qc_trace = 0x1000,
+            optimization_scan_ids = 0x2000
```
nickshulman, line 178: I think this flag is either misnamed or unnecessary.

You would not need to separately say whether there are optimization scan id's-- there would be scan ids if "has_ms1_scan_ids" or "has_frag_scan_ids" were set. (I am not sure why there are separate flags for MS1 and MS2-- maybe there really is a mass spectrometer out there which collects MS1 spectra but uses a triple quad for MS2, or vice-versa).

Maybe this flag should be named "optimization" without the "scan_ids" part.

pwiz_tools/Skyline/Model/Results/OptStepChromatograms.cs
```diff
@@ -58,6 +58,31 @@ public static OptStepChromatograms FromChromatogram(ChromatogramInfo chromatogra
             return new OptStepChromatograms(chromatogramInfo.ProductMz, ImmutableList.Singleton(chromatogramInfo), 0);
         }
 
+        public static IEnumerable<OptStepChromatograms> FromChromatograms(IEnumerable<ChromatogramInfo> chromatogramInfos)
+        {
+            var curGroup = new List<ChromatogramInfo>();
+
+            OptStepChromatograms CurResult()
+            {
+                var result = new OptStepChromatograms(curGroup[curGroup.Count / 2].ProductMz, curGroup, (curGroup.Count - 1) / 2);
```
nickshulman, line 67: Is it correct to be passing in "curGroup[curGroup.Count / 2].ProductMz"?
That means that you are assuming that you have an equal number of chromatogram groups above and below step 0.

With SRM optimization, that ends up not being the case if the collision energy is close to either the upper or lower limit of what the collision energy is allowed to be. Can the same sort of thing also happen with PRM optimization?

It might be that the optimization step integer should be added to ChromGroupHeaderInfo so that we can always tell which optimization step a chromatogram group is for.

kaipot:
@nickshulman Thanks for the review, those are all good suggestions. Especially using raw times instead of interpolated.

nickshulman (approved):
Looks good to me.

PR #2326, opened 2022-10-12
BiblioSpec pLink and vendor support in Proxl XML

chambm:
- added support for Proxl XML files from pLink (requested by Zexin and Daniela)
- added support for vendor file extensions in Proxl XML reader


chambm:
@brendanx67 Should I cherry pick this?

brendanx67:
It doesn't seem critical to me. The audience is relatively small, and we can push them to use Skyline-daily.

PR #2331, opened 2022-10-20
Add detail to CachedFileInfo assertions for TestThermoRatio

chambm:
* changed assertions about serial number and sample id in TestThermoRatio to report expected and actual values when assertion fails


brendanx67 (approved):
Okay. Seems like this should provide more information the next time it fails. Thanks.

chambm:
Yeah when I changed the expected value I got:
` Assert.AreEqual failed. Expected:<TQU004d90>. Actual:<TQU00490>. AssertEx.AreEqual failed, expected: TQU004d90 actual: TQU00490`. That should be a start.

PR #2337, opened 2022-10-26
Add suffix to version string when running Skyline on Wine

chambm:
- added suffix to version string when running Skyline on Wine

### DIFF
--- a/pwiz_tools/Skyline/Util/Install.cs
+++ b/pwiz_tools/Skyline/Util/Install.cs
@@ -63,6 +63,7 @@ namespace pwiz.Skyline.Util
                 Version = string.Empty;
             }
 
+            RunningOnWine = Helpers.RunningOnWine;
         }
         public enum InstallType { release, daily, developer }
 
@@ -80,6 +81,7 @@ namespace pwiz.Skyline.Util
 
         public static bool IsDeveloperInstall { get; }
         public static bool IsAutomatedBuild { get; private set; }
+        public static bool RunningOnWine { get; }
 
         public static bool Is64Bit
         {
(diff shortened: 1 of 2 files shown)


pwiz_tools/Skyline/Util/Install.cs
```diff
@@ -80,6 +81,7 @@ public static InstallType Type
 
         public static bool IsDeveloperInstall { get; }
         public static bool IsAutomatedBuild { get; private set; }
+        public static bool RunningOnWine { get; }
```
brendanx67, line 84: Can we make this property IsRunningOnWine to be consistent with the others?

pwiz_tools/Skyline/Util/Install.cs
```diff
@@ -153,10 +155,11 @@ public static string ProgramNameAndVersion
         {
             get
             {
-                return string.Format(@"{0} ({1}-bit{2}{3}) {4}",
+                return string.Format(@"{0} ({1}-bit{2}{3}{4}) {5}",
```
brendanx67, line 158: Seems like this has an extra space? " {5}" with possible replacements for {5} leading to " " or "  running on Wine" (two spaces)

pwiz_tools/Skyline/Util/Util.cs
```diff
@@ -1918,6 +1919,17 @@ private static void ReportExceptionForRetry(int milliseconds, Exception x, int l
         // in case of dotCover and dotMemory."
         public static bool RunningResharperAnalysis => !string.IsNullOrEmpty(Environment.GetEnvironmentVariable(@"JETBRAINS_DPA_AGENT_ENABLE"));
 
+        [DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
```
brendanx67, line 1922: Do we really not build any differently for the Wine Docker container version? It is really just the 64-bit binaries running on Wine? I was expecting this to be something controlled at build time. Pretty cool, if the binaries are not different in any way. But, running on Wine is still useful information, since there have been cases where the code works slightly differently on Wine... Though, it does raise the question whether we should just make it a system version value rather than part of the software version text.

brendanx67 (changes requested):
Not quite ready to approve this yet for the version text. Helpers.RunningOnWine() would be nice for my newest test.

pwiz_tools/Skyline/Util/Install.cs
chambm: No extra space (" running on Wine" was assigned to {4} inside the parens, not {5}).

brendanx67 (approved):

PR #2343, opened 2022-10-27
Guard against possible null value

bspratt:
I can't actually see a control path that would allow this to be set null, but it can't hurt to add a check to make sure it's EMPTY rather than null. 

Not clear that this is actually the problem, but inspired by https://skyline.ms/announcements/home/issues/exceptions/thread.view?entityId=f15c02de-375f-103b-b47c-22f53556c296&_anchor=57340


nickshulman (approved):
Looks good to me.
The one spot which I found where it looks like the ion mobility could be null is in TransitionDocNode.cs where it says "chromInfo?.GetIonMobilityFilter()", which might mean this bug is my fault and has something to do with CE optimization (or, that line of code might not have anything to do with this bug).

PR #2348, opened 2022-10-31
Add "Skyline 22.2" to the list of available formats in the "File > Share" dialog.

nickshulman:
Also, remove the localized strings "Skyline 21.1" etc. from Resources.resx since they are the same in all languages and so it makes adding new version names easier.
Add "TestMostRecentReleaseFormatIsSupportedForSharing" which makes sure that the most recent version is in the list of available formats to save as.


brendanx67 (approved):
Looks good. Thanks.

PR #2350, opened 2022-11-01
Graceful handling of user editing a molecule in ways that make no sense with child precursor adducts

bspratt:
Handle the issue raised in https://skyline.ms/announcements/home/issues/exceptions/thread.view?entityId=cf3d9fd9-2b8f-103b-83d8-22f53556c1f4&_anchor=57198#row:57198 where user edited a molecule in a way that makes no sense with its child ion adduct descriptions. Now we check for that before closing the molecule edit dialog.


pwiz_tools/Skyline/SettingsUI/EditCustomMoleculeDlg.cs
```diff
@@ -678,6 +681,24 @@ public void OkDialog()
...
+            // If editing the molecule, make sure that the formula makes sense with any child precursors (e.g. not removing more Carbons than are in the molecule)
+            if (_formulaBox.NeutralFormula != null && _childAdducts != null)
+            {
+                foreach (var childAdduct in _childAdducts)
+                {
+                    try
+                    {
+                        childAdduct.ApplyToFormula(_formulaBox.NeutralFormula);
+                    }
+                    catch (InvalidOperationException e)
```
nickshulman, line 694: "ApplyToFormula" can also throw an ArgumentExcepion if there was an error parsing "_formulaBox.NeutralFormula".
So, if I type in a formula which has more than one minus sign in it, then an unhandled ArgumentException might be thrown.

You might want to move this code to some point after you have already made sure that "_formulaBox.NeutralFormula" is a valid formula.
Or maybe change this to catch all exceptions.

bspratt: Excellent point. I always appreciate your reviews.

nickshulman (approved):

PR #2351, opened 2022-11-01
Fix for the red dots for dotp values appear behind bars (issue 909)

rita-gwen:


brendanx67 (approved):
Looks simple enough that I'd like to merge to the release branch for the patch.

PR #2353, opened 2022-11-01
TestRunner manual workerport

chambm:
* overloaded TestRunner's workerport argument to manually set the port the server listens on

Having a manual workerport allows creating a Firewall exclusion based on port instead of on program path.


bconn-proteinms (approved):
This looks good. This is just what we wanted.

PR #2354, opened 2022-11-01
Remove slow code from PeakIntegrator.IntegratePeak

nickshulman:
Crawdad peak finder was being invoked for manually integrated peaks in a misguided attempt to set "Coeluting" to a meaningful value for manually integrated peaks. There were some scenarios where this code was much too slow.

Fixed performance problem with manually integrated peaks (reported by Elena)


brendanx67 (approved):
Good enough for now. We should revisit the best way to replicate this value for manual integration.

PR #2357, opened 2022-11-02
Skyline: Fix issues with library building for DIA-NN .speclib files…

brendanx67:
…for ASMS fall workshop

### DIFF
--- a/pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
+++ b/pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
@@ -658,11 +658,12 @@ bool DiaNNSpecLibReader::parseFile()
                 if (processedRuns.count(run) > 0)
                     continue;
                 currentRun = run;
+                setSpecFileName(run, false);  // make status output report the current spec file
             }
             else if (!bal::equals(currentRun.c_str(), run))
             {
                 // skip rows not from the current runs being processed; another loop iteration will be required
-                hasSkippedRuns = processedRuns.count(run) == 0;
+                hasSkippedRuns = hasSkippedRuns || processedRuns.count(run) == 0;
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


pwiz_tools/Skyline/Model/Lib/BiblioSpecLite.cs
```diff
@@ -311,38 +311,59 @@ public IEnumerable<SpectrumSourceFileDetails> GetDataFileDetails()
                         // Each of these numbers is subdivided by score type.
                         // Also, select "ssf.*" because not all tables have a column "cutoffScore".
```
chambm, line 312: Is this comment no longer accurate? If it is still is, won't the current change break whatever files don't have cutoffScore? Instead of the weird * though we can do a preliminary "HasColumn" to update the query logic. There's already a couple utility functions like that:
```
        private bool HasFileIdColumn(SQLiteConnection connection)
        {
            return SqliteOperations.ColumnExists(connection, @"RefSpectra", @"fileID");
        }
```

pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
```diff
@@ -658,11 +658,12 @@ bool DiaNNSpecLibReader::parseFile()
                 if (processedRuns.count(run) > 0)
                     continue;
                 currentRun = run;
+                setSpecFileName(run, false);  // make status output report the current spec file
             }
             else if (!bal::equals(currentRun.c_str(), run))
             {
                 // skip rows not from the current runs being processed; another loop iteration will be required
-                hasSkippedRuns = processedRuns.count(run) == 0;
```
chambm, line 665: Oops. Looks like it always should have been that way (I must have only tested skipping the last run of the file or something like that).

What about disabling reading the PEP column from the TSV?

brendanx67: We need more involved changes to support the new Bruker format, and I now have a contact who is involved in that dev work we should talk with about future plans. But, I am thinking we may need to check the .tsv file for "Exp.1/K0" / "Lib.1/K0" columns as a way of knowing that we are dealing with the Bruker format which requires a number of changes even to read the .speclib file correctly. So, not in this patch fix.

brendanx67: But, yes, the current code doesn't even read existing .speclib files correctly because of this bug, which we want fixed in the 22.2 patch release.

PR #2366, opened 2022-11-03
Minimal changes to fix the spectrum context menus

rita-gwen:
This updates the Full Spectrum and Library Match menus to this look. Let me know if any other changes are required in this patch.
image


brendanx67 (approved):
Thanks for the limited set of changes to fix the broken menu. Looks good.

PR #2367, opened 2022-11-04
Show library match with multiple precursors

kaipot:


nickshulman:
This looks great.

I wonder whether, when multiple peptides are selected, the dropdown could just show one entry for each peptide and charge, instead of having separate entries for light and heavy precursors.

In the document I am looking at, the Targets tree has light and heavy precursors but the library only has light spectra, so, when I select a different precursor, the title of the graph changes, but the data displayed in the graph stays the same.

image
image

image

If I were to select a particular peptide, then Skyline would not show any dropdown, since it would know that there was only one spectrum available for all of the precursors for the currently selected peptide.

[PrecursorDropdown.sky.zip](https://github.com/ProteoWizard/pwiz/files/9944098/PrecursorDropdown.sky.zip)

PR #2374, opened 2022-11-08
Make find dialog cancellable

nickshulman:
Changed to allow canceling "Edit > Find" operation (reported by Brendan)
Issue 922: Edit > Find needs to use LongWaitDlg and support cancel

Also, fixed some suspicious code in "DuplicatedPeptideFinder" which was comparing hash codes instead of peptide sequences.

image


brendanx67 (changes requested):
This appears to be a busy wait progress indicator, right? It seems like we should be able to show detailed progress by transition count. Getting the transition count from SrmDocument is a constant time operation. It should be possible to pass around a context object rather than just the CancellationToken that allows counting progress and status updates.

brendanx67:
P.S. - I understand that we may find something long before the end of the progress. The progress would tell the user how long it would take to execute a failed search where the thing is not found.

nickshulman:
I will create a new pull request with a simpler way of dealing with progress monitors.

PR #2379, opened 2022-11-10
Fix nightly failure in TestFiguresOfMerit

nickshulman:
The failure was "intermittent" because to speed up the test, TestFiguresOfMerit only tests a random subset of possible sets of settings. This failure happened when calculating the LOD with Bilinear fit.


brendanx67 (approved):
Thanks

PR #2383, opened 2022-11-13
Skyline:  Tester fixes

brendanx67:
- Fix Nightly parsing of perf tests to avoid assigning memory values of the first failure to all prior tests
- Fix TestRunner to avoid reporting 0.0 heap value during perf tests
- Fix SkylineTester parsing to avoid graphing zero memory values when lines contain unexpected output


brendanx67:
The parsing and output fix the problem that was causing the testresults module to show a horizontal line to the first failure.
image
The SkylineTester fix gets rid of missing values in the SkylineTester memory graphs:
image

PR #2385, opened 2022-11-16
Restore Skyline's ability to deal with chemical formulas that contain zero counts (e.g. C12N5H0 - that's a zero, not an Oxygen)

bspratt:
This is a strange construction but we used to allow it, I broke that recently.

Reported by Phil

### DIFF
--- a/pwiz_tools/Skyline/Model/CustomMolecule.cs
+++ b/pwiz_tools/Skyline/Model/CustomMolecule.cs
@@ -459,7 +459,7 @@ namespace pwiz.Skyline.Model
             get { return _formula; }
             protected set
             {
-                _formula = value ?? string.Empty;
+                _formula = BioMassCalc.MONOISOTOPIC.RegularizeFormula(value ?? string.Empty); // e.g. XeC12N1H0 => XeC12N1
                 var unlabeled = string.IsNullOrEmpty(_formula) ? _formula : BioMassCalc.MONOISOTOPIC.StripLabelsFromFormula(_formula);
                 UnlabeledFormula = Equals(_formula, unlabeled) ? _formula : unlabeled;
             }
--- a/pwiz_tools/Skyline/Test/AdductTest.cs
+++ b/pwiz_tools/Skyline/Test/AdductTest.cs
@@ -85,6 +85,12 @@ namespace pwiz.SkylineTest
         private void TestAdductOperators()
         {
             // Test some underlying formula handling for fanciful user-supplied values
+            AssertEx.AreEqual("COOOHN", BioMassCalc.MONOISOTOPIC.RegularizeFormula("COOOHNS0"));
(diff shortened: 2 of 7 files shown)


brendanx67:
Is it intentional or a mistake? Did the person really want it to mean zero
hydrogen atoms?

bspratt:
> Is it intentional or a mistake? Did the person really want it to mean zero hydrogen atoms?

They really did! I should have referenced this:
https://skyline.ms/announcements/home/support/thread.view?rowId=57478

brendanx67:
I don’t read that as a request to continue supporting that notation. The
request to me ends with the user admitting they can live without it. It
doesn’t at all sound like it was intentional with a desired meaning. They
did respond to your request for feedback.

Can we hold off on this revert to the old behavior until some tells us they
want/need it? To me disallowing it still seems like an improvement or a bug
fix. I can see reinstating the bug only if some has a valid need or desire
for it.

bspratt:
It seems like extra work for both us and the user to refuse to handle this admittedly weird input. Dealing with funky homebrewed transition lists is part of our appeal for researchers, I think.

We do have to make some kind of change, in any event - what we're doing now is obviously not great.

nickshulman:
We probably should not be trying to preserve zeroes in the formulas that people type in.
Anytime that Skyline parses a formula, and combines duplicates, Skyline should also remove all zeroes.
For this reason, "ParseCounts" should not take a parameter of whether to preserve zeroes-- they should always be removed.

One way to fix the original exception would be, in "StripLabelsFromFormula" to change:
`atomOrder.Select(atom =>
`
to:
`atomOrder.Intersect(dictAtomCounts.Keys).Select(atom =>`

brendanx67:
I agree that at least removing H0 from the formula is a better way of
handling this as a formula normalization. Note that you felt compelled to
tell readers that H0 was not HO. Since that is a possible mistake, you are
at risk of the problem of silently handling a user mistake in a way that is
hard to catch and ultimately surprising to them. As a principle, we prefer
an error message to silent acceptance with confusing and surprising
results. I worry that you may be headed to the wrong side of this principle
because of one confused user.

If you persist, I agree with Nick that it is better not to persist this
confusing construct, and removing it would seem to give the user more
chance of noticing they did something strange unintentionally.

bspratt:
Excellent. It's already the case that this revised code's support for zero count atoms in a transition list just quietly removes them, as seen in the new test.

In the context of the actual data it's pretty clear that this weirdness is programmatically generated and not typos. (And the user has also told us that they really did mean zero.) In any event, "Zero" vs "Ohh" is not the kind of typo we could really hope to catch in general - CH20 vs CH2O for example. If this is truly a concern our best response is to make sure Skyline uses a font where you can't miss the difference visually.

I was trying for the minimal behavior change but as usual I expect Nick's right about not wanting that switch in ParseCounts. I'll see what if any knock on effects that might have.

PR #2386, opened 2022-11-16
Run EncyclopeDia in Skyline

chambm:
Still some WIP but it's testable. I added a new "EncyclopeDia Search" option under Peptide Search in File/Import.

- added EncyclopeDiaSearchDlg and EncyclopeDiaHelpers for running its search engine inside Skyline
* refactored Prosit model code so it could be used without the Skyline document model
* refactored DDASearchControl as a base class (SearchControl) for EncyclopeDiaSearchControl and DDASearchControl
* fixed max_occurs for peptide_integration Skyline_Current.xsd
* added hard-link utility functions in UtilIO because EncyclopeDia requires input files to be in the same directory and tries to process all files in that directory even when only a subset is desired; hard-links are of course less space-wasting than actually copying the data
* updated C# version in Shared/Common to 8.0
* added SqliteOperations.DumpTable() for quick testing of SQLite file contents
* changed ProcessRunner to take an optional outputAndExitCodeAreGood delegate for detecting errors in processes which don't report exit codes properly (for example, EncyclopeDia!)
* added new option in KeyValueGridDlg for setting additional parameters from a list of valid values (which translates to a drop-down style ComboBox)


chambm:
@brendanx67 @nickshulman
I'm thinking about adding a SHA-1 hash column to SpectrumSourceFiles in BiblioSpec so we can confidently skip regenerating a blib if the input file(s) and settings haven't changed. This would be particularly useful for this PR for unit and functional testing of the Prosit stuff so I can cache the Prosit output and not have to call Prosit if the input file (in this case a CSV file) hasn't changed. Especially for the perf test we don't want to actually run that on Prosit every night for every PR. But maybe do it once a week on master.

If you don't think that's a good idea I can implement an alternative like just assuming the file hasn't changed if it exists during testing (similar to how SimpleFileDownloader changes its behavior when functional testing).

pwiz_tools/Skyline/Model/Lib/BlibData/BlibDb.cs
```diff
@@ -267,11 +419,11 @@ public void Write()
                 int i = 0;
                 var sourceFiles = new Dictionary<string, long>();
                 var proteinTablesBuilder = new ProteinTablesBuilder(session);
-
+                using var spectrumInserter = new SpectrumInserter(session);
                 foreach (var spectrum in listSpectra)
                 {
                     var dbRefSpectrum = RefSpectrumFromPeaks(session, spectrum, sourceFiles);
-                    session.Save(dbRefSpectrum);
+                    spectrumInserter.InsertSpectrum(dbRefSpectrum);
```
chambm, line 426: This SpectrumInserter change more than doubles the performance of this function, but it does disconnect the code from the NHibernate schema. Thus it has to be manually maintained if the schema is changed. Worth it? @brendanx67 @nickshulman

pwiz_tools/Skyline/Util/Extensions/UtilDB.cs
```diff
@@ -119,9 +163,46 @@ public static byte[] Compress(this byte[] uncompressed, int level)
...
+                            if (_workingBuffer.Length - Z.AvailableBytesOut > 0)
+                                ms.Write(_workingBuffer, 0, _workingBuffer.Length - Z.AvailableBytesOut);
+                            continue;
+                        default:
+                            throw new ZlibException(@"deflating: " + Z.Message);
+                    }
+                }
+                while (Z.AvailableBytesIn > 0 || Z.AvailableBytesOut == 0);
+                Z.ResetDeflate();
+
```
chambm, line 205: @nickshulman @brendanx67 A review here would be appreciated. This code is used not just to compress the BlibDb's m/z and intensity arrays, but also the scan id arrays and point arrays in ChromCacheBuilder. It was the latter usage that required me to make the ZlibCodecs thread-static so they were parallel-safe. I haven't added a way to deallocate these static objects/buffers, but since there's only 3 of them, is it worth worrying about? I wasn't able to find a definitive answer for how you're supposed to ensure that thread-static stuff is destroyed when the thread is destroyed, especially when the thread pool is involved.

chambm: The reason for this optimization was excessive time spent initializing zlib buffers (every time an individual m/z or intensity array was compressed for writing a blib), identified by profiling. I discussed it with Nick when I was working on it, but that was before I knew I had to make it parallel-safe because ChromCacheBuilder was using it in parallel.

This change gives a big boost to Compress performance, but I forgot to record timings. The key thing is that profiling shows that the vast majority of time spent in writing blib files is now spent in SQLite insertion I/O, not in NHibernate or zlib. This might even be an appreciable speed-up for the other code using Compress, but I haven't profiled/measured that.

nickshulman: Instead of making it thread-static, can you just make it so the caller can pass the ZLibCodec into the Compress method?

ChromCacheBuilder is thread-safe when it is compressing bytes. One ChromCacheBuilder gets created for each raw file being imported. The ChromCacheBuilder (or actually, its base class "ChromCacheWriter") should create the ZLibCodec at the beginning of the import operation and get rid of it at the end.

(Also, ChromCacheMinimizer should create its own codec)

There are other places in the code where we compress things, but for most of those we don't really care about performance so it's fine if those places have to create a new ZLibCodec for each compression operation.

chambm: How would that work with calling it from a ParallelEx loop?

nickshulman: Good point. Is there a place where we do that? 

We should probably write a variation of ParallelEx.ForEach where you can specify a function to be called per thread, and that function's return value gets passed to the body action.

brendanx67: Couldn't you just use a lambda expression inside the caller that passes in the value?

pwiz_tools/Skyline/Model/Lib/BlibData/BlibDb.cs
```diff
@@ -244,6 +245,157 @@ public void Write()
...
+
+        private class SpectrumInserter : IDisposable
+        {
+
+            // ReSharper disable LocalizableElement
+            static string _insertSpectraSql = "INSERT INTO RefSpectra (id, peptideSeq, peptideModSeq, precursorCharge, " +
+                                              "precursorAdduct, precursorMZ, moleculeName, chemicalFormula, inchiKey, " +
+                                              "otherKeys, prevAA, nextAA, copies, numPeaks, ionMobility, collisionalCrossSectionSqA, " +
+                                              "ionMobilityHighEnergyOffset, ionMobilityType, retentionTime, fileId, SpecIDinFile, score, scoreType) " +
+                                              "VALUES (" + string.Join(",", Enumerable.Repeat("?", 23)) + ")";
```
nickshulman, line 257: You might want to add "; select last_insert_rowid();" to the end of this SQL statement.
The way that you are doing it now where you first query for the highest existing rowid and keep incrementing that number every time you insert a row is a fine way to do it too.

pwiz_tools/Skyline/Model/Lib/BlibData/BlibDb.cs
nickshulman: Yes, it's probably worth it to write custom insert statements (although you might want to use more reflection in the implementation of that inserting code).
Sometimes, when you are worried about two pieces of code that might go out of sync with each other it's a good idea to write a unit test which makes sure they are the same.

pwiz_tools/Skyline/FileUI/PeptideSearch/EncyclopeDiaSearchDlg.cs
```diff
@@ -0,0 +1,611 @@
...
+                MessageDlg.Show(this, Resources.EncyclopeDiaSearchDlg_NextPage_A_FASTA_file_is_required_for_an_EncyclopeDia_Prosit_search_);
+                return;
+            }
+
+            btnNext.Text = Resources.BuildLibraryDlg_btnPrevious_Click__Next__;
+
+            if (CurrentPage == Pages.run_page)
+            {
+                ImportEncyclopediaLibrary();
+                return;
```
chambm, line 301: @nickshulman @brendanx67 Running EncyclopeDia doesn't actually change the document until the import peptide search wizard adds the libraries created in the search. But it would make sense to me to give EncyclopeDia its own entry though: what is the proper way to add an audit log entry when the document doesn't change?

brendanx67: There is really no such thing as this. Again, I think we probably want to expand our understanding of auxiliary files that might get included with a File > Share. So, probably you will want to modify the document with something that points at an EncyclopeDIA product like a parameters file or something that would allow you to re-run the analysis from scratch.

brendanx67: Yeah. Maybe it is worth having a schema element for all searches with name-value pairs for parameters to store how the search was run to make it easier to re-run

pwiz_tools/Skyline/TestFunctional/EncyclopeDiaSearchTest.cs
```diff
@@ -0,0 +1,213 @@
...
+            RunUI(() =>
+            {
+                searchDlg.DefaultCharge = 3;
+                searchDlg.DefaultNCE = 33;
+                searchDlg.MinCharge = 2;
+                searchDlg.MaxCharge = 3;
+                searchDlg.MinMz = 690;
+                searchDlg.MaxMz = 705;
+                searchDlg.ImportFastaControl.MaxMissedCleavages = 2;
+                searchDlg.SetAdditionalSetting("PercolatorTrainingFDR", "0.15");
```
nickshulman, line 84: This line seems to fail in foreign languages.
I'm thinking maybe it should be:
`searchDlg.SetAdditionalSetting("PercolatorTrainingFDR", 0.15.ToString(CultureInfo.CurrentCulture));`
or maybe "SetAdditionalSetting" should be changed to take an object instead of a string.

brendanx67:
Yeah. Never put a decimal point in quotes in a test.

pwiz_tools/Skyline/Model/Prosit/Models/PrositIntensityModel.cs
```diff
@@ -140,25 +152,36 @@ public override PrositMS2Spectra CreateSkylineOutput(SrmSettings settings, IList
...
+            public int PrecursorCharge { get; }
+            public SignedMz PrecursorMz { get; }
+            public ExplicitMods ExplicitMods { get; }
+            public IsotopeLabelType LabelType { get; }
             public int? NCE { get; private set; }
 
             public PeptidePrecursorNCE WithNCE(int nce)
             {
-                return new PeptidePrecursorNCE(NodePep, NodeGroup, LabelType, nce);
+                return new PeptidePrecursorNCE(Sequence, PrecursorCharge, PrecursorMz, ExplicitMods, LabelType, NCE);
```
nickshulman, line 184: This "NCE" in capitals suspicious. "PeptidePrecursorNCE" be constructed with the "nce" parameter that was passed in, instead of the NCE property?
Also, "CreatePrositInputRow" definitely wants a non-null PeptideDocNode, which causes "TestPrositBuildLibrary" to fail.

chambm: That was definitely suspicious. I think I fixed these issues.

nickshulman (approved):
Looks great to me.
All the tests that were previously failing for me seem to be passing now.
I am going to try running all the tests in Japanese and Chinese on macs2, and if they all pass I think this should be merged to master.

(It would be nice if Import Peptide Search warned the user at the beginning if Prosit was not configured correctly. Otherwise, they end up getting an error on about page 4 of the wizard)

chambm:
Let me know if your testing on macs2 works and I'll squash it in.

nickshulman:
I got a failure in both TestEncyclopeDiaSearchTutorial and TestEncyclopeDiaSearchTutorialFullDataset:
```
Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException: Assert.Fail failed. Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException: Assert.Fail failed. Expected target counts <557, 5734, 5734, 40799> do not match actual <557, 5733, 5733, 40777>.
   at TestPerf.EncyclopeDiaSearchTutorialTest.ValidateTargets(Int32[]& targetCounts, Int32 proteinCount, Int32 peptideCount, Int32 precursorCount, Int32 transitionCount, String propName) in D:\Users\nicksh\git_d\sky_impenc2\pwiz_tools\Skyline\TestPerf\EncyclopeDiaSearchTutorialTest.cs:line 480
   at TestPerf.EncyclopeDiaSearchTutorialTest.<>c__DisplayClass10_0.<DoTest>b__12() in D:\Users\nicksh\git_d\sky_impenc2\pwiz_tools\Skyline\TestPerf\EncyclopeDiaSearchTutorialTest.cs:line 336
   at pwiz.SkylineTestUtil.AbstractFunctionalTest.<>c__DisplayClass30_0.<RunUI>b__0() in D:\Users\nicksh\git_d\sky_impenc2\pwiz_tools\Skyline\TestUtil\TestFunctional.cs:line 300
   at System.Windows.Forms.Control.MarshaledInvoke(Control caller, Delegate method, Object[] args, Boolean synchronous)
   at System.Windows.Forms.Control.Invoke(Delegate method, Object[] args)
   at TestPerf.EncyclopeDiaSearchTutorialTest.DoTest() in D:\Users\nicksh\git_d\sky_impenc2\pwiz_tools\Skyline\TestPerf\EncyclopeDiaSearchTutorialTest.cs:line 338
   at pwiz.SkylineTestUtil.AbstractFunctionalTest.RunTest() in D:\Users\nicksh\git_d\sky_impenc2\pwiz_tools\Skyline\TestUtil\TestFunctional.cs:line 1879
   at pwiz.SkylineTestUtil.AbstractFunctionalTest.WaitForSkyline() in D:\Users\nicksh\git_d\sky_impenc2\pwiz_tools\Skyline\TestUtil\TestFunctional.cs:line 1824
------------------------------------------------------
```

chambm:
OK good catch. TC perftests aren't running with the Prosit production config. I'll fix that and the tests.

nickshulman (approved):

PR #2389, opened 2022-11-18
Waters DDA changes

pete-reay-waters:
The Waters DDA changes discussed on a call with Matt and Brendan.

Waters DDA data is now processed with the DDAProcessor in the MassLynxSDK. This handles:
- Lock mass corrected precursor mass (in future this will be expanded to use peak picking, rather than lock mass correcting the set mass)
- Merging of adjacent MS2 scans which relate to the same set mass. The id of such scans is expanded with `merged=23-25`, though for compatibiltiy the `scan=23` is still present, which indicates the first scan in the group.
- Omitting the lock mass function

The biggest difference is that in 'DDA mode' SpectrumList_Waters creates each index entry 'on the fly' as the scan is retrieved. This avoids looking up the values twice in the index.

Currently this uses a beta version of the MassLynx SDK, therefore I have made this a draft pull request, so not to be merged now. We plan to have a new release version in the next couple of weeks, at which point I'll update the PR.


pwiz_tools/commandline/msconvert.cpp
```diff
@@ -337,6 +338,9 @@ Config parseCommandLine(int argc, char** argv)
         ("combineIonMobilitySpectra",
             po::value<bool>(&config.combineIonMobilitySpectra)->zero_tokens(),
             ": write all ion mobility or Waters SONAR bins/scans in a frame/block as one spectrum instead of individual spectra")
+        ("ddaProcessing",
+            po::value<bool>(&config.ddaProcessing)->zero_tokens(),
+            ": combine MS2 spectra referring to the same precursor and survey scan, and calculate accurate precursor masses.")
```
chambm, line 343: Perhaps add a "if support by vendor API" prefix phrase here

pwiz_aux/msrc/utility/vendor_api/Waters/WatersRawFile.hpp
```diff
@@ -425,8 +425,41 @@ struct PWIZ_API_DECL RawData
         }
     }
 
+    void EnableDDAProcessing()
+    {
+        DDAProcessor.SetRawReader(Reader);
```
chambm, line 430: Does this block execution for a significant amount of time?

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.hpp
```diff
@@ -112,6 +115,9 @@ class PWIZ_API_DECL SpectrumList_Waters : public SpectrumListIonMobilityBase
     mutable boost::mutex readMutex;
 
     void createIndex();
+    void createDDAIndex();
+    void getDDAScan(unsigned int index, vector<float>& masses, vector<float>& intensities, IndexEntry& ie) const;
```
chambm, line 119: Unfortunately that IndexEntry reference that gets passed in to this function should have been const. There's no good reason to change the index after it's been generated (the only reason it's mutable is so it can be generated on-demand instead of at construction). This will need to be refactored to return the setMass and precursorMass explicitly (could be floats set by ref or a new struct).

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
```diff
@@ -722,6 +749,46 @@ PWIZ_API_DECL void SpectrumList_Waters::createIndex()
...
+        generate(sink,
+                "function=" << int_ << " process=" << int_ << " scan=" << int_ << " merged=" << int_ << "-" << int_,
+                ie.function + 1, ie.process, ie.scan + 1, startScan + 1, endScan + 1);
+    }
+    idToIndexMap_[ie.id] = ie.index; 
+}
+
+PWIZ_API_DECL void SpectrumList_Waters::createDDAIndex()
+{
+    size_ = rawdata_->GetDDAScanCount();
```
chambm, line 788: The index needs to be created up front. Accessing a single spectrum must mean that all spectra are indexed, which means their ids are generated and SpectrumList_Waters::spectrumIdentity() returns the IndexEntry instantly without having to allocate anything. If this is a costly operation, it can be progress monitored with iterationListenerRegistry. I don't think there's an example right now of using that in a vendor SpectrumList, but we do use it for a couple SpectrumListWrappers.

chambm (changes requested):
Very good integration into our code! Thanks for that! I'm a little torn as to whether this should be a ReaderList::Config option (as you did with ddaProcessing) because that in turn needs a dedicated command-line option, or if it should be added to SpectrumList_ScanSummer as a vendor-specific optimization like we already do for peak picking (i.e. we call vendor peak picking if supported). If it was to go into ScanSummer it would have to be the first filter just like peak picking, but it seems like it would replace peak picking and lockmass refinement for Waters data.

NB: if I decide it should go into ScanSummer I'll of course take on the task of refactoring it that way.

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.hpp
pete-reay-waters: > so it can be generated on-demand instead of at construction
Isn't that exactly what we are doing here - generating on demand?

We need other stuff beyond the setMass and precursorMass, we also populate the function and scan; pretty much everything in the index, in fact.

The alternative is to populate the index up-front (now we just set the size up-front); however, it's doing the same work by generating it as we get each scan - I don't see that's actually degrading performance. Or am I missing something?

pete-reay-waters: As it happens, at one point it *was* generating up front but Martin who owns the MassLynxSDK decided it would be better to also return the `masses` and `intensities` at the point of getting the scan information, which means we have to generate the index 'on-demand' to avoid getting the same information twice.

If we *really* don't want to generate the index on-demand then we could refactor this again.

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
pete-reay-waters: Yes, I was also considering implementing it as a filter. However it seemed a bit complicated by the fact that the LockMassRefiner has a concrete dependency on SpectrumList_Waters, so all the Waters refiners would need to be applied in the right order, and probably aware of each other, so that seemed messier.

chambm: Yes I considered that messiness but doesn't DDAProcessor replace LockMassRefiner (for DDA data, at least)?

pete-reay-waters:
Look to be a couple of issues with the build/checks. I think both are a result of the MassLynxSDK being built a different way to previous releases, presumably with different versions of the C++ runtime.
- We no longer have an x86 build. Is there a way to disable Waters raw support for the x86 build? Out of interest, is anyone still using x86?
- There are some tiny changes in floating point numbers in the expected test results. Would you be happy with updating the expected results, provided we confirm all the changes are very small (they look to be in the 6th significant figure or thereabouts)?

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.hpp
chambm: When a Reader_* is constructed, it also constructs its SpectrumList_* and ChromatogramList_*. Building the index for those can take some time, especially spectra since there are usually more of those. So if the calling code only wants chromatograms, it makes no sense to make them wait for the spectrum list index to be built. So the index is generated on-demand the first time a spectrum, spectrumIdentity, or SpectrumList::size() is called. But that's a vendor-by-vendor thing. I see that in Waters we were NOT doing that because generating the index on construction was never a big delay. Looks like we do it in Agilent:

```
PWIZ_API_DECL size_t SpectrumList_Agilent::size() const
{
    boost::call_once(indexInitialized_.flag, boost::bind(&SpectrumList_Agilent::createIndex, this));
    return size_;
}


PWIZ_API_DECL const SpectrumIdentity& SpectrumList_Agilent::spectrumIdentity(size_t index) const
{
    boost::call_once(indexInitialized_.flag, boost::bind(&SpectrumList_Agilent::createIndex, this));
...snip...
}


PWIZ_API_DECL size_t SpectrumList_Agilent::find(const string& id) const
{
    boost::call_once(indexInitialized_.flag, boost::bind(&SpectrumList_Agilent::createIndex, this));
...snip...
```

The vital thing is the full index must be generated and immutable before the size(), spectrumIdentity(), find(), or spectrum () functions can run. I'm surprised the SpectrumList_Waters test passes because I could of sworn we were testing that spectrumIdentity works in VendorReaderTestHarness. But it looks like I was wrong. Maybe it got buried in a git stash somewhere. :/

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
chambm: > Look to be a couple of issues with the build/checks. I think both are a result of the MassLynxSDK being built a different way to previous releases, presumably with different versions of the C++ runtime.
> 
> * We no longer have an x86 build. Is there a way to disable Waters raw support for the x86 build? Out of interest, is anyone still using x86?
> * There are some tiny changes in floating point numbers in the expected test results. Would you be happy with updating the expected results, provided we confirm all the changes are very small (they look to be in the 6th significant figure or thereabouts)?

Yes we can disable Waters for x86. Or possibly stick with the older SDK. Probably best to just disable though. I think the only x86 users these days would be for extremely old instrument computers. The AB Sciex 4700/4800 TOF/TOF has this weird API I access through their Data Explorer program and it's only available as x86. But yeah, extremely old.

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.hpp
pete-reay-waters: > The vital thing is the full index must be generated and immutable before the size(), spectrumIdentity(), find(), or spectrum () functions can run

Ah ok, I didn't realise this, I will update the code accordingly

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
chambm: Yes, on further thought that does seem pretty complicated because normally all the special indexing logic would be kept in the filter, but in this case it would require close integration with the Waters API which we normally don't do in filters.

pete-reay-waters:
> Yes we can disable Waters for x86.

Don't suppose you could give me a hint on where to do this?

chambm:
> > Yes we can disable Waters for x86.
> 
> Don't suppose you could give me a hint on where to do this?

Don't worry about it, I'll take care of it if you remove the x86 binaries from the 7z. BTW did you add the password to encrypt the 7z?

pete-reay-waters:
Ok, I've made those changes:
- Index is now built at the start, rather than on-demand
- Removed x86 folder from Waters 7z file (and yes I added the password)
- Added "if supported by vendor" to ddaProcessing argument description
- Regenerated expected mzML results for Waters tests

Matt if you could disable the Waters/x86 option then hopefully that will fix the remaining issues.

pwiz_aux/msrc/utility/vendor_api/Waters/WatersRawFile.hpp
pete-reay-waters: No, I believe not in this case, it just takes the handle from the existing reader.

chambm:
Well, that was weird. I used GitHub's CLI to checkout this PR. I made my change, tested it, committed it, and then tried to push it. I figured it would either fail to push to your fork or maybe work if it realized it was for a PR to our repo. I did NOT expect it to push to pwiz's master and apparently count this PR as merged. Oops! I had to force reset master to what it was at. And I guess this PR needs to be reissued? Here's a patch for the 32-bit disabling code. Which GitHub requires me to zip for no good reason!
[disable-32bit-test.zip](https://github.com/ProteoWizard/pwiz/files/10070454/disable-32bit-test.zip)

PR #2390, opened 2022-11-18
BlibBuild: Fixes for DIA-NN speclib parsing

brendanx67:
- may not be ready for merging but hopefully fixes for issues seen at CRG course
- increased robustness by comparing .speclib and .tsv Precursor.IDs to avoid a mix-up situation

### DIFF
--- a/pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
+++ b/pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
@@ -476,9 +476,15 @@ class Library {
         in.read((char*)&iRT_min, sizeof(double));
         in.read((char*)&iRT_max, sizeof(double));
         read_array(in, entries, version);
+        int i = 0;
         for (auto &e : entries)
         {
             e.lib = this;
+            std::string precursor = precursors[i++];
+            if (!bal::equals(e.name.c_str(), precursor.c_str()))
+            {
+                Verbosity::error("Precursor mismatch between %s and %s in speclib file", e.name.c_str(), precursor.c_str());
+            }
             entryByModPeptideAndCharge.emplace(e.name, std::ref(e));
         }
         if (version <= -1 && in.peek() != std::char_traits<char>::eof()) read_vector(in, elution_groups);
@@ -587,7 +593,7 @@ bool DiaNNSpecLibReader::parseFile()
         impl_->specLib.read(specLibStream);
     }
 
-    typedef io::CSVReader<8, io::trim_chars<' ', ' '>, io::no_quote_escape<'\t'>> ReportReaderType;
+    typedef io::CSVReader<9, io::trim_chars<' ', ' '>, io::no_quote_escape<'\t'>> ReportReaderType;
 
     auto diannReportFilepath = bal::replace_last_copy(string(impl_->specLibFile_), "-lib.tsv.speclib", "-report.tsv");
     /*if (diannReportFilepath == impl_->specLibFile_)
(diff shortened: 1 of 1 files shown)


chambm:
So did this fix the 2021KM017-FCB-libr.tsv.speclib error you were able to repro on your laptop?

brendanx67:
Yes. I was able to build on my laptop from the flash drive that has the
data on it.

pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
```diff
@@ -476,9 +476,15 @@ class Library {
         in.read((char*)&iRT_min, sizeof(double));
         in.read((char*)&iRT_max, sizeof(double));
         read_array(in, entries, version);
+        int i = 0;
```
chambm, line 479: This is a good idea but I'll refactor it to use iterators (with a pre-check that the vectors are the same length).

pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
```diff
@@ -622,9 +628,9 @@ bool DiaNNSpecLibReader::parseFile()
             if (kvp.first < 1)
                 break;
 
-            io::CSVReader<1, io::trim_chars<' ', ' '>, io::no_quote_escape<'\t'>> reportReader(kvp.second.string().c_str());
-            reportReader.read_header(io::ignore_extra_column | io::ignore_missing_column, "Precursor.Id");
```
chambm, line 626: Why read "Run" here if it not checking for it below in `if (reportReader.has_column`?

pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
```diff
@@ -669,11 +675,16 @@ bool DiaNNSpecLibReader::parseFile()
 
             auto findItr = speclib.entryByModPeptideAndCharge.find(precursorId);
             if (findItr == speclib.entryByModPeptideAndCharge.end())
+            {
+                // skip contaminant proteins if they are not included in the speclib file
+                if (bal::starts_with(proteinGrp, "contaminant_"))
```
chambm, line 680: Seems pretty arbitrary. This is totally dependent on what's in the FASTA file, right?

chambm (changes requested):
(changes to be made by me)

pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
brendanx67: I added it to be able to process the dataset from the Van Eyk lab. Otherwise, it stops with an error. I suppose another option would be to just skip anything that is missing. This is at least limited to a known case for now. Can it hurt?

pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
brendanx67: I was thinkinkg forward to when I would test for "Run" being absent and "Lib.1/K0" would be present instead of "IMS" and this would indicate Bruker Paser results. Though, I am not totally sure they want to provide .speclib file at all yet instead of just a collection of .tsv files. So, maybe you are right and we should remove "Run" here.

pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
chambm: I can't find where in DIA-NN it has any special handling for contaminant proteins. Maybe it's hard-coded (but I checked the last open source version of diann.cpp and it's not there). @vdemichev could you please comment on how precursorIds could be in the tsv but not the speclib?

chambm:
Also should probably add a test case that fails without this PR. Can I adapt the KM017 set (extracting just a few rows). Not sure how I can subset the speclib though.

PR #2394, opened 2022-11-23
Waters DDA changes 2

pete-reay-waters:
Update to use the Waters DDA processor, if the `--ddaProcessing` command line arg is used.


pete-reay-waters:
@chambm do we perhaps need to disable some tests in the x86 build? Looks like there are failures due to Waters RAW support not being enabled. I'm assuming because we've turned it off

https://teamcity.labkey.org/buildConfiguration/bt36/2133195?buildTab=log&focusLine=6364&logView=flowAware&linesState=546.3924.3971

chambm:
Hmm, I am not seeing that Reader_Waters_Test failure locally with your branch. Do you?

pete-reay-waters:
> Hmm, I am not seeing that Reader_Waters_Test failure locally with your branch. Do you?

No, they pass locally for me. These changes directly correspond to the updated mzML expected values. E.g. [these changes](https://github.com/ProteoWizard/pwiz/pull/2394/commits/65d94bc925a6fbcf1660206240180da094798663#diff-e153b9fce89e9a311c1065f93f651fc7aaacd99324c808275a55e24fe51c346dL990) directly relate to [these test results](https://teamcity.labkey.org/buildConfiguration/bt83/2133468?buildTab=log&focusLine=3857&logView=flowAware&linesState=547.2622.3834) from TeamCity

What do the + and - mean in the teamcity results? Either:
* The tests aren't using the new expected data
* OR whatever has caused the slight differences in floating point numbers on my machine (and yours) in the latest MassLynxSDK doesn't take effect on the TeamCity server. Best guess is a different version of VC++

chambm:
The `+` is the current result (MSData from vendor format), the `-` is the expected result (mzML). I too thought about a different version of VC++. I am not super enthused about tracking it down so I'm trying something in code. Unfortunately this precision error is at the insignificant end of a large number rather than a small one, so our precision-limiting comparison code doesn't work properly. If this doesn't work I'll look at adjusting it for large numbers.

pete-reay-waters:
Hi @chambm I don't suppose you have had a chance to look at the 'numeric tolerance' issue?

If you want to point me towards the code to do this, I can take a look.

chambm:
I worked on it yesterday and I'm almost done I think but we've had an issue with AWS so our TC and S3 hosting is not working. Gotta put out that fire today.

chambm:
I added a test case that uses ddaProcessing but it doesn't create any merged spectra for the file we have. Do you think you can get a super small test .raw that has some spectra that need to be merged?

pete-reay-waters:
I added such a raw folder (a very small DDA one) and a test. I also rebased against `pwiz\master`

pete-reay-waters:
Oh, hope rebasing hasn't broken something, the build is failing again

chambm:
It's failing to identify the new RAW file on Linux because the FUNC files are lowercase (func*.dat). From what I can tell this only happens when it's the SDK writing (or rewriting) the RAW functions. Coming straight off the instruments, it is always uppercase. Can you rename them to uppercase and I'll make the func detection more robust later?

pete-reay-waters:
A I see - so it was me breaking the build! Have renamed those files there

chambm:
Hah have to update the reference mzML too now :)

minimalDDA.mzML
```diff
@@ -0,0 +1,391 @@
...
+          </binaryDataArrayList>
+        </chromatogram>
+      </chromatogramList>
+    </run>
+  </mzML>
+  <indexList count="2">
+    <index name="spectrum">
+      <offset idRef="function=1 process=0 scan=1">6412</offset>
+      <offset idRef="function=1 process=0 scan=2">9032</offset>
+      <offset idRef="function=2 process=0 scan=1 merged=1-2">11652</offset>
```
chambm, line 380: Oops, actually the id should start with merged=, not end with it. And it should be equal to the index of the scan (so it doesn't add information, but that's the convention in the mzML standard). So the 1-2 or 1-3 part should go in the scan= field.

pwiz/data/vendor_readers/Waters/Reader_Waters_Test.data/ATEHLSTLSEK_profile-centroid.mzML
```diff
@@ -102,7 +102,7 @@
         <cvParam cvRef="MS" accession="MS:1000127" name="centroid spectrum" value=""/>
         <cvParam cvRef="MS" accession="MS:1000504" name="base peak m/z" value="341.017578"/>
         <cvParam cvRef="MS" accession="MS:1000505" name="base peak intensity" value="2.36912e05"/>
-        <cvParam cvRef="MS" accession="MS:1000285" name="total ion current" value="1.615975e06"/>
+        <cvParam cvRef="MS" accession="MS:1000285" name="total ion current" value="1.61599e06"/>
         <scanList count="1">
```
chambm, line 106: OK, so I hadn't realized that the SDK update changes the peak values for non-ddaProcessing data. I thought it was only changing the TIC metadata. What is the reason for the peaks changing?

minimalDDA.mzML
pete-reay-waters: I tried putting the range in `scan=` but it appeared to break SeeMS, and I was also worried about other consuming applications which expected a single number too.

Do we need to fix up SeeMS too?

pwiz/data/vendor_readers/Waters/Reader_Waters_Test.data/ATEHLSTLSEK_profile-centroid.mzML
pete-reay-waters: Ah ok I think we discussed this when I updated the test results - my belief is it's probably down to a different version of the C++ runtime, giving slightly different floating point calculations.

This was the thing where it now passes exactly on my machine, and your machine, but was different on the build server.

chambm: But you had to regenerate the test files to get them to pass, right? That should mean the change comes from the SDK changes, not the CRT (I generated the old test files on my own PC, not on a TeamCity AWS agent).

minimalDDA.mzML
chambm: I considered that might be an issue but didn't try it. I thought maybe we should change it to scans= and now I think we have to. The merged= at the beginning should have signified to SeeMS that it wasn't a regular nativeID it could parse though...

pete-reay-waters: Ah ok, maybe I didn't have the `merged=` at the beginning.

So what should the id be, exactly? I'm guessing

```
merged=4 scans=4-7 function=2 process=0
```

pwiz/data/vendor_readers/Waters/Reader_Waters_Test.data/ATEHLSTLSEK_profile-centroid.mzML
chambm: This conversation is anchored right above one of the base64 arrays that changed (I should have anchored it below, sorry). It's the `ATEHLSTLSEK_profile-centroid.mzML` file. It seems to be the only one that changed. Did something in the MassLynx SDK peak picker change?

pete-reay-waters: Hi Matt, just to let you know I'm still investigating this. Looking at the data that's changed, the m/z and intensities are similar, but with slight differences, and there's one spectrum with a couple of missing (fairly low intensity) points.

minimalDDA.mzML
pete-reay-waters: Ok, I fixed that, and SeeMS seems *reasonably* happy. It will load the file, but doesn't show anything in the Function, Process or Scan column. You can see all the info in the id column.

chambm: That is expected. Adding extra fields after merged= is a fairly new development and not really the established convention.

pete-reay-waters:
Hi @chambm Happy New Year! It looks like we've addressed all the issues we had. The differences in centroiding have been fixed so that the results haven't changed from the last release.

Looks like there's one check not complete, but I can't see details of that.
image

Do you think we are in a position to approve/merge now?
Thanks!

chambm:
I'll merge this to master after doing some integration testing with Skyline.

chambm:
I squashed this into an integration branch which we perftested for a few days. Should be ready to merge now. https://github.com/ProteoWizard/pwiz/pull/2458

PR #2397, opened 2022-11-24
Added an "Export CSV" button to the SkylineTester Run Stats tab, creates a more detailed look at performance for use in Excel etc…

bspratt:
Also, allows Summary.log to contain entries for logs that don't originate from the local machine, to enable performance comparison across systems


nickshulman (approved):
Looks good to me.

PR #2398, opened 2022-11-28
* fixed crash in BinaryData when doing swap() on an empty array

chambm:
* fixed crash in BinaryData when doing swap() on an empty array (reported by Joerg)
* 
Fixes issue from https://skyline.ms/announcements/home/support/thread.view?rowId=57502


nickshulman (approved):
Looks good to me.
I think you could simplify the code a bit if you always allocated a new managed array even if the size had not changed. In theory, allocating short lived managed arrays is fast enough that it's rarely worthwhile putting effort into deciding whether you can reuse the array you already have.

PR #2400, opened 2022-11-29
SimpleGridViewDriver fixes

kaipot:
Fixes some bugs with grid view, e.g. pressing delete/backspace when a cell is selected where the underlying property can't be set to `string.Empty`


nickshulman:
The usual way to deal with errors that happen when trying to convert what the user typed in is to handle the "DataError" event on the DataGridView.
If you don't handle that event, then the user sees a message box that looks like this:
image
But if you do handle the event, then you can display a different message.

There are a few examples in the code of places where we handle the DataError event such as EditNoteDlg or EditLinkedPeptidesDlg.

If you were handling the DataError event, would that be simpler than having to call "SetRequiredColumns"?

PR #2405, opened 2022-12-05
Add an assert which makes sure that Skyline.Document can always round-trip to XML during functional tests

nickshulman:
There are two new virtual properties on AbstractFunctionalTest:
```
public virtual bool VerifyDocumentSerializabilityAtEndOfTest { return true; }
public virtual bool ContinuouslyVerifyDocumentSerialization { return false; }
```
My hope is that one day we could make it so that those properties are both true for all tests, but, for now, I worry that "ContinuouslyVerifyDocumentSerialization" would slow down the entire test run too much, and there are two tests which fail if "VerifyDocumentSerializabilityAtEndOfTest" returns true.

When "VerifyDocumentSerializabilityAtEndOfTest", AssertEx.Serializable(SkylineWindow.Document)" gets called at the end of the test.
When ContinuouslyVerifyDocumentSerialization is true, the class "DocumentSerializableVerifier" listens for changes to SkylineWindow.Document and calls AssertEx.Serializable every time the document change.

In order to meaningfully compare the active document with one that has just been read from XML, I needed to add a method "SrmDocument.UnloadDocument" which puts all of the things which had to be read from other files (e.g. ChromatogramCache, Library, BackgroundProteome) back in the state that they would have been in right after the SrmDocument was deserialized.

The most important things that this assert is designed to catch are cases where the user might save their document and then be unable to open it after they exit Skyline. We have had a few of these in the wild, such as when "DocumentReader.ValidateSerializedVsCalculatedProductMz" tells them that the m/z that was written to the XML file was not what it should have been, or when the user is told that one of their peptides has a modification which is not in PeptideSettings.Modifications.

There are other less important things like places where the XML does not match our .xsd file, or where the original document has a string whose value is "" but when we round-trip it to XML that value changes to null.


nickshulman:
Moved to my own fork: https://github.com/nickshulman/pwiz/pull/1

PR #2412, opened 2022-12-06
Skyline: Fix intermittent failure in TestPasteTransitionListAssociate…

brendanx67:
…Proteins related to form visibility


bspratt (approved):
Looks good. 

I find 16 occurrences of the pattern **RunUI(() => blah.OkDialog());** in our code, I'll see about adding to our custom code inspection, with a message that these should probably be replaced with **OkDialog(blah, blah.OkDialog);**

PR #2422, opened 2022-12-08
* added clearing of z:\temp to tc-perftests.bat

chambm:
(I'm not sure why this would be different between the Parag AWS AMI and the MacCoss one, but it's what I noticed taking up space on Z: when I logged in.)

### DIFF
--- a/pwiz_tools/Bumbershoot/idpicker/IDPicker.csproj
+++ b/pwiz_tools/Bumbershoot/idpicker/IDPicker.csproj
@@ -717,12 +717,13 @@
   </Target>
   -->
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


brendanx67:
What was in Z:\temp? Maybe we are leaving stuff in temp or even using temp
accidentally.

chambm:
The OS temp dir is set there rather than to EBS volume. I can't remember if I also did that on the Parag AWS AMI though. MSAmanda does use TEMP, but of course lots of other OS stuff does as well.

With the latest AMI update, I removed VS2017 and it looks like that broke my IDPicker build, ugh.

PR #2424, opened 2022-12-09
Handle exceptions in ImportTransitionListColumnSelectDlg.CheckForErrors

nickshulman:
Fixed "Unhandled Error" that could happen doing Import Transition List if .skyd file is modified (reported anonymously on Exception Web)

### DIFF
--- a/pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
+++ b/pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
@@ -1295,8 +1295,10 @@ namespace pwiz.Skyline.FileUI
                 }
             }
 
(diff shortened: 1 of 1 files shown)


bspratt (commented):
Looks sensible, though I have a notion that we usually try to avoid these catch-all approaches. I'll defer to Brendan on that.

brendanx67:
On the other hand, it looks so generic that maybe we ought to have a property on LongWaitDlg that causes it to handle exceptions in this way? Nick has already added ExceptionUtil.IsProgrammingDefect() as a cover for my general unease with handling this generic, which is that it would otherwise show messages to the user about NREs and IndexOutOfBoundsExceptions, and the like.

brendanx67:
We are still unlikely to find new cases of programming errors that we are accidentally showing to our users, since they are much less likely to report confusing error messages than they are to post to the exception web when an unexpected exception escapes proper handling. Just noting that. I know it has been difficult to maintain the ideal that I imagined starting this project.

PR #2426, opened 2022-12-09
Make it so that "RunDlg" waits until the action which showed the dialog returns

nickshulman:
This change makes it so that "RunDlg" waits until the action that showed the dialog has actually returned. For the "SkylineWindow.ShowPeptideSettingsUi", this means that it will wait until the dialog has finished modifying the document.

Other changes:
1. Add method "RunLongDlg" which can be used for dialogs that show other dialogs.
2. Make the "act" argument to RunDlg required, and change the places that were passing in null to call new method "ShowAndCancelDlg" instead.
3. Add method "ShowAndDismissDlg" for places where the test can no longer use RunDlg because the action which shows the first dialog does not actually return until much later in the test.
4. Remove "[InstantHandle]" annotation on parameter in "ShowDialog", since the action that it is passed really does not finish executing before ShowDialog returns.

I hope this makes things easier to understand. In theory, "RunLongDlg" would make the code in tests like "MetadataRuleTest" easier to follow so that you cannot accidentally interact with a dialog that has been closed because the variable is out of scope.


bspratt:
So this is an alternate approach to my PR [https://github.com/ProteoWizard/pwiz/pull/2425], yes?

nickshulman:
> So this is an alternate approach to my PR [https://github.com/[/pull/2425](https://github.com/ProteoWizard/pwiz/pull/2425)], yes?

No, your code inspection is still important.
It's still totally valid to do "ShowDialog" and then later "OkDialog".

Most dialogs have finished their processing by the time "OkDialog" returns, since it's the dialog itself which does the work.
In the case of PeptideSettingsUi, all the work happens in the function which showed the dialog, so, for that, people have to include a "WaitForDocumentChange", or, now, they could use "RunDlg" instead.

PR #2429, opened 2022-12-10
ReSharper Live Template to insert a property and a ChangeXxx method at the same time

nickshulman:
This is the first Live Template I have written, so I might not be taking advantage of the available features properly at all.
When the cursor is positioned within a class definition, you can press "Ctrl + J" ("Insert Live Template") and see a list of things that can be inserted there:
image
The option that I have added is that one called "ChangeableProperty".
When you choose it, some code gets inserted and the cursor is first positioned where you can fill in the type of the new property:
image
image
After you have finished typing the type name, you can press Tab twice to move on to filling in the property name:
image
And after you hit Enter, you have added a property named Xxx and a ChangeXxx method that goes with it:
image
Here's what the template definition looks like:
image

Did I do this right?


brendanx67:
Ha! I had an active template like this when I was writing the initial immutable objects. It was pretty handy, but I lost track of it in the upgrade from VS 2008 to VS 2013. Nice to see something for this returning to the project.

bspratt:
Let's figure out how to advertise this more widely, and permanently. Seems quite useful but I only heard about it when perusing recent commit logs.

PR #2431, opened 2022-12-12
Add property sheet to the Library Match graph.

rita-gwen:


pwiz_tools/Skyline/Model/Lib/Library.cs
```diff
@@ -34,13 +35,15 @@
...
 using pwiz.Skyline.Model.DocSettings.Extensions;
+using pwiz.Skyline.Model.Hibernate;
 using pwiz.Skyline.Model.Irt;
 using pwiz.Skyline.Model.Lib.ChromLib;
 using pwiz.Skyline.Model.Lib.Midas;
 using pwiz.Skyline.Model.Prosit;
 using pwiz.Skyline.Model.Results;
 using pwiz.Skyline.Model.RetentionTimes;
 using pwiz.Skyline.Properties;
+using pwiz.Skyline.SettingsUI;
```
nickshulman, line 46: This "using" statement is causing the CodeInspectionTest failure.
That test expects there to be exactly 3 places in the model code where it is using something from a UI namespace.
This particular line ends up causing the number of violations to be 4, which causes the test failure.

brendanx67: Any idea why this file is pulling in that "using"? I guess maybe the ViewLibraryPepInfo.

pwiz_tools/Skyline/Model/Lib/Library.cs
```diff
@@ -2508,6 +2511,100 @@ public string FileName
         public double? RetentionTime { get; set; }
         public IonMobilityAndCCS IonMobilityInfo { get; private set; }
         public string Protein { get; private set; } // Also used as Molecule List Name for small molecules
+
```
brendanx67, line 2514: Avoid adding this many consecutive empty lines. Most of our code using single empty line spacing.

pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs
```diff
@@ -57,6 +57,7 @@
 using static pwiz.Skyline.Model.Lib.BiblioSpecLiteLibrary;
 
 
+
```
brendanx67, line 60: Unnecessary empty line.

pwiz_tools/Skyline/Skyline.csproj
```diff
@@ -618,6 +618,7 @@
     <Compile Include="Model\Lib\LibKeyIndex.cs" />
     <Compile Include="Model\Lib\SpectralLibraryExporter.cs" />
     <Compile Include="Model\Lib\SpectrumRanker.cs" />
+    <Compile Include="Model\Lib\ViewLibraryPepInfo.cs" />
```
brendanx67, line 621: Put this back and leave this file unchanged.

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -1132,7 +1132,16 @@ void GraphSpectrum.IStateProvider.BuildSpectrumMenu(bool isProteomic, ZedGraphCo
 
             // Insert skyline specific menus
             var set = Settings.Default;
-            var control = menuStrip.SourceControl.Parent.Parent as IMzScalePlot;
+
+            Func<Control, IMzScalePlot> traverseParents = null;
```
brendanx67, line 1136: Seems like this could be done with a loop. Using a function pointer and even starting with = null and then assigning it seems odd. Best not to leave other developers scratching their heads wondering, "Why did they do that?"

brendanx67 (commented):
A few minor comments. Not yet a full review.

pwiz_tools/Skyline/Model/Lib/Library.cs
brendanx67: One way to make the messages clearer is to actually clean up all of these in the existing code and then report only new ones. I think you own a fair number of the remaining cases, Nick.

pwiz_tools/Skyline/SkylineGraphs.cs
brendanx67: I don't see that you are traversing a tree. Rather you are looping upward through a straight line of parents. Though they are involved in a tree. This is essentially traversing a linked list for your purposes.

Have a look at FormUtil.FindTopLevelOwner() or SequenceTree.GetNodeOfType(). Clearly, Nick and I prefer loops for parent pointer traversal. I don't really see a benefit to using recursion, and I am certain this code could be simpler without it.

pwiz_tools/Skyline/Model/Databinding/Entities/GlobalizedObject.cs
```diff
@@ -0,0 +1,191 @@
+﻿using System;
```
brendanx67, line 1: Needs license and owner.

pwiz_tools/Skyline/Model/Databinding/Entities/GlobalizedObjectResx.resx
```diff
@@ -0,0 +1,187 @@
...
+    <value>File</value>
+  </data>
+  <data name="Category_MatchInfo" xml:space="preserve">
+    <value>Search</value>
+  </data>
+  <data name="Category_PrecursorInfo" xml:space="preserve">
+    <value>Precursor</value>
+  </data>
+  <data name="CCS" xml:space="preserve">
+    <value>CCS</value>
```
brendanx67, line 113: Seems like a strange placement for spectrum properties text.

pwiz_tools/Skyline/Model/Lib/ViewLibraryPepInfo.cs
```diff
@@ -1,221 +1,373 @@
...
+                var precursor = Key.PrecursorMz.GetValueOrDefault();
+                var precursorMono = new TypedMass(precursor, MassType.Monoisotopic);
+                var precursorAverage = new TypedMass(precursor, MassType.Average);
+                var peptide = new Peptide(new CustomMolecule(precursorMono, precursorAverage, precursor.ToString(CultureInfo.CurrentCulture)));
+                transitionGroup = new TransitionGroupDocNode(new TransitionGroup(peptide, Adduct.EMPTY, IsotopeLabelType.light, true, null), null);
+                mods = new ExplicitMods(peptide, new ExplicitMod[0], new TypedExplicitModifications[0]);
+            }
+        }
+
+        // Computes each ModificationInfo for the given peptide and returns a
```
brendanx67, line 305: Use a &lt;summary&gt; tag for header comments.

pwiz_tools/Skyline/Model/Lib/ViewLibraryPepInfo.cs
```diff
@@ -1,221 +1,373 @@
...
+                sequence = peptideLibraryKey.ModifiedSequence;
+            }
+            else
+            {
+                return modList;
+            }
+            int iMod = -1;
+            for (int i = 0; i < sequence.Length; i++)
+            {
+                if (sequence[i] != '[')
```
brendanx67, line 322: Seems hard to believe there is no other code in the product that does this already.

pwiz_tools/Skyline/Properties/Settings.cs
```diff
@@ -1289,6 +1289,7 @@ public NormalizeOption AreaNormalizeOption
                 NormalizeOptionValue = value.PersistedName;
             }
         }
+
```
brendanx67, line 1292: Seems like this could be reverted.

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -1108,6 +1108,12 @@ private void observedMzValuesContextMenuItem_Click(object sender, EventArgs e)
             ToggleObservedMzValues();
         }
 
+        private void showSpectrumPropertiesContextMenuItem_Click(object sender, EventArgs e)
+        {
+            if(_graphSpectrum != null && _graphSpectrum.Visible)
```
brendanx67, line 1113: Space after if

pwiz_tools/Skyline/TestTutorial/TargetedMSMSTutorialTest.cs
```diff
@@ -1133,7 +1133,7 @@ private void TestPropertySheet()
             Assert.IsNotNull(propertyGrid);
 
             // Checks the number of properties displayed is the expected number
-            int expectedPropCount = isSmallMolecules ? 7 : 11;
+            int expectedPropCount = isSmallMolecules ? 6 : 10;
```
brendanx67, line 1136: Why did this change?

pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
```diff
@@ -1396,6 +1413,19 @@ private void DoUpdate()
             }
         }
 
+        public bool ShowPropertiesSheet {
```
brendanx67, line 1416: Put brace on a separate line.

brendanx67 (changes requested):
Minor changes requested. You do also need to add something that tests the new property sheet in the spectrum graph. Passing the ViewLibraryDlg tests written by Aaron and me is not enough.

pwiz_tools/Skyline/Model/Databinding/Entities/GlobalizedObjectResx.resx
rita-gwen: This is where it was originally. Where should they be placed? These strings will be shared between 3 different forms.

brendanx67: I think, on a SpectrumProperties.resx file. GlobalizedObject does not own these properties. It could be subclassed by any other object for which we want to display properties in the future. We want the property name text to live with the objects they describe not accumulate with the base class.

rita-gwen: That means that GlobalizedObject will have to look up the child class resource file at runtime. It's a bit clunky, IMHO, but doable.

brendanx67:
Virtualization? Polymorphism?

pwiz_tools/Skyline/Model/Lib/ViewLibraryPepInfo.cs
rita-gwen: There probably wasn't at the time this code has been written 12 years ago.

pwiz_tools/Skyline/TestTutorial/TargetedMSMSTutorialTest.cs
rita-gwen: Retention time property was missing in action. I fixed it now.

pwiz_tools/Skyline/TestTutorial/TargetedMSMSTutorialTest.cs
```diff
@@ -1188,6 +1199,76 @@ private void TestPropertySheet()
             OkDialog(dlg, () => dlg.Close());
         }
 
+        private void TestLibraryMatchPropertySheet()
+        {
+            var isSmallMolecules = AsSmallMoleculesTestMode != RefinementSettings.ConvertToSmallMoleculesMode.none;
+
+            var expectedPropertiesJson =
+                    "{  \"IdFileName\": \"klc_20100329v_Protea_Peptide_Curve_20fmol_uL_tech1.perc.xml\",  \"FileName\": \"klc_20100329v_Protea_Peptide_Curve_20fmol_uL_tech1.ms2\",  \"FilePath\": \"C:\\\\Users\\\\RitaCh\\\\Workspaces\\\\ProteoWiz\\\\pwiz2\\\\pwiz_tools\\\\Skyline\\\\TestResults\\\\TargetedMSMSMzml_2\\\\TargetedMSMSMzml\\\\Low Res\\\\search\\\\klc_20100329v_Protea_Peptide_Curve_20fmol_uL_tech1.ms2\",  \"LibraryName\": \"BSA_Protea_label_free_meth3\",  \"PrecursorMz\": \"582.319\",  \"Charge\": 2,  \"Label\": \"light\",  \"RetentionTime\": \"46.81\", \"CCS\": null,  \"IonMobility\": null,  \"SpecIdInFile\": \"7901\",  \"Score\": 0.0,  \"ScoreType\": \"Percolator q-value\",  \"SpectrumCount\": 118}";
```
brendanx67, line 1207: This is kind of a mess. Please use consistent ordering between the two cases. Ideally a format that is not quite so heavy on backslashes, if possible, and it looks like this would need a record mode in case anything changes. Look at other tests that do this. I am not clear on why small molecule mode does not preserve things like FileName. Please at least clarify that with a comment.

rita-gwen: Fixed the test data format. The missing file name was a bug, I fixed that too. Waiting for the tests to pass before merging.

pwiz_tools/Skyline/TestTutorial/TargetedMSMSTutorialTest.cs
```diff
@@ -1188,14 +1224,109 @@ private void TestPropertySheet()
             OkDialog(dlg, () => dlg.Close());
         }
 
+        private void TestLibraryMatchPropertySheet()
+        {
+            var isSmallMolecules = AsSmallMoleculesTestMode != RefinementSettings.ConvertToSmallMoleculesMode.none;
+
+            var expectedPropertiesJson =
```
brendanx67, line 1231: I am still left feeling that in our code this should be a Dictionary and we should use NewtonSoft.Json which we already use elsewhere to serialize it to text.

https://www.newtonsoft.com/json/help/html/serializedictionary.htm

brendanx67: Another option would be to use a single string with an insertion point for IdFileName that is either empty or the .perc.xml depending on small molecule status. But, I think I still like the dictionary option better with that value added to the dictionary only if it is not small molecules.

rita-gwen: I can change it to Dictionary, no problem, but I'm not sure about string insertion. IdFileName is not a single difference between the two tests. It's different precursor and different spectrum count.

brendanx67:
I have this on my machine. Hold off

brendanx67 (approved):

PR #2435, opened 2022-12-19
Improvements for importing .MSP library files for GCMS use:

bspratt:
- Allow for entries that give no hint as to parent mass - just use a dummy value of 99 in those cases.
- Use "MW:" value when available
- More tolerant of extra and/or missing spaces in line formatting
- Support "( )" delimiter for fragment peak lists


brendanx67:
Can you provide a bit more info about the source of this and the test case(s)?

bspratt:
Reported by user Brynn

[https://skyline.ms/announcements/home/support/thread.view?rowId=57741](https://skyline.ms/announcements/home/support/thread.view?rowId=57741)

bspratt:
(Sorry, that should have been in initial commit message)

bspratt:
@brendanx67 @nickshulman I'll merge this if one of you would give it the blessing

brendanx67 (commented):
I am really not a fan of the entire 200-line (and growing) parser loop. I could imagine it decreasing a lot of the complexity with a parser class that contains a lookup table (test, RegEx, MatchFunction(line, match)). It feels like this code has really outgrown its original implementation and needs a re-architecture.

That said, this change appears relatively in-line with the rest of the code, and you have added tests and I am sure pass the existing tests. So, I am okay with merging this, with the caveat that a re-architecture at some point seems desirable, much as I eventually had to rearchitect CommandLine and CommandArgs.

bspratt:
Fair point. If you like I can at least bust it up into a few function calls without a lot of effort - something like
                  
                    // Process until the start of the peaks
                    while ((line = reader.ReadLine()) != null)
                    {
                        lineCount++;

                        readChars += line.Length;
                        line = HandleMzVaultLineVariant(line, out var isMzVault);

                        if (isPeptide && line.StartsWith(SYNON, StringComparison.InvariantCultureIgnoreCase)) // Case insensitive
                        {
                            isPeptide = false;
                        }

                        ParseRT()
                        ParseSmallMolecule()
                        ParseComment()
                        ParseNumPeaks()
                        ParseMZVaultDetails()
                        if (line.StartsWith(@"_EOF_", StringComparison.InvariantCultureIgnoreCase)) // Case insensitive
                            ThrowIOException(lineCount, Resources.NistLibraryBase_CreateCache_Unexpected_end_of_file);
                        else if (line.StartsWith(NAME, StringComparison.InvariantCultureIgnoreCase)) // Case insensitive
                            break;
                    } // End parser loop

brendanx67:
I think you would want if statements and "continue"s and you would at least
need to pass in the line and ref the variable getting set. It does seem to
beg for a lookup table implementation.

bspratt:
Easy enough to add a comment.

PR #2437, opened 2022-12-19
Tidy up the zh-CHS translation for "irank" - the translators apparent…

brendanx67:
…ly looked at the comment in

  <data name="TransitionTreeNode_GetLabel_irank__0__" xml:space="preserve">
    <value>irank {0}</value>
    <comment>Isotope rank for transition node</comment>
  </data>

and came up with
    <value>针对离子对节点的同位素排名 {0}</value>

which is literally "Isotope rank for transition node {0}" where what we really want is
    <value>i排名 {0}</value>

which is literally "irank {0}" (note the Roman "i" in "i排名")

One question in my mind - in the Japanese version of this, there is no blank between "irank" and the value. Is that also appropriate in Chinese?


brendanx67:
I do think we want to make this change. The word "irank" only appears in our tutorials in the "PRM" tutorial, where it is accompanied by a screenshot showing the text that gets highlighted in this PR. Strangely the word "irank" does not get ranslated in either the Chinese or the Japanese tutorial text. I will update this PR to include an update to the tutorials.

chambm:
In general Chinese (and I think Japanese as well) doesn't have whitespace between words. So I wouldn't expect them to have whitespace between a label and its value. Korean does have whitespace between words.

brendanx67:
It is a number and our translators have control of this. If they chose to leave that space, I am inclined to leave it for now. If you look at the translation, there are places where they have spaces around insertion points and places where they do not. It may be that all spaces are mistakes, like the use of a three-period ellipsis versus the single-character Asian ellipsis. Always a mistake when the former is used, and translators do make this mistake. In this case, though, we would need to ask our translators to weigh in.

brendanx67:
It is not an easy case, since we are going to include the Roman i and the text itself will be enclosed in Anglo/Euro parentheses, instead of Asian parentheses.

chambm:
I would ask an application-knowledgeable Chinese Skyline user to confirm whether putting the i- suffix in front of a Chinese word makes sense to them, especially with the proportional font making the i barely visible. That looks weird to me. They may prefer a Chinese character or word for intensity there.

Chinese don't _typically_ put whitespace in their writing, but if someone did it for them, AFAIK they'd read it with out noticing any difference. So I suspect that whether the whitespace is there or not is not a matter of being wrong or right, but just not mattering at all.

brendanx67:
It is pretty subtle where this shows up in the UI.
image
To all users now there are two ranking categories distinguished only by an "i". They are also positionally distinct and one is all "precursor" labeled. So, it is a very learnable pattern.
Japanese:
image
English:
image

brendanx67:
Our localization contact agreed to this change. Though, it is not clear how much she consulted native Chinese localizers in doing so.

PR #2443, opened 2022-12-28
Fixed AssumptionException for protein group metadata

chambm:
- fixed AssumptionException for protein group metadata in WritePeptideGroupXml (reported by Rich)


nickshulman (approved):
Looks good to me.
Do you have any idea why more people haven't been running into this bug? Was there something unusual about Rich's document?

chambm:
I think it's because a non-metadata-friendly protein (cRAP contaminant accession) was grouped with metadata friendly proteins (trembl/swissprot).

chambm:
Should this be cherry-picked to release?

nickshulman:
> Should this be cherry-picked to release?

Yes. This fix should probably be cherry picked to release. It's possible we won't ever release another 22.2 patch but if we do release one it would be good for this fix to be in it.

PR #2444, opened 2022-12-29
Fix Result File Rules when importing via commandline

nickshulman:
Fixed applying Result File Rules when importing results using the commandline (reported by Roman)


pwiz_tools/Skyline/TestFunctional/MetadataRuleTest.cs
```diff
@@ -92,7 +92,7 @@ protected override void DoTest()
                         });
                 }, metadataRuleEditor => metadataRuleEditor.OkDialog());
             }, documentSettingsDlg => documentSettingsDlg.OkDialog());
-
+            PauseTest();
```
brendanx67, line 95: Looks like you actually want to revert the changes to this file.

brendanx67 (approved):
Makes sense. Thanks for adding the test. Remember to revert the changes to the test where you left PauseTest. Though, I think TeamCity code inspection should catch this.

PR #2446, opened 2023-01-02
Skyline: Test improvements while working on parallel testing

brendanx67:
- Make SafeDeleteTest safe for parallel testing by saving to the TestResults folder
- Fix generating TestRunner.exe call for parallel testing to work with debug builds
- Add some keyboard mnemonics to SkylineTester for quicker runs without using the mouse


brendanx67:
Thought I would separate these simple improvements from the bulk of my work, which has been on the ConsoleMultiReplicateImportTest, a much more complicated problem with a solution that will clearly risk a broader impact.

chambm (approved):

PR #2448, opened 2023-01-06
Change "AaantivirusTestExclusion" so that it just outputs a warning instead of failing if virus scanning happens to be enabled on the folder.

nickshulman:
This change makes it so that AaantivirusTestExclusion does not fail if virus scanning is turned on (it just writes some text which would be visible in the test log if anyone wanted to look).
The motivation for this change is to allow developers to turn off virus scanning on their nightly folder in case they are interested in discovering whether Skyline ever has trouble saving files on computers with virus protection.


nickshulman:
Closing because I think the name of this branch needs to be "integration".

PR #2450, opened 2023-01-08
Custom filters on Precursors in the Targets tree

nickshulman:
This change allows users to apply custom filters to the transitions in the Targets tree:
image
This brings up the Edit Spectrum Filter dialog where you can specify which spectra should be used for chromatogram extraction:
image
In this demo, I and filtering only for the spectra whose PresetScanConfiguration is "2" (this dataset is from Birgit's CID/EID something or other dataset, but the only distinguishing attribute of the spectra I could find to filter on was "PresetScanConfiguration").

Here I have three different TransitionGroupDocNode's. The first one has no filter, and the other two are filtering for different types of spectra. The first chromatogram is jagged as a result:
image


brendanx67:
Just a note that names have to be put into a RESX file and made more readable (spaces at least) and localizable.

brendanx67:
I don't really understand the Spectrum Grid. It may be helpful for debugging, but it feels like it should at least be buried inside the Edit Spectrum Filter form. Also, spectrum filters should be something you can define in Document Settings (or at least Transition Settings) so that they can impact all of the Peptides/Molecules in a document. The way isotope modifications or precursor charge states do.

Brian Pratt and I had an email discussion going about how this might fit into Document Settings, since a document-wide spectrum filter could impact parameters in the Transition Settings (e.g. fragment ions to filter for, full-scan settings for filter widths, etc.)

nickshulman:
> I don't really understand the Spectrum Grid. It may be helpful for debugging, but it feels like it should at least be buried inside the Edit Spectrum Filter form. Also, spectrum filters should be something you can define in Document Settings (or at least Transition Settings) so that they can impact all of the Peptides/Molecules in a document. The way isotope modifications or precursor charge states do.
> 
> Brian Pratt and I had an email discussion going about how this might fit into Document Settings, since a document-wide spectrum filter could impact parameters in the Transition Settings (e.g. fragment ions to filter for, full-scan settings for filter widths, etc.)

Document-level spectrum filters sounds like a good idea. I think maybe it should be wedged onto the "Instrument" tab in Transition Settings.

Some of the things that you might filter on you would definitely want to share between lots of molecules or peptides in the document. So, people would want to be able to define document-level filters as well as precursor-level filters.

It sounds like we might also want something on the "Transition Settings > Library" tab which says what filters from spectral libraries should be applied.

PR #2452, opened 2023-01-10
Fixed NRE when bringing up context menu on blank spectrum graph

kaipot:


nickshulman (approved):
Looks good to me. I believe this exception came from this support request and the way to reproduce it is to go to Tools > Options > Prosit and blank out the Intensity Model and then right-click on the Library Match window and choose "Prosit".

https://skyline.ms/announcements/home/support/thread.view?rowId=57840

kaipot:
> Looks good to me. I believe this exception came from this support request and the way to reproduce it is to go to Tools > Options > Prosit and blank out the Intensity Model and then right-click on the Library Match window and choose "Prosit".
> 
> https://skyline.ms/announcements/home/support/thread.view?rowId=57840

You're right, it looks like that's the one. I was also able to reproduce it by right-clicking the Library Match window with something selected that had no matching spectrum.

PR #2454, opened 2023-01-10
Tweaks for more stable parallel tests:

bspratt:
I surveyed the current tests with the NoParallelTest attribute. For some tests, I could not determine a reason for exclusion from parallel test so I just removed the attribute. For the others, I updated the attribute with a (newly required) argument intended to explain why a test is not suitable for parallel tests, e.g. "NoParallelTesting(TestExclusionReason.VENDOR_FILE_LOCKING)]"

Some tests that were failing in parallel test just suffered from latent race conditions exposed by the slower VMs, tweaks like WaitForCondition(Equals()) rather than a bare Equals() deal with that conveniently.

I also introduced a new custom attribute "NoNightlyTesting" which is used to exclude tests from the nightly test regimen (formerly these tests just did a no-op when running offscreen, but that precluded any local offscreen testing). The attribute also requires an explanation argument in the same style as the custom attribute NoParallelTesting.

When a test fails on a worker instance, that worker information is now in the error log.

Cleaned up several cases with file lock problems where a file was being accessed in areas shared by parallel test workers, especially the Skyline install directory.

The "Offscreen" button in SkylineTester now greys out when parallel test is selected, since all tests are offscreen in parallel mode.

Also updated the onscreen hint for parallel tests in SkylineTester to read: Be sure to check BOTH public and private options if prompted to "Allow TestRunner to communicate on these networks". See https://skyline.ms/wiki/home/development/page.view?name=Troubleshooting_parallel_mode for troubleshooting tips.


pwiz_tools/Skyline/SkylineTester/SkylineTesterWindow.cs
```diff
@@ -1939,6 +1939,11 @@ private void buttonRunStatsExportCSV_Click(object sender, EventArgs e)
             _tabRunStats.ExportCSV();
         }
 
+        private void runSerial_CheckedChanged(object sender, EventArgs e)
+        {
+            Offscreen.Enabled = runSerial.Checked; // Everything happens offscreen in parallel tests, so don't offer the option
```
brendanx67, line 1944: Could we make this disabled and checked to make it clear that you don't have an option, but the only option is offscreen?

pwiz_tools/Skyline/TestFunctional/RetentionTimeFilterTest.cs
```diff
@@ -44,7 +44,6 @@ public class RetentionTimeFilterTest : AbstractFunctionalTest
     {
         private readonly string extension = ExtensionTestContext.ExtMz5;
         
-        [TestMethod, NoParallelTesting]
```
brendanx67, line 47: Looks like this is removing a test. Why?

pwiz_tools/Skyline/TestFunctional/RunToRunRegressionTest.cs
```diff
@@ -33,7 +33,6 @@ public class RunToRunRegressionTest : AbstractFunctionalTest
     {
         private const int REPLICATES = 5;
 
-        [TestMethod, NoParallelTesting]
```
brendanx67, line 36: Looks like this is removing a test. Why?

pwiz_tools/Skyline/TestRunner/Program.cs
```diff
@@ -669,7 +669,7 @@ private static string LaunchDockerWorker(int i, CommandLineArgs commandLineArgs,
             testRunnerExe = iRelative != -1
                 ? Path.Combine(@"c:\pwiz", testRunnerExe.Substring(iRelative))
                 : @"c:\pwiz\pwiz_tools\Skyline\bin\x64\Release\TestRunner.exe";
-            var testRunnerCmd = $@"{testRunnerExe} parallelmode=client showheader=0 results=c:\AlwaysUpCLT\TestResults log={testRunnerLog}";
+            var testRunnerCmd = $@"{testRunnerExe} parallelmode=client showheader=0 results=c:\AlwaysUpCLT\TestResults_{i} log={testRunnerLog}";
```
brendanx67, line 672: I still don't feel this should be necessary.  These folders should not be shared at all even without an identifying index... Unless the identifying index is more about knowing which client ran the test when looking at the logs. Though, it seems like that could be done more directly, since not all errors report a path.

pwiz_tools/Skyline/TestRunner/Program.cs
```diff
@@ -1113,6 +1114,26 @@ private static void TeamCityFinishTestSuite(CommandLineArgs commandLineArgs)
             bool clientMode = parallelMode == "client" || parallelMode == "server_worker";
             bool asNightly = offscreen && qualityMode;  // While it is possible to run quality off screen from the Quality tab, this is what we use to distinguish for treatment of perf tests
 
+            // If running Nightly tests, remove any flagged for exclusion by the NoNightlyTesting custom attribute
+            if (asNightly)
+            {
+                for (var t = testList.Count; t-- > 0;)
```
brendanx67, line 1120: Could be done with two fairly simple Linq expressions which I prefer to 14 lines of in-place for loops.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
```diff
@@ -212,6 +222,8 @@ private void SetTestDir(TestContext testContext, string resultsDir)
         {
             if (string.IsNullOrEmpty(resultsDir))
                 resultsDir = Path.Combine(GetProjectPath("TestResults"), "TestRunner results");
+            else if (IsParallelClient && resultsDir.Contains("TestResults_"))
```
brendanx67, line 225: Still not seeing the value.

pwiz_tools/Skyline/TestUtil/TestFilesDir.cs
```diff
@@ -254,7 +257,7 @@ public static void CheckForFileLocks(string path, bool useDeletion = false)
             if (useDeletion)
             {
                 RemoveReadonlyFlags(path);
-                Helpers.TryTwice(() => Directory.Delete(path, true));
+                Helpers.TryTwice(() => Directory.Delete(path, true),$@"Directory.Delete(""{path}"", true)");
```
brendanx67, line 260: Space after comma

pwiz_tools/Skyline/TestUtil/TestFilesDir.cs
```diff
@@ -266,23 +269,25 @@ public static void CheckForFileLocks(string path, bool useDeletion = false)
...
             {
-                Helpers.TryTwice(() => Directory.Move(guidName, path));
+                Helpers.TryTwice(() => Directory.Move(guidName, path), $@"Directory.Move(""{guidName}"",""{path}"")");
             }
             catch (IOException)
             {
+                Console.Write($@"# CheckForFileLocks failed to Directory.Move(""{guidName}"",(""{path}"")");
                 // Useful for debugging. Exception names file that is locked.
-                Helpers.TryTwice(() => Directory.Delete(guidName, true));
+                Helpers.TryTwice(() => Directory.Delete(guidName, true),$@"Directory.Delete(""{guidName}"", true)");
```
brendanx67, line 290: Space after comma.

brendanx67 (changes requested):
Mostly minor stuff. I still don't get why you create separate TestResults folders for Docker Containers. I don't think they are ever shared even without this.

pwiz_tools/Skyline/Test/SrmSettingsTest.cs
```diff
@@ -1185,109 +1186,112 @@ public void SerializeProteinAssociationSettingsTest()
...
-            var diff = SrmDocument.EqualsVerbose(oldDoc, newDoc);
-            if (diff != null)
-                Assert.Fail(diff);
-            Assert.AreEqual(newDoc.Settings.TransitionSettings.IonMobilityFiltering.IonMobilityLibrary.Name, "test");
-            Assert.AreEqual(currentDoc.Settings.TransitionSettings.IonMobilityFiltering.IonMobilityLibrary.Name, "test");
-            AuditLogList.IgnoreTestChecks = save;
-
+            // Make sure any generated .imsdb files are not in exe's directory - that interferes with parallel tests
+            TestContext.EnsureTestResultsDir();
+            using (var d = new CurrentDirectorySetter(TestContext.GetTestResultsPath()))
```
chambm, line 1194: You can avoid all this whitespace churn with a using statement:
```
using var d = new CurrentDirectorySetter(TestContext.GetTestResultsPath());
```

pwiz_tools/Skyline/TestRunner/Program.cs
chambm: Nice use of the "[goes to 0 operator](https://stackoverflow.com/questions/1642028/what-is-the-operator-in-c)" though! `--> 0`

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
chambm: He's using the TestResults_n directory to get the worker id. Seems potentially useful to me (especially using it as he is to say which worker is producing the error output) and I'm not seeing a way to get it from environment variables inside the workers. Do you think it would be better to have a dedicated TestRunner parameter for the worker name?

brendanx67: Seems like a kludge to me to have a useful value appended to a folder name where it is not necessary and then parse the path to get the value. Definitely having an explicit TestRunner argument for the value seems like a better route to me.

pwiz_tools/Skyline/SkylineTester/SkylineTesterWindow.cs
bspratt: No problem.

pwiz_tools/Skyline/TestFunctional/RetentionTimeFilterTest.cs
bspratt: Yikes, that was unintentional

pwiz_tools/Skyline/TestRunner/Program.cs
bspratt: It isn't necessary. It just makes for a more informative log.

brendanx67:
Perf test failures make it seem like TeamCity filtering may not be working in all cases.

pwiz_tools/Skyline/TestRunner/Program.cs
bspratt: Yeah, I just copied existing code which also shows my C++ roots. I can tighten that up, sure.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
bspratt: I thought about using another parameter there, yeah. Certainly would want to do that if the info wasn't otherwise encoded, but I do like the richer information in the log from having it in the test directory name so it just seemed like extra work.

brendanx67: Okay. Just add some comments explaining why you did what you did and that it is not actually necessary to keep tests from using the same files.

pwiz_tools/Skyline/Test/SrmSettingsTest.cs
bspratt: Thanks. I'm obviously unclear on the scoping rules around IDisposable.

pwiz_tools/Skyline/TestPerf/PerfImportAgilentSpectrumMillRampedIMSTest.cs
```diff
@@ -50,13 +50,13 @@ public class PerfImportAgilentSpectrumMillRampedIMSTest : AbstractFunctionalTest
     {
         private int _testCase;
 
-        [TestMethod]
```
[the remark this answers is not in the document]
bspratt: Interesting, I'll revisit this, thanks.

pwiz_tools/Skyline/TestFunctional/RunToRunRegressionTest.cs
bspratt: Error, thanks for catching it! Only meant to remove the parallel test prohibition.

brendanx67:
At least the TestDiaPasefFullDataset-en failure seemed to me like a case we
were trying not to run on TeamCity for a reason like this:

NoNightlyTesting(TestExclusionReason.EXCESSIVE_TIME)] // do not run full
filesets for nightly tests

I could be wrong, though.

bspratt:
> At least the TestDiaPasefFullDataset-en failure seemed to me like a case we were trying not to run on TeamCity for a reason like this: NoNightlyTesting(TestExclusionReason.EXCESSIVE_TIME)] // do not run full filesets for nightly tests I could be wrong, though.

Do we need another attribute for "no team city testing", or are the needs of TC and nightlies the same? That is, should TC runs just omit the same tagged tests as nightly?

chambm:
Regarding whether nightly and TC exclusions should be the same, I think TC is probably more restrictive than nightly because quite a few of those perf tests have huge data files to download and extract or the tests run a long time. This is the current skiplist:
AgilentSpectrumMillIMSImportTest
AllVsMz5OptimzeCeImportPerformanceTests
DecoysMatchTest
TestDiaTtofDiaUmpireTutorialFullFileset
TestDiaTtofFullSearchTutorial
TestDiaQeDiaUmpireTutorial
TestDiaQeDiaUmpireTutorialFullFileset
TestDiaQeFullSearchTutorial
TestPeakPerf
TestPerfTicArea
zzzNativeVsMz5_AbDiaChromatogramPerformanceTest
zzzNativeVsMz5_AbIdaChromatogramPerformanceTest
zzzNativeVsMz5_AgilentDdaChromatogramPerformanceTest
zzzNativeVsMz5_AgilentDiaChromatogramPerformanceTest
zzzNativeVsMz5_BrukerFullScanMS1filteringPerformanceTest
zzzNativeVsMz5_BrukerFullScanMSeDataPerformanceTest
zzzNativeVsMz5_BrukerFullScanSWATHDataPerformanceTest
zzzNativeVsMz5_ThermoDdaVChromatogramPerformanceTest
zzzNativeVsMz5_ThermoDiaChromatogramPerformanceTest
zzzNativeVsMz5_WatersChromatogramPerformanceTest

So if we do move away from the batch file to a single TestRunner execution, then yeah probably we need another attribute for annotating skipping on TeamCity.

chambm:
Yesterday I set perftests to run with a single TestRunner invocation (not using the standalone skiplist) and unsurprisingly it failed when trying to run the very long tests. Although I did forget to set offscreen=1. When I set that, more tests passed, but it still ultimately timed out. It's back to running using the script now.

brendanx67:
This seems to be creeping. Can we try to hold the scope and get it committed before adding more? I just run into the TestIonMobility failure and considered fixing it but then remembered this PR probably had the fix.

chambm:
Right. The perftests on TC can be handled in another PR.

PR #2455, opened 2023-01-10
PeptideProphet Proxl BiblioSpec support

nickshulman:
Enabled building BiblioSpec libraries from PeptideProphet .proxl.xml files (requested by Lindsey)


chambm (changes requested):
The mzML should be compressed. Other than that, looks good.

nickshulman:
> The mzML should be compressed. Other than that, looks good.



> The mzML should be compressed. Other than that, looks good.

Is there a standard set of commandline options that I should pass to msconvert.exe to create test files like this?
I usually use "--noindex" and then use a text editor to remove things like the folder name that the raw file was in.

PR #2456, opened 2023-01-11
Add optimization step to ChromKey/ChromTransition

kaipot:


pwiz_tools/Skyline/Model/Results/ChromatogramCache.cs
```diff
@@ -1230,6 +1231,91 @@ public IEnumerable<ChromKeyIndices> GetChromKeys(MsDataFileUri msDataFilePath)
             }
         }
 
+        private ChromatogramCache UpdateOptimizationSteps(SrmDocument doc)
```
nickshulman, line 1234: I think this function should only process MsDataFileUri's which belong to a ChromatogramSet with a non-null OptimizationFunction.

nickshulman (commented):
The code to save as an older format still needs to be written, right?
When saving as an older format, the ChromTransition product needs to be shifted by OptimizationStep times 0.01.
That conversion code would probably go in a new function that gets called by ChromatogramCache.OptimizeToPath.

pwiz_tools/Skyline/Model/Results/ChromHeaderInfo.cs
```diff
@@ -2677,7 +2656,7 @@ public static bool IsDerivative(this TransformChrom transformChrom)
     public class ChromatogramInfo
     {
         public const double OPTIMIZE_SHIFT_SIZE = 0.01;
-        private const double OPTIMIZE_SHIFT_THRESHOLD = 0.001;
+        private const double OPTIMIZE_SHIFT_THRESHOLD = 0.0015;
```
nickshulman, line 2659: Why was this change necessary? Is there a particular Skyline document whose optimization chromatograms did not have the expected product m/z's?

pwiz_tools/Skyline/Model/Results/ChromatogramCache.cs
```diff
@@ -1230,6 +1231,104 @@ public IEnumerable<ChromKeyIndices> GetChromKeys(MsDataFileUri msDataFilePath)
             }
         }
 
+        private ChromatogramCache UpdateOptimizationSteps(SrmDocument doc)
+        {
+            if (doc?.MeasuredResults?.Chromatograms == null || Version >= CacheFormatVersion.Seventeen)
```
nickshulman, line 1236: Can "doc" ever actually be null here? I couldn't find a code path that might be passing null in here, but I might have missed it.

pwiz_tools/Skyline/Model/Results/ChromatogramCache.cs
```diff
@@ -1230,6 +1231,104 @@ public IEnumerable<ChromKeyIndices> GetChromKeys(MsDataFileUri msDataFilePath)
             }
         }
 
+        private ChromatogramCache UpdateOptimizationSteps(SrmDocument doc)
+        {
+            if (doc?.MeasuredResults?.Chromatograms == null || Version >= CacheFormatVersion.Seventeen)
+                return this;
+
+            // Get the set of MsDataFileUris belonging to ChromatogramSets with optimization functions.
+            var optFiles = new HashSet<MsDataFileUri>();
```
nickshulman, line 1240: I would probably make an array of booleans so that you can quickly check whether ChromGroupHeaderInfo.FileIndex was optimized or not, instead of having to look a MsDataFileUri up in a HashSet.
There probably won't be enough ChromGroupHeaderInfo's for that to make a difference, though.

pwiz_tools/Skyline/Model/Results/ChromatogramCache.cs
```diff
@@ -1342,7 +1441,16 @@ public IEnumerable<ChromKeyIndices> GetChromKeys(MsDataFileUri msDataFilePath)
                         int start = lastEntry.StartTransitionIndex;
                         int end = start + lastEntry.NumTransitions;
                         for (int j = start; j < end; j++)
-                            listKeepTransitions.Add(_rawData.ChromTransitions[j]);
+                        {
+                            var chromTransition = _rawData.ChromTransitions[j];
+                            if (formatVersion < CacheFormatVersion.Seventeen && chromTransition.OptimizationStep != 0)
+                            {
+                                // Convert to old format with shifted m/z.
+                                chromTransition.Product += chromTransition.OptimizationStep * ChromatogramInfo.OPTIMIZE_SHIFT_SIZE;
```
nickshulman, line 1449: Could you make a method on ChromTransition that returns a new ChromTransition with the Product and Optimization Step set to the correct value.
It would be nice if ChromTransition could remain immutable so that people don't accidentally modify instances that don't belong to them.

nickshulman (commented):
This looks really good.

Do you have any idea why a couple of tests are failing in the Tutorial/Perf teamcity run? It looks like some of the numbers are slightly different, but I can't see what in this PR might have resulted in numbers being different.

pwiz_tools/Skyline/Model/Results/ChromHeaderInfo.cs
kaipot: This was from trying to load `MRM Triggered MRM data imported.sky` in `TestExportMethodDlg`. Most of the m/zs were within the threshold, but a small percentage were very slightly over 0.001. I didn't look too much into it, I'm just assuming it was some floating point issue going on.

pwiz_tools/Skyline/Model/Results/ChromatogramCache.cs
kaipot: I don't think it can be null, but better to check for it than throw an exception here?

pwiz_tools/Skyline/Model/Results/ChromatogramCache.cs
kaipot: Sure, I'll change that.

pwiz_tools/Skyline/Model/Results/ChromatogramCache.cs
kaipot: I agree, though currently it isn't actually immutable. The flags have setters, e.g. in `ChromCacheBuilder.WriteChromDataSet` there is this:

`chromTran.MissingMassErrors = true;`

brendanx67:
So, just checking in that the goal is not to have the changes in the tutorial logs and other tests. If you end up deciding some tests need to change, please explain why.

nickshulman:
> So, just checking in that the goal is not to have the changes in the tutorial logs and other tests. If you end up deciding some tests need to change, please explain why.

The reason that most of the numbers are changing is that there is currently a bug in ChromDataSet.MarkOptimizationData.
Whenever there are a pair of transitions whose product m/z is approximately 0.01 units apart, one of those transition's chromatograms ends up getting ignored during peak finding. This happens regardless of whether the user has asked for CE optimization or anything.

The other reason that some of the numbers are changing (in "TestExportMethodDialog") is that "ChromatogramInfo.GetAllTransitionInfo" was willing to accept groups of transitions which differed by something other than "OPTIMIZE_SHIFT_SIZE" so long as they were still within "MzMatchTolerance".

PR #2462, opened 2023-01-19
Skyline: Fix ReSharper 22.3.1 warnings

brendanx67:


pwiz_tools/Skyline/Model/DdaSearch/MSAmandaSearchWrapper.cs
```diff
@@ -266,7 +266,7 @@ public override bool Run(CancellationTokenSource tokenSource, IProgressStatus st
                         }
                         finally
                         {
-                            SearchEngine?.Dispose();
+                            SearchEngine.Dispose();
```
bspratt, line 269: I imagine there's a reason for the change, but on the face of it this seems less cautious than before?

bspratt (approved):
Seems good - I had just a couple of questions that I imagine are non-issues.

pwiz_tools/Skyline/Model/DdaSearch/MsFraggerSearchEngine.cs
```diff
@@ -341,7 +341,10 @@ private void FixMSFraggerPin(string cruxInputFilepath, string cruxFixedInputFile
 
         private void EnsureFastaHasDecoys()
         {
-            _progressStatus = _progressStatus.ChangeMessage(string.Format(Resources.EnsureFastaHasDecoys_Detecting_decoy_prefix_in__0__, Path.GetFileName(fastaFilepath)));
+            Assume.IsNotNull(_fastaFilepath);
```
chambm, line 344: Consider putting this after the string.Empty assignment and then check Assume.IsFalse(string.IsNullOrWhitespace()) since it certainly shouldn't be empty (or even a non-existing filepath) coming in to this function.

pwiz_tools/Skyline/Model/DdaSearch/MSAmandaSearchWrapper.cs
brendanx67: We can choose either this, or we can test for null above where it gets used without a null check in the main try block. I suppose we could test for null above and then throw an exception that the SearchEngine was not set. It didn't look to me like the code ever really expected it to be null. Maybe we should ask Matt to weigh in since this is his code.

pwiz_tools/Shared/Common/SystemUtil/StreamEx.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            if (bytesTransferred != bytesToTransfer)
+            {
+                throw new InvalidDataException(string.Format(
+                    @"Tried to transfer {0} bytes, but actual byte count transferred was {1}", bytesToTransfer,
+                    bytesTransferred));
+            }
+        }
+
+        public static void ReadOrThrow(this Stream stream, [In, Out] byte[] buffer, int offset, int count)
```
nickshulman, line 55: I think this should be called something like "ReadComplete" or "ReadFully".
The problem is that "ReadOrThrow" does not give an indication of what would cause it to throw.

Also, in addition to this method, I kind of wish we also had a method which returned a new buffer of a particular length:
public static byte[] ReadByteCount(this Stream stream, int byteCount)
which returned a new buffer of the correct length.

pwiz_tools/Skyline/TestUtil/TestUtil.csproj
```diff
@@ -120,9 +120,9 @@
     <Reference Include="System.Core">
       <RequiredTargetFramework>3.5</RequiredTargetFramework>
     </Reference>
-    <Reference Include="System.Data.SQLite, Version=1.0.105.2, Culture=neutral, PublicKeyToken=db937bc2d44ff139, processorArchitecture=$(PLATFORM)">
+    <Reference Include="System.Data.SQLite, Version=1.0.105.2, Culture=neutral, PublicKeyToken=db937bc2d44ff139, processorArchitecture=$(Platform)">
       <SpecificVersion>False</SpecificVersion>
-      <HintPath>..\..\..\libraries\SQLite\$(PLATFORM)\System.Data.SQLite.dll</HintPath>
+      <HintPath>..\..\..\libraries\SQLite\$(Platform)\System.Data.SQLite.dll</HintPath>
```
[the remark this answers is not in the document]
brendanx67: \libraries\SQLite\ still contains x86 and x64 subfolders. And the .csproj file uses $(Platform) everywhere else for these values. It may be the ReSharper is case sensitive in its warnings where MSBuild is not.

pwiz_tools/Skyline/Model/DdaSearch/MSAmandaSearchWrapper.cs
brendanx67: @chambm Can you pay special attention to this one and see what you think? Should we add a test for if (SearchEngine == null) above, or just remove the test for it here?

chambm: I basically agreed with Brian so I didn't comment earlier. SearchEngine is set in InitializeEngine. If that function throws before SearchEngine is set, it will be null in the finally block. If it throws after SearchEngine is set or doesn't throw at all, it won't be null. And if it doesn't throw at all, the rest of the try block runs and it won't be null so those SearchEngine calls don't need a null check. What am I missing?

bspratt: As Brendan explained it in conversation at my desk, ReSharper notices that we're checking for null in one place (that "?.") but not others nearby, so it barks. Check in neither and it's placated. Sounds like Brendan's change is the right one.

chambm: No, I'm quite certain I put that nullsafe dereference there in response to seeing a NRE there. To appease ReSharper the simplest thing is to just do nullsafe derefs above as well. ReSharper is being too conservative here but that's par for the course (better safe than sorry). Another way to appease ReSharper that would also make the code clearer would be to return SearchEngine from InitializeEngine so it's clear in the finally block that SearchEngine is set after that line (as long as the function can't return null, which it can't, but I'm not sure if ReSharper can know that without an explicit attribute).

chambm:
```

[NotNull]
        private MSAmandaSearch InitializeEngine(CancellationTokenSource token, string spectrumFileName)
        {
            _outputParameters = new OutputParameters();
            _outputParameters.FastaFiles = FastaFileNames.ToList();
            _outputParameters.DBFile = FastaFileNames[0];
            //2 == mzid
            _outputParameters.SetOutputFileFormat(2);
            _outputParameters.IsPercolatorOutput = true;
            _outputParameters.SpectraFiles = new List<string>() { spectrumFileName};
            Settings.GenerateDecoyDb = true;
            Settings.ConsideredCharges.Clear();
            foreach(var chargeStr in AdditionalSettings[CONSIDERED_CHARGES].Value.ToString().Split(','))
                Settings.ConsideredCharges.Add(Convert.ToInt32(chargeStr));
            Settings.ChemicalData.UseMonoisotopicMass = true;
            Settings.ReportBothBestHitsForTD = false;
            Settings.CombineConsideredCharges = false;
            //Settings.WriteResultsTwice = true;
            //Settings.ForceTargetDecoyMode = false;
            //Console.WriteLine("\nReportBothBestHitsForTD CombineConsideredCharges WriteResultsTwice ForceTargetDecoyMode");
            //Console.WriteLine($@"{Settings.ReportBothBestHitsForTD}       {Settings.CombineConsideredCharges}        {Settings.WriteResultsTwice}       {Settings.ForceTargetDecoyMode}");
            mzID.Settings = Settings;
            var searchEngine = new MSAmandaSearch(helper, _baseDir.DirPath, _outputParameters, Settings, token);
            searchEngine.InitializeOutputMZ(mzID);
            Settings.LoadedProteinsAtOnce = (int) AdditionalSettings[MAX_LOADED_PROTEINS_AT_ONCE].Value;
            Settings.LoadedSpectraAtOnce = (int) AdditionalSettings[MAX_LOADED_SPECTRA_AT_ONCE].Value;
            return searchEngine;
        }


...

SearchEngine = InitializeEngine(tokenSource, rawFileName.GetSampleLocator());
```

pwiz_tools/Shared/Common/SystemUtil/StreamEx.cs
brendanx67: I don't know about the latter. In cases where we have an expected byte count, isn't it usually better to throw if we can't get that number of bytes because the file ends unexpectedly?

I did think first of "ReadAll" but that seemed too ambiguous. And I actually preferred being explicit that the main difference was that a failure to read the requested bytes would result in a thrown Exception. I am still not sure I think your proposed names are better.

pwiz_tools/Skyline/Model/DdaSearch/MSAmandaSearchWrapper.cs
brendanx67: Okay. I came up with a no-op assignment that silenced ReSharper and which should have no impact on the logic of this code.

pwiz_tools/Skyline/Model/DdaSearch/MsFraggerSearchEngine.cs
brendanx67: Sure

PR #2464, opened 2023-01-19
Update old BuildMethod projects to .NET Framework 4.7.2

kaipot:
Visual Studio complains about projects targeting .NET Framework 4.0 not being supported. Also, 4.0 targeting pack is really hard to find anywhere for download at this point (the only way I could get it was by installing it through the Visual Studio 2017 installer).


brendanx67 (changes requested):
Don't we at the same time want to update the EXE files under Method? This seems like it will update the target in those EXE files the next time we build them but doesn't actually do that for our next release.

kaipot:
> Don't we at the same time want to update the EXE files under Method? This seems like it will update the target in those EXE files the next time we build them but doesn't actually do that for our next release.

Updating the actual binaries at the same time is a good idea. I rebuilt them with the new configs.

brendanx67 (approved):

nickshulman (approved):

PR #2467, opened 2023-01-24
Change framework version to .Net 4.8

nickshulman:
This change makes it so that nearly all of our C# projects use .Net Framework 4.8.
I also deleted a few folders such as "LocalizationHelper" or "PwizConvert" which are not used anymore and which I could not get to build because the code has changed so much.

I did not change the .Net Framework versions for the things in pwiz_tools\Executables\BuildMethod because that would also require upgrading the Visual Studio project version (maybe this would be easier after PR #2464 is committed)


brendanx67:
This is an interesting page that seems to indicate that Windows 11 may not have .NET 4.7.2 by default, but both Windows 10 (since May 2019) and Windows 11 have .NET 4.8. Windows 10 has had .NET 4.7.2 since April 2018.

https://learn.microsoft.com/en-us/dotnet/framework/migration-guide/versions-and-dependencies

nickshulman:
> This is an interesting page that seems to indicate that Windows 11 may not have .NET 4.7.2

Windows 11 comes with .Net Framework 4.8.
For the .Net Framework versions from 4.0 to 4.8, if you already have a newer .Net Framework version than the one that is required, you do not need to install the older version.

pwiz_tools/SeeMS/Properties/Settings.Designer.cs
```diff
@@ -12,7 +12,7 @@ namespace seems.Properties {
     
     
     [global::System.Runtime.CompilerServices.CompilerGeneratedAttribute()]
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.VisualStudio.Editors.SettingsDesigner.SettingsSingleFileGenerator", "15.9.0.0")]
```
chambm, line 15: Revert changes here I think.

chambm (changes requested):
Looks good when it's all green.

chambm:
My escapades with upgrading MSConvertGUI/SeeMS/bindings/dependencies to .NET 7 made me realize it would probably break Skyline on Docker/Wine if I did get the pwiz bindings to compile on .NET 7 (it's not actually working as smoothly as I alluded to on Monday). That reminded me of this PR since if Docker/Wine only had 4.7.2 this PR would break it. But since the Docker/Wine TC test passed for this PR I was confused. Looking into I saw I had already upgraded the image to .NET 4.8. Hooray for proactive upgrading!

From preliminary reading it does appear that .NET 5+ will be easier to install on Wine than .NET < 5. So that's something I have to look forward to?

nickshulman:
> My escapades with upgrading MSConvertGUI/SeeMS/bindings/dependencies to .NET 7 made me realize it would probably break Skyline on Docker/Wine if I did get the pwiz bindings to compile on .NET 7 (it's not actually working as smoothly as I alluded to on Monday). That reminded me of this PR since if Docker/Wine only had 4.7.2 this PR would break it. But since the Docker/Wine TC test passed for this PR I was confused. Looking into I saw I had already upgraded the image to .NET 4.8. Hooray for proactive upgrading!
> 
> From preliminary reading it does appear that .NET 5+ will be easier to install on Wine than .NET < 5. So that's something I have to look forward to?

I am not sure I understand this comment but it would be awesome if we could use .Net 5 or greater for Skyline.
I have been barely able to figure out how to create a .Net 5 project, and I was completely stumped as to how to convert a .Net Framework project into something using 5+.

chambm:
> 
> I am not sure I understand this comment but it would be awesome if we could use .Net 5 or greater for Skyline. I have been barely able to figure out how to create a .Net 5 project, and I was completely stumped as to how to convert a .Net Framework project into something using 5+.

There's an upgrade-assistant tool that can do a lot of the work for converting MSBuild project files to Project SDK style (ugh that is a terrible name). But there are a lot of caveats and cleanup to do afterward, even for projects as small as SeeMS and MSConvertGUI. For Skyline it'll be a considerable PITA. We almost certainly will want to disable the new default behavior of "compile and embed everything in the same directory as the project file and all its subdirectories" because the Skyline project has all its subprojects as subdirectories instead of siblings.

PR #2469, opened 2023-01-25
Declare MSBuildAssemblyName to quiet a new ReSharper inspection warning

bspratt:
ugh, I messed up the branch naming convention, sorry


chambm:
That's a bit of a headscratcher. It's defining the property but only if it's empty/undefined? Why doesn't that line itself trigger the ReSharper warning? Maybe that's a ReSharper bug. Did it not work as expected without the condition? Did you test with the Jamfile build with and without the branch set to 0 to see if the logic still works both ways?

bspratt:
Yeah, I'm not convinced it's the best way - I was looking at
` <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>`
which means "if Platform isn't defined externally, set it to AnyCPU locally"
and trying for "if MSBuildAssemblyName isn't defined externally, set it empty locally" 
maybe` <MSBuildAssemblyName />` does the same thing. I don't really speak the language, and I don't know how to exercise the mechanism it's supporting. Perhaps I should hand off to you. I was thinking the TC run would prove it out, sounds like maybe not.

bspratt:
If you're happy with this now, let's go ahead and merge it.

chambm:
@brendanx67 should this be cherry picked to avoid the warning in release or we're beyond patching?

brendanx67:
I hope we are beyond patching, but we are definitely beyond trying to keep warnings to zero in new ReSharper versions. So, no need to cherry-pick.

PR #2471, opened 2023-01-27
Read isolation window using Waters MassLynx SDK

pete-reay-waters:
Adds the ability to specify the DDA MSMS isolation window upper and lower offsets in a Waters RAW file. This is a new feature in the recently added MassLynxSDK which uses a very simple file format to include the isolation window.

This allows generation of an mzML file with correct isolation window offset values, which is required for some downstream processing applications.

Added DDA_IsolationWindow.raw to test this functionality. It only applies to DDA.


pwiz_aux/msrc/utility/vendor_api/Waters/WatersRawFile.hpp
```diff
@@ -457,6 +457,21 @@ struct PWIZ_API_DECL RawData
         return success;
     }
 
+    bool GetIsolationWindow(float& lowerOffset, float& upperOffset) {
+        MassLynxParameters parameters = DDAProcessor.GetParameters();
+        float lowerOffsetParam = stof(parameters.Get(DDAParameter::LOWEROFFSET));
+        float upperOffsetParam = stof(parameters.Get(DDAParameter::UPPEROFFSET));
```
chambm, line 463: Thanks for the PR!

Just a few things changes would be nice here:
- existing style is open brace on new line and one line if statements don't need braces
- use lexical_cast<float> instead of stof (it's actually overloaded to use strtod but we can consistently check for bad_lexical_cast)
- how big is this parameters object that's going to get copied for every scan? Are those parameters the same for every scan?

chambm (changes requested):

chambm:
Is creating this file something the user is expected to do prior to data conversion/reading?

pwiz_aux/msrc/utility/vendor_api/Waters/WatersRawFile.hpp
Simon-PJ: Hi, these were my changes that Pete has created the PR for. On the first two points, yep all fair, I'll make those changes. On the last point we believe that current instruments it will be the same for each scan, but there's the possibility of upcoming instruments varying by scan. The parameters file only contains the two values seen here so we'd prefer to leave as is.

Simon-PJ:
> Is creating this file something the user is expected to do prior to data conversion/reading?

This file is something that is added to a masslynx file by our software prior to conversion. The user won't be aware of it or have to do anything.

pwiz/data/vendor_readers/Waters/Reader_Waters_Test.data/DDA_IsolationWindow-ddaProcessing.mzML
```diff
@@ -0,0 +1,193 @@
...
+              </scanWindow>
+            </scanWindowList>
+          </scan>
+        </scanList>
+        <precursorList count="1">
+          <precursor>
+            <isolationWindow>
+              <cvParam cvRef="MS" accession="MS:1000828" name="isolation window lower offset" value="10.0" unitCvRef="MS" unitAccession="MS:1000040" unitName="m/z"/>
+              <cvParam cvRef="MS" accession="MS:1000829" name="isolation window upper offset" value="999.90002" unitCvRef="MS" unitAccession="MS:1000040" unitName="m/z"/>
+              <cvParam cvRef="MS" accession="MS:1000827" name="isolation window target m/z" value="500.0" unitCvRef="MS" unitAccession="MS:1000040" unitName="m/z"/>
```
chambm, line 103: Looking at this again, this may not be what's intended. The offsets are relative to the target m/z, so this is indicating the isolation window is from 490 to 1499.9. Is that correct?

pete-reay-waters: Hi Matt, you're correct - that's what those values are indicating. I took the values from a unit test we had, but it's definitely not realistic. Let us discuss and get back to you.

pete-reay-waters:
Hi @chambm we've addressed the PR comments, let us know if anything else needs changing. Cheers! cc @Simon-PJ

chambm:
Hey guys I tested the DDA processing mode on an old DDA file I have. Is it it supposed to do peak picking? When I try to use Waters MassLynx peak picking, I get profile data that pwiz thinks are centroids. If I use CWT peak picking I get sensible peaks. But we probably need to get that sorted out.

pete-reay-waters:
You mean so you can do centroiding and DDA processing on the same run? Yes, we do want that to happen

chambm:
Yes. In the meantime I can tell users to use CWT peak picking.

PR #2473, opened 2023-01-27
Automatically create "Skyline_X.Y.xsd" for the current document format version for official builds

nickshulman:
Change "SchemaDocumentsTest.TestDocumentFormatSchemaFiles" so that, for official builds, it automatically creates "Skyline_X.Y.xsd" for the current document format version if it does not already exist, and adds that file to git.

Also, change TestUtil.csproj so that it uses a filename wildcard to make it so that all xsd files in TestUtil\Schemas folder get added to project.


pwiz_tools/Skyline/TestUtil/TestUtil.csproj
```diff
@@ -233,238 +233,7 @@
...
-    <EmbeddedResource Include="Schemas\Skyline_22.13.xsd">
-      <SubType>Designer</SubType>
-    </EmbeddedResource>
-    <EmbeddedResource Include="Schemas\Skyline_22.2.xsd">
-      <SubType>Designer</SubType>
-    </EmbeddedResource>
-    <EmbeddedResource Include="Schemas\Skyline_22.21.xsd">
-      <SubType>Designer</SubType>
-    </EmbeddedResource>
+    <EmbeddedResource Include="Schemas\*.xsd" />
```
chambm, line 236: Cool! Didn't know you could do that.

PR #2477, opened 2023-01-31
Skyline: Get rid of async/task/await use in DDASearchControl in favor…

brendanx67:
… of ActionUtil.RunAsync
- which has Exception handling and LocalizationHelper.InitThread
- and avoids the .NET thread pool which can make tracking down leaks challenging
- this also will hopefully make nightly test failures in this code easier to decipher

### DIFF
--- a/pwiz_tools/Skyline/FileUI/PeptideSearch/DDASearchControl.cs
+++ b/pwiz_tools/Skyline/FileUI/PeptideSearch/DDASearchControl.cs
@@ -19,13 +19,13 @@
 using System;
 using System.Collections.Generic;
 using System.Threading;
-using System.Threading.Tasks;
 using System.Windows.Forms;
 using MSAmanda.Utils;
 using pwiz.Common.SystemUtil;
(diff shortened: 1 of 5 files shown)


brendanx67:
I got rid of "Conversion finished". If this showed up in the log, that might be a mistake. The old code never showed "Starting search" because it used IProgressStatus.ChangeMessage() without assigning the result and a ProgressStatus is immutable. If these messages don't show up in the log, then there is not enough time between the two messages for a user to see the first message.

brendanx67:
For testing, I ran all the DDA tests, and I ran the TestDdaTutorial test in screenshot mode and performed a number of cancels during the DDA progress, which seemed to work.

chambm (approved):
If it works and is easier to debug when the Run function throws, sounds like a great change. I just extended what Viktoria implemented.

brendanx67:
I figured that was the case. We have no other async/task/await use in our code. So, not really any thought given to supporting the Task class.

PR #2478, opened 2023-02-02
- updated to .NET 7 (currently just MSConvertGUI, SeeMS, and their de…

chambm:
This is just a draft for discussion. The motivation for this was a bit silly: I wanted access to
`<ApplicationDefaultFont>Segoe UI, 8.25pt</ApplicationDefaultFont>` in the project file to override the OS default font for CJK Windows users for MSConvertGUI, so that what I see in the MSConvertGUI form designer would also be what CJK users would see (since there's no i18n). It turned out that I didn't actually need to compile the CLI bindings with .NET core: the .NET 5+ GUIs can use .NET Framework DLLs *as long as* the .NET Framework DLLs don't use .NET APIs that have been removed in .NET 5+. It also turned out that many of our vendor DLLs, including Thermo RawFileReader, Agilent MHDAC (but apparently not MIDAC), and the WIFF1 and WIFF2 SDKs, do use such APIs, so none of them work with this new version. All the C/C++ vendor DLLs still work. Fixing this would require upgrades to the vendor DLLs. I'm not sure MHDAC is still being maintained, and I'm certain the WIFF1 SDK isn't. RawFileReader probably does have an updated version with .NET 5+ support.

So yeah, the font should look nice on CJK Windows version, but it can't convert Thermo/Sciex/Agilent data. 🤣 

NB: Since I upgraded CommonUtil but not any other Skyline dependencies or Skyline itself, the Skyline build is expected to be broken here.

- updated to .NET 7 (currently just MSConvertGUI, SeeMS, and their dependencies)
- set default font in MSConvertGUI to Segoe UI (this should also apply to non-English Windows versions)
- fixed incorrect icon for MSConvertGUI
* moved MSConvertGUI project to its own subdirectory (otherwise the project includes files from all subprojects, or must explicitly exclude them in the project file)
* removed obsolete attributes from ZedGraph
* removed SpyTools not supported in .NET 5+


chambm:
@nickshulman Look at pwiz_tools/SeeMS/seems.csproj to see a basic example of the conversion from MSBuild to Project SDK.

PR #2479, opened 2023-02-02
Change interpolation algorithm so that it preserves the integrated area

nickshulman:
There have been a few cases where Skyline's chromatogram interpolation algorithm produces unexpected results.
Most recently, there was this support request:
https://skyline.ms/announcements/home/support/thread.view?rowId=59209
Here is what the raw chromatogram looks like:
image
and here is what the current implementation of the interpolation algorithm looks like:
image
and here is what the interpolated chromatogram looks like with this pull request:
image

The current interpolation algorithm sometimes produces unexpected results if the interpolated time values happen to fall in strange places in between the existing points.
The new algorithm is guaranteed to preserve the integrated area over all ranges where the y-values of the end points are unchanged.


brendanx67:
Have you tested the new algorithm for performance? My memory of a prior change to the algorithm to preserve areas was that it was considerably slower.

nickshulman:
I made a new pull request in my own fork: https://github.com/nickshulman/pwiz/pull/2

PR #2481, opened 2023-02-03
Add "Peptide Spectrum Match Count" column

nickshulman:
Add "Peptide Spectrum Match Count" to the PeptideResult in the Document Grid
(requested by Paul)

image
image

What do we think of the name "Peptide Spectrum Match Count"?
Should we maybe call it "PSM Count"? "Peptide Spectrum ID Count"?

What do we think of the tooltip:
_Number of spectra in the spectral library which match the peptide and Result File._

(still need to write a unit test)


pwiz_tools/Skyline/Model/Databinding/Entities/PeptideResult.cs
```diff
@@ -267,6 +267,15 @@ public FiguresOfMerit BatchFiguresOfMerit
             }
         }
 
+        public int PeptideSpectrumMatchCount
+        {
+            get
+            {
+                return SrmDocument.Settings.GetRetentionTimes(GetResultFile().ChromFileInfo.FilePath,
```
brendanx67, line 274: I can imagine wanting the total number of matches for a peptide regardless of matching the result file, can't you? This is what I thought you had implemented until I read the code. Now, I definitely think both are interesting and convey different useful information. It would definitely make naming trickier, because you would want this name for the total and something even longer for this like "Peptide Spectrum Match In File Count" or "Peptide Spectrum Match Count In File"

pwiz_tools/Skyline/Skyline.csproj
```diff
@@ -472,12 +472,12 @@
       <DesignTime>True</DesignTime>
       <DependentUpon>ColumnCaptions.resx</DependentUpon>
     </Compile>
-    <Compile Include="Model\DdaSearch\MsconvertDdaConverter.cs" />
     <Compile Include="Model\Databinding\Entities\ColumnToolTips.Designer.cs">
       <AutoGen>True</AutoGen>
       <DesignTime>True</DesignTime>
       <DependentUpon>ColumnToolTips.resx</DependentUpon>
     </Compile>
+    <Compile Include="Model\DdaSearch\MsconvertDdaConverter.cs" />
```
brendanx67, line 480: Seems like an extraneous change.

brendanx67 (changes requested):
I am not convinced this is ready to merge, since it may use a name that would be better used as the total number of PSMs for the peptide rather than the number in a result file. Also, it should impact at least one test to make sure the number is as expected.

brendanx67:
Should "Protein Result" be showing up in "Peptide Result"? I know this is how lookups work, but we don't usually do this, since the user can already access Protein Result from higher in the tree.

nickshulman:
> Should "Protein Result" be showing up in "Peptide Result"? I know this is how lookups work, but we don't usually do this, since the user can already access Protein Result from higher in the tree.

"Protein Result" shows up there because this is the Customize Report window for the Results Grid.

If you were customizing a report on the Document Grid, then Protein Results would be hanging off of the "Protein" in the tree, since all of the columns that you are choosing are property paths that begin at a property on "SkylineDocument".

In the Results Grid, you are always starting with properties from a class such as "PeptideResult", "PrecursorResult", or "TransitionResult".

I am not sure whether this property should be the one named "Peptide Spectrum Match Count", or whether we should reserve that name for a property on the Peptide which tells you the total number of PSMs across all of the replicates.
When would the user be interested in the total value across all replicates?
If they were quantifying using spectral counting, then the only number which is interesting is the per-replicate PSM count.
I was imagining that the total value across all replicates is obscure enough that users would need to use the Pivot Editor in order to see that value.

brendanx67:
Well, it is essentially shown in the chromatogram graphs. You are showing the count of lines that show up with ID annotations and I was suggesting the full count of all vertical ID lines shown in the graph. It is also what the user might get at by selecting the peptide and dropping the dropdown in the Library Match graph when ID redundancy is allowed. It is not the same as the total of the values you present because there are many cases where library IDs do not match any files.

Again, thinking of the chromatogram graphs, we have IDs, aligned IDs, and other IDs. Your new value makes the first easy to get, the second possible with the pivot editor, but leaves the last impossible to access.

brendanx67:
Let's just leave it until it gets requested. I guess what I am talking about would most appropriately get added as Precursor.LibraryMatchCount along with the library score values. More insight into how the target shows up in the library. But, this is in response to a user request looking for something quite different, and no one has requested what I am suggesting.

brendanx67 (approved):

PR #2487, opened 2023-02-07
Fix for problems with Agilent GC-TOF data:

bspratt:
- Fixed a logic flaw where Agilent GC-EI data was mistaken for non-GC all ions data
- Also, wasn't extracting chromatograms unless both MS1 and MS2 were enabled in Full Scan settings, really only MS2 setting matters for GC

Reported by user LeeAnn


nickshulman (approved):
Looks good to me.

PR #2488, opened 2023-02-08
Fix WIFF SIM/SRM RTWindows bug

chambm:
- added automatic doubling of SIM/SRM RTWindows if a WIFF file was collected by SCIEX OS version less than 3.1 (reported by Phillip)
* removed very old conditional code for .NET < 4

[ProteoWizard installer (MSConvert/SeeMS)](https://teamcity.labkey.org/viewLog.html?buildId=2288892&buildTypeId=bt83&tab=artifacts&guest=1)
[SkylineTester.zip](https://teamcity.labkey.org/guestAuth/repository/download/bt209/2288896:id/SkylineTester.zip)


chambm:
@brendanx67 I have been trying to decide whether I want to leave this ignoreScheduledLimitsForChromatograms flag as an option or make it unconditional (i.e. always return the biggest chromatogram possible for scheduled MRM data). Turning it on does break one Skyline test (TestOptimization). But I'm not actually sure what the test difference is reporting:
```
Diff found at line 1:
450.695949,524.246337,20,peptides1.FNDDFSR.+2y4.light,80,21.1,17.00
>
450.695949,524.246337,20,peptides1.FNDDFSR.+2y4.light,80,21.1,16.50
#    at pwiz.SkylineTestFunctional.OptimizeTest.CovOptimizationTest() in C:\dev\pwiz\pwiz_tools\Skyline\TestFunctional\OptimizeTest.cs:line 704
```
What is that last column? Compensation voltage?

The user that had this problem wasn't using Skyline, but if they had, they would have the same problem this new option fixes: they would not have been able to see all the data they expected to see. That's because the API's default chromatogram extraction method uses too narrow RT window boundaries for some scheduled methods. So I wonder actually how many users this has affected without them realizing it.

nickshulman:
@chambm,
I could not figure out how to get "TestOptimization" to fail.
I tried changing the C++ code so that "ignoreScheduledLimitsForChromatograms" was always true, and as far as I could tell, TestOptimization still passed.
If you get this pull request into a state where tests are failing, I'll try to figure out how to fix them.

chambm:
> @chambm, I could not figure out how to get "TestOptimization" to fail. I tried changing the C++ code so that "ignoreScheduledLimitsForChromatograms" was always true, and as far as I could tell, TestOptimization still passed. If you get this pull request into a state where tests are failing, I'll try to figure out how to fix them.

I'm actually getting 3 failures with `ignoreScheduledLimitsForChromatograms = true` in the MsDataFileImpl constructor: TestSrmTutorialLegacy,TestPeakPickingTutorial,TestAsymCEOpt

None are TestOptimization. I'm not sure what I did to cause that one.

nickshulman:
I think we should just update the expected values and audit log files for "TestSrmTutorialLegacy" and "TestPeakPickingTutorial".
I'll try to figure out what is going on with "TestAsymCEOpt".

chambm:
If we're not considering making user-configurable in Skyline, I'm inclined to just always enable it in pwiz and remove the option. Did you determine that the new behavior is preferable for those tutorials?

I'm a bit worried that I didn't see those test failures the first time I was working on this PR and set the parameter to always be true; maybe I didn't have it set to run the tutorials, but that doesn't explain why only TestAsymCEOpt failed now and only TestOptimization failed back then.

nickshulman:
> Did you determine that the new behavior is preferable for those tutorials?
I would not say "preferable".
The new behavior did not seem to raise any red flags.

Yes, getting rid "ignoreScheduledLimitsForChromatograms" sounds like a good idea.
I will try running the entire test suite on macs2 and make sure everything passes.

PR #2492, opened 2023-02-09
Fix MSFragger C-terminal mods

chambm:
- fixed MSFragger searches with C-terminal mods giving error "'c' is not an amino acid" (reported by jalbin)

Closes https://skyline.ms/announcements/home/support/thread.view?rowId=59296 (if we used GitHub for Skyline issues it would close the issue when I merged the PR... :) )


pwiz_tools/Skyline/Model/DdaSearch/MsFraggerSearchEngine.cs
```diff
@@ -92,6 +92,7 @@ public MsFraggerSearchEngine(double percolatorQvalueCutoff)
         private Enzyme _enzyme;
         private int _ntt, _maxMissedCleavages;
         private int _maxVariableMods = 2;
+        private List<Tuple<StaticMod, double, string>> _variableMods;
```
nickshulman, line 95: I would recommend making a private class called "CruxModification" that you could use instead of this Tuple, and would also have the methods equivalent to "getCruxTerminusOrdinal" and "getCruxResidues".

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/Model/DdaSearch/MsFraggerSearchEngine.cs
brendanx67: Agreed. Tuples are barely acceptable when they are entirely local to a function, but they should never, never, never by used by member variables or passed back as a return value from a function. The resulting code is poorly self-documenting.

brendanx67: We should also discuss use of sub-functions as a team and whether camel case is appropriate to indicate a sub-function or we should use something else, and whether there are any other considerations for when and where to use them.

chambm: "Never passed back as a return value from a function" would ding a lot of Python's standard library functions and a few from the C++ standard library's as well. Sometimes they are just more convenient than making named class: e.g. when the named class doesn't add any meaningful encapsulation. But having specific property names certainly does make a named class self-documenting in a way a tuple can never be. In this case it was a close call for my personal tastes (obviously Brendan's are different and more important in this context :) ). In this refactored version I feel moving the getCruxResidues and getCruxTerminusOrdinal functions into member functions made the code harder to understand. IMO better here would be to leave them as local functions that take a CruxModification (i.e. still refactor the tuple).

brendanx67: Your tastes are based on how you feel about reading your own code. I promise a named class is easier for an unfamiliar reader of your code to get up to speed on what you are trying to achieve.

chambm: Apparently with C# 7 (we're on 8) there are tuple types. So I could have done:
```
private List<(StaticMod Mod, double Mz, string Residues)> _variableMods;
```
That's pretty neat. But with the named class I could put a real XmlDoc on the properties to explain what they represent. That's impossible to beat for documentation.

brendanx67: That's cool. It seemed like they would eventually need to do this. Don and I discussed this a fair bit back in the early 20-teens, and he thought F# already had a good solution for property naming on dynamic types. Great that it finally made it to C#.

PR #2493, opened 2023-02-09
Skyline: Add a wait for document change and loaded to PerfUniquePepti…

brendanx67:
…desTest in an area causing an intermittent failure

- Seems like this at least stands a chance of fixing the failure for which the Assume.DebugOnFail() was added at line 77


bspratt (approved):
I like it - I wonder how many similar situations there may be?

brendanx67:
> I like it - I wonder how many similar situations there may be?

It is probably not always as critical, and Nick's new dialog box waiting variants aim at the same problem that forms can disappear before all of their processing is complete.

brendanx67:
I was surprised to see cases where clicking the OK button did not have an impact in this test. My first attempt was simply to put in the using (new WaitDocumentChange()), but that caused 2 of the tests to hang.

PR #2495, opened 2023-02-13
Feature detection via Hardklör integration with Skyline

bspratt:
This adds a new menu item "Files > Import... >Feature Detection..." which invokes a workflow employing a lightly modified version of Hardklor to detect possible peptides.

The workflow runs Hardklor on selected mass spec data, generates a library from the Hardklor results, adds the contents of the library to the Skyline document, and performs chromatogram extraction on the mass spec data that Hardklor was examining.

Most of the modifications to Hardklor have to do exposing the chemical formula and mass shift that Hardklor used to generate the isotope envelope for each reported detection. For Skyline's purposes just having the mass and charge as reported by standard Hardklor isn't enough, we want the entire envelope. The other significant change to Hardklor is that we don't emit matches on fewer than three isotopic peaks - it's just too noisy to be useful, otherwise.

BlibBuild is extended to read this Hardklor format and produce .blib files that contain precursor information only (this is an unusual use of the .blib format, which generally is about fragments along with precursors). 

This also introduces an extension to Skyline's idea of chemical formulas - they may now contain information about unexplained mass shifts, e.g. "C12H5O3[+1.2]" (unexplained additional mass of 1.2 AMU),  "C12H5O3[-1.3]" (unexplained reduced mass of 1.3 AMU),  "C12H5O3[+1.2/1.21]" (unexplained additional mass of 1.2 mono / 1.21 average) etc. The + or - sign is required, the value is always Invariant formatted ("." decimal separated) and all must be enclosed in square brackets. Normally such descriptions appear at the end of the formula but may appear anywhere, and there may be more than one (Skyline will do the math internally).


pwiz_tools/Shared/Common/Chemistry/Formula.cs
```diff
@@ -292,9 +353,78 @@ public static T ParseExpression(String formula)
...
+                    massValueAverage += ParseMassModification(kvp.Key, true) * kvp.Value;
+                    result.Remove(kvp.Key);
+                }
+
+                var newMassMod = FormatMassModification(massValueMono, massValueAverage, observedDecimals);
+                result.Add(newMassMod, 1);
+            }
+        }
+
+        public bool HasMassModifications => Keys.Any(k => k.StartsWith(MASS_MOD_START));
```
nickshulman, line 410: This does not belong in this class.
A "Molecule" should never have mass modifications on it. If you want a Molecule with mass modifications you should have to use the class "MoleculeMassOffset".
I think the code for parsing formulas that might have mass offsets in the middle of them should be confined to the stuff that deals with HardKlor. By the time the formula gets written to something like a .blib file, it should look like what MoleculeMassOffset.ToString() produces.

bspratt: Can't argue with that semantically - molecules are strictly speaking made of atoms and isotopes thereof, nothing more. 

But if we want to use a special class for the idea of "thing that results in an isotopic distribution", that would be a pretty extensive change to the codebase. I'd think that we'd wind up using that new class nearly everywhere that we currently use Molecule.

As to tolerating the offset description being randomly placed, consider this: if we understand that simply concatenating "C4H9O" and "H3O2N" is meaningful, then I'd think we'd want to understand that concatenating "C4H9O[+1.07]" and "H3O2N[-2.3]" is meaningful too.

pwiz_tools/BiblioSpec/src/DelimitedFileReader.h
```diff
@@ -233,37 +259,73 @@ template <typename STORAGE_TYPE> class DelimitedFileReader {
         std::string line;
         bool parseSuccess = true;
         std::string errorMsg;
-        
+
+        // Boost tokenizer is overkill for some formats
```
chambm, line 263: What is the rationale for this added complexity? Was the tokenizer a performance bottleneck?

pwiz_tools/BiblioSpec/tests/reference/Small_Library-Positive-ions_CE-Merged.check
```diff
@@ -36,3 +36,4 @@ id	scoreType	probabilityType
 17	BYONIC SCORE	PROBABILITY_THAT_IDENTIFICATION_IS_INCORRECT
 18	PEPTIDE SHAKER CONFIDENCE	PROBABILITY_THAT_IDENTIFICATION_IS_CORRECT
 19	GENERIC Q-VALUE	PROBABILITY_THAT_IDENTIFICATION_IS_INCORRECT
```
chambm, line 38: This is annoying. Do we even need to output this table for every check file, requiring that adding a new score changes every check file? If not then we can take the opportunity here to change all files by removing them all instead of changing all files adding one more row that is irrelevant to all the files except one.

pwiz_tools/Shared/Common/Chemistry/Formula.cs
nickshulman: The class for "Thing that can result in an isotope distribution" is "MoleculeMassOffset".
"C4H9O[+1.07]" and "H3O2N[-2.3]" are both MoleculeMassOffset objects.
When you combine them you get another MoleculeMassOffset.

Molecules happen to have the property that if you concatenate their ToString() values, you get something that can still be parsed as a Molecule. We do not necessarily need to preserve that behavior for the class MoleculeMassOffset.

pwiz_tools/BiblioSpec/tests/reference/Small_Library-Positive-ions_CE-Merged.check
bspratt: Yeah, Brendan and I were musing on this too. It doesn't seem very high value - all it really checks for is did somebody accidentally change something from ROBABILITY_THAT_IDENTIFICATION_IS_CORRECT to ROBABILITY_THAT_IDENTIFICATION_IS_INCORRECT etc.

pwiz_tools/BiblioSpec/src/DelimitedFileReader.h
bspratt: Yeah, though that may have been in debug build, I don't recall for sure. Hardklor makes crazy big delimited files and this was slowing me down.

pwiz_tools/Shared/Common/Chemistry/Formula.cs
bspratt: I guess I'd have to take a shot at doing all the necessary replacements before I can pursue the cost/benefit argument. Brendan didn't think the current implementation was too nutty, at first blush anyway.

pwiz_tools/BiblioSpec/src/DelimitedFileReader.h
chambm: Boost tokenizer is slow if it's having to allocate new strings. Can you verify that it was a pure debug build that was slow(er) enough to justify this? Because if the `iterator-debugging=off runtime-debugging=off` fast-debug build avoids that problem, I don't think we need the extra complexity.

chambm:
New files will need to be added to Product.wxs-template for the admin installer.

pwiz_tools/BiblioSpec/src/DelimitedFileReader.h
bspratt: My debug builds are iterator-debugging=off , not sure about runtime-debugging=off. What's the latter do?

It really was painful.

chambm: Runtime debugging uses the debug CRT which can help for memory-related errors but slows down memory allocation a lot as well. How far is your strtok approach from just being able to handle the tokenizer case as well? I have no particular allegiance to tokenizer so if what you've got is faster then let's use that and only that. I've been looking at alternatives but subbing those in would probably be more work than getting strtok to work for all the cases.

bspratt: The simple strtok-based reader really is simple - no support for quotes etc. That's sufficient for reading Hardklor output, but not for a lot of other stuff, I don't think we can dump tokenizer without reimplementing it ourselves, which of course would be silly.

pwiz_tools/Shared/Common/Chemistry/Formula.cs
brendanx67: I didn't look at the specifics, but I probably agree with Nick that it should be something other than a molecule. Maybe a "MoleculeRoughGuess". That's what it really is. We are just saying that probably the chemical formula we have is close enough to give us a good guess at the isotope distribution which will be dominated by the number of Carbon atoms in the molecule. So, whether those are exactly so many Ns, Os, and Hs and knowing that exactly is less important that knowing roughly the number of Carbon atoms and a highly precise center for the m/z distributions or centroids. Mass accuracy and precision are too high for us to just settle for a rough guess on the molecule mass, but for the isotope distribution a rough guess on the number of atoms, especially Carbons, may be just fine.

bspratt: Let me make one more argument for my current implementation then I'll drop it:

We already have a peculiar syntax for mass shifts in molecular formulas (the  ' and " thing), so it seems wrong to be hung up on one more like "[+1.1/1.12]". If we were going to be truly purist about Molecule/Formula we'd need to go back and revise everything to use standard isotope notation like 13C and 15N. But of course we'll never do that because the current notation is convenient for us and easy enough for users to understand, and it certainly wouldn't be worth the code churn. 

I don't think it's worth the code churn of replacing Formula and Molecule virtually everywhere they're used in Skyline for this reason either. (Looks like at least 88 and 140 places in the code, respectively)

brendanx67: That is just a notation difference for specifying a thing that truly exists in nature. They have very clear definitions as molecules, even if the notation is less purist. They really are formulae and molecules.

pwiz_tools/Shared/Common/SystemUtil/ProcessRunner.cs
```diff
@@ -147,7 +147,7 @@ public class ProcessRunner : IProcessRunner
                             if (double.TryParse(percentPart.Substring(0, percentPart.Length - 1), out percent))
                             {
                                 percentLast = (int)percent;
-                                status = status.ChangePercentComplete(percentLast);
+                                status = status.ChangePercentComplete(percentLast).ChangeMessage(percentPart);
```
nickshulman, line 150: This "ChangeMessage(percentPart)" looks suspicious. Isn't that going to just be a number (it's everything between the last space and the percent sign). Wouldn't it be better to leave the message unchanged?

pwiz_tools/Skyline/Util/FormulaWithMassModification.cs
```diff
@@ -0,0 +1,628 @@
...
+using pwiz.Common.Chemistry;
+using pwiz.Common.Collections;
+
+namespace pwiz.Skyline.Util
+{
+    /// <summary>
+    /// Based on Formula class in this same namespace, this adds the ability to describe a mass
+    /// offset of unknown origin.
+    /// </summary>
+    public class FormulaWithMassModification : IDictionary<string, int>, IComparable
```
nickshulman, line 35: I don't think this source file is mentioned in Skyline.csproj

nickshulman (commented):
Looks good to me.
We will need to make sure that all developers have checked the "Recursive" checkbox in TortoiseGit so that they get the Hardklor and MSToolkit submodules.

pwiz_tools/Skyline/Executables/Hardklor/Jamfile.jam
```diff
@@ -0,0 +1,108 @@
+#
+# $Id$
+#
+#
+# Original author: Brian Pratt <bbspratt at uw.edu>,
```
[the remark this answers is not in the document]
bspratt: Hah, thanks

pwiz_tools/Skyline/Util/FormulaWithMassModification.cs
bspratt: Yeah, that's cruft. I'll delete it, thanks

pwiz_tools/Shared/Common/SystemUtil/ProcessRunner.cs
bspratt: Yeah, there's a bug in how percent change is handled that resulted in the progress bar not moving in response to such messages, but that wasn't the best fix. Thanks for pointing that out.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -97,16 +98,25 @@ public class DDASearchPage : IFormView { }
 
         private readonly Stack<SrmDocument> _documents;
 
-        public ImportPeptideSearchDlg(SkylineWindow skylineWindow, LibraryManager libraryManager)
+        public ImportPeptideSearchDlg(SkylineWindow skylineWindow, LibraryManager libraryManager,
+            Workflow workflowType = (Workflow)(-1))
```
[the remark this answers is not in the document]
bspratt: Fair enough

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -97,16 +98,25 @@ public class DDASearchPage : IFormView { }
...
+        public ImportPeptideSearchDlg(SkylineWindow skylineWindow, LibraryManager libraryManager,
+            Workflow? workflowType)
         {
             SkylineWindow = skylineWindow;
             _documents = new Stack<SrmDocument>();
             SetDocument(skylineWindow.Document, null);
 
-            ImportPeptideSearch = new ImportPeptideSearch();
+            ImportPeptideSearch = new ImportPeptideSearch()
+                { IsFeatureDetection = (workflowType??Workflow.dda) == Workflow.feature_detection };
```
nickshulman, line 109: You can do:
`IsFeatureDetection = workflowType == Workflow.feature_detection`
[For the comparison operators if one or both operands are null, the result is false](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types#lifted-operators).

bspratt: Good point. Tidied that up, thanks.

pwiz_tools/Skyline/FileUI/PeptideSearch/SearchSettingsControl.cs
```diff
@@ -247,13 +258,16 @@ public void InitializeEngine()
 
         private void LoadComboboxEntries()
         {
-            LoadFragmentIonEntries();
-            LoadMs2AnalyzerEntries();
+            if (!ImportPeptideSearch.IsFeatureDetection)
```
chambm, line 261: Less code churn and indentation if you invert this if (if feature detection, return early).

bspratt: Thanks, I'll take a look

bspratt: yeah, better thanks

bspratt:
Full scan and Hardklor settings are now on different tabs (formerly combined in a confusing manner)
image
image

pwiz_tools/Skyline/SettingsUI/FullScanSettingsControl.cs
```diff
@@ -905,8 +907,8 @@ public double TimeAroundPrediction
                 textAt.Visible = false;
                 textRes.Enabled = true;
                 textRes.Text = resCurrent.HasValue && (analyzerTypeCurrent == analyzerTypeNew)
-                                  ? resCurrent.Value.ToString(LocalizationHelper.CurrentCulture)
-                                  : TransitionFullScan.DEFAULT_CENTROIDED_PPM.ToString(LocalizationHelper.CurrentCulture);
+                                  ? resCurrent.Value.ToString(resolvingPowerFormat)
+                                  : TransitionFullScan.DEFAULT_CENTROIDED_PPM.ToString(resolvingPowerFormat);
```
brendanx67, line 911: PPM values don't need this special format. They are only allowed to range between 1 and 1000, and people rarely use factional values, or if they do, I would expect them to be no finer-grained than 0.5

brendanx67 (changes requested):

pwiz_tools/Skyline/SettingsUI/FullScanSettingsControl.cs
bspratt: OK, I'll write a single function that formats based on value type, thanks

brendanx67: Sounds like a good idea.

pwiz_tools/BiblioSpec/src/BlibMaker.cpp
```diff
@@ -130,6 +131,21 @@ int BlibMaker::parseNextSwitch(int i, int argc, char* argv[])
         exit(0);
     } else if (switchName == 't') {
         scoreLookupMode_ = true;
+    }
+    else if (switchName == 'z') {
+        string charges((++i < argc) ? argv[i] : "");
```
chambm, line 136: This line takes too much thinking to understand what it does. Why not just check i+1 >= argc and if it is throw an error like an argument is required? Unless precursorCharges_ can be empty, but that's the same as not passing -z at all?

pwiz_tools/BiblioSpec/src/BuildParser.cpp
```diff
@@ -23,12 +23,13 @@
...
 #include "SpecData.h"
+#include "SslReader.h"
 
 namespace BiblioSpec {
 
 BuildParser::BuildParser(BlibBuilder& maker,
-                         const char* filename,
-                         const ProgressIndicator* parentProgress_)
+                             const char* filename,
+                             const ProgressIndicator* parentProgress_)
```
chambm, line 32: Spurious whitespace change?

.gitmodules
chambm, line 1: Doesn't this make the submodules unpinned so they always point to the latest HEAD? If I go into the submodule and change its branch, and forget I did so, the submodules may never update. Seems they should be set to branch=. or branch=master?

chambm: And that's not even considering pinning to a specific commit in the branch. Which is a serious consideration for Skyline's release branch.

bspratt: Maybe it's better to just straight up copy the code into the pwiz repo?

pwiz_tools/BiblioSpec/src/BuildParser.cpp
bspratt: Yeah, that's evil, thanks for pointing it out

pwiz_tools/BiblioSpec/src/BlibMaker.cpp
bspratt: Fair enough, it should just follow the pattern of the rest of the arglist handling.

.gitmodules
chambm: The submodule approach is more efficient because it means that all the non-Skyline builds for TeamCity won't have to checkout all the code in the submodules.

What I didn't realize before is that changes to the submodules will show up as changes in the parent tree, e.g. what branch is checked out or what commit is checked out. This is what it looks like after I changed the submodule's origin back to mhoopmann and picked a branch:
image

So I think my concerns were misplaced. I'm not sure where this information is getting saved since it's not in .gitmodules, but it is saved.

pwiz_tools/BiblioSpec/src/PSM.h
```diff
@@ -107,20 +107,49 @@ struct PSM{
...
+  }
+
   bool IsCompleteEnough() const
   {
       return (specKey >= 0 || specIndex >= 0 || !specName.empty()) && 
           (smallMolMetadata.IsCompleteEnough() ||
            (charge != 0 && !unmodSeq.empty()));
   }
+  
+  int idAsInteger() const
```
nickshulman, line 136: I could not find this method being used anywhere. I probably would have named it "getHashCode".

pwiz_tools/Shared/CommonUtil/SystemUtil/ProcessRunner.cs
```diff
@@ -84,23 +86,24 @@ public static bool GoodIfExitCodeIsZero(string stderr, int exitCode)
...
 
             _messageLog.Clear();
+            var cmd = $"{psi.FileName} {psi.Arguments}";
 
             // Optionally create a subdir in the current TMP directory, run the new process with TMP set to that so we can clean it out afterward
             _tmpDirForCleanup = forceTempfilesCleanup ? SetTmpDirForCleanup(psi) : null;
 
             Process proc = null;
-            var msgFailureStartingCommand = @"Failure starting command ""{0} {1}"".";
+            var msgFailureStartingCommand = @"Failure starting command ""{cmd}"".";
```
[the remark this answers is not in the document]
bspratt: Oops, thanks.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -810,6 +860,159 @@ private void NextPage()
...
+                    docTmp = docTmp.ChangeSettings(docTmp.Settings.ChangeTransitionSettings(newTransitionSettings)); // Update the settings
+                    // Now apply auto pick so that M+1, M+2 etc precursors get created 
+                    docTmp = ImportPeptideSearch.ChangeAutoManageChildren(docTmp, PickLevel.precursors | PickLevel.transitions, true);
+
+                    // Copy the resulting nodes to the actual doc, and update its settings too
+                    docNew = ((SrmDocument)docNew.Add(docTmp.MoleculeGroups.First())).
+                        ChangeSettings(docTmp.Settings.ChangeTransitionSettings(newTransitionSettings));
+                }
+
+                progressMonitor.UpdateProgress(status.ChangePercentComplete(100));
```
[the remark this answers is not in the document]
bspratt: I think it's correct as written - it's trying to determine what new nodes would be added to the original doc on a per-new-library basis, and iteratively adding those to a new aggregate doc.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -810,6 +860,159 @@ private void NextPage()
...
+            ImportPeptideSearch.SearchEngine.SetEnzyme(Document.Settings.PeptideSettings.Enzyme,
+                Document.Settings.PeptideSettings.DigestSettings.MaxMissedCleavages);
+            ImportPeptideSearch.SearchEngine.SetSpectrumFiles(BuildPepSearchLibControl.DdaSearchDataSources);
+            ImportPeptideSearch.DdaConverter?.SetSpectrumFiles(BuildPepSearchLibControl.DdaSearchDataSources);
+            ImportPeptideSearch.SearchEngine.SetFastaFiles(ImportFastaControl.FastaFile);
+            SearchControl.SearchFinished -= SearchControlSearchFinished;
+            SearchControl.SearchFinished -= SearchControlSearchStepFinished;
+            if (ImportPeptideSearch.RemainingStepsInSearch > 1)
+            {
+                SearchControl.SearchFinished += SearchControlSearchStepFinished; // Two step search, e.g. Hardklor then Bullseye
```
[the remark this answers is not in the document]
bspratt: This is ideally all going to get reverted anyway in favor of a totally separate parallel execution pattern, so I think I'll not put any more time into the current logic (which is, yes, messy, but was already pretty messy going in).

pwiz_tools/BiblioSpec/src/BuildParser.cpp
```diff
@@ -329,6 +330,28 @@ sqlite3_int64 BuildParser::insertProtein(const Protein* protein) {
...
 
+// Optionally sort the psms before writing
+void BuildParser::OptionalSort(PSM_SCORE_TYPE scoreType)
+{
+    if (scoreType == PSM_SCORE_TYPE::HARDKLOR_CORRELATION_SCORE)
+    {
+        // Sort PSMs by mass before writing to library
+        std::stable_sort(psms_.begin(), psms_.end(), [](PSM* a, PSM* b)
+        {
+            if (a == NULL || b == NULL)
```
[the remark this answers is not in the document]
bspratt: The thinking is that there's no point moving a NULL relative to anything else since it's ultimately ignored. I'll improve the comment.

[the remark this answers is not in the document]
bspratt: OK, on reflection I see your point - consistency is necessary. I'll fix that, thanks.

pwiz_tools/BiblioSpec/src/PSM.h
bspratt: Yeah, that's cruft, thanks

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -810,6 +860,159 @@ private void NextPage()
             UpdateButtons();
         }
 
+        private void AddDetectedFeaturesToDocument(IProgressMonitor progressMonitor)
+        {
+            // Add the library molecules to the document
+            var status = new ProgressStatus(PeptideSearchResources.ImportPeptideSearchDlg_NextPage_Adding_detected_features_to_document);
+            progressMonitor.UpdateProgress(status);
+            Assume.IsTrue(!Equals(Document.Settings.PeptideSettings.Libraries.LibrarySpecs, _existingLibraries.LibrarySpecs));
```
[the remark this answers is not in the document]
bspratt: Yes, better.

pwiz_tools/BiblioSpec/tests/working-with-tests.txt
```diff
@@ -12,11 +12,13 @@ in detail.
...
-All input files are in inputs.tar.bz2.  They are automatically
+Most input files are in inputs.tar.bz2 although recently we have been
+adding them directly to git.  Those that are in the bz2 file are automatically
 unpacked into BiblioSpec/tests/inputs by bjam as part of the
-proteowizard build process. If you add or change an input file, be
+proteowizard build process. If you change an input file from the bz2 file, be
 sure to update the tar file and check the new version into the
-repository.
+repository. For new work just add the file directly to git, it's much easier
+to track changes that way.
```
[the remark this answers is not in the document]
bspratt: Agreed

brendanx67:
This does look way better. Retention time alignment looks great. I think it is still relatively easy for a trained eye to see where things are going wrong, but that is also possible with DDA data that has the benefit of searched MS/MS spectra. I would really like to see this merged next week, and then we can get on to fixing polish issues and making the initial feature finding run in parallel for multiple files, as Matt did for EncyclopeDIA.

image

rita-gwen:
It looks like this PR changed some default settings values in the Settings.settings file, but it did not include the app.config changes. I'm having this warnings from Visual Studio that wants to update the settings file with the outdated default values it found in app.config.

bspratt:
Hi Rita,

I don't understand the issue (it's not you, it's me) - can you show a
screenshot? Or, if you understand the problem already, maybe offer a fix if
it's not too complicated?

Thanks

Brian

PR #2499, opened 2023-02-14
No zip files in SkylineTester.zip

chambm:
* commented out inclusion of zip files in SkylineTester.zip to decrease its size from 600MB to 100MB

### DIFF
--- a/pwiz_tools/Skyline/SkylineTester/CreateZipInstallerWindow.cs
+++ b/pwiz_tools/Skyline/SkylineTester/CreateZipInstallerWindow.cs
@@ -186,8 +186,12 @@ namespace SkylineTester
                             AddFile(file, zipFile);
                     }
 
+                    // MCC 2/14/2023: commented out adding test zips after discussion with Brendan that if we need
+                    // to test Skyline outside a source tree, we can find a way to get the zip files on the fly or
+                    // have a separate artifact for that. This will cut SkylineTester.zip from ~600MB to ~100MB.
+
                     // Add test zip files.
-                    var zipFilesList = new List<string>();
+                    /*var zipFilesList = new List<string>();
                     FindZipFiles(solutionDirectory, zipFilesList);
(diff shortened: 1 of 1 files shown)


pwiz_tools/Skyline/SkylineTester/CreateZipInstallerWindow.cs
```diff
@@ -186,8 +186,12 @@ public static void CreateZipFile(string zipPath)
                             AddFile(file, zipFile);
                     }
 
+                    // MCC 2/14/2023: commented out adding test zips after discussion with Brendan that if we need
+                    // to test Skyline outside a source tree, we can find a way to get the zip files on the fly or
+                    // have a separate artifact for that. This will cut SkylineTester.zip from ~600MB to ~100MB.
+
                     // Add test zip files.
-                    var zipFilesList = new List<string>();
+                    /*var zipFilesList = new List<string>();
```
brendanx67, line 194: Instead, please add an argument to the function for whether or not to include the test files, and have the command-line pass in false. That will make it simpler to connect functionality that re-enables this feature, either by adding a command-line argument to make the value true, or by connecting this form and adding a checkbox for whether or not to include test data.

brendanx67 (changes requested):
I was hoping we could keep the functionality available to a developer, but avoid doing this on TeamCity. This change no requires a developer to find and uncomment this code. I can see that the UI around this is not fully implemented, but leave the code uncommented with a flag controlling whether or not test data is included for a clearer future direction for the feature you are removing.

pwiz_tools/Skyline/SkylineTester/CreateZipInstallerWindow.cs
chambm: OK. Do you want me to implement the UI, or just add the function param?

brendanx67:
Just the function parameter is fine. I just don't want to see the code
commented out, and leave it with an obvious path to re-implementing the
feature.

brendanx67 (approved):
Looks good.

PR #2501, opened 2023-02-15
Associate proteins description fix

nickshulman:
Fix bug where protein descriptions assigned by "Associate Proteins" would disappear when the document was saved and reloaded (reported by Mike)

[Replace class "FastaData" with a new class "FastaParser" which returns an enumeration of FastaSequence objects. FastaParser gets used by both "File > Import Fasta" and "Refine > Associate Proteins"]


chambm (approved):
Looks great, thanks!

PR #2503, opened 2023-02-15
Stop rounding off predicted retention times to 2 decimal places

nickshulman:
Fixed to preserve all decimal places of predicted retention times (requested by Will)

Will requested that we stop truncating predicted retention times to two decimal places:
https://skyline.ms/announcements/home/support/thread.view?rowId=59574

I can't think of a reason that we would want to be truncating these values to two decimal places. The majority of places in the code were not passing in "true" for fullPrecision to the GetRetentionTime methods. The few places that actually were going to display the time to the user already had a mechanism in place to format the value to fewer decimals. I cannot think of any reason we would want to be doing our calculations on truncated predicted retention times, so I think we should get rid of the "fullPrecision" parameter.


brendanx67 (approved):
Seems fine to me as long as I don't end up seeing full precision values in the UI. Though, I agree that that should be handled with display formatting and not in a way that might impact internal calculations.

PR #2505, opened 2023-02-16
Improved support for working with paths that contain spaces, or legal but unusual characters like caret and ampersand.

bspratt:
Among other things, while running tests, when Skyline generates temporary filenames potentially troublesome characters like space, caret and ampersand are included in the filename.

I also changed the name of my test data downloads directory to include these characters.

We now escape such characters when passing arguments to msconvert's diaUmpire filter, as it requires nested quotes.


nickshulman:
I think we should write a class called "ArgumentList" which is a list of unescaped arguments and whose ToString() method produces a string will all of the escaped arguments in a row.

pwiz/analysis/spectrum_processing/SpectrumListFactory.cpp
```diff
@@ -668,6 +688,26 @@ UsageInfo usage_precursorRecalculation = {"","This filter recalculates the precu
...
 
+inline void unescapeQuotedPath(string &path)
+{
+    if ((bal::starts_with(path, "\'") && bal::ends_with(path, "\'")) || (bal::starts_with(path, "\"") && bal::ends_with(path, "\"")))
+    {
+        // Remove the first and last characters
+        path = path.substr(1, path.length() - 2);
+    }
+    // Look for the ^ escape character
+    for (size_t escapePos = 0;;)
```
chambm, line 699: Is this necessary? I would expect the Windows shell to take the `^`s out (but not the escaped quotes). That's the case with that ugly titleMaker filter escaping I sent to you.

pwiz_tools/Skyline/Util/UtilIO.cs
```diff
@@ -946,6 +946,21 @@ public static string GetElapsedTimeExplanation(DateTime startTime, DateTime endT
...
+        /// Like Path.GetRandFileName(), but in test context adds some legal but unusual characters for robustness testing
+        /// </summary>
+        /// <returns>a random filename</returns>
+        public static string GetRandomFileName()
+        {
+            var result = Path.GetRandomFileName();
+            if (Program.UnitTest)
+            {
+                // Introduce some potentially troublesome characters like space, caret, ampersand to test our handling
+                result = @"test c^ha&rs_" + result;
```
chambm, line 960: I like this function. How about some Unicode as well: 试验

brendanx67:
> I think we should write a class called "ArgumentList" which is a list of unescaped arguments and whose ToString() method produces a string will all of the escaped arguments in a row.

Sounds okay to me, as long as we don't over-quote the arguments, e.g. putting quotes around every argument, which would work, but would look really lame when we show the command-lines to users.

pwiz/analysis/spectrum_processing/SpectrumListFactory.cpp
bspratt: This is for dealing with what's between the nested quotes, which windows shell doesn't mess with. But I suppose that implies that we maybe didn't need to escape any of that stuff while constructing the command line. I'll have to review.

pwiz_tools/Skyline/Util/UtilIO.cs
bspratt: Great idea!

bspratt:
> I think we should write a class called "ArgumentList" which is a list of unescaped arguments and whose ToString() method produces a string will all of the escaped arguments in a row.

Not following you, sorry

pwiz/analysis/spectrum_processing/SpectrumListFactory.cpp
chambm: You DO need to escape it when making the command-line in Skyline, but I'm hoping you don't need to process it inside the called exe (msconvert). Here's the titleMaker escaping even though it's inside the nested quotes:
```
--filter "titleMaker <RunId>.<ScanNumber>.<ScanNumber>.<ChargeState> File:"""^<SourcePath^>""", NativeID:"""^<Id^>""""
```
What msconvert sees: 
```
titleMaker <RunId>.<ScanNumber>.<ScanNumber>.<ChargeState> File:"<SourcePath>", NativeID:"<Id>"
```
But I concede it could be different if it's nested inside another level of quoting.

brendanx67: The escape characters should not show up inside the program. They are to direct the shell to give the correct characters to the program. If you handle them again side the program, you will be double-escaping, I think.

brendanx67: double-unescaping, I meant

bspratt:
Oh, OK I get it. Though I'm not sure how to generalize the weirdness of the arguments to msconvert, with the nested quotes and all.

brendanx67:
I think you would just have the functions "QuoteIfNecessary()" and
"EscapeIfNecessary()". If the Argument text is:

dia-umpire params="C:\Users\Test & Dev\Tmp\foobar.txt"

Then these functions would turn it into:

"dia-umpire params=\"C:\Users\Test ^& Dev\foobar.txt\""

Though, even the first part might be best handled by a
MsConvertFilterArgumentsList class with a FilterName and argument type
classes like NameAndValue.

Nick makes a good point that classes and polymorphism could do a lot here,
as they have for our own CommandArgs inside the Skyline Command Line
Interface.

bspratt:
Just wanted to emphasize "I tried also adding some unicode (e.g. 试验, means "test") but it just breaks too many tests that involve unicode-ignorant 3rd party tools, mz5 files, etc."

I suppose we could add a flag for tests that we know won't work with unicode for reasons that are out of our control...

chambm:
> Just wanted to emphasize "I tried also adding some unicode (e.g. 试验, means "test") but it just breaks too many tests that involve unicode-ignorant 3rd party tools, mz5 files, etc."
> 
> I suppose we could add a flag for tests that we know won't work with unicode for reasons that are out of our control...

How many tests were broken that could be reasonable fixed (we control all the code) vs. those that can't? If the latter is a much greater number, then yeah it's probably not worth it.

bspratt:
At least a half dozen - MSFragger and anything using mz5 were the main culprits I observed. I can give it a shot.

pwiz_tools/Shared/Common/SystemUtil/ProcessRunner.cs
```diff
@@ -203,5 +204,15 @@ public IEnumerable<string> MessageLog()
         {
             return _messageLog;
         }
+
```
brendanx67, line 207: Feels like this should go into a new class named "ProcessEx". It is not really appropriate here. This class is about running another process not for asking questions about the currently running process.

pwiz_tools/Skyline/Util/Xml.cs
```diff
@@ -961,6 +961,18 @@ public static string UnescapeNonPrintableChars(this string str)
             return sb.ToString();
         }
 
+        // Inspect a file path for characters that must be escaped for use in XML (currently just "&")
```
brendanx67, line 964: I could imagine putting this on PathEx also as EscapePathForXml() with EscapePathForCommandLine()... Or moving EscapePathForCommandLine() to ProcessRunner.EscapePath(). Feels unbalanced the way it is.

brendanx67 (changes requested):
A few comments about where things live and definitely need to consider unescaping of ^ inside msconvert.exe. I think that escaping is all about getting the right text into the process. When that is right, then it should not contain any extra ^ characters that need to be removed.

pwiz_tools/Shared/Common/SystemUtil/ProcessRunner.cs
bspratt: OK, will do.

pwiz_tools/Skyline/Util/Xml.cs
bspratt: I can do that, sure.

pwiz_tools/Skyline/Util/UtilIO.cs
```diff
@@ -946,6 +946,23 @@ public static string GetElapsedTimeExplanation(DateTime startTime, DateTime endT
                 return elapsedSpan.TotalMilliseconds + @" milliseconds";
             return deltaTicks + @" ticks";
         }
+
+        /// <summary>
+        /// Like Path.GetRandFileName(), but in test context adds some legal but unusual characters for robustness testing
+        /// </summary>
+        /// <returns>a random filename</returns>
+        public static string GetRandomFileName()
```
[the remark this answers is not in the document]
bspratt: OK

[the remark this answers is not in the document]
bspratt: Hmm, problem is we don't normally get fancy with random filenames unless we're in a test. And that's not information PathEx has, over in pwiz.common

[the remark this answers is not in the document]
brendanx67: That makes me a little queasy that we are choosing code location to access internal test settings. I guess I would prefer something like a static on PathEx that is the RandomFile root and if it is not set, we just pass through to the normal Path.GetRandomFileName(). And the test framework sets it. That way these fairly low level classes don't need to reach outside themselves to look at test settings. Instead, the test framework pokes them into doing what it needs.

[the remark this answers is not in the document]
bspratt: OK, I'll see what I can do.

bspratt:
I've set this as the Integration branch, and set BSPRATT-UW3 to run it as its Nightly.

bspratt:
@nickshulman @brendanx67 @chambm last chance to comment before I merge this, with this commit message:

**Improved support for working with paths that contain spaces, or legal but unusual characters like caret and ampersand, and for the issue of stray files being left behind by various tests and 3rd party tools**

* Made changes to the Skyline automated testing scheme to improve our support for users with windows usernames containing spaces an other legal but unusual characters. This is achieved by setting the TMP environment variable to point at a directory whose name contains such characters. Normally that value is based on the user name, so even on a system with username "ihavenospacesinmyname" we can test running with a TMP dir with a name that contains space, caret, and ampersand.

* We now properly escape quotes when passing arguments to msconvert's diaUmpire filter, as it requires nested quotes.

* Additional cleanup for Crux pepXML output that doesn't properly escape the & in filenames

* Added a custom attribute for tests that don't do well with unicode paths, usually due to mz5 or 3rd party tool usage. Tests without the NoUnicodeTesting(string reason) will run with TMP and results directories whose names include unicode characters.

* Check for cleanup of our temp directory on test exit

* Explicitly clean up MSAmanda's tempfiles, which are left on purpose but which violate the general expectations of test cleanup.  Added a couple of TODO comments about how we need to think more about MSAmanda leaves tempfiles around.

* ProcessRunner.Run now accepts an optional argument "forceTempfilesCleanup" which, when true, runs the process with an altered environment in which TMP is a subdirectory of the original temp directory. Afterward the directory is deleted.

* added pwiz.Common.SystemUtil.TempDir class to make it easier to clean up after things like ExcelReaderFactory that leave tempfiles behind

* Custom tools that write to the immediate window weren't cleaning up their temp files.

* Do a better job of cleaning up tempfiles when an MSFragger search is interrupted

pwiz_tools/Skyline/TestPerf/DdaTutorialTest.cs
```diff
@@ -41,7 +41,7 @@ namespace TestPerf
     [TestClass]
     public class DdaTutorialTest : AbstractFunctionalTest
     {
-        [TestMethod, NoParallelTesting(TestExclusionReason.RESOURCE_INTENSIVE)]
+        [TestMethod, NoParallelTesting(TestExclusionReason.RESOURCE_INTENSIVE), NoUnicodeTesting("MsFragger does not handle unicode paths")]
```
brendanx67, line 44: These should probably be constants on TestExclusionReason. They appear to get repeated.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
```diff
@@ -332,6 +338,34 @@ public bool Run(TestInfo test, int pass, int testNumber, string dmpDir, bool hea
                     Environment.SetEnvironmentVariable(@"SKYLINE_TESTER_PARALLEL_CLIENT_ID", ParallelClientId); // Accessed in pwiz_tools\Skyline\Util\Util.cs
                 }
 
+                // Set the temp file path to something peculiar - helps guarantee support for
```
brendanx67, line 341: Put this in a function instead of making a BIG function bigger.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
```diff
@@ -517,29 +554,55 @@ public bool Run(TestInfo test, int pass, int testNumber, string dmpDir, bool hea
...
+        }
+
         /// <summary>
         /// Resets the test directory to empty when cleanupLevel is 'all'.
         /// </summary>
         /// <returns>A list of the entries, if any, that were removed</returns>
-        private List<string> CleanUpTestDir()
+        private List<string> CleanUpTestDir(string tmpTestDir, bool final)
         {
+            CleanupMSAmandaTmpFiles();
```
brendanx67, line 579: Put this in the tests that need it or put a TODO comment on it, if these still get created all the time.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
```diff
@@ -517,29 +554,55 @@ public bool Run(TestInfo test, int pass, int testNumber, string dmpDir, bool hea
...
             var allEntries = new List<string>();
             // If everything is supposed to be cleaned up, then check for any left over files
-            if (_cleanupLevelAll && Directory.Exists(TestContext.TestDir))
+            if (_cleanupLevelAll)
             {
-                allEntries.AddRange(Directory.EnumerateFileSystemEntries(TestContext.TestDir));
-                if (allEntries.Count > 0)
+                foreach (var dir in new []{TestContext.TestDir, tmpTestDir})
```
brendanx67, line 584: Use sub-functions instead of loops around an existing body of code.

brendanx67 (changes requested):
Definitely getting there! Lots of issues solved. Still have a few suggestions.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
chambm: If this gets moved to a function it needs to be a functor with a Dispose call that resets the TMP variable (as is done later in this function). But having that environment-restoring behavior is probably a good idea anyway if the code throws between the time TMP is redirected and when it is restored.

brendanx67: Or the function needs to return tmpTestDir?

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
bspratt: There's already one in the function itself, FWIW, I'll add another here.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
bspratt: Fair enough. Updated.

pwiz_tools/Shared/Common/SystemUtil/ProcessRunner.cs
```diff
@@ -53,23 +53,48 @@ public class ProcessRunner : IProcessRunner
...
             psi.RedirectStandardOutput = true;
             psi.RedirectStandardError = true;
             psi.RedirectStandardInput = stdin != null;
             if (OutputEncoding != null)
                 psi.StandardOutputEncoding = psi.StandardErrorEncoding = OutputEncoding;
+            string tmpDirForCleanup = null;
 
             _messageLog.Clear();
 
+            if (forceTempfilesCleanup)
```
brendanx67, line 74: This appears to be another place where a sub-function might improve readability rather than adding a new chunk of code to an already large function dedicated to a single issue.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
```diff
@@ -517,34 +530,98 @@ public bool Run(TestInfo test, int pass, int testNumber, string dmpDir, bool hea
...
+            CleanupMSAmandaTmpFiles();  // TODO(MattC): tidy up MSAmanda implementation so that we can distinguish intentional uses of tmp dir (caching potentially re-used files) from accidental directory creation and/or not-reused files within
+            var abandonedFilesList = new List<string>();
             // If everything is supposed to be cleaned up, then check for any left over files
-            if (_cleanupLevelAll && Directory.Exists(TestContext.TestDir))
+            if (_cleanupLevelAll)
             {
-                allEntries.AddRange(Directory.EnumerateFileSystemEntries(TestContext.TestDir));
-                if (allEntries.Count > 0)
+                CleanupAbandonedFiles(TestContext.TestDir, true, abandonedFilesList);
+                CleanupAbandonedFiles(tmpTestDir, !final, abandonedFilesList);
```
brendanx67, line 596: I guess I wonder about this placement. It would seem to imply that we are fine with leaving temp files around when _cleanupLevelAll is false, as we really are with leaving around the files tests produce, since they can be very helpful artifacts when a test fails. I could imagine we would want to see the temp files go away no matter what the clean-up level. Are we expecting normal use to leave temp files around?

brendanx67 (commented):
Thanks for the improvements. Couple last issues.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
chambm: It would not be atypical program behavior at least. Just look at all the cruft that builds up in %TEMP%. :) But I think if Skyline is creating a temp file that is supposed to be cleaned up during execution but maybe isn't due to a bug or error, it's ok to clean it up always. If we want to keep a file around for understanding why an error is happening, it's better to have it be a non-temp file (like I will move the DDA params files to the output directory instead of TEMP).

brendanx67: Right. I think that means that the tmpTestDir check should go outside the if block, so that we check for clean-up regardless of whether the test settings are set to clean up all files.

bspratt: I agree.

pwiz_tools/Shared/Common/SystemUtil/ProcessRunner.cs
bspratt: No problem.

bspratt:
OK, most recent push should address these concerns.

brendanx67 (approved):

PR #2508, opened 2023-02-16
Update WIFF2 SDK to support all(?) instruments

chambm:
I don't know if it's really ALL instruments that can make WIFF2s, but it's a lot more than just the X500 now.

Although the new SDK can read MRM chromatograms, it can't tell you the Q3 mass for the chromatograms it gives you. I have reported that defect to Sciex. In the meantime I added a warning that WIFF2 MRM/SIM data should point at the WIFF file instead, which led me to discover that MSConvertGUI wasn't displaying the stderr warnings that msconvert.exe shows. I spent a lot of time trying to read the stderr on the .NET side without modifying the C++ side; it might be possible but I couldn't figure it out, so there's a new C function in the binding that will redirect cerr/cout output to a callback function. This might also be useful in Skyline when importing results.

### DIFF
--- a/pwiz/data/msdata/SpectrumListBase.cpp
+++ b/pwiz/data/msdata/SpectrumListBase.cpp
@@ -24,6 +24,7 @@
 
 #include "SpectrumListBase.hpp"
 #include "pwiz/utility/misc/String.hpp"
+#include "pwiz/utility/misc/Stream.hpp"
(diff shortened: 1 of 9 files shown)


chambm:
@bspratt @nickshulman Can you set this as the integration branch for a few days to see if it causes any leaks or other new issues?

PR #2513, opened 2023-02-24
Fixes a medium size annoyance for developers using parallel test in Skyline Tester…

bspratt:
…. I, at least, often realize that I've kicked of the wrong selection of tests. So, Stop the tests, make the selection, then either wait for (or manually kill) the workers so a new set can be spun up. Or forget to wait and get a bunch of errors.

With this change, each round of parallel tests uses unique worker names so this conflict never happens.

image

### DIFF
--- a/pwiz_tools/Skyline/TestRunner/Program.cs
+++ b/pwiz_tools/Skyline/TestRunner/Program.cs
@@ -658,8 +658,8 @@ namespace TestRunner
         private static string LaunchDockerWorker(int i, CommandLineArgs commandLineArgs, ref string workerNames, bool bigWorker, long workerBytes, int workerPort, StreamWriter log)
         {
             var pwizRoot = Path.GetDirectoryName(Path.GetDirectoryName(GetSkylineDirectory().FullName));
-            var invocationDetail = GetInvocationDetail(); // Avoids conflicts between this and any previous invocation
-            string workerName = bigWorker ? $"docker_big_worker{invocationDetail}_{i}" : $"docker_worker{invocationDetail}_{i}";
+            // Adding timestamp to worker name voids conflicts between this and any previous invocation
+            string workerName = bigWorker ? $"docker_big_worker{GetTestRunTimeStamp()}_{i}" : $"docker_worker{GetTestRunTimeStamp()}_{i}";
             string dockerRunRedirect = string.Empty;
             string testRunnerLog = @$"c:\AlwaysUpCLT\TestRunner-{workerName}.log";
             if (commandLineArgs.ArgAsBool("keepworkerlogs"))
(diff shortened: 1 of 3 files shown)


pwiz_tools/Skyline/TestRunner/Program.cs
```diff
@@ -675,7 +677,7 @@ private static string LaunchDockerWorker(int i, CommandLineArgs commandLineArgs,
             testRunnerCmd = AddPassThroughArguments(commandLineArgs, testRunnerCmd);
             testRunnerCmd += $" workerport={workerPort}";
 
-            string dockerArgs = $"run --name {workerName} -it --rm -m {workerBytes}b -v {PathEx.GetDownloadsPath()}:c:\\downloads -v {pwizRoot}:c:\\pwiz {RunTests.DOCKER_IMAGE_NAME} \"{testRunnerCmd} workername={workerName}\" {dockerRunRedirect}";
```
chambm, line 678: I don't think a new TestRunner argument is needed. You can change the workerName to include the TestRunner process ID:
```
var pid = Process.GetCurrentProcess().Id;
string workerName = bigWorker ? $"docker_big_worker_{pid}_{i}" : $"docker_worker_{pid}_{i}";
```
Theoretically the PID could be re-used from the last TestRunner but that seems very unlikely. But if that's a concern, could use a timestamp instead:
```
var timestamp = DateTime.UtcNow.ToShortTimeString();
string workerName = bigWorker ? $"docker_big_worker_{timestamp}_{i}" : $"docker_worker_{timestamp}_{i}";
```

chambm (changes requested):

pwiz_tools/Skyline/TestRunner/Program.cs
bspratt: Simpler is better, I'll try this out.

brendanx67:
I was hoping it was giong to fix the issue I sometimes see with Docker
Containers hanging "forever" or until I manually closed and even delete
them. This still seems to happen and I would say more frequently at the end
of a full test run followed by this case of stopping tests in the middle.

This is obviously an improvement, though. So, once Matt is happy with it,
that works for me.

bspratt:
@chambm can you give this one more look before I merge? Thanks

pwiz_tools/Skyline/TestRunner/Program.cs
```diff
@@ -656,7 +658,8 @@ private static void CheckDocker(CommandLineArgs commandLineArgs)
         private static string LaunchDockerWorker(int i, CommandLineArgs commandLineArgs, ref string workerNames, bool bigWorker, long workerBytes, int workerPort, StreamWriter log)
         {
             var pwizRoot = Path.GetDirectoryName(Path.GetDirectoryName(GetSkylineDirectory().FullName));
-            string workerName = bigWorker ? $"docker_big_worker_{i}" : $"docker_worker_{i}";
+            var invocationDetail = GetInvocationDetail(); // Avoids conflicts between this and any previous invocation
```
chambm, line 661: GetInvocationDetail() seems like a pretty vague name. Why not `var timestamp = GetWorkerNameTimestamp()`" or something self-documenting like that? But this is nitpicking.

chambm (approved):

pwiz_tools/Skyline/TestRunner/Program.cs
bspratt: I was going for not exposing implementation details, but your way is good too, I'll go ahead and call it something more descriptive.

PR #2514, opened 2023-02-25
fix an annoyance in SkylineTester when switching back and forth from parallel mode …

bspratt:
… - formerly the "Offscreen" setting would be forced True. It's left alone now, and its control is replaced with a text hint when parallel mode is selected


bspratt:
Feel free to tweak any language then merge to master

brendanx67 (commented):
Could this possibly be missing changes to a RESX file? I can't see anything that would set the text of the new label. What am I missing?

bspratt:
Yes, I must have messed up when trying to manually select just the meaningful changes in the salad that the IDE makes when changing a form's contents. I'll sort this out, thanks

PR #2516, opened 2023-02-27
Fix hard crash when reading corrupt SQLite file

chambm:
- fixed hard crash when reading corrupt SQLite file (reported by An)

Throwing exceptions in destructors...very troublesome! And yes Nick that SQLite TDF file is corrupt.


chambm:
Still working on this. Not really happy with the exception throwing logic changes: I don't have a way to unset the throwingException_ flag after an exception is handled...

Why trigger a TC build?

PR #2517, opened 2023-02-28
Fix intermittent failure in TestMidas test.

nickshulman:
Change the code so that it passes a SettingsChangeMonitor into the call to "ChangeMeasuredResults".

The SettingsChangeMonitor notices when the SrmDocument in the IDocumentContainer changes, and will cancel the operation. This prevents the settings update code from trying to read from a scale .skyd file.

### DIFF
--- a/pwiz_tools/Skyline/Model/Lib/Library.cs
+++ b/pwiz_tools/Skyline/Model/Lib/Library.cs
@@ -26,6 +26,7 @@ using System.Text;
 using System.Xml;
 using System.Xml.Schema;
 using System.Xml.Serialization;
+using JetBrains.Annotations;
 using pwiz.BiblioSpec;
 using pwiz.Common.Chemistry;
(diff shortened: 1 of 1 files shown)


bspratt:
Most elegant! Might there be other applications for this - other background loaders etc?

nickshulman:
> Most elegant! Might there be other applications for this - other background loaders etc?

Interesting idea. I see that several other BackgroundLoader's do call "ChangeSettings" or "ChangeMeasuredResults", but most of them are not passing in a SettingsChangeMonitor. (ChromatogramManager does pass in one).
This is only an issue if the ChangeSettings call is going to result in all of the results in the TransitionGroupDocNode's being updated. Usually a ChangeSettings is so fast that it does not need a SettingsChangeMonitor.

brendanx67 (approved):

PR #2521, opened 2023-03-02
Avoid absurd high energy offset values in ion mobility library "Find Results"

bspratt:
For drift tube IM, these can only ever be negative values. For PASEF, this isn't normally used at all, but if user insists limit offset to within +-5% of MS1 IM value. For FAIMS we already just don't allow for it.

Prompted by user Premy in https://skyline.ms/announcements/home/support/thread.view?rowId=59707
Also fixes issue https://skyline.ms/issues/home/issues/details.view?issueId=751&_docid=issue%3A751 which refers to https://skyline.ms/announcements/home/support/thread.view?rowId=48100


bspratt:
@chambm could you eyeball this for me, please? Nothing too controversial here, other than what I did for PASEF which probably doesn't matter anyway.

chambm (approved):

PR #2526, opened 2023-03-06
Try to fix a race condition that only shows up on TeamCity code coverage build…

bspratt:
…, by not removing temp files associated with Skyline Immediate window until that window is Disposed. Originally we didn't try to delete them at all, my recent commit deleted them when the tool writing to the window exited but that apparently introduced a race condition. So now they will accumulate during a Skyline run but should not remain afterward, which is still an improvement.

### DIFF
--- a/pwiz_tools/Skyline/Controls/ImmediateWindow.cs
+++ b/pwiz_tools/Skyline/Controls/ImmediateWindow.cs
@@ -299,12 +299,6 @@ namespace pwiz.Skyline.Controls
                     if (_indexesInUse.IndexInRange(index) && _indexesInUse.Get(index))
                         _indexesInUse.Set(index, false);
                 }
-
-                foreach (var tempFile in _tempFiles)
-                {
-                    FileEx.SafeDelete(tempFile, true); // Don't throw if file can't be deleted
-                    DirectoryEx.SafeDelete(tempFile); // In case that was actually a directory name 
-                }
             }
         }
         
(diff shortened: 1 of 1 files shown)


bspratt:
Note that this won't truly be tested until merged to master, since that's the only branch on which TC runs the code coverage test. But first let's make sure nothing breaks on this branch.

chambm:
I added this PR to the branch filter for TC.

bspratt:
> I added this PR to the branch filter for TC and manually triggered a build.

Thanks!

bspratt:
> Fixed the broken test but broke two new ones?

Maybe? Though I don't see how, flow of control doesn't go into the area I changed. (Verified with breakpoints on my local machine)

chambm:
The pwiz core was clearly a sporadic error that resolved with a rerun. I'll rerun the code coverage build to see if those were sporadic too.

bspratt:
> The pwiz core was clearly a sporadic error that resolved with a rerun. I'll rerun the code coverage build to see if those were sporadic too.

I'd done the same, can't hurt to have two looks at the ball though.

bspratt:
@chambm They both passed, I'm going to go ahead and merge to master if you'll sign off on the changes.

chambm:
We seem to be getting a lot more intermittent failures on TC lately. I wonder if it has to do with using different instance types and/or spot fleet.

chambm (approved):

PR #2529, opened 2023-03-14
Add property sheet to the FullScanViewer

rita-gwen:


pwiz_tools/Skyline/Model/GlobalizedObject.cs
```diff
@@ -139,7 +141,13 @@ public override bool Equals(object obj)
...
                 .ToDictionary(prop => prop.Name, prop => prop.GetValue(other));
 
-            return ArrayUtil.EqualsDeep(thisProps, otherProps);
+            var joinedValues = (from t in thisProps
+                join o in otherProps on t.Key equals o.Key
+                select new { t = t.Value, o = o.Value }).ToList();
+            if (joinedValues.Count != thisProps.Count)
+                return false;
+            var res = joinedValues.Any(tuple => !tuple.t.Equals(tuple.o));
+            return !res;
```
nickshulman, line 150: We usually do not override "Equals" and "GetHashCode" for objects that are mutable.
That is, a mutable object is considered equal to something else if and only if they are the same object.
Weird bugs happen if you use a mutable object as the key in a hash map and the object gets changed later.

(It looks like this "Equals" method is only used for testing. I would recommend writing a different method "GetPropertyValues" which the property names and values as an ImmutableList sorted by name, and you can use ImmutableList equals method.

pwiz_tools/Skyline/Model/GlobalizedObject.cs
```diff
@@ -156,12 +164,35 @@ public override int GetHashCode()
         }
 
         public string Serialize()
+        {
+            StringBuilder sb = new StringBuilder();
+            StringWriter sw = new StringWriter(sb);
```
nickshulman, line 169: You don't need the separate StringBuilder here variable here. You can just call the zero-argument StringWriter constructor and return stringWriter.ToString() at the end.

pwiz_tools/Skyline/Model/FullScanPropertiesRes.resx
```diff
@@ -277,8 +277,20 @@
...
+  </data>
+  <data name="HighEnergyOffset" xml:space="preserve">
+    <value>High Energy Offset</value>
+  </data>
+  <data name="Description_HighEnergyOffset" xml:space="preserve">
+    <value>High Energy Offset used to obtain the MS2 spectrum.</value>
+  </data>
   <data name="dotp" xml:space="preserve">
-    <value>dotp</value>
+    <value>Library dotp</value>
```
nickshulman, line 293: I think this should be "Library Dot Product", and the string for "idotp" down below should be "Isotope Dot Product".

nickshulman (commented):
Looks pretty good.
I see a "Spectrum Properties" menu item when I right-click on the Full Scan graph, but that menu item does not seem to work because the event handler in SkylineGraph.cs thinks it's supposed to do something with the Library Match window.

pwiz_tools/Skyline/Model/GlobalizedObject.cs
rita-gwen: There is a very little chance that these objects will be used as dictionary keys. Property values can be objects themselves, so the ImmutableList approach doesn't fit very well. Should I just use Serialize method and compare JSON strings?

rita-gwen:
Good catch on the menu, thank you!

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -1095,6 +1115,11 @@ private MsDataSpectrum GetSpectrum(Spectrum spectrum, int spectrumIndex)
                         msDataSpectrum.Precursors = ImmutableList.ValueOf(GetMs1Precursors(spectrum));
                     }
 
+                    msDataSpectrum.SourceFilePath = FilePath;
+                    msDataSpectrum.InstrumentInfo = GetInstrumentConfigInfoList().FirstOrDefault();
```
nickshulman, line 1119: The file might have more than one instrument configuration in it so you should get the instrument configuration that is specific for this spectrum.

You can get the "InstrumentConfiguration" object by looking at:
`spectrum.scanList.scans[0].instrumentConfiguration`
but then you would need to turn that into a MsInstrumentConfigInfo.

(There is some example code for looking at the instrument configuration on a spectrum in "pwiz_tools\SeeMS\Dialogs\SpectrumListForm.cs")

pwiz_tools/Skyline/Model/FullScanProperties.cs
```diff
@@ -0,0 +1,141 @@
...
+            if(spectrum == null)
+                return null;
+            var res = new FullScanProperties();
+            res.FilePath = spectrum.SourceFilePath;
+            res.FileName = Path.GetFileName(spectrum.SourceFilePath);
+            if (spectrum.PrecursorsByMsLevel.Any())
+            {
+                var precursor = spectrum.Precursors.FirstOrDefault();
+                res.PrecursorMz = precursor.PrecursorMz.HasValue ? precursor.PrecursorMz.Value.Value.ToString(Formats.Mz) : null;
+                res.Charge = spectrum.NegativeCharge ? @"-" : @"+" + precursor.ChargeState;
```
nickshulman, line 66: I noticed "precursor.ChargeState" is actually always null.
The following line of code should to be added to "MsDataFileImpl.CreatePrecursor":
`ChargeState = GetChargeStateValue(p),
`

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -540,6 +518,112 @@ private void CreateIonMobilityHeatmap()
...
+                        spectrumProperties.HighEnergyOffset = _msDataFileScanHelper.CurrentTransition?.IonMobilityInfo?.HighEnergyIonMobilityOffset.ToString();
+                }
+                else
+                {
+                    spectrumProperties.MzCount = spectra[0].Mzs.Length;
+
+                    var parts = _msDataFileScanHelper.MsDataSpectra[0].Id.Split('.'); // Check for merge.frame.start.stop from 3-array IMS data
+                    var id = parts.Length < 4
+                        ? _msDataFileScanHelper.MsDataSpectra[0].Id
+                        : string.Format(@"{0}.{1}-{0}.{2}", parts[1], parts[2], parts[3]);
```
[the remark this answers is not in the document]
rita-gwen: This was copied from the original code that put Id range into the graph title. I presume it was correct.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -540,6 +518,115 @@ private void CreateIonMobilityHeatmap()
...
+                    .Select(pre=> pre.PrecursorCollisionEnergy).Where(ce => ce.HasValue).Select(ce => ce.Value)
+                    .Distinct().ToArray();
+                if (ces.Length == 1)
+                    spectrumProperties.CE = ces[0].ToString(Formats.OPT_PARAMETER);
+
+                var precursor = _msDataFileScanHelper.ScanProvider.Transitions
+                    .FirstOrDefault(t => (t.Id as Transition)?.IonType == IonType.precursor);
+                if (precursor != null)
+                {
+                    spectrumProperties.PrecursorMz = precursor.PrecursorMz.Value.ToString(Formats.Mz);
```
nickshulman, line 542: `FullScanProperties.CreateProperties(spectra[0])` already set the precursor m/z based on the precursor in the spectrum.
It looks like this line of code is replacing that precursor value with the precursor of something in the document.
I don't think that's what users will want: they want to know what the mass spectrometer said the precursor was for the spectrum that they are looking at.
(same with collision energy and charge state)

pwiz_tools/Skyline/Model/FullScanProperties.cs
```diff
@@ -0,0 +1,141 @@
...
+        [Category("PrecursorInfo")] public string RetentionTime { get; set; }
+        [Category("PrecursorInfo")] public string CCS { get; set; }
+        [Category("PrecursorInfo")] public string IonMobility { get; set; }
+        [Category("PrecursorInfo")] public string IsolationWindow { get; set; }
+        [Category("AcquisitionInfo")] public string IonMobilityRange { get; set; }
+        [Category("AcquisitionInfo")] public string IonMobilityFilterRange { get; set; }
+        [Category("PrecursorInfo")] public string HighEnergyOffset { get; set; }
+        [Category("AcquisitionInfo")] public string ScanId { get; set; }
+        [Category("AcquisitionInfo")] public string CE { get; set; }
+        [Category("AcquisitionInfo")] public string MSStage { get; set; }
```
nickshulman, line 119: Everywhere else in Skyline, (e.g. "Settings > Peptide Settings > Quantification") I think we call this "MS Level" instead of "MS Stage".

pwiz_tools/Skyline/Model/FullScanProperties.cs
rita-gwen: This change actually breaks the DdaSearchTest. The number of proteins changes, so the assertion fails.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
rita-gwen: This is a "philosophical" question. Since the full scan viewer is not synced with the targets tree the user needs to know for what precursor in the document this spectrum was recorded. Maybe both should be shown, one in the Acquisition and the other on the Precursor sections.

pwiz_tools/Skyline/Model/FullScanProperties.cs
chambm: Oops. It seems like we must have never been passing charge states in the raw data through to MSAmanda. That would definitely change the results (and it should be better). Can you update the numbers? Use IsRecordMode => true to get all the numbers so it doesn't just fail at the first assert. Or I can do it.

pwiz_tools/Skyline/Model/FullScanProperties.cs
rita-gwen: Yes, I think MS Stage came from the SeeMS since it provided initial inspiration for this.

pwiz_tools/Skyline/Model/FullScanProperties.cs
rita-gwen: I think I revert it for now.

rita-gwen: Yes, sure.

nickshulman (commented):
I think this looks really good but that the values in the property sheet which are coming from the Skyline document should be removed until we can figure out how to implement them perfectly.
The properties that I would recommend removing for now are "Label", "HighEnergyOffset", "dotp", "idotp" and "rdotp".

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -540,6 +518,114 @@ private void CreateIonMobilityHeatmap()
...
+                    {
+                        var chromSet = stateProvider.DocumentUI.Settings.MeasuredResults.Chromatograms.FirstOrDefault(chrom =>
+                            chrom.ContainsFile(
+                                _msDataFileScanHelper.ScanProvider.DataFilePath
+                            ));
+                        spectrumProperties.ReplicateName = chromSet?.Name;
+                        var resultsIndex = stateProvider.DocumentUI.Settings.MeasuredResults.Chromatograms.IndexOf(chromSet);
+                        var nodePath = DocNodePath.GetNodePath(_msDataFileScanHelper.CurrentTransition?.Id, _documentContainer.DocumentUI);
+
+                        //CONSIDER: calculate dot products for the current spectrum rather than whole peak
```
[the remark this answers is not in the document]
brendanx67: Agreed. If we show dotp values in this grid it should be only ones calculated from the spectra themselves against... Hmmm. I think:
- dotp = contrast angle between the library spectrum and the MS/MS spectrum being viewed in the Full-Scan viewer
- idotp = contrast angle between the expected isotope dot product and the MS1 spectrum being viewed in the Full-Scan viewer
- rdotp = contrast angle between the peak areas from a reference standard precursor and the spectrum being viewed in the Full-Scan viewer.

The last is the only case where we should be using peak areas, and those are peak areas for a different precursor than the one associated with the spectrum in the Full-Scan viewer. It doesn't make sense to calculate these with peak areas for the same precursor. That should almost always be very close to 1.0, and no one would usually care how well a single spectrum correlates to the entire integrated peak. What they want to know is how well a spectrum correlates in the same way that the peak areas get scored.

brendanx67:
When I add MS/MS chromatograms to the MS1 tutorial and use those to look at MS/MS spectra in the Full-Scan plot, the Precursor m/z value does not appear to change from the calculated precursor m/z that shows up in the Targets view, while the isolation window does change. In this case, I was expecting a precursor m/z associated with the spectrum, rather than just showing the value in the TransitionGroupDocNode.

brendanx67:
Advancing through spectra with Instrument components expanded causes them to collapse, even when nothing is changing.

brendanx67:
For TestOrbiPrmTutorial, clicking on chromatograms to show the MS/MS spectra shows "Charge: +" for both doubly and triply charged precursors.

brendanx67:
Wonder if we should add a thousands separator to the Mz Count format. The numbers can get large in IMS data, e.g. 105119 in the one I am looking at now.

brendanx67:
> For TestOrbiPrmTutorial, clicking on chromatograms to show the MS/MS spectra shows "Charge: +" for both doubly and triply charged precursors.

TestHiResMetabolomics also shows "Charge: -" without any number.

brendanx67:
I don't see any mention of collision energy, which is an especially big loss if you add a PauseTest() to OptimizeTest.cs line 828, PrmCeOptimizationTest, show its optimization chromatograms and click on one to bring up a spectrum. The CE value used to be possible to show in the title. Another bug is that the "Show Collision Energy" menu item is still present but it does nothing. It seems to be like almost every spectrum I look at should have a collision energy, but none do. One way to test which should and which shouldn't would be to test an older version and use the menu item to show it in the title, or have a look in SeeMS.

brendanx67:
> I don't see any mention of collision energy, which is an especially big loss if you add a PauseTest() to OptimizeTest.cs line 828, PrmCeOptimizationTest, show its optimization chromatograms and click on one to bring up a spectrum. The CE value used to be possible to show in the title. Another bug is that the "Show Collision Energy" menu item is still present but it does nothing. It seems to be like almost every spectrum I look at should have a collision energy, but none do. One way to test which should and which shouldn't would be to test an older version and use the menu item to show it in the title, or have a look in SeeMS.

TestOrbiPrmTutorial is probably the easiest way to see a spectrum with a CE that used to show up in the title:
image

rita-gwen:
- Precursor section of the property sheet shows information on the document precursor this spectrum is matched to. Without it user has no way to identify what spectrum they are looking at. This is intentional. Maybe I should add this to the property descriptions. I guess we can show both the document and the spectrum precursor mz.
- Advancing through the spectra (I presume you mean left and right arrow buttons) reloads the full scan viewer and the property pages as a result. At least scan Id must change in this situation.

rita-gwen:
Actually, Collision energy is shown in the same OrbiPrmTutorial. I will investigate why it is missing in the OptimizeTest and remove the menu item.
image

rita-gwen:
This commit should address most of the comments except the second one, because I don't think it's a bug. When the displayed spectrum changes the whole properties sheet is reloaded, not only the changed properties, this is how this control works. I can keep the state of the Instrument section and restore it, but not sure it's worth the effort.

brendanx67 (approved):
Looks pretty good to me. Let's get it merged. I think it would be nice to include the instrument serial number in the instrument properties. We are making that serial number more and more important on Panorama. So, increasing its visibility in our UI would help users realize when something is amiss with it. You can get to it through the document grid, of course. It would still be nice to have here also.

rita-gwen:
Instrument serial number is there. If it's available through the ProteoWizard interface it should be visible.
image

PR #2531, opened 2023-03-14
MSGF+ mods as string

chambm:
* changed MSGF+ mods file to be created as a string and only write it to disk when running MS-GF+ so it can be cleaned up in the finally block

I made this change in the course of looking for the tempfile-left-over bug. Still makes sense to me even after you fixed the bug.


bspratt (approved):
Looks good.

PR #2541, opened 2023-03-22
Added an error message box for JsonReaderException

spallanck:
If the JSON cannot be parsed an error message is displayed that says, "Error parsing JSON data" and displays the error message underneath.


nickshulman:
I was thinking that the fix for this exception was going to need to go inside the method "SkylineWindow.PublishToSavedUri" where it already catches a couple of different kinds of exceptions.

spallanck:
Okay, I can go ahead and change that.

spallanck:
Was this more what you were thinking of?

pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -3344,6 +3344,11 @@ public void ShowPublishDlg(IPanoramaPublishClient publishClient)
             {
                 return false;
             }
+            catch (Exception e)
+            {
+                MessageDlg.ShowWithException(Program.MainWindow, TextUtil.LineSeparate(Resources.WebPanoramaPublishClient_SupportedVersionsJson_Error_parsing_JSON_data, e.Message), e);
```
nickshulman, line 3349: You should use "this" instead of "Program.MainWindow".
The error message should be less specific since at the spot where we are catching we don't know what might have caused it. I see that there is already a string in Resources.resx which you can use: Resources.RemoteSession_FetchContents_There_was_an_error_communicating_with_the_server__ which just says "There was an error communicating with the server: "

nickshulman:
Looks good. You should revert the changes to "Resources.resx", "Resources.designer.cs" and "PanoramaUtil.cs".

nickshulman (approved):

PR #2546, opened 2023-03-31
I don't know if it's the recent update to Docker Desktop or something else, but lately the default of 15 seconds for parallel test "workertimeout" is too aggressive.

bspratt:
…  Bumping it to 30 seconds to give workers time to fire up and connect.


chambm:
There didn't use to be a timeout. It does seem odd for it to take more than 15 seconds for the first worker to connect. Did it start working for you with the higher timeout?

bspratt:
Yeah, it's fine at 30.

You're saying the whole idea of a timeout is new? When was that added?

bspratt:
Ah, I see, quite recently added, yes.

chambm (approved):

PR #2548, opened 2023-04-01
Support crosslinked peptide sequences in .ssl files

nickshulman:
The spectrum sequence list (.ssl) reader currently gives you an error if the peptide sequences have crosslinks in them.
I was able to write something which parses crosslinked peptide sequences and adds the correct modifications to the Modifications table.

My C++ coding is very rusty, so it's definitely possible I made a mistake such as forgetting to delete an object that I did not realize I created or misunderstanding whether a vector is mutable.


pwiz_tools/BiblioSpec/src/SslReader.cpp
```diff
@@ -228,6 +239,56 @@ SslReader::SslReader(BlibBuilder& maker,
     }
   }
 
+  /**
+   * Parses a crosslinked peptide sequence, adds the modifications from the first crosslinked peptide 
+   * to the "mods" vector, and returns the unmodified sequence.
+   * The modifications that get added to the mods vector will include a modification representing the
+   * mass of the crosslinker plus the masses of all of the linked peptides.
+   */
```
chambm, line 247: Would be good documentation to have an example of the crosslinked format this is parsing.

chambm (changes requested):
Looks good except for atof/i.

brendanx67:
Nick, do we have a wiki page on skyline.ms that documents the peptide
cross-linking sequence format we support? If not, you should create one and
then point code where it is supported at the URL for it.

PR #2550, opened 2023-04-03
Fix for ion mobility view does not work anymore

rita-gwen:
Original issue: https://skyline.ms/announcements/home/support/thread.view?rowId=60276


brendanx67 (approved):
Looks good. Thanks, Rita.

PR #2552, opened 2023-04-06
Refactoring for more extensive use of Nick's MoleculeMassOffset class in Skyline

bspratt:
- More use of MoleculeMassOffset instead of strings for internal representation of molecules. Increasingly we need to track mass offset in addition to chemical formula (e.g. crosslinking, feature finding) this tides that up
  - Centralized molecule string parsing code including handling of formula math (e.g. "C12H5NO-H2O"). Now it's all done in BioMassCalc (formerly there were a few other places doing string manipulation)
  - Moved TypedMass to pwiz_tools/Shared/Common/Chemistry for use with MoleculeMassOffset  mass offset values
  - Skyline/Util/ParsedMoleculeMassOffset extends MoleculeMassOffset for molecules whose formulas need to be represented to users in their original form (wasn't a problem when we just carried strings around, but just showing the elements in alpha sort order is confusing in many cases (e.g. user input N1C12S3H but we show "C12HNS3" instead of "N1C12S3H" - funky order but presumably it means something to them for downstream tools table lookups or whatever)
  - TypedMass and MoleculeMassOffset ctors are now private - using Create() and Change() methods instead prevents duplication of objects - especially empty ones
  - BioMassCalc ctor is now private, with a GetBioMassCalc() function that provides already-constructed objects for reuse
  - Tweaks to BioMassCalc to make it cheaper to ask about presence of isotopes in formulas

This is in support of the feature detection work where formula+offset is important, though that's not included here for clarity of purpose.

### DIFF
--- a/pwiz_tools/Shared/Common/Chemistry/Formula.cs
+++ b/pwiz_tools/Shared/Common/Chemistry/Formula.cs
@@ -19,6 +19,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Text;
 using pwiz.Common.Collections;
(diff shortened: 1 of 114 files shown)


pwiz_tools/Shared/Common/Chemistry/MoleculeMassOffset.cs
```diff
@@ -29,73 +29,197 @@ namespace pwiz.Common.Chemistry
...
         {
-            Molecule = molecule;
+            Molecule = molecule ?? Molecule.Empty;
             MonoMassOffset = monoMassOffset;
             AverageMassOffset = averageMassOffset;
         }
 
-        public MoleculeMassOffset(Molecule molecule) : this(molecule, 0, 0)
+        public Molecule Molecule { get; private set; }
+        public TypedMass MonoMassOffset { get; protected set; }
```
nickshulman, line 48: I am not sure I understand what effect changing this from a double to a TypedMass is supposed to do.
Are we trying to keep track of whether the mass is heavy, and/or is a massH?
What does it mean to add two massH's together? It looks like "operator +" in "TypedMass" has complicated rules about what to do when you the two operands differ in terms of heavy and massH.

pwiz_tools/Shared/Common/Chemistry/Formula.cs
```diff
@@ -197,146 +198,230 @@ public bool Remove(string key)
...
+                    resultDict.Add(kvpOther.Key, kvpOther.Value);
                 }
             }
 
             return FromDict(resultDict);
         }
 
+        public bool IsEmpty()
+        {
+            return Dictionary == null || Dictionary.Count == 0;
```
nickshulman, line 384: Is "Dictionary" really allowed to be null here? That often leads to cases where values end up slightly different when you round-trip to XML.

pwiz_tools/Shared/Common/Chemistry/TypedMass.cs
```diff
@@ -0,0 +1,334 @@
...
+using System.Globalization;
+using System.Diagnostics.Contracts;
+
+
+namespace pwiz.Common.Chemistry
+{
+    /// <summary>
+    /// There are many places where we carry a mass or massH and also need to track how it was derived
+    /// </summary>
+    public class TypedMass :  IComparable<TypedMass>, IEquatable<TypedMass>, IFormattable
```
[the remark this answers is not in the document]
bspratt: Nullability is convenient, particularly where it replaced nullable doubles. Are those extra bytes a problem?

pwiz_tools/Shared/Common/Chemistry/MoleculeMassOffset.cs
bspratt: Yes, tracking the massH thing (for historical peptide reasons), and the Heavy (for small molecule mass-only description reasons).

Adding two massH together is fine, it's only important when interacting with protonated adducts to determine m/z. That's all historical accident and I wish we didn't need that distinction at all, but we do.

pwiz_tools/Shared/Common/Chemistry/Formula.cs
bspratt: Unsure - just a belt-and-suspenders thing. But I'm not sure how that affects any string representation?

pwiz_tools/Shared/Common/Chemistry/MoleculeMassOffset.cs
```diff
@@ -29,73 +29,197 @@ namespace pwiz.Common.Chemistry
...
+        public string ToString(int desiredDigits)
+        {
+            var stringBuilder = new StringBuilder();
+            stringBuilder.Append(ChemicalFormulaString());
+            stringBuilder.Append(FormatMassModification(MonoMassOffset, AverageMassOffset, desiredDigits)); // Use our standard notation e.g. [+1.23/1.24], [-1.2] etc
+            return stringBuilder.ToString();
         }
 
-        protected bool Equals(MoleculeMassOffset other)
+        public bool Equals(MoleculeMassOffset other)
```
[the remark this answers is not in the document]
bspratt: Hmm, how do you satisfy IEquitable<> then?

nickshulman:
I think the following:
The "MonoMassOffset" and "AverageMassOffset" should be changed back from "TypedMass" to "double". It is best to think of "MoleculeMassOffset" as "part of a molecule", and I do not think it makes sense to talk about whether a part of a molecule is "heavy" or "massH" etc.

"TypedMass" should be changed back into a "struct" and moved back into Skyline. One of the advantage of a struct is that code can explicitly declare whether it does or does not handle null by whether it uses "Nullable" (although, yes, Nullable&lt;TypedMass> is rather inefficient in that it's twice the size of the original struct, so it should usually only be used for local variables, not fields).

"ParsedMoleculeMassOffset" should be changed so that it has a MoleculeMassOffset as a member, instead of being a subclass.

We should write a static method which parses a formula and returns IEnumerable&lt;KeyValuePair&lt;string, int>> and the "Molecule.Parse" and "ParsedMoleculeMassOffset" would both make use of that method, but would do slightly different things with them.

bspratt:
Fair enough. Perhaps there's a better name than "ParsedMoleculeMassOffset", too - the point of that class I'm trying produce here is: 
thing that that interacts with an adduct to yield an m/z value, and
has some combination of chemical formula and/ or mass declaration, and 
remembers enough about its origin to be represented in a way that will visually match a user's excel document or whatever source it came from originally

nickshulman:
> Perhaps there's a better name than "ParsedMoleculeMassOffset": thing that that interacts with an adduct to yield an m/z value, and has some combination of chemical formula and/ or mass declaration, and remembers enough about its origin to be represented in a way that will visually match a user's excel document or whatever source it came from originally

I was thinking that should be called "ParsedMolecule". The fact that it has a MoleculeMassOffset in it and not just a Molecule is just an implementation detail: it's the thing that you get when you parse a formula and want to remember what it used to look like.

bspratt:
I feel like that's going to annoy Brendan, who was pretty insistent that "Molecules" consist of atoms and nothing else.

bspratt:
"MassBasis" or something awkward like that

bspratt:
Brendan says he's fine with "ParsedMolecule", I'll roll with that. If you don't love my next cut at this I'll let you have a shot, but I think we're pretty close to iterating to a conclusion. Thanks for all the review effort!

pwiz_tools/Shared/Common/Chemistry/MoleculeMassOffset.cs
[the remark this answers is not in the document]
nickshulman: Good question.

This [web page](https://www.c-sharpcorner.com/article/story-of-equality-in-net-part-three/) seems to imply that the `IEquatable<T>` interface is probably only useful for "struct" types. There is some overhead in converting a struct to an object, so for struct types, it is faster to call the `IEquatable<T>.Equals` method if it exists instead of the regular `Equals(Object)` method.
This apparently makes `Dictionary<TKey, TValue>` faster. Dictionary uses `System.Collections.Generic.EqualityComparer<T>.Default` which checks whether "T" implements `IEquatable<T>`.

I'm thinking we should only be implementing IEquatable&lt;T> on our struct's, and if we do implement it on a class we need to either make it virtual or make sure it calls a virtual method because otherwise it will be ignoring fields declared in subclasses.

PR #2558, opened 2023-04-13
Change BiblioLiteSpectrumInfo from a struct into a class so that more of them can fit in a List.

nickshulman:
Increased limit on number of spectra that could be in a BiblioSpec spectral library from 16 million to more than 100 million (requested by Ian).

Fixed problem when .blib cache file (.slc) was more than 2GB in size.
Fix NullReferenceException in ViewLibraryDlg.UpdateUI


brendanx67:
Maybe use a blocked list? Seems like this would slow downloading from disk,
an important aspect of the .slc files. We have solved this for .SKYD files
with BlockedList

PR #2563, opened 2023-04-18
Deal with an unhandled exception in peptide transition list import. 

bspratt:
I haven't come up with a repro but the problem is pretty obvious.

This has been popping up a fair amount in ExceptionWeb, as seen in https://skyline.ms/announcements/home/issues/exceptions/thread.view?entityId=e3314598-bf38-103b-a2f7-22f5355603be&_anchor=60600#row:60600

### DIFF
--- a/pwiz_tools/Skyline/Model/Import.cs
+++ b/pwiz_tools/Skyline/Model/Import.cs
@@ -1077,7 +1077,15 @@ namespace pwiz.Skyline.Model
                 if (PeptideColumn == -1)
                     return new TransitionImportErrorInfo(Resources.MassListRowReader_NextRow_No_peptide_sequence_column_specified, null, lineNum, line);
 
-                ExTransitionInfo info = CalcTransitionInfo(lineNum);
+                ExTransitionInfo info;
+                try
+                {
+                    info = CalcTransitionInfo(lineNum);
+                }
+                catch (LineColNumberedIoException e)
+                {
+                    return new TransitionImportErrorInfo(e);
+                }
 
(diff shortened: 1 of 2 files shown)


nickshulman:
There's something suspicious going on in "SrmDocument.ImportMassList" (which was further up the callstack on the reported exception):
There is one code path where LineColNumberedIoException's get rethrown (which is what causes the reported exception):
```
                catch (LineColNumberedIoException x)
                {
                    throw new InvalidDataException(x.Message, x);
                }
```
And there's a different block of code in "SrmDocument.ImportMassList" which adds it to the error list:
```
                catch (LineColNumberedIoException x)
                {
                    var line = (lines != null && x.LineNumber >=0 && x.LineNumber < lines.Count ? lines[(int)x.LineNumber] : null)?.
                        Replace(TextUtil.SEPARATOR_TSV_STR, @" ");
                    errorList.Add(new TransitionImportErrorInfo(x.PlainMessage, x.ColumnIndex, x.LineNumber + 1, line));  // CONSIDER: worth the effort to pull row and column info from error message?
                }
```
I think the correct fix for this exception is going to be to make it so that none of the code paths in "SrmDocument.ImportMassList" rethrow exceptions.

nickshulman (approved):

brendanx67:
Looked at the error at least. Would be nice if the fix actually reported
the problem text that is in an invalid format.

bspratt:
Turns out the message already does include the offending peptide description, so the problem in this particular case seems to be that there just wasn't one. I think with this fix in place it should be possible for a user to figure that out.

PR #2564, opened 2023-04-19
Somewhere along the way, EditIonMobilityLibraryDlg's use of TargetResolver got broken

bspratt:
So in the IMSDB creation grid you'd see, for example, "#$#Sulfamethizole$C9H10N4O2S2$" instead of "Sulfamethizole". This fixes that, and adds a test to ensure that it stays fixed.

Issue was raised by Brendan after watching someone deliver a small molecule tutorial in Dortmund.


pwiz_tools/Skyline/TestPerf/SmallMolDriftTimePredictorTest.cs
```diff
@@ -104,6 +104,10 @@ protected override void DoTest()
                 editIonMobilityLibraryDlg.CreateDatabaseFile(databasePath); // Simulate user click on Create button
                 editIonMobilityLibraryDlg.GetIonMobilitiesFromResults();
             });
+
+            // Make sure we haven't broken this dialog's use of TargetResolver (if we have, we'll probably get the serializable format)
+            AssertEx.AreEqual(@"Sulfamethizole", editIonMobilityLibraryDlg.GetTargetDisplayName(0));
```
nickshulman, line 109: I would recommend putting this Assert inside of the RunUI right above.
I usually end up with infrequent intermittent failures if I write code which interacts with a DataGridView from the test thread.

bspratt: Good tip, thanks.

bspratt:
@nickshulman otherwise OK with this?

nickshulman (approved):

PR #2567, opened 2023-04-21
When explicit ion mobility values are used, the reported CCS value should be adjusted appropriately in the resulting chromatogram extractions

bspratt:
In situations where Ion Mobility is known but no CCS is given, or where ion mobility is explicitly given such that CCS may not agree with the IM value, users reasonably expect a chromatogram's CCS value to be (re)calculated when possible. Without this change, incorrect or simply missing CCS values would be associate with the chromatogram.

per https://skyline.ms/announcements/home/support/thread.view?rowId=57855 and https://skyline.ms/issues/home/issues/details.view?issueId=929

Reported by user Premy


pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
```diff
@@ -1064,14 +1070,15 @@ public static IonMobilityFilter GetSonarMzIonMobilityFilter(double mz, double wi
             LibraryIonMobilityInfo libraryIonMobilityInfo,
             double ionMobilityMax)
         {
+            IonMobilityFilter result = null;
```
brendanx67, line 1073: Would it work to just use IonMobilityFilter.Empty here? Then you could just skip all the result != null logic and return result. Wouldn't IonMobilityFilter.Empty fail result.HasIonMobilityValue?

pwiz_tools/Skyline/Model/DocSettings/TransitionIonMobilityFiltering.cs
```diff
@@ -930,6 +930,12 @@ public IonMobilityFilter ChangeIonMobilityUnits(eIonMobilityUnits units)
 
         }
 
+        public IonMobilityFilter ChangeCollisionCrossSection(double ?ccs)
+        {
+            return Equals(CollisionalCrossSectionSqA, ccs) ? this :
```
brendanx67, line 935: I don't know about this check for equality. It doesn't seem like a bad idea per se, but we don't do it everywhere. So, it may just give a false sense of security that the change will be free of impact, when elements above it in the tree of objects may still get cloned and a new tree created.

And in this case, I don't see how it gets used with an equal value, since you are only using it to provide a value when you started with none.

You are not simply following a general style guideline in our code, but sort of make it a bit more confusing what our accepted best practice is. Should every ChangeProp() be doing this?

brendanx67 (approved):
Looks good. A couple of minor comments you can choose to heed or not.

pwiz_tools/Skyline/TestPerf/PerfMobilIonTest.cs
```diff
@@ -80,6 +81,25 @@ void Test(string docFile, int groups, int peptides, int tranGroups, int transiti
...
+            {
+                editMoleculeDlgA.IonMobility = 308.71;
+                editMoleculeDlgA.IonMobilityUnits = eIonMobilityUnits.drift_time_msec;
+                editMoleculeDlgA.CollisionalCrossSectionSqA = null;
+            });
+            OkDialog(editMoleculeDlgA, editMoleculeDlgA.OkDialog);
+            document = WaitForDocumentChange(document);
+            var im = document.MoleculePrecursorPairs.First().NodeGroup.ExplicitValues;
+            AssertEx.AreEqual(308.71, im.IonMobility, .01); // As set above
+            AssertEx.IsFalse(im.CollisionalCrossSectionSqA.HasValue); // As set during chromatogram extraction
```
nickshulman, line 101: I'm confused about what this line of code is checking.
The value that gets set during chromatogram extraction is "TransitionChromInfo.IonMobility.CollisionalCrossSectionSqA".
This assert is checking TransitionDocNode.ExplicitValues.CollisionalCrossSectionSqA which wouldn't change during chromatogram extraction.

nickshulman (commented):
One thing I noticed while I was working on PR#2405 (assert that documents can round-trip to XML) is that the TransitionChromInfo CollisionalCrossSectionSqA never gets written to the .sky file. This means that the document is slightly different when you first open it, compared to what it was after you extracted chromatograms.
Should some code be added to "DocumentWriter.WriteTransitionChromInfo" to write out the CCS value?

pwiz_tools/Skyline/Model/DocSettings/TransitionIonMobilityFiltering.cs
bspratt: The idea is just to avoid unneeded copying of things that aren't actually changing, there probably are a number of places where this would be a good thing to do.

It's true that the only current use of this newly created method is a case where the current value is none, but I didn't want to make that assumption in the implementation.

pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
bspratt: Yes, but this works just as well, and is more robust in case anything in between does return a null.

bspratt:
> One thing I noticed while I was working on PR#2405 (assert that documents can round-trip to XML) is that the TransitionChromInfo CollisionalCrossSectionSqA never gets written to the .sky file. This means that the document is slightly different when you first open it, compared to what it was after you extracted chromatograms. Should some code be added to "DocumentWriter.WriteTransitionChromInfo" to write out the CCS value?

I'll check that out, thanks.

pwiz_tools/Skyline/TestPerf/PerfMobilIonTest.cs
bspratt: Misleading comment, sorry. Cut and paste cruft, I'll fix it.

pwiz_tools/Skyline/Model/Serialization/DocumentWriter.cs
```diff
@@ -878,6 +878,7 @@ private void WriteTransitionChromInfo(XmlWriter writer, TransitionChromInfo chro
                 writer.WriteAttribute(ATTR.retention_time, chromInfo.RetentionTime);
                 writer.WriteAttribute(ATTR.start_time, chromInfo.StartRetentionTime);
                 writer.WriteAttribute(ATTR.end_time, chromInfo.EndRetentionTime);
+                writer.WriteAttributeNullable(ATTR.ccs, chromInfo.IonMobility.CollisionalCrossSectionSqA);
```
nickshulman, line 881: You also need to add code to read this in in "DocumentReader.TransitionInfo.ReadTransitionPeak".

Also, you need to add a field to the "TransitionPeak" protocol buffer in "pwiz_tools\Skyline\ProtocolBuffers\SkylineDocument.proto".
And then the code to read and write this will go in "TransitionDocNode.GetTransitionPeakProtos" and "TransitionChromInfo.FromProtoTransitionResults".

(Once I finish PR#2405, tests will fail if things like this are missing)

bspratt: I'll do the protocol buffer thing, thanks.  

But I don't understand your comment about ReadTransitionPeak - isn't that happening at line 492?

nickshulman: Oops. Yes. You're right. I didn't see that line.

bspratt: @nickshulman ok, protobuf stuff done. Wasn't sure if order mattered in adding that new field (that is, add to the end, or to the end of the numbered fields area?)

nickshulman (approved):

PR #2570, opened 2023-04-22
Allow changes to molecule name (or anything that doesn't change its mass) without losing the association to any existing chromatograms.

bspratt:
Users expect that just changing the name won't make chromatograms go away.

Observed by Brendan while watching a small molecule tutorial being presented in Dortmund


nickshulman:
We should think about how this work meshes with PR #2450.

In PR #2450 the serialized string (Chromatogram Text Id) in the .skyd file gets replaced by something that gets serialized to a protocol buffer, and which is a "ChromatogramGroupId":
[ChromatogramGroupId.cs](/ProteoWizard/pwiz/blob/Skyline/work/20220511_nicksh_SpectrumGrid/pwiz_tools/Skyline/Model/Results/ChromatogramGroupId.cs)

bspratt:
@nickshulman Should be no big deal to merge the two.  My changes are relatively lightweight and largely superseded by yours - as you suspected, pretty much everywhere that I've changed from using "ModifiedTarget" to "ChromatogramTarget" you've already changed to "ChromatogramGroupId".

Really the only thing that needs to survive from my work is using this:
        Peptide.OriginalMoleculeTarget ?? ModifiedTarget
instead of peptideDocNode.ModifiedTarget in ChromatogramGroupId.ForPeptide().

So the changes I'd need to make to your work are even smaller than the ones I'd need to make to the current code.

I can branch your branch if you'd like to see what this would look like. 

If your code is likely to be integrated any time soon then we should probably save the churn and just put my stuff into yours and abandon my branch.

bspratt:
@nickshulman see my PR 2574 for details.

nickshulman:
I wonder whether it would be better to fix this by making it so that when looking for chromatograms in the .skyd file, if an exact match on the TextId is not found, then it tries to find the TextId which most closely matches.
The logic for finding chromatograms in the .skyd file is in "ChromatogramCache.ChromatogramIndexesMatching".

One of the problems with changing it so that chromatograms are identified by the original name of the Molecule is that you would then have weird scenarios where two molecules might have a conflict in the .skyd file because two molecules which currently have different names used to have the same name.

Maybe Skyline should not remember the original names of molecules if there are no chromatograms in the document. Also, the original names would be forgotten if the user told Skyline to reimport all chromatograms.

bspratt:
"find the TextId which most closely matches" sounds pretty wiggly. If the user manages to tweak a molecule such that its name and formula/mass and child ion adduct and IM match more than one chromatogram, does it really matter which chromatogram gets referenced?  Don't we already deal with this ambiguity, given that a molecule (or peptide) can exist in more than one place in the targets tree?

>> Maybe Skyline should not remember the original names of molecules if there are no chromatograms in the document. 

We could do that - it's not useful information in that case.

>> Also, the original names would be forgotten if the user told Skyline to reimport all chromatograms.

That would be kind of weird, forcing an identity change based on reimport.

nickshulman (approved):
Looks good to me.

PR #2571, opened 2023-04-24
Fix the Show Chromatogram logic for spectrum viewers

rita-gwen:


pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
```diff
@@ -1501,6 +1502,11 @@ public MzRange Range
 
         public LibraryRankedSpectrumInfo SpectrumInfo => GraphItem?.SpectrumInfo;
 
+        public bool HasChromatogramData
+        {
+            get { return _spectrum?.LoadChromatogramData() != null; }
```
nickshulman, line 1507: This "LoadChromatogramData" call really does read the chromatogram from the sqlite database (in "Chromatogram.ReadLibrary"). I wonder whether GraphSpectrum should be hanging onto the chromatogram data so that it does not have to keep getting reloaded.

PR #2577, opened 2023-04-26
fixes for Issue 914: "Proteomics features still showing in small molecule mode"

bspratt:
From the issue:

During the latest iteration of Skyline-Online for small molecules, the following proteomics features were found to show up when Skyline is in small molecule UI mode:
1. Reports that no longer show up in the Document Grid - Reports (Peptide Quantification, Peptide Ratio Results, Peptide RT Results, etc.) show up in the list under File > Export > Report. If you click the "Edit list..." button in this form, you see a list with certain items grayed out. I think the grayed out items should no appear in the "Report:" list under File > Export > Report. Possibly, we should categorize tools by whether they are small molecule or proteomics tools and hide/show their reports based on this.
2. In the Molecule Settings - Library tab, clicking the "Build..." button shows a form with "Data source" for "Files" and "Prosit", but Prosit is a Proteomics-only tool. So, the small molecule form probably should have the "Data source" box hidden, and controls below it shifted up to its y-position.
3. If you give your library a name in the Build Library wizard and then click the "Add Files..." button, the file types drop-down list contains "Matched Peptides" and a very extensive list of peptide spectrum matching tools that produce output formats we recognize for library building. This should be changed to "Matched Molecules (*.ssl)" (adding any other output formats that we know we support for small molecules.


bspratt:
image

bspratt:
image

bspratt:
image
image

pwiz_tools/Skyline/Util/ModeUIAwareForm.cs
```diff
@@ -257,15 +279,16 @@ internal void AdjustMenusForModeUI(ToolStripItemCollection items)
...
 
-            public bool MenuItemHasOriginalText(string name)
+            public bool MenuItemHasOriginalText(ToolStripMenuItem toolStripMenuItem)
             {
                 foreach (var item in _modeUIExtender.GetOriginalToolStripText().Keys)
                 {
                     if (item is ToolStripMenuItem menuItem)
                     {
-                        if (Equals(name, menuItem.Text))
+                        if (Equals(toolStripMenuItem.Text, menuItem.Text) && Equals(toolStripMenuItem.ToolTipText, menuItem.ToolTipText))
```
nickshulman, line 288: You could take advantage of the fact that the struct "MenuTextAndToolTip" which gives you .Equals for free.
```
  public bool MenuItemHasOriginalText(ToolStripMenuItem toolStripMenuItem)
  {
      var textAndTooltip = new ModeUIExtender.MenuTextAndToolTip(toolStripMenuItem);
      foreach (var keyValuePair in _modeUIExtender.GetOriginalToolStripText())
      {
          if (keyValuePair.Key is ToolStripMenuItem menuItem)
          {
              if (Equals(new ModeUIExtender.MenuTextAndToolTip(menuItem), textAndTooltip))
              {
                  return Equals(keyValuePair.Value, textAndTooltip);
              }
          }
      }
      return false;
  }

```

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/Util/ModeUIAwareForm.cs
bspratt: Interesting thought. The only disadvantage to that approach is that it uses reflection to implement the memberwise comparisons, so it's slower than the hand rolled version. Probably doesn't really matter in this case, though. At any rate this works and is just as easy to understand so I'm going to leave it alone.

bspratt:
Oops, not quite there yet...  CalibrationScenariosTest was happily using proteomic reports "CalibrationCurves" and
"PeptideResultQuantification" on small molecule data, now the test hangs since those tests don't get listed for small mol documents:

image

@nickshulman can I get one more review? Thanks

nickshulman:
> Oops, not quite there yet... CalibrationScenariosTest was happily using proteomic reports "CalibrationCurves" and "PeptideResultQuantification" on small molecule data, now the test hangs since those tests don't get listed for small mol documents:

The reports for CalibrationScenariosTest come from a .skyr file inside of "CalibrationScenariosTest.zip".
I have updated that .zip file with a new .skyr file which has uimode="mixed" but otherwise produces identical results to the old report (this required actually setting the caption on a few columns so it says "Protein" instead of "MoleculeList").

bspratt:
> > Oops, not quite there yet... CalibrationScenariosTest was happily using proteomic reports "CalibrationCurves" and "PeptideResultQuantification" on small molecule data, now the test hangs since those tests don't get listed for small mol documents:
> 
> The reports for CalibrationScenariosTest come from a .skyr file inside of "CalibrationScenariosTest.zip". I have updated that .zip file with a new .skyr file which has uimode="mixed" but otherwise produces identical results to the old report (this required actually setting the caption on a few columns so it says "Protein" instead of "MoleculeList").

Thanks - are we good to go with this change, then?

nickshulman (approved):

PR #2578, opened 2023-04-27
Move TreeViewStateRestorer to Common/Controls

vagisha:
TreeViewStateRestorer is moved to Common/Controls so that it can be used in other projects (e.g. PanoramaClient).
A subclass, TreeViewMSStateRestorer, is added to Skyline/Controls.  This class operators on TreeViewMS. 
To retain line history in both files, the strategy described here was used: https://devblogs.microsoft.com/oldnewthing/20190916-00/?p=102892


nickshulman (approved):
It looked like "TreeViewStateRestorerBase" with the type parameter was overly complicated, so I made it so "TreeViewStateRestorer" was the base class, and the subclass has a "new" property "TreeView" whose type is "TreeViewMS".

PR #2579, opened 2023-04-27
Remove references to UI classes from Model

nickshulman:
Move external tool installing code into "Skyline\Controls\Tools"
Split peak boundary message code into separate "PeakBoundaryMessages" class.


brendanx67 (changes requested):
This is not really the point. The point is to not have code in Model actually using any UI elements. Model code is not allowed to show a MessageDlg or a LongWaitDlg. These changes simply sneak around the static analysis to achieve what we are attempting to prevent.

brendanx67:
Let's close this PR. I have a clean PR that solves 2 of the cases. I think the last may be more complicated, but I would prefer to first merge the PR where I really feel we know the right solutions and do the final change in a separate PR.

pwiz_tools/Skyline/CommandArgs.cs
```diff
@@ -38,8 +38,8 @@
 using pwiz.Skyline.Model.Irt;
 using pwiz.Skyline.Model.Results;
 using pwiz.Skyline.Model.Results.Scoring;
-using pwiz.Skyline.Model.Tools;
 using pwiz.Skyline.Properties;
+using pwiz.Skyline.ToolsUI;
```
brendanx67, line 42: This should be enough to tell you, these changes have a problem. You are not allowed to include a UI namespace in CommandArgs.

pwiz_tools/Skyline/Skyline.csproj
```diff
@@ -1754,9 +1757,9 @@
     <Compile Include="Model\ModificationMatcher.cs" />
     <Compile Include="Model\StandardType.cs" />
     <Compile Include="Model\Themes\ColorScheme.cs" />
-    <Compile Include="Model\Tools\ToolDescription.cs" />
-    <Compile Include="Model\Tools\ToolInstaller.cs" />
-    <Compile Include="Model\Tools\ToolMacros.cs" />
+    <Compile Include="ToolsUI\ToolDescription.cs" />
```
brendanx67, line 1760: You can't move all of these to ToolsUI, because we need to support tool installation from the command-line so that a 3rd party tool can install a tool on the Skyline Tools menu.

PR #2582, opened 2023-04-27
Skyline: Move some remaining UI code out of the Model folder

brendanx67:
This gets us down to just one place where a UI element is used or accessed in the Model folder. The last use may be more challenging, but I think these changes are pretty clean and the right solutions for these cases.


nickshulman (approved):
Looks good to me.

PR #2583, opened 2023-04-27
A couple of tweaks to TestRunner's tempdir scheme …

bspratt:
…(we use it to make Skyline more robust on systems with unusual user names):

- do a better job of cleaning up the tempdir at the end - was deleting "Skyline\~&TMP ^\(testname)" but leaving "Skyline\~&TMP ^"
- unit tests (as opposed to functional tests) normally don't involve any tempdir use, so don't mess with setting/creating/deleting tmpdir for those


chambm:
When/if I add DDA search to the command-line, those will probably be unit tests in practice (technically functional tests) because Skyline's functional test scaffolding always shows the GUI.

bspratt:
> When/if I add DDA search to the command-line, those will probably be unit tests in practice (technically functional tests) because Skyline's functional test scaffolding always shows the GUI.

Yes,  may want to add a custom code attribute that says "this test wants the tmpfile cleanup check anyway". I'll do a quick scan to see if we already have some like that.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
```diff
@@ -349,8 +347,18 @@ public bool Run(TestInfo test, int pass, int testNumber, string dmpDir, bool hea
                 LocalizationHelper.CurrentCulture = LocalizationHelper.CurrentUICulture = Language;
                 LocalizationHelper.InitThread();
 
+                // Unit tests normally don't do disk access, so don't mess around with tempdir creation for those
+                var exerciseTMP = test.TestClassType?.BaseType?.Name != "AbstractUnitTest" &&
+                                  test.TestClassType?.BaseType?.Name != "AbstractUnitTestEx";
+                if (exerciseTMP)
```
brendanx67, line 353: Seems cleaner to just use the test in this if clause and then test for tmpTestDir != null after that. Currently you have 2 if clauses below, one which reuses this value and one which tests for tmpTestDir != null.

bspratt: Yeah, that's tighter.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
```diff
@@ -349,8 +347,18 @@ public bool Run(TestInfo test, int pass, int testNumber, string dmpDir, bool hea
                 LocalizationHelper.CurrentCulture = LocalizationHelper.CurrentUICulture = Language;
                 LocalizationHelper.InitThread();
 
+                // Unit tests normally don't do disk access, so don't mess around with tempdir creation for those
+                var exerciseTMP = test.TestClassType?.BaseType?.Name != "AbstractUnitTest" &&
```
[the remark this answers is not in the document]
bspratt: Good point, made the change. Though it's looking at assembly rather than namespace as the cue, that seemed to be what was available.

bspratt:
> > When/if I add DDA search to the command-line, those will probably be unit tests in practice (technically functional tests) because Skyline's functional test scaffolding always shows the GUI.
> 
> Yes, may want to add a custom code attribute that says "this test wants the tmpfile cleanup check anyway". I'll do a quick scan to see if we already have some like that.

OK, made the change so that anything not in Test.dll gets the special TMP treatment.

PR #2586, opened 2023-05-03
Add new toolstrip menu item to open remote Panorama files in Skyline

spallanck:
In the file dropdown section of Skyline, added a new menu item 'Open From Panorama' that opens a file browser dialog which allows a user to explore Panorama files and open them in Skyline


pwiz_tools/Shared/PanoramaClient/DirectoryPicker.cs
```diff
@@ -0,0 +1,110 @@
...
+namespace pwiz.PanoramaClient
+{
+    public partial class DirectoryPicker : Form
+    {
+
+        private FolderBrowser folders;
+        public string State;
+        public string Selected;
+
+        public DirectoryPicker(Uri server, string user, string pass, bool showCheckBox, string state, bool showSkyFolders = false)
```
vagisha, line 14: server, user, and pass are not getting used so can be removed?

pwiz_tools/Shared/PanoramaClient/DirectoryPicker.cs
```diff
@@ -0,0 +1,110 @@
...
+            folders.NodeClick += DirectoryPicker_MouseClick;
+            up.Enabled = false;
+            back.Enabled = false;
+            forward.Enabled = false;
+            checkBox1.Visible = showCheckBox;
+        }
+
+        public string Folder { get; set; }
+        public string OKButtonText { get; set; }
+        public string Server { get; set; }
```
vagisha, line 29: This is also not getting used anywhere as far as I can tell.

pwiz_tools/Shared/PanoramaClient/FilePicker.cs
```diff
@@ -0,0 +1,897 @@
...
+using System.Linq;
+using Newtonsoft.Json.Linq;
+using System.Windows.Forms;
+
+
+namespace pwiz.PanoramaClient
+{
+    public partial class FilePicker : Form
+    {
+        private static string _peptideInfoQuery;
```
vagisha, line 12: Looks like _peptideInfoQuery is getting build each time a folder node is clicked. So maybe this does not need to be a class variable?

pwiz_tools/Shared/PanoramaClient/FilePicker.cs
```diff
@@ -0,0 +1,897 @@
...
+        }
+
+
+        /// <summary>
+        /// Builds a string that will be used as a URI to find all .sky folders
+        /// </summary>
+        /// <returns></returns>
+        private static string BuildQuery(string server, string folderPath, string queryName, string folderFilter, string[] columns, string sortParam, string equalityParam)
+        {
+            var query =
```
vagisha, line 136: I think you should use PanoramaUtil.CallNewInterface here.  Perhaps something like this? 
```
var columnsQueryParam = columns != null ? "&query.columns=" + string.Join(",", columns) : string.Empty;
var sortQueryParam = !string.IsNullOrEmpty(sortParam) ? "&query.sort={sortParam}" : string.Empty;
var allQueryParams = $"schemaName=targetedms&query.queryName={queryName}&query.containerFilterName={folderFilter}{columnsQueryParam}{sortQueryParam}";
var queryUri = PanoramaUtil.CallNewInterface(serverUri, @"query", folderPath, @"selectRows", allQueryParams);
```

I don't think `equalityParam` is used.  Can it be removed?

pwiz_tools/Shared/PanoramaClient/FilePicker.cs
```diff
@@ -0,0 +1,897 @@
...
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Displays the latest versions of all files in a particular folder
+        /// </summary>
+        /// <param name="path"></param>
+        private void GetLatestVersion(string path)
```
vagisha, line 212: Do you need another query here? I think you should have all the information you need to display the latest versions from the `AddQueryFies` method since you are including all the required columns in that query.

pwiz_tools/Shared/PanoramaClient/FilePicker.cs
spallanck: I went ahead and changed the BuidQuery method based on your suggestion and removed the equalityParam that wasn't being used

pwiz_tools/Shared/PanoramaClient/FilePicker.cs
spallanck: I decided the keep the _peptideInfoQuery a class variable and just reference that query from this method so I only end up making this query one time instead of re-querying.

pwiz_tools/Shared/PanoramaClient/DirectoryPicker.cs
spallanck: I went ahead and removed the unused parameters and class variables that weren't being used.

pwiz_tools/Shared/PanoramaClient/PanoramaFormUtil.cs
```diff
@@ -0,0 +1,272 @@
...
+        panorama,
+        labkey,
+        chrom_lib,
+        folder
+    }
+
+    internal class PanoramaFormUtil
+    {
+
+        public string Server;
```
vagisha, line 22: I don't think Server, User and Pass need to be instance variables.  They are used only in the `AddChildFiles` method but that method is duplicated here.  I think the one you are using is in `FilePicker`.

pwiz_tools/Shared/PanoramaClient/PanoramaFormUtil.cs
```diff
@@ -0,0 +1,272 @@
...
+        {
+            var treeNode = new TreeNode(server.URI.ToString());
+            Server = server.URI.ToString();
+            User = server.Username;
+            Pass = server.Password;
+            treeView.Invoke(new Action(() => treeView.Nodes.Add(treeNode)));
+            treeView.Invoke(new Action(() => AddChildContainers(treeNode, folderJson, false, true)));
+        }
+
+        private static bool ContainsTargetedMSModule(IEnumerable<JToken> modules)
```
vagisha, line 51: Not used?

pwiz_tools/Shared/PanoramaClient/PanoramaFormUtil.cs
```diff
@@ -0,0 +1,272 @@
...
+
+        private JToken GetJson(string query, string user, string pass)
+        {
+            var queryUri = new Uri(query);
+            var webClient = new WebClientWithCredentials(queryUri, user, pass);
+            JToken json = webClient.Get(queryUri);
+            return json;
+        }
+
+        private static string BuildQuery(string server, string folderPath, string queryName, string folderFilter, string[] columns, string sortParam, string equalityParam)
```
vagisha, line 205: Duplicate method? This is in FilePicker too.

pwiz_tools/Shared/PanoramaClient/RemoteFileDialog.cs
```diff
@@ -0,0 +1,1668 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Newtonsoft.Json.Linq;
+using System.Windows.Forms;
+
+
+namespace pwiz.PanoramaClient
+{
+    public partial class RemoteFileDialog : Form
```
vagisha, line 10: It doesn't look like RemoteFileDialog is being used in the project.  Can it be removed, along with RemoteFileDialog.resx and RemoteFileDialog.Designer.cs?

pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -894,6 +895,146 @@ private bool CheckResults(SrmDocument document, string path, FormEx parent)
...
+
+        public void OpenFromPanorama()
+        {
+            var servers = Settings.Default.ServerList;
+            if (servers.Count == 0)
+            {
+                DialogResult buttonPress = MultiButtonMsgDlg.Show(
+                    this,
+                    TextUtil.LineSeparate(
+                        Resources.SkylineWindow_ShowPublishDlg_There_are_no_Panorama_servers_to_upload_to,
```
vagisha, line 936: This message has to be different.  It can't say "There are no Panorama servers to upload to".  We also need to be able to add a server without providing a username and password.  I will take a look at that.

pwiz_tools/Shared/PanoramaClient/PanoramaUtil.cs
```diff
@@ -0,0 +1,1096 @@
...
+
+        /// <summary>
+        /// Given the name of a file, allows a user to select
+        /// which folder on their machine to download the file to
+        /// and returns the path of the selected folder
+        /// </summary>
+        /// <param name="fileName"></param>
+        /// <param name="lastPath"></param>
+        /// <returns></returns>
+        public string SaveFile(string fileName, string lastPath)
```
vagisha, line 732: Since this is not actually saving the file can it be renamed so something like SelectDownloadPath? Have you tried using SaveFileDialog instead of FolderBrowserDialog?

pwiz_tools/Skyline/SkylineFiles.cs
```diff
@@ -894,6 +895,146 @@ private bool CheckResults(SrmDocument document, string path, FormEx parent)
...
+                {
+                    var folderPath = string.Empty;
+                    if (!string.IsNullOrEmpty(Settings.Default.LastFolderPath))
+                    {
+                        folderPath = Settings.Default.LastFolderPath;
+                    }
+                    var curServer = dlg.ActiveServer;
+                    var panoramaClient = new WebPanoramaClient(curServer.URI);
+
+                    var downloadPath = panoramaClient.SaveFile(dlg.FileName, folderPath);
```
vagisha, line 991: Since SaveFile is not interacting with the Panorama server perhaps it should not be a method in WebPanoramaClient. You could try replacing it here with:
```
string downloadPath;
using (var saveAsDlg = new SaveFileDialog
                                     {
                                          FileName = dlg.FileName,
                                          DefaultExt = SrmDocumentSharing.EXT_SKY_ZIP,
                                          Filter = TextUtil.FileDialogFilter(Resources.SrmDocument_FILTER_DOC_Skyline_Documents, SrmDocumentSharing.EXT_SKY_ZIP),
                                          InitialDirectory = folderPath,
                                          OverwritePrompt = true,
                                      })
 {
     if (saveAsDlg.ShowDialog(this) != DialogResult.OK)
      {
          return;
      }
      downloadPath = saveAsDlg.FileName;
 }
```

pwiz_tools/Skyline/SkylineFiles.cs
spallanck: What do you think a better message would be? I changed it to say 'There are no Panorama servers to download from' although maybe it should be more general like 'There are no Panorama servers to connect to'?

pwiz_tools/Skyline/SkylineFiles.cs
spallanck: I talked to Brendan about SaveFileDialog vs FolderBrowserDialog last week and he said for now FolderBrowserDialog is fine, but I went ahead and took the SaveFile method out of WebPanoramaClient and added the code you suggested into SkylineFiles so now it's using the SaveFileDialog. Hopefully that is fine but it would be an easy change to go back to FolderBrowserDialog if Brendan prefers that!

brendanx67:
Sounds good. I think SaveFileDialog makes it clearer what the action will
do. So, Vagisha is right, it is a better option.

pwiz_tools/Shared/PanoramaClient/PanoramaFolderBrowser.cs
```diff
@@ -0,0 +1,753 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Initializes the JSON that will be
+        /// used to build the TreeView of folders
+        /// </summary>
+        private void InitializeServers()
+        {
+            if (_serverList != null)
```
nickshulman, line 149: When the entire body of a function is inside of a single "if" statement, I usually like to invert the if and put a "return". The only advantage of this is that the rest of the code in the function does not have to be indented as far:
```
if (_serverList == null) 
{
   return;
}
// rest of the function goes here

```

nickshulman (commented):
Here are some comments.

Also, I imagine that maybe we should move "AlertDlg" into the "CommonUtil" project.

brendanx67:
I still feel like the testing could use some work to reduce special code in the form that runs only during testing.
See: PanoramaFilePicker.InitializeDialog()
But, it is getting there.

PR #2587, opened 2023-05-04
Ensure "Isotope Dist Proportion" has the correct value for newly added small molecule transition

nickshulman:
Fixed incorrect isotope dot product for newly imported small molecule precursor transition (reported by Noortje)

There might be a better way to fix this involving sharing code for creating transitions between SmallMoleculeTransitionListReader and TransitionGroupDocNode.


bspratt:
I agree, that seems like a lot of local and probably duplicated code for something that presumably ought to happen whenever such a node is created. Could there be a common wrapper? And are there other places where that wrapper ought to be employed?

PR #2588, opened 2023-05-07
Fix MS1 chromatogram extraction when doing PRM CE Optimization

nickshulman:
Fixed bug preventing MS1 chromatogram extraction when doing CE Optimization

With this fix, the MS1 chromatograms that Skyline extracts for each optimization step end up being identical:
image

One could imagine instead making it so that the MS1 chromatogram only exists for "step=0", but I think it would be a bigger code change to make work.


brendanx67 (changes requested):
How about a test that fails with the old code but succeeds with this fix?

PR #2589, opened 2023-05-09
Make find dialog cancellable

nickshulman:
Changed to allow canceling "Edit > Find" operation (reported by Brendan)
Issue 922: Edit > Find needs to use LongWaitDlg and support cancel

Also, fixed some suspicious code in "DuplicatedPeptideFinder" which was comparing hash codes instead of peptide sequences.

image


brendanx67 (commented):
Seems like there should be something more for testing this new work. Like a test that actually cancels a search? Maybe you could force it to use a test finder that just waits 5 seconds and then fails the test if that point is reached while the test is expected to cancel before that time is reached, i.e. immediately.

brendanx67 (approved):

PR #2590, opened 2023-05-09
Support for DIA-NN output from FragPipe

chambm:
- added support for DIA-NN output from FragPipe (always named diann-output/diann-output.tsv; NB the msfragger-diann test is not actually from FragPipe, it's a copy of the diann-swath test files with different names)


chambm:
@brendanx67 Need this for FragPipe to be able to use SkylineCmd.

PR #2594, opened 2023-05-11
* fixed ConsoleInvalidValuesTest

chambm:
@nickshulman Is there a better way to do this (get localized system error messages) that I missed?

### DIFF
--- a/pwiz_tools/Skyline/TestData/CommandLineRefineTest.cs
+++ b/pwiz_tools/Skyline/TestData/CommandLineRefineTest.cs
@@ -22,8 +22,6 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
-using System.Resources;
-using System.Security.Permissions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using pwiz.Skyline;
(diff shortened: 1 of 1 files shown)


nickshulman:
> @nickshulman Is there a better way to do this (get localized system error messages) that I missed?

I can't think of a better way.
Normally if we actually cared what error message was being shown to the user we would catch the exception and throw a new exception with our own message.
In this case it's only important to the user that they are told which path has the problem. After that, I don't think we need to worry about whether the rest of the message says "The given path's format is not supported." or "Segmentation fault (core dumped)"

chambm:
> > @nickshulman Is there a better way to do this (get localized system error messages) that I missed?
> 
> I can't think of a better way. Normally if we actually cared what error message was being shown to the user we would catch the exception and throw a new exception with our own message. In this case it's only important to the user that they are told which path has the problem. After that, I don't think we need to worry about whether the rest of the message says "The given path's format is not supported." or "Segmentation fault (core dumped)"

Then do you think I should rework that test so that it only checks that the path is mentioned in the error message? That would definitely simplify it, and might catch cases where the path is NOT mentioned in the error message, which we'd probably want to fix.

PR #2600, opened 2023-05-23
Switch dependency ordering between SkylineCmd and Skyline projects

nickshulman:
Make it so that Skyline project depends on SkylineCmd project instead of the other way around.
As far as I can tell this fixes the problem where doing a build immediately after having done a clean build fails.

### DIFF
--- a/pwiz_tools/Skyline/Jamfile.jam
+++ b/pwiz_tools/Skyline/Jamfile.jam
@@ -342,23 +342,7 @@ if [ modules.peek : NT ] && --i-agree-to-the-vendor-licenses in [ modules.peek :
         This is a fake file so Skyline can point at both msparser.dll and msparserD.dll unconditionally.
         ) > "$(<)"
     }
-    
-    
-    rule make_fake_SkylineCmd ( targets + : sources * : properties * )
-    {
-        return [ build-properties $(targets) : $(sources) : $(properties) ] ;
-    }
 
-    actions make_fake_SkylineCmd
-    {
-        @($(STDOUT):E=
-        This is a fake SkylineCmd so Skyline can point at it without a circular dependency.
-        ) > "$(<)"
-    }
-
-    make SkylineCmd.exe : : @make_fake_SkylineCmd : <conditional>@build-location ;
-    explicit SkylineCmd.exe ;
-    
     if --official in [ modules.peek : ARGV ]
     {
         constant SKYLINE_PROSIT_CONFIG : "$(SKYLINE_PATH)/Model/Prosit/Config/PrositConfig_production.xml" ;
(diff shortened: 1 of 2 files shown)


chambm:
Hmm, that must be because the clean is deleting the placeholder SkylineCmd that gets created by quickbuild. I inverted this dependency because working on SkylineCmd, or debugging with it, was quite annoying that it wouldn't automatically rebuild when there were changes to Skyline or its dependencies.

nickshulman:
> because working on SkylineCmd, or debugging with it, was quite annoying that it wouldn't automatically rebuild when there were changes to Skyline or its dependencies.

I just discovered the setting "Only build startup projects and dependencies on Run" at "Tools > Options > Project and Solutions > Build and Run" which I imagine would be a different way to solve this.

chambm:
> > because working on SkylineCmd, or debugging with it, was quite annoying that it wouldn't automatically rebuild when there were changes to Skyline or its dependencies.
> 
> I just discovered the setting "Only build startup projects and dependencies on Run" at "Tools > Options > Project and Solutions > Build and Run" which I imagine would be a different way to solve this.

According to https://learn.microsoft.com/en-us/visualstudio/ide/building-and-cleaning-projects-and-solutions-in-visual-studio?view=vs-2022#to-build-only-the-startup-project-and-its-dependencies:

> Select the Only build startup projects and dependencies on Run check box.
> 
> When this check box is selected, only the current startup project and its dependencies are built when you choose Debug > Start (F5) or Debug > Start Without Debugging (Ctrl+F5).
> 
> When you choose Build > Build Solution (Ctrl+Shift+B), the entire solution is built.
> 
> When this check box is cleared, all projects, their dependencies, and the solution files are built when you run any of the preceding commands.

The problem is that the old (and your PR) way of having Skyline depend on SkylineCmd means that Skyline is not a dependency of SkylineCmd (because that would be circular). Do you think there's a way to prevent the Clean command from ever deleting SkylineCmd.exe (or perhaps replacing it with a placeholder on Clean)?

chambm:
@bspratt Want to test whether this branch fixes the debug/release build issue?

bspratt:
> @bspratt Want to test whether this branch fixes the debug/release build issue?

I'll give it a go, sure

bspratt:
> @bspratt Want to test whether this branch fixes the debug/release build issue?



> > @bspratt Want to test whether this branch fixes the debug/release build issue?
> 
> I'll give it a go, sure

OK, can confirm this does solve the issue. I can once again clean, bjam build release, do stuff in IDE, switch to Debug in IDE and do other stuff, no problem. Where before, I had to do a clean debug build which was not at all convenient.

PR #2602, opened 2023-05-23
Update "quantitation_2.xsd"

nickshulman:
Fixed building library from Mascot .dat files from newer version (reported by Evy)

Copied "quantitation_2.xsd" and "unimod_2.xsd from the newest Matrix Science that I downloaded.
(I don't actually know how this code works but this seems to fix the bug)


chambm:
Might this be better integrated as a BiblioSpec test? That would test the contents of the library (but on the other hand wouldn't actually test that Skyline can import it). Was the reported error a BiblioSpec error, or a Skyline one?

nickshulman:
> Might this be better integrated as a BiblioSpec test? That would test the contents of the library (but on the other hand wouldn't actually test that Skyline can import it). Was the reported error a BiblioSpec error, or a Skyline one?

Good point. I will write a BiblioSpec test that tests the same file.
I did want to make sure to have a Skyline test because a month ago I thought I had managed to fix this in BiblioSpec, but the fix didn't seem to work in the BlibBuild.exe was included in Skyline.

chambm:
And yes I can update the TC AMI with the latest Mascot Parser.

PR #2603, opened 2023-05-24
Issue a more informative message when SkylineNightly runs into trouble with XML parsing

bspratt:
It was choking on a backspace character in a log file, but not being very helpful in identifying the problem

### DIFF
--- a/pwiz_tools/Skyline/SkylineNightly/Xml.cs
+++ b/pwiz_tools/Skyline/SkylineNightly/Xml.cs
@@ -38,15 +38,12 @@ namespace SkylineNightly
             return _rootNode;
         }
 
-        public static void ShowXmlExceptionDetails(Exception e, string hint)
+        // Caller isn't expected to handle this - these exceptions are rare, and this isn't
+        // user facing so writing a bunch of error recovery code isn't a good use of time.
+        public static void ShowXmlExceptionDetailsAndDie(Exception e, string hint)
         {
-            var message = $@"XML exception ""{e.Message}"" \nat\n{e.StackTrace}\n while handling this text:\n{hint}";
-            if (e.InnerException != null)
-            {
-                message += $@"\nInner Exception:\n{e.InnerException.Message}";
-            }
-
-            throw new Exception(message);
+            var message = $@"XML exception while handling this text:\n{hint}";
+            throw new Exception(message, e); 
         }
 
         // Load from XML string.
@@ -59,7 +56,7 @@ namespace SkylineNightly
             }
             catch (Exception e)
             {
-                ShowXmlExceptionDetails(e, xmlString);
+                ShowXmlExceptionDetailsAndDie(e, xmlString);
             }
             return new Xml(doc.FirstChild);
         }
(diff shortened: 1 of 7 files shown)


pwiz_tools/Skyline/SkylineNightly/Xml.cs
```diff
@@ -38,19 +38,44 @@ private XmlNode GetRoot()
             return _rootNode;
         }
 
+        public static void ShowXmlExceptionDetails(Exception e, string hint)
+        {
+            var message = $@"XML exception ""{e.Message}"" \nat\n{e.StackTrace}\n while handling this text:\n{hint}";
```
nickshulman, line 43: Usually you would throw an exception with just a little more information in its message and preserve "e" as the inner exception:
throw new Exception("XML exception while handling this text:\n"" + hint, e);

Is there a reason that you are trying to throw an exception with a big message and no nested exceptions?
Is the problem that somewhere else, all that is getting written to the log file is the message of the exception and no information about the inner exceptions?
Can we fix the other place so it outputs all of Exception.ToString()?

Also, if you do need to convert the inner exception to text it's best to use it's ToString() method because that walks down all of the nested exceptions and makes sure that the whole message contains ever exception's Message and StackTrace, and it even does the correct thing for "AggregateException".

bspratt: Excellent points, thanks. I'll tidy this up.

bspratt: Didn't properly answer your question, sorry - this is in response to SkylineNightly behavior like this:
image
just throwing that exception wasn't very helpful in tracking down the problem. The point here is definitely to get the trouble text onto the screen.

bspratt: Keeping in mind that it's going to appear in a system message box at program early exit. There's nothing to handle the rethrown exception.

brendanx67: I think normally you would not put the stack trace in the Exception message. So:

var message = $@"XML exception while handling this text:\n{hint}";

Then the caller that catches this would be responsible for appropriately presenting both the stack trace of this throw and the stack trace of the embedded exception.

bspratt: There isn't anything catching this, though. This is SkylineNightly's dying gasp.  We could do something fancier but I don't think it's worth the effort. Though perhaps it should be called ShowXmlExceptionDetailsAndDie()

bspratt: So:

        // Caller isn't expected to handle this - these exceptions are rare, and this isn't
        // user facing so writing a bunch of error recovery code isn't a good use of time.
        public static void ShowXmlExceptionDetailsAndDie(Exception e, string hint)
        {
            var message = $@"XML exception while handling this text:\n{hint}";
            throw new Exception(message, e); 
        }

PR #2606, opened 2023-05-30
Add a logo to the TurnoveR external tool

nickshulman:
Bump version number to 1.0.0.3
The new version of the tool with the logo has already been uploaded to the Skyline External Tool store.


brendanx67 (commented):
What is up with the weird line endings in the info.properties file?

PR #2608, opened 2023-05-31
Added MSstatsShiny External Tool

NateMG4:
Added an external tool for the MsstatsShiny app. Tool will open link to https://msstatsshiny.com/app/MSstatsShiny, and allow users to export skyline data in MSstats format.


brendanx67:
Let's get rid of the .sky.view and .skyl files. They are not adding anything important.

brendanx67 (approved):
Looks good. Thanks for cleaning up, Nick.

PR #2609, opened 2023-06-01
Fix BiblioSpec to read MaxQuant files with UTF-8 byte order mark

chambm:
- fixed BiblioSpec to read MaxQuant msms.txt files with UTF-8 byte order mark


nickshulman (approved):
Looks good to me.
I noticed there is similar code in a class called "LineReader" in "/libraries/csv.h" which does the same thing of skipping the UTF-8 byte order mark.

chambm:
Whew, glad debug build caught that one. Might have made for some really confusing Undefined Behavior bugs.

PR #2613, opened 2023-06-09
Add "Library Retention Times" column to the Document Grid

nickshulman:
This change adds a "Library Retention Times" to the Peptide level in the Document Grid.
The "Library Retention Times" column contains a comma-separated list of all of the retention times for the peptide from any library.
It has sub-properties "MinLibraryRetentionTime", "MaxLibraryRetentionTime" and "MeanLibraryRetentionTime".
image

image

This fixes Issue 138: Add report value for spectral library retention time

Does this seem like a reasonable way to deal with the fact that peptides usually have several retention times in the libraries?


brendanx67:
Hmmm... Our other library values are at the Precursors level. It feels like rolling everything up to the peptide level is losing information about when precursors were actually identified. And staying at the precursor level allows you to promote the "best" spectrum time for that precursor, maybe.

nickshulman:
> Hmmm... Our other library values are at the Precursors level. It feels like rolling everything up to the peptide level is losing information about when precursors were actually identified. And staying at the precursor level allows you to promote the "best" spectrum time for that precursor, maybe.

Good points. Sometimes people would want to know the whole range of retention times without regard to charge state or isotope label type. Someone at ASMS wanted to use these times when exporting numbers that would eventually be used in a method.
I think we should eventually have a LibraryRetentionTimes on the Precursor too.
I'm thinking I should rename this one that I want to put on the Peptide to be "PeptideLibraryRetentionTimes" so that when I add a similar column to the Precursor it can just be named "LibraryRetentionTimes".

It seems to be surprisingly difficult to add this column to "Precursor" right now. I was able to use the existing method "GetUnalignedRetentionTimes" to get the times across all files and all charge states and label types. But, if I wanted to get the times across all files and for a particular charge state and label type, that would be a lot of calls to "Library.TryGetLibraryRetentionTimes".
I am going to take a look at consolidating all of the methods "TryGetRetentionTimes" methods on "Library" so that things like this are easier to implement.

brendanx67:
It may be worth keeping in mind that we will want to make this even more complicated to allow spectral libraries with properties that match the new precursor filters. e.g. being able to match spectra in a .blib to either the Orbitrap filter or the LIT filter, but not necessarily both.

I can see how we would not be well prepared for light v. heavy, since this distinction doesn't impact fragmentation, and we go out of our way to use either for transition selection for the other. I would expect it to be easier to differentiate by charge state, since that does impact fragmentation greatly. But, in the case of the Document Grid, I think you would want to pay attention to light v. heavy and strictly match spectra to the exact precursor used for a row.

nickshulman:
People would probably want to get the retention times on the PeptideResult (or PrecursorResult).
Putting them on the Peptide like this is useful if you don't care which Replicate the retention times came from, but that's probably an obscure use case.

PR #2618, opened 2023-06-12
Make TestRunner choose first unused port above 9810 if running parallel with no workerport specified

nickshulman:
This change makes it so that when running parallel tests, if the workerport commandline argument has not been specified, TestRunner chooses the first unused port above 9810.

The server macs2.gs.washington.edu is configured to be able to use any port between 9810 and 9820 inclusive.
If TestRunner cannot find any unused port in that range, it will keep looking all the way up to 65535.

The new class "UnusedPortFinder" uses a lot of code that I copied from the pinvoke website.


chambm:
This would only be necessary if multiple people are running it at the same time, right? Which would possibly cause things to get quite congested, but I suppose that's worth testing.

chambm:
Also a better solution might just be to pick the first port between 9810 and 9820 that isn't being used (I'd have to look up how to do that of course).

nickshulman:
> Also a better solution might just be to pick the first port between 9810 and 9820 that isn't being used (e.g. no new UI needed, but I'd have to look up how to do that of course).

Yes, that would be a better solution. I assume we would make it so that TestRunner always does that if a port has not been specified.
I do not know how to find an unused port within a range either.

PR #2619, opened 2023-06-13
Skyline: Spectral library fixes for EncyclopeDIA from ASMS fall workshop

brendanx67:
These are a set of fixes I came up with from working with EncyclopeDIA .elib files during the ASMS fall workshop. They push the concept of a score down to the SpectrumHeaderInfo base class to get rid of the need for type-checking and casting. Also the LibraryType gets reported using polymorphism rather than a switch...case statement based on the derived types. EncyclopeDIA gets its own derived type instead of using ChromLibSpectrumHeaderInfo, which was never really appropriate use. And there are some other fixes inside EncyclopeDiaLibrary such as correctly reporting the LibraryDetails.


pwiz_tools/Skyline/Model/Lib/EncylopeDiaLibrary.cs
```diff
@@ -159,9 +158,61 @@ public override LibraryDetails LibraryDetails
...
+                            // Query for the source files detail information.
+                            select.CommandText = @"select one.SourceFile, BestSpectra, MatchedSpectra
+from (
+  select SourceFile, count(*) as MatchedSpectra
+  from entries as s
+  group by SourceFile) as one
+inner join (
+  select SourceFile, count(*) as BestSpectra
+  from entries as e
+  where Score = (select min(Score) from entries where e.PeptideModSeq = PeptideModSeq)
```
nickshulman, line 180: I think you also need "e.Charge = Charge" in the WHERE clause here.
The compound key that uniquely identifies a row in the "entries" table is "PeptideModSeq, Charge, SourceFile".

This query also looks like it would be very slow because there are no indexes in the .elib. It sounded like maybe you had convinced Brian to add some indexes but we still need to handle older .elib files.

Also, Score = (select min(Score)...) means that there could be more than one best spectrum for a given peptide if there is a tie for the lowest score. Wouldn't we want to say that if there is a tie for the best score, the best spectrum is the one that we found first, or something, so that you don't end up with more best spectra than peptides?

(I'm thinking this query should be changed to be something that reads all the necessary rows from the tables and puts them in some dictionaries and we count things using C# code)

pwiz_tools/Skyline/Model/Lib/EncylopeDiaLibrary.cs
```diff
@@ -769,29 +820,69 @@ public static EncyclopeDiaLibrary Deserialize(XmlReader reader)
...
+
+            // ReSharper disable once MemberHidesStaticFromOuterClass
+            public static ElibSpectrumHeaderInfo Deserialize(XmlReader reader)
+            {
+                return reader.Deserialize(new ElibSpectrumHeaderInfo());
+            }
+
+            public override void ReadXml(XmlReader reader)
+            {
+                // Consume tag
```
nickshulman, line 878: I think the "Consume tag" comment belongs on the next line because it's referring to the "reader.Read()" call, not the "base.ReadXml(reader)" call.

PR #2627, opened 2023-06-22
Allow changing score threshold for Waters MSe

nickshulman:
Fixed to allow changing score threshold when building library from Waters MSe final_fragment.csv files (reported by Matt)

A recent support request noticed that with Waters MSe peptide search results, Skyline does not allow you to change the score cutoff from its default value of 6:
https://skyline.ms/announcements/home/support/thread.view?rowId=61440
I believe the reason that it was implemented that way is that back when the only way to change the cutoff was with the "-c" BiblioSpec command line argument, we had no idea how to convert that probability argument into a Waters MSe score.
Now that the user can specify different cutoffs for different file types, we can allow the user to enter any number between 0 and 10 for Waters MSe data.


chambm (approved):
Where does the 0-10 range come from? A comment might be a good idea for anything that's not 0-1 or 0-100.

PR #2633, opened 2023-06-27
Attempt to fix "TestAssayLibraryImport"

nickshulman:
Add a few calls to "GC.Collect".
If the problem is that a stream is not being closed, then garbage collecting might prevent the test failure.


nickshulman:
Doesn't seem to make a difference.

chambm:
Perhaps:

GC.Collect();
GC.WaitForPendingFinalizers();
GC.Collect();

Just Collect() by itself wouldn't wait for finalizers to run. Did you conclude that my EncyclopeDia PR caused this?

PR #2635, opened 2023-06-27
Fix problem where parallel testing wouldn't work unless you passed in…

nickshulman:
… a workerport.

The code path where it chose the first unused port was missing the line that binds the port.


bspratt:
Definitely resolves the issue on my machine.

PR #2636, opened 2023-06-28
* fixed errors caused by assuming lockmass function is not IMS

### LINKED ISSUE #2623: "error in thread: Invalid Scan Number" for IMS
Hello, 

First thanks a lot and congrats for this tool. 

I'm trying to convert Ion mobility spectrometry (IMS) data but got a : 
```
2023-06-19T13:55:14.627609472Z [SpectrumWorkerThreads::work] error in thread: Invalid Scan Number
```

I'm converting with the latest docker and with the most minimal args possibles: 

```
wine msconvert my_super_IMS_file.raw --outdir /out_data
```

As I've seen the `combineIonMobilitySpectra` I guess Ion Mobility is handled by msconvert. 

I might also be using the wrong parameter for this task or perhaps the file is corrupted but it happened with many of them...

Thanks in advance for any help. 
Have a great day.

chambm:
* fixed errors caused by assuming lockmass function is not IMS: Invalid Scan Number exceptions (reported by Pierre)

Brian, do you remember what was the impetus for that particular usage of the isLockMassFunction condition? Why not rely on the `IonMobilityByFunctionIndex()` metadata?

Closes #2623


bspratt:
> Brian, do you remember what was the impetus for that particular usage of the isLockMassFunction condition? Why not rely on the `IonMobilityByFunctionIndex()` metadata?

Pretty sure we were acting on Waters advice at the time. Things have likely changed. I wonder how this will play with older files.

bspratt (commented):
Why drop the check for config_.ignoreCalibrationScans?

chambm:
> Why drop the check for config_.ignoreCalibrationScans?

Because AFAIK it's redundant. If ignoreCalibrationScans is set, the lockmass function was never included in the index.

bspratt:
> > Why drop the check for config_.ignoreCalibrationScans?
> 
> Because AFAIK it's redundant. If ignoreCalibrationScans is set, the lockmass function was never included in the index.

That feels a little wiggly, but I guess it would come out in testing if not true.

chambm:
> > > Why drop the check for config_.ignoreCalibrationScans?
> > 
> > 
> > Because AFAIK it's redundant. If ignoreCalibrationScans is set, the lockmass function was never included in the index.
> 
> That feels a little wiggly, but I guess it would come out in testing if not true.

The logic seems simple to me in createIndex:
```
for(int function : rawdata_->FunctionIndexList())
    {
        if (config_.ignoreCalibrationScans && isLockMassFunction(function))
        {
            continue;
        }
```
So unless the result of isLockMassFunction can change, I don't see the problem. If it CAN change, that's probably a problem by itself. :)

However it doesn't appear the ignoreCalibrationScans has a test case in Reader_Waters_Test. My bad for letting that get added without testing.

bspratt:
As long as that index is always consulted, seems good. I don't know/remember the code well enough to assert that it's always the case.

chambm:
> As long as that index is always consulted, seems good. I don't know/remember the code well enough to assert that it's always the case.

All pwiz spectra are accessed by zero-based index. Calling code can try to search for any id it wants but if it's not in the index it'll get index.size() (i.e. not found), which would trigger the first out of bounds error in SpectrumList_*::spectrum().

bspratt:
OK, I'm convinced, thanks!

PR #2637, opened 2023-06-28
TestPrositOutputToEncyclopediaLibraries was failing on parallel clients due to very small numerical differences in resulting TSV files…

bspratt:
…. Added some tolerances, and made the tolerance check able to handle doubles with "." or "," decimal separators.

Also added some quotes to TestRunner's Docker invocations.

### DIFF
--- a/pwiz_tools/Skyline/Model/Lib/NistLibSpec.cs
+++ b/pwiz_tools/Skyline/Model/Lib/NistLibSpec.cs
@@ -1086,11 +1086,11 @@ namespace pwiz.Skyline.Model.Lib
                             string mzField = linePeak.Substring(0, iSeperator1++);
                             string intensityField = linePeak.Substring(iSeperator1, iSeperator2 - iSeperator1);
 
-                            if (!TryParseFloatUncertainCulture(mzField, out var mz))
+                            if (!TextUtil.TryParseFloatUncertainCulture(mzField, out var mz))
                             {
                                 ThrowIoExceptionInvalidPeakFormat(lineCount, i, sequence);
                             }
-                            if (!TryParseFloatUncertainCulture(intensityField, out var intensity))
+                            if (!TextUtil.TryParseFloatUncertainCulture(intensityField, out var intensity))
                             {
                                 ThrowIoExceptionInvalidPeakFormat(lineCount, i, sequence);
                             }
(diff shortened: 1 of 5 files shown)


chambm:
What was your parallel test list that saw failures and how common were they? I'm not able to repro yet.

bspratt:
It was TestPrositOutputToEncyclopediaLibraries, and the TSV files produced had very small differences in the numerical values in columns 0 and 5. Didn't happen on the server instance of Testrunner, just the clients. Different java? Different .net? Not sure. Note that I had deleted and rebuilt my images because I was trying to understand why I couldn't get parallel testing working at all, which turned out to be something Nick did (would have been nice to have a heads up on that)

bspratt:
To be clear, didn't need any more than just the one test to consistently reproduce. Ran it in all all languages to ensure that parallel clients were involved.

chambm:
Huh. The Java version should be the same. Only differences I can think is processor (mine's AMD), OS (Windows 11 here), and base docker image (Windows 11 can use the newer base image).

The time tolerance (column 5) is small enough I'm not worried about it, but the precursor m/z one seems big enough that I can imagine the bigger EncyclopeDia tests failing in parallel, which I have not tried.

bspratt:
I'll check those too, and adjust if needed. Though if they're big they're probably already tagged as NoParallelTest?

bspratt:
The tolerance for column zero is more generous than it needs to be, I was seeing differences out at the last decimal place (well beyond any realistic measurement).

nickshulman:
I think we should try to debug the underlying problem and figure out why it's possible for these .tsv files to end up being different.
Can you send me the "pan_human_library_690to705-z3_nce33.dlib" that gets generated by a failing test?

bspratt:
> I think we should try to debug the underlying problem and figure out why it's possible for these .tsv files to end up being different. Can you send me the "pan_human_library_690to705-z3_nce33.dlib" that gets generated by a failing test?

That's a bit tricky since it only fails on a parallel client. I wonder if you can repro on your system? Select just TestPrositOutputToEncyclopediaLibraries, enable all languages, 5 or more parallel workers.

bspratt:
I've just pushed a change tightening up the overly generous tolerance:

`var columnTolerances = new Dictionary<int, double>() { { 0, 0.0000000001 }, { 5, .000000001 } };  // Allow some numerical wiggle in columns 0 and 5`

I don't think it's worth pursuing, it really just comes down to tiny differences in formatting doubles.

chambm:
Those tolerances are fine with me, I can't imagine any difference that small having a knockon effect for the columns in question. One concern is that the tsv files are actually composites of multiple table dumps. Not every table has the same columns. If a column can't be parsed as a double is it going to be compared as text?

bspratt:
> Those tolerances are fine with me, I can't imagine any difference that small having a knockon effect for the columns in question. One concern is that the tsv files are actually composites of multiple table dumps. Not every table has the same columns. If a column can't be parsed as a double is it going to be compared as text?

Yes, though I could certainly write it such that there's a default tolerance for columns that can be understood as numbers.

chambm:
Were you running that test by itself? What languages? I was able to run it locally by itself with all languages (5 workers) and didn't see any errors. Odd

bspratt:
I've just pushed a change that applies that very narrow tolerance to any column that parses as a double.

As to MACS2, that's unexplored country for me.

chambm (approved):

PR #2645, opened 2023-07-05
Skyline: Changes to prepare for a 23.1 feature complete release...

brendanx67:
of Skyline-daily


pwiz_tools/Skyline/Test/FeatureTooltipTest.cs
```diff
@@ -41,8 +41,8 @@ public void TestFeatureTooltips()
                 var tooltip = calculator.Tooltip;
                 if (string.IsNullOrEmpty(tooltip))
                 {
-                    Console.Out.WriteLine("<data name=\"{0}\" xml:space=\"preserve\"><value></value></data>",
-                        calculator.FullyQualifiedName);
+                    // Console.Out.WriteLine("<data name=\"{0}\" xml:space=\"preserve\"><value></value></data>",
```
nickshulman, line 44: This line of code only executes once it has been determined that the test is going to fail, and its job is to output text that a developer could copy and paste into "FeatureTooltips.resx".

nickshulman (commented):
I would recommend reverting the changes to DocumentFormat.cs and delete "Skyline_23.xsd".
Those changes were only made to try to fix "TestMostRecentReleaseFormatIsSupportedForSharing" which we decided to fix a different way.
Also, it would be good to revert "ReatureTooltipTest.cs".

Other than that, looks good to me.

PR #2648, opened 2023-07-06
Fix IndexOutOfRangeException if a chromatogram has zero points in it

nickshulman:
This bug was introduced as part of the "Spectrum Filter" checkin.
There has not yet been a Skyline-daily released with this bug in it.


chambm:
Cherry pick action hadn't been updated yet. Now it is.

PR #2651, opened 2023-07-06
Fix ConsoleChangePredictTranSettingsTest on Team City release branch

nickshulman:
Audit log code is insistent on some objects being Reference Equal to each other, so make sure to always use the actual OptimizationLibrary.NONE rather than something that might have been round-tripped to XML.


pwiz_tools/Skyline/Util/Xml.cs
```diff
@@ -75,7 +75,7 @@ public XmlSchema GetSchema()
             return null;
         }
 
-        public void ReadXml(XmlReader reader)
+        public virtual void ReadXml(XmlReader reader)
```
brendanx67, line 78: There is already a virtual method ValidateLoad() in which you could do this. I also think we should probably do it for everything in GetDefaults() so that we are guaranteed that all of the defaults in the list that are Equals a value returned from GetDefaults() are also ReferenceEquals().

brendanx67 (changes requested):
I would prefer a more thorough change that doesn't add the new virtual.

pwiz_tools/Skyline/Properties/Settings.cs
```diff
@@ -1980,6 +1980,26 @@ public void EnsureDefault()
         public override string Label { get { return Resources.OptimizationLibraryList_Label_Optimization_Database; } }
 
         public override int ExcludeDefaults { get { return 1; } }
+
+        protected override void ValidateLoad()
```
brendanx67, line 1984: This should go into SettingsListBase where it can do this for all our settings lists. I can't see why we would ever want deserialized values over newly created defaults. That seems like it is just asking for this same confusion to happen elsewhere.

nickshulman: I was trying to do the minimal change to fix this test.

The code in ReflectorEnumerator.cs is very complicated, so I have not figured out which line of code causes this test failure.
I had hoped it was this line in "Reflector.ProcessDefaults", but changing it did not fix this test:
`return !property.IgnoreDefaultParent && defaults.Any(d => ReferenceEquals(d, objectInfo.NewObject));`

In general, it's really wrong for there to be code that behaves differently (other than performance) when objects are Equal but not ReferenceEqual to each other. If we had that behavior with strings, that would be quite confusing.

I was planning on figuring out what the real fix in the Audit Log reflection code should be in the master branch.

nickshulman:
Tests are currently passing on TeamCity release branch without this fix.

PR #2664, opened 2023-07-14
MSFragger DDA PASEF search and better BiblioSpec errors

chambm:
- fixed MSFragger DDA search to handle DDA PASEF data correctly
- added log output of the count of BiblioSpec PSMs that do not pass score filter and improved error message when score filter is the cause of no PSMs being added
* moved Skyline BlibBuild stdin file to input file directory and set it to only be deleted if library build succeeds (so the file will still be available to rerun the BlibBuild command if the library build fails)
* added ProcessRunner lines starting with "warning" to error log

### DIFF
--- a/pwiz/data/msdata/MSData.cpp
+++ b/pwiz/data/msdata/MSData.cpp
@@ -24,7 +24,8 @@
 
 #include "MSData.hpp"
 #include "pwiz/utility/misc/Std.hpp"
-#include <boost/lexical_cast.hpp>
+#include <boost/range/adaptor/map.hpp>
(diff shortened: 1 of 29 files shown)


nickshulman:
I think the "Check detailed output" part of the error message makes it more confusing when the error is displayed in Skyline.
image
Is the "detailed output" the lines saying "Warning: did not find spectrum..." or is that referring to some other output that I might see if I ran BiblioSpec from the commandline?

chambm:
Both, but I guess it is redundant for the Skyline case if the warnings are already shown. I can remove that bit.

chambm:
I plan to let this run overnight before also merging to release.

PR #2671, opened 2023-07-25
Improved sort order for adducts …

bspratt:
…: formerly they just sorted by charge then an alphanumeric sort on descriptions, which was generally OK but could lead to unexpected ordering like this:

[M-H] [M2C13-H] [M3C13-H] [MC13-H]
which resulted in a targets tree where m/z values did not necessarily appear in the expected (increasing) order. Now the sort is:
polarity (- then +),
then by increasing absolute charge,
then by m/z when applied to an arbitrary mass,
then by text description.

So now that sort order would be [M-H] [MC13-H] [M2C13-H] [M3C13-H] as expected. This more closely matches the sort order Skyline uses for peptide transitions.

Reported by user Maud


nickshulman (approved):
Looks good to me.

PR #2672, opened 2023-07-26
Replace the target tree's peak indicator icons with shapes for red/green colorblind help

bspratt:
Colors remain the same, but now there are shape hints for those who have trouble distinguishing color. The "Peak" (green) and "Keep" (yellow) icons both feature a checkmark to indicate the presence of useful peaks, but where the "Peak" icon is round like a traffic light the Keep icon has the shape of a "caution" sign. The "NoPeak" (red) icon has the form of a "do not enter" sign.

image


bspratt:
(We started messing with this earlier this year, the subject came up again in the recent Skyline training)

bspratt:
"Keep" has horizontal bar instead of checkmark
"NoPeak" has X instead of horizontal bar
tweak to checkmark shape in "Peak"

image

bspratt:
better vertical centering of horizonatal in "Keep" icon

image

bspratt:
Made the X a little bolder
image

PR #2674, opened 2023-07-27
Point dot product property implementation

rita-gwen:


pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -598,13 +604,125 @@ private void PopulateProperties()
                     var chromSet = stateProvider.DocumentUI.Settings.MeasuredResults.Chromatograms.FirstOrDefault(
                         chrom => chrom.ContainsFile(_msDataFileScanHelper.ScanProvider.DataFilePath));
                     spectrumProperties.ReplicateName = chromSet?.Name;
+                    if (_peaks != null && _peaks.Length > 0)
+                    {
+                        var nodePath = DocNodePath.GetNodePath(_msDataFileScanHelper.CurrentTransition?.Id,
```
nickshulman, line 609: I think you need to check whether nodePath is null, especially because of that "CurrentTransition?.Id"

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -598,13 +604,125 @@ private void PopulateProperties()
                     var chromSet = stateProvider.DocumentUI.Settings.MeasuredResults.Chromatograms.FirstOrDefault(
                         chrom => chrom.ContainsFile(_msDataFileScanHelper.ScanProvider.DataFilePath));
                     spectrumProperties.ReplicateName = chromSet?.Name;
+                    if (_peaks != null && _peaks.Length > 0)
+                    {
+                        var nodePath = DocNodePath.GetNodePath(_msDataFileScanHelper.CurrentTransition?.Id,
+                            _documentContainer.DocumentUI);
+
+                        if (!nodePath.Precursor.Id.Equals(_precursor?.DocNode?.Id))
+                            _precursor = new GraphSpectrum.Precursor(_documentContainer.DocumentUI.Settings, null,
```
nickshulman, line 613: (Not important)
This should have curly braces around it. Whenever the block of an "if" statement is more than one line of code, either because it's a long statement or because there's a comment, it should be surrounded by curly braces.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -598,13 +604,125 @@ private void PopulateProperties()
                     var chromSet = stateProvider.DocumentUI.Settings.MeasuredResults.Chromatograms.FirstOrDefault(
                         chrom => chrom.ContainsFile(_msDataFileScanHelper.ScanProvider.DataFilePath));
                     spectrumProperties.ReplicateName = chromSet?.Name;
+                    if (_peaks != null && _peaks.Length > 0)
```
nickshulman, line 607: (Not important)
If you feel like it, you could simplify this to:
`if (_peaks?.Length > 0)`
(there's a similar if statement down below that's doing the same with with "_precursor.Spectra")

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -598,13 +604,125 @@ private void PopulateProperties()
...
+                            _precursor = new GraphSpectrum.Precursor(_documentContainer.DocumentUI.Settings, null,
+                                nodePath.Peptide, nodePath.Precursor);
+
+                        if (_msDataFileScanHelper.Source == ChromSource.fragment) // Calculate library dotp
+                        {
+                            if (_precursor.Spectra != null && _precursor.Spectra.Count > 0)
+                            {
+                                var libSpectrum = _precursor.Spectra[0].SpectrumPeaksInfo;
+
+                                // Rank the library spectrum with the most generic options. We need it to identify ion types
```
nickshulman, line 622: This is interesting.
I was imagining that you would calculate the dot product between the raw file spectrum and the library spectrum only based on the m/z's of the TransitionDocNode's that are in the document.
Instead, it looks like you are going to be calculating the dot product using all of the m/z's that could be identified as a particular fragment.

I think it is better to only look at the m/z's that are in the document because the user has told us m/z's they care about. They might have deleted a transition from their document because they knew it had interference and so they would not want it included in any dot product.

nickshulman (commented):
Here are some comments.
Some of these comments are not important so I marked them as such.

I think when you calculate the dot product you should only be looking at the m/z's of the transitions that are in the document, but I might be wrong about that.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
rita-gwen: I guess you didn't look through all the changes. I do use document transitions only. I use the SpectrumRanker to mark up the library peaks because I want to make sure that the peaks used for dotp calculation are same as the ones used for the area dotp. Also there is a possibility to add another dotp value for the full set of annotated peaks and the marked up spectrum will be useful there.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -598,13 +604,124 @@ private void PopulateProperties()
...
+
+        private Dictionary<TransitionFullScanInfo, double> GetPeakIntensities(
+            IEnumerable<TransitionFullScanInfo> transitions)
+        {
+            // aggregate the peak intensities a in single pass through the spectrum
+            var spectrumIndex = 0;
+            var transitionIntensities = transitions.ToDictionary(t => t, t => 0.0);
+            foreach (var docTransition in transitionIntensities.Keys.OrderBy(t => t.ProductMz))
+            {
+                while (_peaks[spectrumIndex].Mz <
```
nickshulman, line 709: I think you need to add "_spectrumIndex < _peaks.Length" to the while clause:
```
while (spectrumIndex < _peaks.Length && _peaks[spectrumIndex].Mz <
                       docTransition.ProductMz - docTransition.ExtractionWidth / 2)
```

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
nickshulman: Oh. I think I understand now.

The TransitionDocNode's know what their intensities from the library spectrum are. It's stored in "TransitionDocNode.LibInfo.Intensity".
It might make more sense to use the TransitionDocNode.LibInfo.Intensity values to calculate the dot product, instead of reading the library spectrum from the .blib file again.

rita-gwen:
Nick, I did the changes you suggested, could you have a look to make sure this is what you had in mind? The build errors are in UninstallSkylineTest, so they are probably unrelated.

pwiz_tools/Skyline/Controls/Graphs/GraphFullScan.cs
```diff
@@ -598,13 +605,107 @@ private void PopulateProperties()
...
+                            _precursor = new GraphSpectrum.Precursor(_documentContainer.DocumentUI.Settings, null,
+                                nodePath.Peptide, nodePath.Precursor);
+                        }
+
+                        if (_msDataFileScanHelper.Source == ChromSource.fragment) // Calculate library dotp
+                        {
+                            if (_precursor.Spectra?.Count > 0)
+                            {
+                                var libIntensities = _precursor.DocNode.Children.Select(t => t as TransitionDocNode)
+                                   .Where(t => t?.HasLibInfo == true && t.Id is Transition tran && tran.IonType != IonType.precursor)
```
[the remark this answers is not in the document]
rita-gwen: It looks like they are filtered on IsMs1 elsewhere. See here, for example: [TransitionGroupDocNode.cs:2082](https://github.com/ProteoWizard/pwiz/blob/c72ce1644e9623dd5d56bf558ec7640e778f1a2f/pwiz_tools/Skyline/Model/TransitionGroupDocNode.cs#L2082)

rita-gwen:
Brendan, would you like to have a look at this PR or should I merge it?

PR #2682, opened 2023-08-04
Adding a new check for dialogs that should be declared Localizable

bspratt:
Fix up a couple of dialogs caught by the new test
Fix a bug in the code inspection test logic that could result in a file not being checked for all conditions

internal test issue, not reported by any user


bspratt:
@nickshulman you might want to check the automated violence done to the dialogs I set to Localizable in the designer UI

@brendanx67 currently SharedBatch and SkylineBatch are exempted from this test, but I think that's probably not the intent?

nickshulman (approved):

PR #2683, opened 2023-08-04
Automate the maintenance of msconvert online documentation…

bspratt:
… by adding a build step which updates a checked-in text file with the current result of "msconvert --help".

Once this branch is merged to master, the page
https://proteowizard.sourceforge.io/tools/msconvert.html should be updated to replace all detail with a note telling the user to follow this link: https://raw.githubusercontent.com/ProteoWizard/pwiz/master/pwiz_tools/commandline/msconvert-help.txt and the content will never go stale.

N.B. removed some explicit mid-paragraph line breaks in help text, they didn't play well with console wrapping.


chambm:
Do you think I should change the default screen width (if it can't be detected) to something really wide so this txt file won't have hard line wrapping? It would make diffs better, at least.

bspratt:
Yes, it would be ideal in situations where we know the output is not a terminal. On Aug 8, 2023, at 9:05 AM, Matt Chambers ***@***.***> wrote:﻿
Do you think I should change the default screen width (if it can't be detected) to something really wide so this txt file won't have hard line wrapping? It would make diffs better, at least.

chambm (changes requested):
Can you change the get_console_bounds default width to something really high like 10000 (I'm not sure you can make it unlimited) and regenerate the help file? As we discussed that'll probably work better in browsers with adjustable width.

bspratt:
Hmm, currently the usage statement is presented by throwing an exception, so it ends up on stderr, which is unusual for --help responses, I think?

I'm going to change that to a simple output to stdout followed by an exit(0), and do what needed to avoid line wrap when we're aware that the output  !isatty

chambm:
I'm not a big fan of using exit() in "production" code. I don't have a strong opinion on whether --help should go to stdout or stderr. But use *os_ so that cerr is used when outputFile=="-".

bspratt:
Yeah, currently writing to os_ then throwing an empty exception to force an exit.

Though I'm now thinking we should actually force a wrap at, say, 60 rather than no wrap at all. The reason being we lose the logic that neatly breaks things like this
`  --merge                            : create a single output file from multiple input files by merging file-level metadata and concatenating spectrum lists`

into neat lines aligned under the colon
`  --merge                    : create a single output file from multiple input files by`
 `                               merging file-level metadata and concatenating spectrum lists`

chambm:
You could use a different default width for od_config and *ListFactory::wrap to get a reasonable compromise of both approaches. Ideally we'd output to HTML and have that right column as a separate table cell. :/

pwiz_tools/commandline/msconvert.cpp
```diff
@@ -400,98 +471,57 @@ Config parseCommandLine(int argc, char** argv)
...
     else
     {
         // append options description to usage string
         usage << od_config;
 
         // extra usage
-        usage << SpectrumListFactory::usage(detailedHelp, "(run this program with --help to see details for all filters)", consoleBounds.first);
-        usage << ChromatogramListFactory::usage(detailedHelp, nullptr, consoleBounds.first) << endl;
+        usage << SpectrumListFactory::usage(detailedHelp, "(run this program with --help to see details for all filters)", wrapWidth);
+        usage << ChromatogramListFactory::usage(detailedHelp, nullptr, wrapWidth) << endl;
```
chambm, line 498: This wrapWidth should be unlimited for --doc so instead of:
```
Many filters take 'int_set' arguments.  An 'int_set' is a list of intervals of
the form [a,b] or a[-][b].
For example '[0,3]' and '0-3' both mean 'the set of integers from 0 to 3
inclusive'.
'1-' means 'the set of integers from 1 to the largest allowable number'.  
'9' is also an integer set, equivalent to '[9,9]'.
'[0,2] 5-7' is the set '0 1 2 5 6 7'. 
```

We get:
```
Many filters take 'int_set' arguments.  An 'int_set' is a list of intervals of the form [a,b] or a[-][b].
For example '[0,3]' and '0-3' both mean 'the set of integers from 0 to 3 inclusive'.
'1-' means 'the set of integers from 1 to the largest allowable number'.
'9' is also an integer set, equivalent to '[9,9]'.
'[0,2] 5-7' is the set '0 1 2 5 6 7'.
```

bspratt: OK, as previously suggested it's now using the default width for the first section to get the nice columization, and a crazy wide width for the rest so the viewer can wrap as appropriate.

pwiz_tools/commandline/msconvert-help.txt
```diff
@@ -118,78 +118,42 @@ Options:
...
-
-Many filters take 'int_set' arguments.  An 'int_set' is a list of intervals of
-the form [a,b] or a[-][b].
-For example '[0,3]' and '0-3' both mean 'the set of integers from 0 to 3
-inclusive'.
+Note: Filters are applied sequentially in the order that you list them, and the sequence order can make a large difference in your output.  In particular, the peakPicking filter must be first in line if you wish to use the vendor-supplied centroiding algorithms since these use the vendor DLLs, which only operate on raw untransformed data.
+
+Many filters take 'int_set' arguments.  An 'int_set' is a list of intervals of the form [a,b] or a[-][b].
+For example '[0,3]' and '0-3' both mean 'the set of integers from 0 to 3 inclusive'.
 '1-' means 'the set of integers from 1 to the largest allowable number'.
```
chambm, line 125: Yes, this is much better, thanks!

bspratt: @chambm OK, I'll go ahead and merge if you're happy with it, and leave it to you to update
https://proteowizard.sourceforge.io/tools/msconvert.html
so that it functions mostly as a place to point users at
https://raw.githubusercontent.com/ProteoWizard/pwiz/master/pwiz_tools/commandline/msconvert-help.txt
(possibly with an iframe, up to you)

chambm (approved):

PR #2685, opened 2023-08-08
fix: quameter TIC: avoid summing of double and integer values

### LINKED ISSUE #2684: BUG: TIC calculation in quameter mixes double and integer values
The code in `quameter` to calculate the TIC of MS1 spectra adds the (`double`) intensity values to the initial (`integer`) `0` value (https://github.com/ProteoWizard/pwiz/blob/master/pwiz_tools/Bumbershoot/quameter/quameter.cpp#L906). The resulting TIC values are thus wrong.

The quantiles for one of files tested are for example (quantiles calculated in `R`, TIC values calculated using the same code as used in `quameter`):

```r
           0%           25%           50%           75%          100% 
-2147483648.0      273489.2      929755.5     2312925.5  2146555655.0 
```

Negative TIC values obviously don't make sense. Quantiles for the same file, but calculated with `double TIC = accumulate(intensV.begin(), intensV.end(), 0.0);`:

```r
          0%          25%          50%          75%         100% 
         0.0     320060.7    1011606.3    2573847.2 3195396659.9 
```

jorainer:
- Use a `double` as initial value for the `accumulate` call to calculate the TIC in `quameter`. Closes #2684 .

### DIFF
--- a/pwiz_tools/Bumbershoot/quameter/quameter.cpp
+++ b/pwiz_tools/Bumbershoot/quameter/quameter.cpp
@@ -903,7 +903,7 @@ namespace quameter
                             info.chromatogram.MS1RT.push_back(curRT);
                         } // done with unidentified MS2 scans
                         
-                        double TIC = accumulate(intensV.begin(), intensV.end(), 0);
+                        double TIC = accumulate(intensV.begin(), intensV.end(), 0.0);
                         ms1TICs.push_back(TIC);
                         totalTIC += TIC;
                     }


chambm:
Thanks for the PR. Really amazing how such a simple oversight can completely change the results of an addition loop. I guess the accumulate was casting the doubles to int before adding them up?

jorainer:
I'm not an C/C++ developer, but I guess yes - and then, I assume, there was probably a value overflow for the `int` because we've seen differences between `quameter` and R, but only for some files.

Based on [this post](https://stackoverflow.com/questions/3221812/how-to-sum-up-elements-of-a-c-vector) I assumed it might be due to ints and double.

Thanks for merging!

tnaake:
Many thanks for merging this so quickly @chambm and thanks for spotting, @jorainer. Do you know how fast the bug fix can be incorporated in a stable version and propagated to the version accessible via https://proteowizard.sourceforge.io/download.html?

chambm:
About an hour after the merge it'll be available for download. So it's already available now.

PR #2687, opened 2023-08-09
Make audit log use "Equals" instead of "ReferenceEquals" when deciding whether properties have changed

nickshulman:
Here is what I think that we should do to fix the failure in ConsoleChangePredictTranSettingsTest on the Team City release branch.

The Audit Log was using "ReferenceEquals" to decide whether a property had changed, which leads to some unexpected behavior. I changed that to "Equals".
The only test failure that this change caused was "TestPeakPickingTutorial", which turned out to be caused by a bug in "PeptideDocNode.Equals" where the "GlobalStandardType" property was not being included in ".Equals", but that property was marked with "[Track]".


brendanx67:
Please post profiling statistics on some large documents to show that this does not slow down making changes in a large document. I can imagine wanting to use (ReferenceEquals() || Equals()) to make sure that reference equality short-circuits the test.

I guess ReSharper generated equality comparers already include the line:

if (ReferenceEquals(this, other)) return true;

At the top, which should be sufficient. But, humor me and show actual profiler statistics of a change on one of the DIA full data tutorial documents. Thanks.

nickshulman:
> At the top, which should be sufficient. But, humor me and show actual profiler statistics of a change on one of the DIA full data tutorial documents. Thanks.

Here's a breakdown of how much time we spend in "ModifyDocumentInner" during "TestDiaQeFullSearchTutorial" without this change:
image
Here's what it looks like with this change:
image

It's really impossible to observe any difference in timings in "EnumerateDiffNodes".

PR #2689, opened 2023-08-09
Move source files for CommonUtil into separate folder so not interleaved with Common source files.

nickshulman:
When we first created CommonUtil, we were using Subversion and we wanted to minimize the size of the source code diff, so all of the source files were left in the same folders as the source files for Common.

Now that we are adding a lot more source files to CommonUtil, it's really confusing having these two projects in the same folder.


chambm:
Remember you can merge without updating from master. :) To save time on the build agents, try not to trigger rebuilds if you don't expect the changes from master to effect your branch.

PR #2693, opened 2023-08-11
Fix for Issue 967: Create matching small molecule versions of all standard peptide reports in Document Grid

bspratt:
Our change to remove peptide-only features in small molecule mode resulted in the loss of some useful reports. There is currently no equivalent to:
"Peptide RT Results", which has headers
Peptide,Protein,Replicate,Predicted Retention Time,Peptide Retention Time,Peptide Peak Found Ratio and
"Transition Results", which has headers
Peptide,Protein,Replicate,Precursor Mz,Precursor Charge,Product Mz,Product Charge,Fragment Ion,Retention Time,Area,Background,Peak Rank


bspratt:
@nickshulman  I went ahead and added equivalent reports for Molecule mode. Feels like these could have been handled automatically though maybe that's too much effort.

nickshulman:
The "Peptide RT Results" report might not be important enough to deserve a small molecule version.

For the "Molecule Transition Results" report, I wonder whether we should hide the "Fragment Ion" and "Peak Rank" columns because they are unlikely to be useful in a small molecule document.

bspratt:
Sorry, I don't follow - is there something unique to peptides that makes the RT results report useful there?

Also don't understand your second point - why would users not want to know about fragments? Or peak rank?

nickshulman:
> Sorry, I don't follow - is there something unique to peptides that makes the RT results report useful there?
> 
> Also don't understand your second point - why would users not want to know about fragments? Or peak rank?

"Peptide RT Results" report gets mentioned in the "Advanced Peak Picking" tutorial.
That tutorial does not get used by small molecule people because there is no such thing as a decoy for a small molecule target.
I am not sure that "Peptide RT Results" is an important enough report to deserve to have a small molecule version (but I might be wrong).

For the Molecule Transition Results, I was trying to imagine what we might want to do to make the report particularly suited for small molecules. What I had in my head was that most people do not give their fragment ions names, so the "Fragment Ion" column ends up not being as useful for molecules as it is for peptides (but, you are right, it's still pretty useful to distinguish precursors, precursor isotopes and fragments).

Maybe the report should have the "Precursor Adduct" and "Product Adduct" columns instead of "Precursor Charge" and "Product Charge"?

These default reports are our opportunity to demonstrate the best practices for building a report. Are there any other columns that would be particularly useful to include on a Molecule Transition Results report?

bspratt:
> "Peptide RT Results" report gets mentioned in the "Advanced Peak Picking" tutorial. That tutorial does not get used by small molecule people because there is no such thing as a decoy for a small molecule target. I am not sure that "Peptide RT Results" is an important enough report to deserve to have a small molecule version (but I might be wrong).

I feel like we want parity, even if it's not mentioned in any tutorial. Users shouldn't wonder why a report goes missing when the UI mode is changed (unless it something whose content is just completely proteomic, but this isn't that)

> For the Molecule Transition Results, I was trying to imagine what we might want to do to make the report particularly suited for small molecules. What I had in my head was that most people do not give their fragment ions names, so the "Fragment Ion" column ends up not being as useful for molecules as it is for peptides (but, you are right, it's still pretty useful to distinguish precursors, precursor isotopes and fragments).

It's true that fragments are often nameless, but not always. Easier for a user to them to ignore than add.

> Maybe the report should have the "Precursor Adduct" and "Product Adduct" columns instead of "Precursor Charge" and "Product Charge"?

Good point - I'd say "additionally" though, rather than "instead of". I've added those.
 
> These default reports are our opportunity to demonstrate the best practices for building a report. Are there any other columns that would be particularly useful to include on a Molecule Transition Results report?

Ideally a report would also have things like SMILES, InCHi, CAS etc but different folks use different combinations (or, often as not, none)  of those, so it's probably a matter for customization unless we can think of something clever like columns that are omitted when empty.

PR #2694, opened 2023-08-11
Associate Proteins Keep unmapped peptides

chambm:
- changed Associate Proteins dialog to keep unmapped peptides in an "Unmapped Peptides" list, and hide min peptides per protein option when called from Refinement menu

- changed Associate Proteins dialog to use final document counts for proteins and peptides in the "pro, pep, pre, tran" string (so it includes decoys, iRT, and the unmapped peptide group in the count)
- fixed MSAmanda to write PeptideEvidence with isDecoy attribute
- fixed IdentData mzIdentML parser to be case-insensitive on isDecoy attribute
* fixed Docker parallel testing check (the faster cmd /c method I had changed to does not seem to actually catch the "Login Failed" case)
* changed lexical_cast<bool> to be case-insensitive


pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
```diff
@@ -805,24 +807,61 @@ public SrmDocument CreateDocTree(SrmDocument current, IProgressMonitor monitor)
...
             {
-                // Get non-peptide children
-                var nonPeptideNodes = nodePepGroup.Children.Where(node => (node as PeptideDocNode)?.Peptide.Target.IsProteomic == false).ToList();
+                var peptideDocNodes = nodePepGroup.Children.Where(node => node is PeptideDocNode).Cast<PeptideDocNode>().ToList();
 
-                // Ignore old groups with no non-peptide children
-                if (nonPeptideNodes.Count == 0)
+                // drop all empty nodes
+                if (peptideDocNodes.Count == 0)
+                    continue;
```
nickshulman, line 823: I think this should be:
`appendPeptideLists.Add(nodePepGroup);`
Any peptide list which was empty before they did "Associate Proteins" should be ignored and remain in the document.

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
```diff
@@ -805,24 +807,61 @@ public SrmDocument CreateDocTree(SrmDocument current, IProgressMonitor monitor)
...
+                    // If a peptide list peptide is unmapped, leave it in the peptide list but remove it from the global unmapped list
+                    var peptidesByMappedStatus = peptideDocNodes.ToLookup(node => _peptideToProteins.Contains(kvp => kvp.Key.Target == node.Target), node => node);
+                    var unmappedPeptides = peptidesByMappedStatus[false].ToHashSet();
+                    unmappedPeptideNodes.ExceptWith(unmappedPeptides);
+
+                    // If it was mapped, remove it from the peptide list
+                    var mappedPeptides = peptidesByMappedStatus[true];
+                    var mappedPeptideIndexes = mappedPeptides.Select(node => node.Peptide.GlobalIndex);
+                    var newPeptideList = nodePepGroup.RemoveAll(mappedPeptideIndexes.ToList()) as PeptideGroupDocNode;
+                    appendPeptideLists.Add(newPeptideList);
```
nickshulman, line 855: I think this should be:
```
if (newPeptideList.Children.Count != 0)
{
   appendPeptideLists.Add(newPeptideList);
}
```
If the peptide list used to have some peptides and after mapping it has zero peptides, it should be removed from the document.

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
```diff
@@ -175,10 +175,12 @@ public void UseBackgroundProteome(BackgroundProteome backgroundProteome, Func<Fa
             
             Assume.IsTrue(localResults.ProteinsMapped + localResults.ProteinsUnmapped > 0);
 
+            var distinctPeptideDocNodes = _peptideToPath.SelectMany(kvp => kvp.Value).Distinct().ToList();
+            int distinctTargetPeptideCount = distinctPeptideDocNodes.Where(p => !p.IsDecoy).Select(p => p.Peptide.Target).Distinct().Count();
```
nickshulman, line 179: I would write this as:
```
int distinctTargetPeptideCount = _peptideToPath.SelectMany(kvp => kvp.Value)
    .Where(p => !p.IsDecoy)
    .Select(p => p.Peptide.Target).Distinct().Count();
```
This avoids the calling "Distinct" on the collection of PeptideDocNode's on the previous line, which is redundant because you eventually call "Distinct" on the Target's anyway.
(PeptideDocNode.GetHashCode and PeptideDocNode.Equals are both big slow functions so it's best to avoid calling them if you can)

nickshulman (commented):
Here are some comments.

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
chambm: So remove a peptide list if it was populated before association and empty after association, but keep a peptide list if it was empty before association? That seems a bit inconsistent, what do you think @brendanx67 ?

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
chambm: I think I had it that way to speed up interactive debugging and forgot to change it back. Thanks for catching that.

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
chambm: My comment was addresses both cases. My inclination is, to be consistent, to remove empty peptide lists regardless of whether they had peptides in them to begin with. Alternately to keep empty peptide lists regardless of whether they had peptides in them to begin with (users can always delete them afterward). But like Brendan I'm interested to hear if you have an example in mind of "the user had some reason to have it in the document in the first place".

brendanx67: I would recommend a thorough unit test of all the interesting cases to ensure they are and stay what we want and expect.

brendanx67: I side with deleting them always. I can't think of a use case where a user would want them. So, keeping them always is likely to annoy the user who has to delete them after this operation.

brendanx67: The point of this feature was to take a potentially complex document full of lists, but with no protein sequence backings, and allow the user to specify the FASTA file containing the sequences, and have Skyline magically make everything as if they had used File > Import > FASTA from the start. Imagine a case where the user started with File > Import > Assay Library with no protein sequence information. They may end up with many thousands of peptide lists representing proteins. Now they want to "Associate Proteins" (sequence information) with their peptides. This is one of our most important use cases. Please don't lose sight of that in noodling over the details in documents with a few peptides in them.

brendanx67: The rat plasma document is actually a reasonably good example. About half of its proteins have sequence information and the rest are peptide lists standing in for proteins. This feature should be able to leave the user with an ideal document when all they do is supply the rat FASTA.

PR #2698, opened 2023-08-14
"failed" is a useful log search term for dealing with build/test issues, but "doctest" uses the word even when it succeeds ("0 failed").…

bspratt:
I've changed it to report "0 failures". (This has been bugging me for years)

This change will be lost next time we update the file from its parent GitHub repo but that doesn't happen very often.


chambm:
I just search for "...failed" but I have no objections to this change.

chambm (approved):

PR #2699, opened 2023-08-16
Skyline: clean-up and fix leak in new panorama client code

brendanx67:
Use for LongWaitDlg was causing a Thread handle leak because was creating a control on a background thread and adding event handlers to that control. I don't think the LongWaitDlg was serving its intended purpose. So, I just removed it. I am left wondering how this code should actually be querying the web server for its JSON information in a way that involves a LongWaitDlg that can actually be canceled. But, I left that for another time. The existing code didn't do that, and it wasn't doing anything inside LongWaitDlg.PerformWork() that would ever be slow enough to require the progress form to show, and it had not support for canceling.

I did a bit of code clean up and simplification along the way also.

### DIFF
--- a/pwiz_tools/Shared/PanoramaClient/PanoramaDirectoryPicker.cs
+++ b/pwiz_tools/Shared/PanoramaClient/PanoramaDirectoryPicker.cs
@@ -49,13 +49,28 @@ namespace pwiz.PanoramaClient
             {
                 FolderBrowser = new LKContainerBrowser(servers, state, false, selectedPath);
             }
-            FolderBrowser.Dock = DockStyle.Fill;
-            FolderBrowser.NodeClick += DirectoryPicker_MouseClick;
-            folderPanel.Controls.Add(FolderBrowser);
-            up.Enabled = false;
             SelectedPath = selectedPath;
-            back.Enabled = false;
-            forward.Enabled = false;
+
+            InitializeDialog();
+        }
+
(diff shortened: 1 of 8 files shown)


nickshulman (approved):
Looks good to me.
I am guessing it's those using statements for the dialog boxes which is what fixes the leaks.

nickshulman:
It looks like the SkylineBatch build is broken here. That looks simple to fix so I'll push a fix for it.

brendanx67:
No. It wasn’t the using. It was a Thread leak that was fixed by removing
LongWaitDlg. I was, like you, quite hopeful about the missing using, but it
didn’t fix the handle leak, which wasn’t even in user+gdi.

nickshulman:
> It was a Thread leak that was fixed by removing LongWaitDlg.

Oh, yes, I see. That LongWaitDlg was being used inappropriately. Controls should never be constructed on a different thread from the thread that they are going to be shown on.

If it's ever actually necessary to do something like that, the class "LongOperationRunner" can be used.
LongOperationRunner shows a LongWaitDlg on a background thread and executes the time consuming function on the event thread.

brendanx67:
Ugh. I may have gotten this wrong. It looks like LKContainerBrowser() does
end up constructing a WebClient and making requests on it. So, that this a
Control class.

I probably need to go back and use LongOperationRunner as you describe.

PR #2700, opened 2023-08-16
The latest ReSharper (2023.2.20230731.174521) gives hundreds of new warnings…

bspratt:
﻿* **Initialize object properties inside the 'using' statement to ensure that the object is disposed if an exception is thrown during initialization**

﻿﻿216 of these - ReSharper wants to change this, for example
```
                using (var graph = new GraphRegression(new[] { _graphData }) { Width = 800, Height = 600 })
                {
                    graph.ShowDialog(this);
                }
```
to
```
                using (var graph = new GraphRegression(new[] { _graphData }))
                {
                    graph.Width = 800;
                    graph.Height = 600;
                    graph.ShowDialog(this);
                }
```
No important effect in most cases, but if in this example assigning Width or Height involved creating a disposable object, that object would not have been disposed if GraphRegression() ctor threw an exception. I've cleaned these up. 

* **Local variable 'xxxx' is only used to discard 'out' parameter value.** 

204 of these. 

Resharper is saying that rather than declaring a variable we could just use the "out _" form in a function call. While the compiler is probably optimizing this away, I've gone ahead and cleaned up the unneeded declarations.

* **Redundant dictionary 'ContainsKey' before adding to the collection.** 

7 cases like this:
```
                if (!_dict.ContainsKey(a))
                {
                    _dict.Add(a, value);
                }
                else
                {
                    _dict[a] = value;
                }
```
where it could just be
`                _dict[a] = value;`
I've cleaned these up. Possibly a little performance boost here.

* **Dictionary lookup can be simplified with TryGetValue** 

23 cases like this
```
            return Matches.ContainsKey(key)
                ? Matches[key]
                : (AAModMatch?)null;
```
which could just be
```
            return Matches.TryGetValue(key, out var match)
                ? match
                : (AAModMatch?)null;
```
I've cleaned these up. Possibly a little performance boost here.

### DIFF
--- a/pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
+++ b/pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
@@ -26,7 +26,6 @@ using pwiz.CLI.cv;
 using pwiz.CLI.data;
 using pwiz.CLI.msdata;
 using pwiz.CLI.analysis;
-using pwiz.CLI.util;
 using pwiz.Common.Chemistry;
 using pwiz.Common.Collections;
 using pwiz.Common.Spectra;
@@ -988,14 +987,14 @@ namespace pwiz.ProteowizardWrapper
 
         private double[] GetIonMobilityArray(Spectrum s)
         {
-            BinaryDataDouble data = null;
+            double[] data = null;
             // Remember where the ion mobility value came from and continue getting it from the
             // same place throughout the file. Trying to get an ion mobility value from a CVID
             // where there is none can be slow.
             if (_cvidIonMobility.HasValue)
             {
                 if (_cvidIonMobility.Value != CVID.CVID_Unknown)
-                    data = s.getArrayByCVID(_cvidIonMobility.Value)?.data;
+                    data = s.getArrayByCVID(_cvidIonMobility.Value)?.data?.Storage();
             }
             else
             {
@@ -1031,16 +1030,16 @@ namespace pwiz.ProteowizardWrapper
                     _cvidIonMobility = CVID.CVID_Unknown;
             }
 
-            return data?.Storage();
+            return data;
         }
 
-        private BinaryDataDouble TryGetIonMobilityData(Spectrum s, CVID cvid, ref CVID? cvidIonMobility)
+        private double[] TryGetIonMobilityData(Spectrum s, CVID cvid, ref CVID? cvidIonMobility)
         {
-            var data = s.getArrayByCVID(cvid)?.data;
+            using var data = s.getArrayByCVID(cvid)?.data;
             if (data != null)
                 cvidIonMobility = cvid;
 
-            return data;
+            return data?.Storage();
         }
 
         private MsDataSpectrum GetSpectrum(Spectrum spectrum, int spectrumIndex)
--- a/pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
+++ b/pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
@@ -1036,7 +1036,7 @@ namespace pwiz.ProteowizardWrapper
 
         private BinaryDataDouble TryGetIonMobilityData(Spectrum s, CVID cvid, ref CVID? cvidIonMobility)
         {
-            using var data = s.getArrayByCVID(cvid)?.data;
+            var data = s.getArrayByCVID(cvid)?.data;
             if (data != null)
                 cvidIonMobility = cvid;
 
@@ -1217,8 +1217,7 @@ namespace pwiz.ProteowizardWrapper
                 
                 for (var i = 0; i < len; i++)
                 {
-                    int index;
-                    var id = GetChromatogramId(i, out index);
+                    var id = GetChromatogramId(i, out _);
 
                     if (IsSingleIonCurrentId(id))
                         return true;
--- a/pwiz_tools/Shared/zedgraph/ZedGraph/Axis.cs
+++ b/pwiz_tools/Shared/zedgraph/ZedGraph/Axis.cs
@@ -839,7 +839,7 @@ namespace ZedGraph
 			this.MinSpace = 0;
 			// Calculate the space required for the current graph assuming scalefactor = 1.0
 			// and apply the bufferFraction
-            float space = this.CalcSpace( g, pane, 1.0F, out _ ) * bufferFraction;
+			float space = this.CalcSpace( g, pane, 1.0F, out _ ) * bufferFraction;
 			// isGrowOnly indicates the minSpace can grow but not shrink
 			if ( isGrowOnly )
 				space = Math.Max( oldSpace, space );
--- a/pwiz_tools/Shared/zedgraph/ZedGraph/Axis.cs
+++ b/pwiz_tools/Shared/zedgraph/ZedGraph/Axis.cs
@@ -799,8 +799,8 @@ namespace ZedGraph
 				SetTransformMatrix( g, pane, scaleFactor );
 
 				double baseVal = _scale.CalcBaseTic();
-				float topPix, rightPix;
-				_scale.GetTopRightPix( pane, out topPix, out rightPix );
+				float topPix;
+				_scale.GetTopRightPix( pane, out topPix, out _ );
 
 				shiftPos = CalcTotalShift( pane, scaleFactor, shiftPos );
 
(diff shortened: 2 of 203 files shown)


pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -1036,7 +1036,7 @@ private double[] GetIonMobilityArray(Spectrum s)
 
         private BinaryDataDouble TryGetIonMobilityData(Spectrum s, CVID cvid, ref CVID? cvidIonMobility)
         {
-            using var data = s.getArrayByCVID(cvid)?.data;
+            var data = s.getArrayByCVID(cvid)?.data;
```
nickshulman, line 1039: I think "TryGetIonMobilityData" should be changed so that it returns a double[] like this:
```
        private double[] TryGetIonMobilityData(Spectrum s, CVID cvid, ref CVID? cvidIonMobility)
        {
            using var data = s.getArrayByCVID(cvid)?.data;
            if (data != null)
                cvidIonMobility = cvid;

            return data?.Storage();
        }
```
I am a little confused, though, as to why the current code works and doesn't crash. It might be that "BinaryDataDouble" does not actually do anything.

bspratt: Yeah, this was a "why did that ever work?" thing for sure. 

@chambm what do you think of Nick's suggestion?

pwiz_tools/Skyline/TestFunctional/EditCustomThemeTest.cs
```diff
@@ -155,6 +155,7 @@ protected void TestEditCustomThemes()
                 AssertColor(colors[18], Color.RoyalBlue);
                 OkDialog(editCustomThemeDlg, editCustomThemeDlg.save); // Now save changes
                 Assert.IsNull(ColorScheme.ColorSchemeDemo);
+                // ReSharper disable once HeuristicUnreachableCode
```
nickshulman, line 158: This is weird. Is this a ReSharper bug?

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -2603,72 +2602,78 @@ public static IList<DbRefSpectra> GetRefSpectra(SQLiteConnection connection)
...
-                var iAdduct = reader.GetOrdinal("precursorAdduct");
-                var iIonMobility = reader.GetOrdinal("ionMobility");
-                var iIonMobilityHighEnergyOffset = reader.GetOrdinal("ionMobilityHighEnergyOffset");
-                var iCCS = reader.GetOrdinal("collisionalCrossSectionSqA");
-                var noMoleculeDetails = reader.GetOrdinal("moleculeName") < 0; // Also a cue for presence of chemicalFormula, inchiKey, and otherKeys
-                while (reader.Read())
+            using (var select = new SQLiteCommand(connection))
+            {
+                select.CommandText = "SELECT * FROM RefSpectra";
+                using (var reader = select.ExecuteReader())
```
nickshulman, line 2608: It might be good to have a using declaration here instead of a using statement with curly braces, just to reduce the amount that the code has to be indented.

pwiz_tools/Skyline/Util/SSRCalc3.cs
```diff
@@ -1046,9 +1046,9 @@ private double helicity1(string sq)
                 {
                     hc6 = hc.Substring(i, 6);
                     sc6 = 0.0;
-                    if (HlxScore6.ContainsKey(hc6))
+                    if (HlxScore6.TryGetValue(hc6, out var value))
```
nickshulman, line 1049: This if block could be simplified to:
`HlxScore6.TryGetValue(hc6, out sc6);`
because TryGetValue will set the out parameter to zero if it returns false

pwiz_tools/Shared/zedgraph/ZedGraph/Axis.cs
```diff
@@ -839,8 +839,7 @@ internal void DrawGrid( Graphics g, GraphPane pane, float scaleFactor, float shi
 			this.MinSpace = 0;
 			// Calculate the space required for the current graph assuming scalefactor = 1.0
 			// and apply the bufferFraction
-			float fixedSpace;
-			float space = this.CalcSpace( g, pane, 1.0F, out fixedSpace ) * bufferFraction;
+            float space = this.CalcSpace( g, pane, 1.0F, out _ ) * bufferFraction;
```
nickshulman, line 842: Indenting looks wrong here, because this line uses spaces but all the other lines in "Axis.cs" use tabs.
I imagine this line should use tabs too, or we could replace the tabs on all of the other lines in this file.

nickshulman (commented):
Everything looks good.
Here are a few suggestions.
Also, the indentation in some of the ZedGraph source files looks weird.

pwiz_tools/Skyline/Controls/Graphs/GraphChromatogram.cs
```diff
@@ -1652,9 +1645,7 @@ private void SetErrorGraphItem(UnavailableChromGraphItem graphItem)
             }
             float end = tranPeakInfo.EndRetentionTime;
             float start = tranPeakInfo.StartRetentionTime;
-            double[] allTimes;
-            double[] allIntensities;
-            chromatogramInfo.AsArrays(out allTimes, out allIntensities);
+            chromatogramInfo.AsArrays(out _, out _);
```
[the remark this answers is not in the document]
bspratt: Noted, thanks.

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -3295,15 +3289,16 @@ public void CreateRegression()
...
             {
+                dlg.Regression = regression;
                 dlg.ShowPeptides(true);
                 if (dlg.ShowDialog(this) == DialogResult.OK)
                 {
                     regression = dlg.Regression;
                     listRegression.Add(regression);
 
-                    ModifyDocument(string.Format(Resources.SkylineWindow_CreateRegression_Set_regression__0__, regression.Name),
+                    ModifyDocument(string.Format(Resources.SkylineWindow_CreateRegression_Set_regression__0__, regression?.Name),
```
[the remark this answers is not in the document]
bspratt: Noted, thanks

pwiz_tools/Skyline/TestFunctional/EditCustomThemeTest.cs
bspratt: Yeah, it seemed that way to me. Not a great heuristic, clearly

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
bspratt: Can do, though I was trying to minimize code thrash. Do you feel it's important?

pwiz_tools/Skyline/Util/SSRCalc3.cs
bspratt: Fair enough, there are a couple more like it in there I'll deal with as well.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
nickshulman: Not important, but if you were to replace one using statement with a using declaration things would end up being indented the same amount as they were before.

pwiz_tools/Shared/zedgraph/ZedGraph/Axis.cs
bspratt: Noted. It's a bit jarring here so I've converted those spaces to tabs, but I don't think we want to mess with the entire directory just for formatting purposes.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
bspratt: Ah, I see. Yes, actually less code thrash that way. Thanks!

nickshulman (approved):

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
nickshulman: It looks like the reason that the original did not crash is that "data.owner_" is always non-null, and the destructor for "BinaryData" does not do anything in that case:
`    virtual ~BinaryData() { if (owner_ == nullptr && base_ != NULL) delete base_; }
`

chambm: It's worth trying as double[]. None of the vendors are providing ion mobility arrays in a managed context, so Storage() will always copy the array from native memory to managed. I'm not sure whether that copying would be worse than iterating over the native memory, so definitely compare the IMS perftests before and after. (But certainly that using declaration was wrong even if it was a no-op.)

nickshulman: I don't think there can be any performance impact from this change. The caller ("GetIonMobilityArray") used to always immediately call "Storage()" on what was being returned by this method. This change just makes it so that the call to "Dispose()" happens after instead of before the call to "Storage()".

bspratt: I read this the same way as Nick. I think it's good to go, thanks all.

PR #2702, opened 2023-08-17
Fix an exception thrown when reading a small molecule transition list  lacking anything like a "name" Column…

bspratt:
Also, add the Skyline version to the stack trace information behind the "More info" button for exception alert dialogs (we already do this for unhandled exceptions, the version number gives important context to the line numbers in the stack)

Not reported by any user


bspratt:
This should be backported to the release branch, too

nickshulman (approved):
Looks good to me. Having the version number above the stack trace looks classy.
image


I think only the SmallMoleculeTransitionListReader.cs part of this fix should be ported back to the release branch at this stage.

PR #2704, opened 2023-08-18
Convert Advanced Peak Picking tutorial to markdown (.md) and automatically generate screenshots

nickshulman:
I have been thinking that things might be better if our tutorials were written in markdown instead of Microsoft Word, and if the necessary screenshots were automatically generated by the tutorial test.

For one thing, if we wanted to point someone at the latest version of a tutorial test, we could send them to a page on github:
https://github.com/ProteoWizard/pwiz/blob/Skyline/work/20230818_MarkdownTutorialTests/pwiz_tools/Skyline/Documentation/Tutorials/markdown/PeakPicking/AdvancedPeakPicking.md

We could potentially also have a copy of the tutorial in Skyline, so that older versions of Skyline could display their version of the tutorial in a web browser control that is accessed from the "Help" menu.

Lots of details to still work out, like exactly when these screenshots should be updated and committed.


chambm:
I have been wanting this since I had to write my first tutorial. Bravo!

nickshulman:
> I think it would be ideal if we had something that can embed the pictures into a single file for each tutorial.

I had been working on a plan for that.
I found a dll called "MarkDig" which can convert markdown to HTML, and there was some way to process the image tags and base64 encode the data with src="data:image/png;base64...

nickshulman:
Obsolete. We will be converting tutorials to HTML instead.

PR #2715, opened 2023-09-10
Feature/analog devices

xjasg:
Cleanup / Refactoring - add prefix analog to the chromatogram string
trim only left and right


pwiz/data/msdata/IO.cpp
chambm, line 1: Because there will be no namePrefix (or changes to MSData::chromatogram either, for the same reason), this all needs to be reverted and the logic put in pure userParams.

pwiz/data/msdata/IO.cpp
```diff
@@ -2557,6 +2583,13 @@ void write(minimxml::XMLWriter& writer, const ChromatogramList& chromatogramList
         // write the chromatogram
         if (chromatogram->index != i) throw runtime_error("[IO::write(ChromatogramList)] Bad index.");
         chromatogram->index -= skipped;
+
+        if(!chromatogram->additionalChannelInfo.empty())
+        {
+            chromatogram->userParams.push_back(UserParam("ChannelType",
+                                                         chromatogram->additionalChannelInfo));
+        }
+
```
chambm, line 2592: If the purpose of additionalChannelInfo is just to create a userParam, it seems simpler to just add this userParam at the point additionalChannelInfo is set.

pwiz/data/msdata/MSData.hpp
chambm, line 1: For the same reason that ParamType can't have arbitrary new members, spectrum and chromatogram can't either.

pwiz/data/vendor_readers/Waters/ChromatogramList_Waters.cpp
```diff
@@ -329,9 +353,49 @@ PWIZ_API_DECL void ChromatogramList_Waters::createIndex() const
...
+    s.erase(std::find_if(s.rbegin(), s.rend(),
+        std::not1(std::ptr_fun<int, int>(std::isspace))).base(), s.end());
+}
+
+// trim from both ends (in place)
+void ChromatogramList_Waters::trim(std::string& s)
+{
+    rtrim(s);
+    ltrim(s);
+}
```
chambm, line 398: These trim members are redundant. Use `boost::algorithm::trim` (namespace alias `bal`).

pwiz/data/vendor_readers/Waters/ChromatogramList_Waters.cpp
```diff
@@ -329,9 +353,49 @@ PWIZ_API_DECL void ChromatogramList_Waters::createIndex() const
...
+        ie.function = -1;
+        ie.analogChannel = ch;
+        ie.offset = 0;
+        ie.Q1 = 0;
+
+        std::string temp = rawdata_->AnalogChannelNames()[ch];
+
+        trim((temp));
+        ie.id = temp;
+        ie.chromatogramType = MS_Analog_chromatogram;
```
chambm, line 372: As mentioned above, chromatogramType should be a specific type (like EMR chromatogram or flow rate chromatogram). If the type isn't known, then use `MS_chromatogram_type`.

Also, the id must be unique. Do we know for sure whether AnalogChannelNames are unique?

pwiz_aux/msrc/utility/vendor_api/Waters/WatersRawFile.hpp
```diff
@@ -192,6 +220,8 @@ struct PWIZ_API_DECL RawData
                 hasProfile_ = hasProfile_ || Info.IsContinuum(itr.first);
         }
 
+        ReadAnalogChromatograms();
```
chambm, line 223: If this function is always and only called by the constructor, make it a private function `readAnalogChromatograms`. Do you know what kind of performance impact this has? If it's more than a small fraction of a second, this initialization needs to be done on-demand so client code that doesn't care about analog signals, like Skyline, can open a file and get a single spectrum as instantly as possible.

chambm (changes requested):
Hi @xjasg @Andre99999999 , thanks for the PR! Sorry it took a while to review it, I was out of town this week. What kind of chromatograms are in the analog channels? UV? PDA? Pressure pump traces? We'll need to add a small test file with some analog channel content.

pwiz/data/common/ParamTypes.hpp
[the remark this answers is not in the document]
xjasg: Done :-)

pwiz/data/common/cv.hpp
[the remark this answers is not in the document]
xjasg: Done.

pwiz/data/msdata/IO.cpp
xjasg: Okay, I changed it.

pwiz_aux/msrc/utility/vendor_api/Waters/WatersRawFile.hpp
xjasg: okay.

xjasg:
Hi, I´ve changed all issues. Hope it´s fine now. Thank´s a lot for that.

pwiz_aux/msrc/utility/vendor_api/Waters/WatersRawFile.hpp
xjasg: I forgot about this point: I wouldn't expect the performance for this method to be extremely poor, but I can measure that.

xjasg:
Hi, sorry, I was not aware of closing the PR.

pwiz/data/vendor_readers/Waters/ChromatogramList_Waters.cpp
```diff
@@ -329,10 +355,27 @@ PWIZ_API_DECL void ChromatogramList_Waters::createIndex() const
...
+        ie.index = index_.size() - 1;
+        ie.function = -1;
+        ie.offset = ch;
+        ie.Q1 = 0;
+
+        std::string temp = rawdata_->AnalogChannelNames()[ch];
+
+        boost::algorithm::trim(temp);
+        ie.id = temp;
+        ie.chromatogramType = MS_chromatogram;
```
chambm, line 373: I'm still not sure what kind of chromatograms these are. We also should (bordering on must) have an example file to include in Reader_Waters_Test.data to test this code.

pwiz_aux/msrc/utility/vendor_api/Waters/WatersRawFile.hpp
```diff
@@ -524,6 +538,26 @@ struct PWIZ_API_DECL RawData
         }
     }
 
+    private:
+
+	void readAnalogChromatograms()
```
chambm, line 543: This is a stray tab instead of space indentation.

I do want to see performance measurement for this because it's reading all the analog data just to open the file even if the user isn't interested in the analog data.

pwiz/data/vendor_readers/Waters/ChromatogramList_Waters.cpp
xjasg: I´m on it.

pwiz_aux/msrc/utility/vendor_api/Waters/WatersRawFile.hpp
xjasg: Done.

pwiz/data/common/ParamTypes.hpp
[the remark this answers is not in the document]
xjasg: Done.

pwiz_aux/msrc/utility/vendor_api/Waters/WatersRawFile.hpp
xjasg: Performance measurement still needs to be done. I´m working on it.

xjasg:
I have to apologize,  but I'm not that familiar with your code. Therefore, the review efforts are certainly a bit higher. I will still rework the rest, but I am currently only part-time on this project, so it may take another 1-2 days. Thanks a lot.

chambm:
> I have to apologize, but I'm not that familiar with your code. Therefore, the review efforts are certainly a bit higher. I will still rework the rest, but I am currently only part-time on this project, so it may take another 1-2 days. Thanks a lot.

No hurry. Let me know if there's something about the code you don't understand (like adding a new test file).

xjasg:
Hi Matt, maybe some help on "how to add a new test file" would be very helpful. I copied my file container to the folder ....\vendor_readers\waters\Reader_Waters_Test.data. After that I added a new test function to ....\pwiz\data\vendor_readers\Waters\Reader_Waters_Test.cpp, but after creating the program there is no mzml file. Do you have a description how to do it normally? Many thanks!

xjasg:
Hi Matt, I have since found out some more. I have now copied my test folder to the test data directory. Previously I had created a mzml file via MsConvertGui and copied it there as well. But it seems that the analog chromatograms are not recognized during testing. At least it brings error messages. I then noticed that there is a command line parameter that can be used to create an mzml file when building. I have done that. The result mzml file looks exactly like my file built by MSConvertGui. But the following error messages appear in the .out file:

+
  msdata:
    id: QC_LCMS2-2_23_268-1-1
    version: 1.1.0
    run:
      id: QC_LCMS2-2_23_268-1-1
      spectrumList (0 spectra)
      chromatogramList (4 chromatograms)
-
  msdata:
    id: QC_LCMS2-2_23_268-1-1
    version: 1.1.0
    run:
      id: QC_LCMS2-2_23_268-1-1
      spectrumList (0 spectra)
      chromatogramList (4 chromatograms)
+
  chromatogram:
    index: 1
    id: System Pressure
    defaultArrayLength: 0
    userParam: find() result, 5
-
  chromatogram:
    index: 1
    id: System Pressure
    defaultArrayLength: 0
    userParam: find() result, 1
+
  chromatogram:
    index: 2
    id:  A
    defaultArrayLength: 0
    userParam: find() result, 5
-
  chromatogram:
    index: 2
    id:  A
    defaultArrayLength: 0
    userParam: find() result, 2
+
  chromatogram:
    index: 3
    id: (1) CM Selected Column Temp
    defaultArrayLength: 0
    userParam: find() result,

....

It seems like the Testprogramm does not create these new chromatograms because the index found is equal to the size of the chromatogram list. I will check in my test files so you can take a look if you like.

xjasg:
In the meantime I have found the problem and fixed it. The performance measurement remains. I continue to work on it. Thank you.

pwiz/utility/misc/VendorReaderTestHarness.cpp
```diff
@@ -387,12 +402,15 @@ void testRead(const Reader& reader, const string& rawpath, const bfs::path& pare
         // test SpectrumList::min_level_accepted() for vendors
         if (msd.run.spectrumListPtr && msd.run.spectrumListPtr->size() > 0)
         {
-            DetailLevel msLevelDetailLevel = msd.run.spectrumListPtr->min_level_accepted([](const Spectrum& s) { return s.hasCVParam(MS_ms_level); });
+            DetailLevel msLevelDetailLevel = msd.run.spectrumListPtr->min_level_accepted([](const Spectrum& s) { return !ContainsMsInfo(s)
```
chambm, line 405: Good catch, I didn't realize we didn't have any existing tests with non-MS spectra which would have already triggered this edge case. I think we can simplify the predicate though and avoid the need for `ContainsMsInfo`:
```
DetailLevel msLevelDetailLevel = msd.run.spectrumListPtr->min_level_accepted([](const Spectrum& s) { return s.hasCVParamChild(MS_mass_spectrum) ? s.hasCVParam(MS_ms_level) : boost::indeterminate;});
```
(indeterminate means it needs to try another spectrum, false means it will try a higher detail level; the latter would be wrong for an EMR spectrum because even the highest detail level won't have an MS level)

If the spectrumList doesn't have ANY MS spectra, then this would still fail. But that's not the case for your example file, right?

pwiz_aux/msrc/utility/vendor_api/Waters/WatersRawFile.hpp
chambm, line 1: After you do your performance measurement just let me know what the results are. We don't need to persist this measurement code in the source.

pwiz/utility/misc/VendorReaderTestHarness.cpp
xjasg: Thank you very much. Okay, I'll take a look at that. I'm not sure if there is another Ms spectrum in my sample. I will try to find out if there is one.

QC_LCMS2-2_23_268-1-1.mzML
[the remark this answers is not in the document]
xjasg: Yes this was not intended. Thank you.

QC_LCMS2-2_23_268-1-1.raw/_CHRO001.DAT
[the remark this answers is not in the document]
xjasg: Yes, it's already there, but there was some test data for my tests with MsConvertGui. But these I did not want to check in.

pwiz/data/vendor_readers/Waters/Reader_Waters_Test.data/QC_LCMS2-2_23_268-1-1.raw/_CHRO001.DAT
xjasg, line 1: Actually the test data was not checked in because of .gitignore. But I tested only with this data.

pwiz/utility/misc/VendorReaderTestHarness.cpp
xjasg: My idea was to extend the list for other spectra which don´t have a ms spectrum. But, sure, your solution is more general. But I still have to check if my example has other ms spectra.

xjasg: I took another closer look at the code and I think I understand it better now. My example would find a Ms1 Spectrum, even if none would be present if we had an EMR Spectrum. That would certainly not be correct. The way you suggested it should work if there was at least 1 Ms1 Spectrum. Seems to work now. I´ll check it in.

chambm:
Can you give me push rights to your repo? I can fix the compile issue and possibly other things.

chambm:
Your QC file does have MS spectra so that min_level_accepted test should work fine. There's a lambda syntax thing I have a fix for (and also help with any other issues).

PR #2718, opened 2023-09-14
Add "CommonAlertDlg" in CommonUtil project

nickshulman:
This change adds a form "CommonAlertDlg" in the "CommonUtil" project which can be used by SkylineBatch and AutoQC in addition to Skyline.
It's in the package "pwiz.Common.GraphicalUserInterface" (is that a good name for the package?)
I also added "pwiz.Common.GraphicalUserInterface" to the list of things that Model code is not allowed to reference in CodeInspectionTest.

Notes:
SkylineBatch would show different icons on the message boxes depending on whether it was an Error, Warning, or Question. That functionality was not preserved with this change. All message boxes look the same.
Things such as the name of the application, or whether a functional test is running are now stored in a class called "CommonApplicationSettings" in CommonUtil.
Localizable strings that are expected to be used in a lot of places such as "OK", "Yes", etc. are now in a .resx file called "GeneralTerms.resx".

Let me know if it seems that something should work differently or be named differently or be moved somewhere else.


vagisha (approved):
This looks OK to me.  Package name could be pwiz.Common.GUI if you want it to be shorter.

PR #2719, opened 2023-09-15
Feature/mzMLb

chambm:
Adds support for mzMLb format, an HDF5-based format that stores the binary arrays in HDF5 datasets but keeps the mzML metadata in XML. So it supports the 3-array ion mobility format and is apparently easier for Java and Python-based HDF5 parsers to handle due to different choices about which HDF5 features to use.


pwiz/data/msdata/mzmlb/Connection_mzMLb.cpp
```diff
@@ -0,0 +1,612 @@
...
+            }
+            
+            hid_t atype = H5Aget_type(aid);
+            {
+                hid_t atype_mem = H5Tget_native_type(atype, H5T_DIR_ASCEND);
+                {
+                    char* ver = new char[H5Tget_size(atype)];
+                    H5Aread(aid, atype_mem, ver);
+                    std::string version = ver;
+                    if (version != "mzMLb 0.6")
```
mobiusklein, line 97: Is this the version number that will be used when this branch is merged?

pwiz/data/msdata/mzmlb/Connection_mzMLb.cpp
```diff
@@ -0,0 +1,612 @@
...
+        // write version string
+        hid_t aid = H5Screate(H5S_SCALAR);
+        {
+            hid_t atype = H5Tcopy(H5T_C_S1);
+            {
+                H5Tset_size(atype, 10);
+                H5Tset_strpad(atype, H5T_STR_NULLTERM);
+                hid_t attrid = H5Acreate(mzML_.dataset, "version", atype, aid, H5P_DEFAULT, H5P_DEFAULT);
+                {
+                    std::string version_out = "mzMLb 0.6";
```
mobiusklein, line 174: Again, is this the value going forward? This should probably be factored out as class static member or macro-defined?

PR #2722, opened 2023-09-18
Don't add 0.01 to Q3 m/z values for Agilent CE Opt transition lists

nickshulman:
When exporting Agilent CE Optimization Transition Lists, stop adding 0.01 to the Q3 m/z values (requested by Marilyn)

### DIFF
--- a/pwiz_tools/Skyline/Model/Export.cs
+++ b/pwiz_tools/Skyline/Model/Export.cs
@@ -3046,7 +3046,9 @@ namespace pwiz.Skyline.Model
             writer.Write(FieldSeparator);
             writer.Write(@"Unit");   // MS1 Res
             writer.Write(FieldSeparator);
-            writer.Write(GetProductMz(SequenceMassCalc.PersistentMZ(nodeTran.Mz), step).ToString(CultureInfo));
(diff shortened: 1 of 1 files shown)


pwiz_tools/Skyline/Model/Export.cs
```diff
@@ -3046,7 +3046,9 @@ protected override void WriteHeaders(TextWriter writer)
             writer.Write(FieldSeparator);
             writer.Write(@"Unit");   // MS1 Res
             writer.Write(FieldSeparator);
-            writer.Write(GetProductMz(SequenceMassCalc.PersistentMZ(nodeTran.Mz), step).ToString(CultureInfo));
+            // For Agilent we do not call GetProductMz because we want all of the Q3 m/z values to be the same
+            // for all of the optimization step chromatograms
+            writer.Write(SequenceMassCalc.PersistentMZ(nodeTran.Mz).ToString(CultureInfo));
```
brendanx67, line 3051: I would have just replace "step" with 0. What if GetProductMz() somehow gets expanded in the future?

brendanx67 (approved):
Thanks, Nick.

PR #2731, opened 2023-09-27
Skyline: Include process output in ProcessRunner exceptions when it…

brendanx67:
…would otherwise be lost
- for diagnosing method export failures


brendanx67:
Not sure we will merge this to master, but it may be useful in diagnosing a current support request on a failure exporting Waters methods.

brendanx67:
I think we could merge this now. It puts the output into a nested IOException so that the message doesn't change but the output is available if the user clicks More Info. One thing I am unsure about is that I did this only for IOException, since I didn't want to write another WrapAndThrowException and I didn't want to lose the exception type in the throw that wraps the exception.

pwiz_tools/Shared/CommonUtil/SystemUtil/ProcessRunner.cs
```diff
@@ -215,6 +222,13 @@ public static bool GoodIfExitCodeIsZero(string stderr, int exitCode)
                 }
 
             }
+            catch (IOException ex)  // CONSIDER: Should we handle more types like WrapAndThrowException does?
+            {
+                if (sbOutput != null)
```
nickshulman, line 227: I would probably make this:
`if (sbOutput != null && sbOutput.Length > 0)`
or
`if (sbOuput?.Length > 0)`
so that we don't bother wrapping the exception unless there's at least one character of captured output.

pwiz_tools/Shared/CommonUtil/SystemUtil/ProcessRunner.cs
```diff
@@ -215,6 +222,13 @@ public static bool GoodIfExitCodeIsZero(string stderr, int exitCode)
                 }
 
             }
+            catch (IOException ex)  // CONSIDER: Should we handle more types like WrapAndThrowException does?
```
nickshulman, line 225: I would probably catch all exception types here, and not worry about the fact that the type of the exception might be getting changed to IOException from whatever its original type might have been.
But this way is fine too.

nickshulman (approved):
Looks good to me.

It's unfortunate that the way "CommandLine.ExportInstrumentFile" is currently written, it will catch IOException and throw away this new extra information:
```
            catch (IOException x)
            {
                _out.WriteLine(Resources.CommandLine_ExportInstrumentFile_Error__The_file__0__could_not_be_saved___Check_that_the_specified_file_directory_exists_and_is_writeable_, args.ExportPath);
                _out.WriteLine(x.Message);
                return false;
            }
```

pwiz_tools/Shared/CommonUtil/SystemUtil/ProcessRunner.cs
brendanx67: After considering, I guess I see value in having "Output:" in the exception details to make it completely clear that Skyline was collecting output but didn't get any before the exception was thrown. Otherwise, I can imagine wondering if the code to collect output failed for some reason. So, I am going to leave this as is.

brendanx67:
> Looks good to me.
> 
> It's unfortunate that the way "CommandLine.ExportInstrumentFile" is currently written, it will catch IOException and throw away this new extra information:
> 
> ```
>             catch (IOException x)
>             {
>                 _out.WriteLine(Resources.CommandLine_ExportInstrumentFile_Error__The_file__0__could_not_be_saved___Check_that_the_specified_file_directory_exists_and_is_writeable_, args.ExportPath);
>                 _out.WriteLine(x.Message);
>                 return false;
>             }
> ```

So, you are suggesting that we use _out.WriteLine(x) instead?

pwiz_tools/Shared/CommonUtil/SystemUtil/ProcessRunner.cs
brendanx67: Okay. That makes sense. I was thinking that catching Exception you would throw Exception, but if you catch exception and throw IOException that seems a bit better to me.

PR #2736, opened 2023-10-03
Deal with Waters data that has lockmass channel, but leads with measurements in a non-MS channel…

bspratt:
… which confused our lockmass channel detection logic


chambm (commented):
I think this is a reasonable backup for mzML files as you mentioned. However, I suspect for files which do have EMR spectra, they could easily have dozens or hundreds of EMR spectra before the first MS spectrum. Is this just a performance optimization to avoid enumerating those spectra, or will getting this wrong mean that those spectra are actually used for extraction?

bspratt:
I think the "skip functions greater than the presumed lockmass function" log is pretty much an optimization since they'd be ignore anyway with an MS level. But ideally, yes, they'd never get passed in at all, nor would lockmass. There's already a flag for ReaderConfig "ignoreCalibrationScans" but it doesn't seem to be doing anything in this case.

chambm:
Huh, somehow I had read the new `for` loop as only looking at the first 5 spectra. That's why I mentioned there perhaps being dozens or hundreds of EMR spectra. If you're potentially checking the whole list, that's not an issue.

chambm (approved):

PR #2747, opened 2023-10-13
Skyline (23.1): Update small molecule tutorials to remove the word...

brendanx67:
…"peptide" and update screenshots


bspratt:
Looks good. Missed opportunity to show the updated red/green/yellow indicators in the targets tree, but that's not nearly as important and the updated document grid screenshots.

brendanx67:
This is just updating to 23.1 which does not have the new indicators.

Issue #2748, opened 2023-10-13
 Spectral Library Explorer filter, "StartsWith" vs "Contains"


meowcat:
https://skyline.ms/announcements/home/support/thread.view?rowId=62625

PR #2749, opened 2023-10-13
Skyline: Filter spectral library using Contains not StartsWith (#2748)

meowcat:
This is a first suggestion according to #2748. The implementation is naive but on a (small molecule) library with 13k entries I have no observable performance impact. Now I don't know how large libraries in the proteomics world are.

As a first-time contributor I appreciate any pointers and suggestions.


pwiz_tools/Skyline/SettingsUI/ViewLibraryPepInfoList.cs
```diff
@@ -29,6 +29,26 @@
...
+                throw new ArgumentNullException("substring",
+                                             "substring cannot be null.");
+            else if (!Enum.IsDefined(typeof(StringComparison), comp))
+                throw new ArgumentException("comp is not a member of StringComparison",
+                                         "comp");
+
+            return str.IndexOf(substring, comp) >= 0;
+        }
+    }
+
```
bspratt, line 51: This is nice, but probably overkill - just using IndexOf directly is cleaner and easier to understand.

pwiz_tools/Skyline/SettingsUI/ViewLibraryPepInfoList.cs
```diff
@@ -278,8 +313,8 @@ public IList<int> Filter(string filterText, string filterCategory)
             // We have to deal with the UnmodifiedTargetText separately from the adduct because the
             // adduct has special sorting which is different than the way adduct.ToString() would sort.
 
-            // Find the indices of entries that have a field that could match the search term if something was appended to it 
-            var filteredIndices = PrefixSearchByProperty(filterText);
+            // Find the indices of entries that have a field that contain the search term
+            var filteredIndices = ContainsSearchByProperty(filterText);
```
bspratt, line 317: This would be a performance hit for the conventional use case. But changing PrefixSearchByProperty to return and empty list when filterText starts with "*" would be useful.

pwiz_tools/Skyline/SettingsUI/ViewLibraryPepInfoList.cs
```diff
@@ -304,7 +339,8 @@ public IList<int> Filter(string filterText, string filterCategory)
                         StringComparison.OrdinalIgnoreCase));
                 // Return the elements from the range whose DisplayText actually matches the filter text.
                 return ImmutableList.ValueOf(new RangeList(range).Where(i => _allEntries[i].DisplayText
-                    .StartsWith(filterText, StringComparison.OrdinalIgnoreCase)));
+                    .Contains(filterText, StringComparison.OrdinalIgnoreCase)));
+                    //.StartsWith(filterText, StringComparison.OrdinalIgnoreCase)));
```
bspratt, line 343: I'd leave this alone and have a seperate clause for filterText starts with "*", something like

                if (filterText.StartsWith(@"*") && !_continuousFields.Contains(_selectedFilterCategory))
                {
                    // Wildcard
                    var pattern = filterText.Substring(1);
                    return ImmutableList.ValueOf((new RangeList(0, _allEntries.Count-1)).Where(i => _allEntries[i].DisplayText
                        .IndexOf(pattern, StringComparison.OrdinalIgnoreCase)>=0));
                }

meowcat: I removed the wildcard search for this separate "if nothing found" case for now, since this seems to be a special case for peptides, or rather I don't really understand when it is triggered.

meowcat:
Note: some of the ViewLibraryDlg changes are just what my VS2022 does to it. I somehow can't make it leave it alone.

bspratt:
> Tooltip is still missing localization

As long as it's in the resx, it will get localized later - nothing you have to worry about.

But I'm going to go through and make sure the changes are minimal and agree with our team practices, if that's OK. Can I just push to your branch?

meowcat:
> But I'm going to go through and make sure the changes are minimal and agree with our team practices, if that's OK. Can I just push to your branch?
Sure! I just added you to the repo. 

If you have overall recommendations, I would be happy for input, since I have one more PR upcoming.

bspratt:
I've branched your branch in your repo, have a look.

meowcat:
Thanks for the changes! From my point of view we can merge. 

A note: The current tooltip
`Use * to search within descriptions, e.g. *PEET will match PEETID and AAPEETB`
might suggest to users that they can do `AAP*TB`. We can obviously also implement this with some glob-to-regex, but at some point the performance hit is going to be significant.

bspratt:
Fair point. Maybe "Filters beginning with * search within descriptions, e.g. *PEET will match PEETID and AAPEETB"?

meowcat:
Sounds good. I'll push the change once I finally get the tests to work on my computer.

nickshulman:
If we wanted to avoid introducing the new "*" syntax for filtering the list of peptides, a different way to implement this could be to make it so that when you filter for a particular string, all of the peptides that contain that string show up, but the ones which match "Starts With" appear before the ones that only matched with "Contains".
It would be similar to how ReSharper orders the matches when you use "Search Everywhere", and as you go down the list the matches become more and more fuzzy.

bspratt:
I like that idea, but there are performance implications for Contains that perhaps are better understood by the user when they've requested it explicitly.

brendanx67:
You could try the Pan Human assay library or generate something massive
with Prosit.

bspratt:
Peptide libraries can be quite large - if you examine the original StartsWith code you can see a fair bit of trickery to deal with performance issues.  If we wanted to make Contains the defult we'd have to make sure that the search is easily user interruptable, I'd think. At least in the current proposed code the user has asked for the extra depth of search and should not be surprised by some extra wait time.

brendanx67:
Is the search done on a background thread? I could imagine the UI Nick
describes if the search were performed on a background thread. Then we
could show whatever we can in the first 500 ms and follow that with the
rest when it becomes available.

bspratt:
The current implementation is not multithreaded, unfortunately.

bspratt:
I think this is good within the confines of the current single threaded architecture. Once Nick or Brendan signs off on it, we'll accept the PR.

brendanx67 (approved):
Looks good. Thanks for the new feature.

bspratt:
Excellent, I'll get it merged into master.

PR #2752, opened 2023-10-16
Add validation for cutoff score in ImportPeptideSearchDlg

bspratt:


pwiz_tools/Skyline/FileUI/PeptideSearch/BuildPeptideSearchLibraryControl.cs
```diff
@@ -643,5 +656,15 @@ private void comboInputFileType_SelectedIndexChanged(object sender, EventArgs e)
         {
             UpdatePerformDDASearch();
         }
+
+        #region test_support
+
+        // For test purposes, lets us test error handling
+        public void SetCutoffControlText(string cutoffControlText)
```
chambm, line 663: If properties like CutOffScore are only supposed to be used by tests and not by other Skyline application code, then I wonder whether it should be a string that gets/sets textCutoff.Text. Other than that, PR looks good.

bspratt: I don't understand. The user can only deal in strings, we have to convert strings like textCutoff to values like CutoffScore, and gracefully deal with users typing in strings that don't make sense. We're pretty good about that, but for whatever reason this particular control never got any such validation code. The problem has been lurking for at least ten years.

bspratt: I see. Keep in mind that the test would have to think about L10N when trying to represent fractional values, though. Probably even more awkward.

chambm: Yes, devs would have to remember to do the `.ToString(CultureInfo.CurrentCulture)` bit after the number (or use a helper function). Which I admit is easy to forget (and I have done so).

However the `double CutoffScore` is much closer to the way the rest of Skyline does this testing. The difference is things like transition settings UI use a nullable property like `double? CutoffScore` and test that when setting CutoffScore to null, it triggers the `Resources.MessageBoxHelper_ValidateDecimalTextBox__0__must_contain_a_decimal_value` error. Is there a reason not to use `ControlUtil.ValidateDecimalTextBox` here? Testing a null control value is probably close enough to an invalid value like "12%@$@": they both should generate the same error.

bspratt: Yes, I should use that more centralized code for consistency, will do.

bspratt:
Looks like it would be quite a lot of refactoring to centralize parsing of doubles - typical pattern is to use ValidateDecimalTextBox to verify that a text box can be understood as a double, then parse again later with a bare double.Parse(). So, a more L10N-tolerant handling of doubles is a nice idea but perhaps not worth the effort.

chambm:
And handling it as an optional input seems tricky too with just a double?. Using null as "invalid input" means it can't also be used for "user left field empty".

brendanx67:
Please don’t merge this until I can have a look at it.

pwiz_tools/Skyline/FileUI/PeptideSearch/BuildPeptideSearchLibraryControl.cs
```diff
@@ -281,13 +281,27 @@ public void AddSearchFiles(IEnumerable<string> fileNames)
             SearchFilenames = BuildLibraryDlg.AddInputFiles(WizardForm, SearchFilenames, fileNames, PerformDDASearch);
         }
 
+        // For test purposes, lets us test error handling
+        public string CutOffScoreText
```
brendanx67, line 285: Let's implement this as an internal private _cutoffScore that starts as null (default) and only gets its double value in the Next() function with code like:

var helper = new MessageBoxHelper(this.ParentForm);
if (!helper.ValidateDecimalTextBox(textCutoff, out _cutoffScore))
    return;

public double? CutoffScore would be get only to get the _cutoffScore and CutoffScoreText could remain to set the control text.

brendanx67 (changes requested):
Close, but I would prefer a little code clean-up to make it clearer that the standard pattern is being employed. Put the MessageBoxHelper in the NextPage() function and use it the set a private double? _cutoffScore.

PR #2754, opened 2023-10-19
Add support for reading chemical formulas with numeric subscripts, e.g. C₁₀H₁₄N₅O₈P

bspratt:
As seen in a recent ExceptionWeb report


pwiz_tools/Skyline/TestFunctional/PasteMoleculesTest.cs
```diff
@@ -645,6 +646,7 @@ private void TestLegitimatePaste(string line2start, string[] fullColumnOrder)
             {
                 docEmpty = NewDocument();
                 line1 = BuildTestLine(imTypeIsDrift);
+                line1 = line1.Replace(caffeineFormula+"\t", caffeineFormulaUnicode + "\t"); // Test with unicode subscript numbers
```
chambm, line 649: Doesn't this mean the original caffeineFormula isn't being tested anymore?

chambm:
Interesting! What about subscripts used to indicate isotope like <sub>13</sub>C? I suppose when used like that, the element counts have to be normal size?

bspratt:
> Interesting! What about subscripts used to indicate isotope like 13C? I suppose when used like that, the element counts have to be normal size?

Actually for isotopes those are normally superscripts, I think - but yes, worth considering. Mostly we don't deal with that nomenclature, though - we use the C' and C" thing.

chambm:
> > Interesting! What about subscripts used to indicate isotope like 13C? I suppose when used like that, the element counts have to be normal size?
> 
> Actually for isotopes those are normally superscripts, I think - but yes, worth considering. Mostly we don't deal with that nomenclature, though - we use the C' and C" thing.

Oops! It's good they don't overlap.

bspratt:
I could imagine superscripts as a potential issue in adduct parsing, e.g. [M2C13+H] (labeled with two heavy carbon) might be presented as [M2C¹³+H].

I could also imagine the general formula parser being expected to understand H2¹³C5 (or H₂¹³C₅) as meaning H2C'5.

brendanx67:
Remember that it is entirely appropriate to simply report an error message
to a user that they can solve and that making it to the exception reporting
handler is a sign of poorly protected code for error handling even when it
catches something we might be able to write new code to interpret correctly.

bspratt:
Indeed - this shouldn't have been passed along to us, that bug was fixed last week. But I'm glad this was passed along, a bit of luck there.

nickshulman:
It sounds like the unhandled error problem will be fixed in the next Skyline-daily by PR #2741 

Using non-ascii digits in a chemical formula is usually in the "doctor it hurts when I do this" category. I would recommend holding off on implementing support for this until a user tells us they need it.

bspratt:
Respectful disagreement - the user did tell us that they need it. Granted that was made easier for them due to a broken exception handler, but it's proof of this kind of data in the wild.

It's not like they typed in garbage: anyone looking at that would know what it means and be confused by Skyline's inability to do the same. Probably it was a copy and paste from a document or webpage - the world becomes less ASCII and more Unicode day by day. 

I'm fine with holding off worrying about the isotope superscript thing since we haven't seen that in the wild.

bspratt:
> Like I wish we could do better with tables from PDFs. 

Yes, so frustrating. For some reason the end of line doesn't get encoded in the copy.

>  I think this is fine.

I'll take that as formal approval, then?

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/TestFunctional/PasteMoleculesTest.cs
```diff
@@ -123,6 +123,7 @@ private static string ToLocalText(string text)
         const string caffeineSMILES = "Cn1cnc2n(C)c(=O)n(C)c(=O)c12";
         const string caffeineKEGG = "C07481";
         const string caffeineFormula = "C8H10N4O2";
+        const string caffeineFormulaUnicode = "C\u2088H\u2081\u2080N\u2084O\u2082"; // Unicode subscripts
```
brendanx67, line 126: I think a good idea to add something with a 9 in it to cover the full range from 0 to 9. I like that you have the H10 which covers both multiple characters and zero.

brendanx67 (approved):
Yes. I am okay with this fix. I would like to see a test of \u2089 added to cover the full range that is supposed to work.

PR #2763, opened 2023-10-31
Update unimod.xml and UniModData.cs

brendanx67:
Thermo requested TMTpro be added to our modifications. So, after 10+ years, this is what an update to unimod.xml looks like to make that happen.


nickshulman (approved):
Looks good to me.
If we wanted, we could get rid of all of the "PrecisionRequired" stuff.
I think that stuff was added by mistake as part of the high precision bibliospec modification work, but was never actually necessary for the implementation that we eventually ended up going with.

PR #2764, opened 2023-10-31
Gene level parsimony option in protein association dialog

chambm:
- added gene level parsimony option in protein association dialog
* bumped schema to 23.11 for gene level parsimony
* updated audit logs
* fixed Skyline build system to copy vendor DLLs to both Debug and Release directories so vendor reader tests don't fail in Debug config
* removed MinPeptidesPerProtein test from TestHugeAssociateProteins because it's now hidden in the Refine -> Associate Proteins dialog


pwiz_tools/Skyline/CommandArgs.cs
```diff
@@ -1276,6 +1276,8 @@ public bool ImportingSearch
 
         public static readonly Argument ARG_AP_GROUP_PROTEINS = new Argument(@"associate-proteins-group-proteins",
             (c, p) => c.AssociateProteinsGroupProteins = p.IsNameOnly || bool.Parse(p.Value));
+        public static readonly Argument ARG_AP_GENE_LEVEL = new Argument(@"associate-proteins-gene-level-parsimony",
```
nickshulman, line 1279: Instead of having a name-only argument, do you think we should have an argument called "associate-proteins-parsimony-level" where the valid values are "protein" (default) or "gene"?

pwiz_tools/Skyline/CommandArgs.cs
```diff
@@ -1286,16 +1288,18 @@ public bool ImportingSearch
             (c, p) => c.AssociateProteinsMinPeptidesPerProtein = p.ValueInt) { WrapValue = true };
 
         private static readonly ArgumentGroup GROUP_ASSOCIATE_PROTEINS = new ArgumentGroup(() => Resources.CommandLine_AssociateProteins_Associating_peptides_with_proteins, false,
-            ARG_AP_GROUP_PROTEINS, ARG_AP_SHARED_PEPTIDES, ARG_AP_MINIMAL_LIST, ARG_AP_MIN_PEPTIDES, ARG_AP_REMOVE_SUBSETS)
+            ARG_AP_GROUP_PROTEINS, ARG_AP_GENE_LEVEL, ARG_AP_SHARED_PEPTIDES, ARG_AP_MINIMAL_LIST, ARG_AP_MIN_PEPTIDES, ARG_AP_REMOVE_SUBSETS)
             { LeftColumnWidth = 45 };
 
         public bool? AssociateProteinsGroupProteins { get; private set; }
+        public bool? AssociateProteinsGeneLevelParsimony { get; private set; }
```
nickshulman, line 1295: I think if you add "[DefaultValuesFalse]" above this property definition, the Audit Log won't write it out when it is equal to false, so you won't need to check in new copies of the tutorial log files in this pull request.

chambm: I think I was copying what I did for Group proteins, where I wanted it to show false even though that's the default (so users don't have to think about whether grouping proteins is the default behavior or not). But that makes less sense for gene level parsimony since it should be pretty obvious to users that the default will be by protein. So yeah, I may get to undo all those audit log changes...

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
```diff
@@ -561,11 +601,61 @@ public void ApplyParsimonyOptions(bool groupProteins, bool findMinimalProteinLis
...
+                        return proteinToPeptides.Keys.First();
+
+                    var longestProtein = proteinToPeptides.OrderByDescending(kvp2 => kvp2.Key.Sequence.Sequence.Length).First().Key;
+                    var allPeptides = proteinToPeptides.Values.SelectMany(o => o).Distinct().ToList();
+                    if (allPeptides.All(node => longestProtein.Sequence.Sequence.Contains(node.Peptide.Sequence)))
+                        return longestProtein;
+
+                    return new FastaRecord(longestProtein.RecordIndex, 0,
+                        new FastaSequence(longestProtein.Sequence.Name, longestProtein.Sequence.Description, null,
+                            string.Join("", proteinToPeptides.Keys.Select(o => o.Sequence.Sequence))), longestProtein.Metadata);
```
chambm, line 652: @brendanx67 This is what I came up with to pick/make a protein sequence that contains all the member peptides. It will use the longest sequence of the group if it contains all the peptides, and if it doesn't then it'll concatenate all the protein sequences together. Some alternatives might be: concatenate all the peptides instead of the proteins, or only concatenate enough proteins to contain all the peptides (i.e. a parsimonious approach to concatenating).

Or throw this hacky workaround away and go back to the drawing board, making gene group a new FastaSequenceGroup that allows its peptides to map to only some of the parent's FastaSequenceList. I honestly don't know how much work that would be though.

chambm:
@brendanx67 Good to go for this PR?

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
```diff
@@ -561,11 +610,65 @@ public void ApplyParsimonyOptions(bool groupProteins, bool findMinimalProteinLis
             _finalResults.FinalSharedPeptideCount = sharedPeptidesRemaining.Values.Sum();
         }
 
-        private Dictionary<IProteinRecord, PeptideAssociationGroup> CalculateProteinGroups(MappingResultsInternal results, ILongWaitBroker broker)
+        public class GeneLevelEqualityComparer : EqualityComparer<IProteinRecord>
```
nickshulman, line 613: Instead of writing an EqualityComparer, I probably would have written a new class called "GeneOrProteinSequence" which I used as the Key in the dictionary.
The reason is that it's easier to understand what is going on if the dictionary has a completely different key type, compared to if it has a special EqualityComparer.

But, this way is fine, too.

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
```diff
@@ -561,11 +610,65 @@ public void ApplyParsimonyOptions(bool groupProteins, bool findMinimalProteinLis
...
+                    if (proteinToPeptides.Count == 1)
+                        return proteinToPeptides.Keys.First();
+
+                    var longestProtein = proteinToPeptides.OrderByDescending(kvp2 => kvp2.Key.Sequence.Sequence.Length).First().Key;
+                    var allPeptides = proteinToPeptides.Values.SelectMany(o => o).Distinct().ToList();
+                    if (allPeptides.All(node => longestProtein.Sequence.Sequence.Contains(node.Peptide.Sequence)))
+                        return longestProtein;
+
+                    // each protein's individual metadata is kept, but all protein sequences are replaced by the concatenated sequence
+                    var concatenatedSequence = string.Join(string.Empty, proteinToPeptides.Keys.Select(p => p.Sequence.Sequence));
```
nickshulman, line 660: You could do "string.Concat(" instead of "string.Join(string.Empty,"

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
```diff
@@ -561,11 +610,65 @@ public void ApplyParsimonyOptions(bool groupProteins, bool findMinimalProteinLis
...
+                    return new FastaRecord(longestProtein.RecordIndex, 0,
+                        new FastaSequenceGroup(longestProtein.Sequence.Name,
+                            proteinToPeptides.Keys.Select(p => new FastaSequence(p.Sequence.Name,
+                                p.Sequence.Description, p.Sequence.Alternatives, concatenatedSequence)).ToList()),
+                        new ProteinGroupMetadata(proteinToPeptides.Keys.Select(p => p.Metadata).ToList()));
+                }
+                proteinOrGeneToPeptideGroup = geneToPeptides.ToDictionary(kvp => GenerateConcatenatedSequenceIfNecessary(kvp.Value),
+                    kvp => new PeptideAssociationGroup(kvp.Value.Values.SelectMany(o => o).Distinct().ToList()));
```
nickshulman, line 668: You might want to make this:
`kvp => new PeptideAssociationGroup(kvp.Value.Values.SelectMany(o => o).Distinct((IEqualityComparer<PeptideDocNode>)ReferenceValue.EQUALITY_COMPARER).ToList())`
in order to avoid calling "Equals" on the PeptideDocNode's

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
```diff
@@ -561,11 +610,65 @@ public void ApplyParsimonyOptions(bool groupProteins, bool findMinimalProteinLis
...
+                        geneToPeptides.Add(kvp.Key, new Dictionary<IProteinRecord, List<PeptideDocNode>> { { kvp.Key, kvp.Value.Peptides } });
+                    else
+                        geneToPeptides[kvp.Key][kvp.Key] = kvp.Value.Peptides;
+                // now pick the protein with the longest sequence if it contains all the peptides, or a concatenation of all of the sequences if not
+                IProteinRecord GenerateConcatenatedSequenceIfNecessary(Dictionary<IProteinRecord, List<PeptideDocNode>> proteinToPeptides)
+                {
+                    if (proteinToPeptides.Count == 1)
+                        return proteinToPeptides.Keys.First();
+
+                    var longestProtein = proteinToPeptides.OrderByDescending(kvp2 => kvp2.Key.Sequence.Sequence.Length).First().Key;
```
nickshulman, line 654: This function is checking whether the longest protein sequence contains all of the peptide sequences.
It's true that the longest protein sequence is the one that is most likely to contain all of the peptide sequences, but it is theoretically possible that a shorter protein sequence in the group might happen to contain all the peptide sequences.
(This probably doesn't matter).

nickshulman (commented):
Looks very nice.

Here are some comments.
The only comments that I actually care about are the two where "Distinct" is being called on a list of PeptideDocNode objects.
Feel free to ignore all my other comments.

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
```diff
@@ -561,11 +610,65 @@ public void ApplyParsimonyOptions(bool groupProteins, bool findMinimalProteinLis
...
+                    var allPeptides = proteinToPeptides.Values.SelectMany(o => o).Distinct().ToList();
```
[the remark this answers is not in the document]
chambm: Oops, I missed this when I was reviewing your reversion of my overzealous ReferenceValue refactoring. But it seems I wouldn't need to change these usages if we kept my overzealous approach, right? It might just be simpler to use ReferenceValue<PeptideDocNode> *everywhere* in ProteinAssociation with an explicit comment that it's to "simplify" the code and make it clear the nodes are almost never mutated.

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
```diff
@@ -561,11 +610,65 @@ public void ApplyParsimonyOptions(bool groupProteins, bool findMinimalProteinLis
...
+        private Dictionary<IProteinRecord, PeptideAssociationGroup> CalculateProteinOrGeneGroups(MappingResultsInternal results, bool geneLevel, ILongWaitBroker broker)
         {
             var _peptideGroupToProteins = new Dictionary<PeptideAssociationGroup, List<IProteinRecord>>();
 
-            foreach(var kvp in AssociatedProteins)
+            var proteinOrGeneToPeptideGroup = AssociatedProteins;
+            if (geneLevel)
+            {
+                // gene to protein to peptides; the top level dictionary uses the GeneLevelEqualityComparer
+                var geneToPeptides = new Dictionary<IProteinRecord, Dictionary<IProteinRecord, List<PeptideDocNode>>>(new GeneLevelEqualityComparer());
```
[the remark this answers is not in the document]
chambm: Whenever I find myself writing code that checks whether a dictionary contains a key before adding an item to the associated value's container, I wish I had C++ map semantics whose operator[] default constructs the container if it's missing. :( So no need to ever have that if/else or GroupBy, just:
```
foreach (var kvp in AssociatedProteins)
    geneToPeptides[kvp.Key][kvp.Key] = kvp.Value.Peptides;
```

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
chambm: After I wrote this, I realized I could have just done a string compare with the gene name (or protein name if gene isn't set). Does that seem sensible to you as well?

pwiz_tools/Skyline/Model/Proteome/ProteinAssociation.cs
chambm: It seems very strange to me that you can cast Foo<object> to Foo<PeptideDocNode>.

PR #2765, opened 2023-10-31
Skyline: Clean-up "StaticMod.PrecisionRequired" changes

brendanx67:
- These changes ended up not getting used and just add confusion to the code
- Revert UniModCompiler/InputFiles/unimod.xml to what was actually used to generate UniModData.cs
- Add a few changes to UniModTemplate.cs to make it match UniModData.cs


brendanx67:
Nick, I decided to do a bit of clean-up of the "RequiredPrecision" code before applying my unimod.xml update in the hope that doing this will make the update easier to understand without all the changes to precision required for the modifications. Glad you final solution did not require this attribute. It does make a bit of a mess out of updates.

nickshulman (approved):
Looks good to me.
I fixed the line endings on Program.cs using a checkout which had the AutoCRLF git setting set to "false".
It looks like back in 2016 that file had some unix style line endings, so when you reverted to it, you got those line endings.

PR #2770, opened 2023-11-06
Waters dda and lockmass fixes

ianmorns:
1. Update to MassLynxSDK v4.11.0 which fixes:
  a Failure to export Waters DDA data if none-MS functions are present
  b Can't lockmass correct precursor m/z of Waters DDA where mass spectra are lockmass corrected in the instrument
  c allows other fixes below
2. Don't read binary data to build DDA index
  a When we build the "DDAIndex" prior to exporting data using the -ddaProcessing option we used to read but not use all of the spectra binary data. We've made use of a new endpoint to the SDK which makes it possible to read the necessary meta data (RT, setmass) without also reading the binary data
3. Ensure DDA precursor m/z are lockmass corrected for Waters data
  a The intention of the ddaprocessor is that the precursor is the lockmass corrected setmass. This often doesn't work because the precursor m/z values were calculated before lockmass correction was applied. We've changed it so precursor m/z is calculated just before they are needed.
4. Allow vendor peak picking for waters DDA data
  a Uses a new SDK endpoint which allows you to specify that spectra produced by the DDAProcessor should be centroided.
5. Use MassLynxSDK to identify lockmass function
  a Switch to using an endpoint in the MassLynx SDK to identify the lockmass function. This fixes issues where lockmass spectra were being exported even though 'ignoreCalibrationScans' option was being used.


pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
```diff
@@ -293,9 +275,10 @@ PWIZ_API_DECL SpectrumPtr SpectrumList_Waters::spectrum(size_t index, DetailLeve
     if (msLevel > 1 && isMS)
     {
         double setMass = 0;
+        double ddaPrecursorMass = 0;
         if (useDDAProcessor_)
         {
-            setMass = ie.setMass;
+            getDDAPrecursorMasses(index, setMass, ddaPrecursorMass);
```
chambm, line 281: Does this mean only the DDA processing mode will get lockmass-corrected m/zs? I suppose that is the most important case but it would be nice to be consistent to always provide all lockmass-corrected m/zs for all m/zs when lockmass is applied. Was an SDK function that just takes an m/z and returns the lockmass-corrected m/z not feasible?

ianmorns: I think the only other cases where you could lm correct the setmass would be for SONAR data. If you applied the correction to SONAR data then you would introduce a change that the SONAR quad window positions would vary slightly from spectrum to spectrum as lockmass correction varies over time.

There is an SDK function which returns a lockmass correction
`int getLockMassCorrection(const CMassLynxBaseProcessor mlLockMassProcessor, const float retentionTime, float* pGain);`
though that gives a 'gain' you have to multiply with your m/z to get the lockmass corrected version, a method which takes the uncorrected m/z and returns the corrected version would be nicer. The correction we are applying isn't quite as simple as that though. The lockmass correction which is used is the correction for the RT of the survey scan proceeding the MSMS spectrum, though in practice this will only make a tiny difference. The rationale behind this is that the setmass was selected based on measurements of the survey scan.

chambm: Good to know about the SDK function. I'd like to use for the setMass even when DDA processing mode hasn't been applied. But you're right that's probably the only situation it makes sense. It would probably be undesirable for the m/z bounds of DIA isolation windows to change, for example.

chambm:
The lockmass refiner test fails as it should. I'll handle updating that reference data.

PR #2773, opened 2023-11-09
exception handler for user attempting to modify base molecule in targets tree with an invalid chemical formula

bspratt:


nickshulman (approved):

nickshulman (approved):
Looks good to me.

There are a couple of support requests where the user managed to corrupt their document (i.e. it could not be opened after being saved) when using the Edit Custom Molecule dialog:
https://skyline.ms/announcements/home/support/thread.view?rowId=62755
https://skyline.ms/announcements/home/support/thread.view?rowId=62822

While you're in this area of the code, it might make sense to throw an exception if the adduct is not valid for the molecule, which would take care of the second of those two support requests.

bspratt:
Yeah, nice, I'll verify that this keeps users from hurting themselves in the ways described in those support requests. Thanks!

PR #2774, opened 2023-11-13
Avoid running TestToolService on machines where it is consistently flaky (just BRENDANX-UW7 for now)

bspratt:
N.B. you can argue that placing the logic in AbstractUnitTest instead of the test itself is overly ornate, but I wanted it near other machine-specific test logic.


nickshulman (approved):
Looks good to me.
I probably would have put all the logic for skipping this test in ToolServiceTest.cs, but I understand why you did it this way.

PR #2777, opened 2023-11-14
Move many of the resources from "Resources.resx" into separate .resx files

nickshulman:
I wrote a tool to figure out which source files were using which resources from Resources.resx.
For resources that were only used by source files in a particular folder, the resources get moved into a new .resx file "FOLDERNAMEResources.resx".

The tool which performed the refactor is in pwiz_tools\Skyline\Executables\AssortResources

How does the layout of these new resources look?
It would not be difficult to change the way the tool works if it turns out we'd like the resources organized differently.


chambm:
What's the idea behind doing this? The main file was too big for operations to run quickly on?

nickshulman:
> What's the idea behind doing this? The main file was too big for operations to run quickly on?

The main file is too big and changes too often. Every time that someone commits a change to Resources.resx, I believe our repository grows by 3MB because a new version of "Resources.resx", "Resources.designer.cs" etc. need to be stored for all time.

chambm:
I'm not sure about the 3MB thing once a non-binary file gets compressed into a git pack file, but just having to do that compression and decompression (and deltas) probably slows it down, so it does seem preferable to avoid churn on large files as much as possible. I tried to figure out how much of our repository is due to resource string updates (taking into account pack file compression) but my git-fu isn't up to the task. All I got is summing up the size of the pwiz_tools/Skyline/Properties/Resources.resx blobs, which is about 5GB. The repo's not that big, so the pack file compression is working. I guess we could take a scratch copy and rewrite the history removing all the pwiz_tools/Skyline/Properties/Resources.resx files and see how much smaller the resulting repo is...

chambm:
With help from GPT-4 I ran the commands to remove pwiz_tools/Skyline/Properties/Resources.resx from the history of a copy of the repo. It trimmed 800mb from the repo's overall size (4.2GB to 3.4GB). That is a sizeable chunk to be sure. I wonder if we should archive the repo at some point and start fresh (or past some arbitrary date)? Devs would keep a read-only copy of the archive around to check logs if necessary.

PR #2781, opened 2023-11-20
attempt to fix an intermittent failure in TestDetectionsPlot

bspratt:


nickshulman:
I wonder if @rita-gwen has any ideas about how this could be fixed.
In theory that `DetectionPlotData.GetDataCache().Status == DetectionPlotData.DetectionDataCache.CacheStatus.idle` is supposed to wait until the cache has no more work to do, so it might be that something in there is not working the way that we would expect.

bspratt:
> I wonder if @rita-gwen has any ideas about how this could be fixed. In theory that `DetectionPlotData.GetDataCache().Status == DetectionPlotData.DetectionDataCache.CacheStatus.idle` is supposed to wait until the cache has no more work to do, so it might be that something in there is not working the way that we would expect.

Yeah, I was thinking that we might be overrunning that - asking for its status before its even aware that there's work to be done

nickshulman (approved):

PR #2782, opened 2023-11-20
- added most peptide filter settings to CommandArgs/CommandLine

chambm:


brendanx67 (changes requested):
Looks good. Please add tests that check some boundary conditions with these new arguments, i.e. using these constants to set values just beyond the valid range and verify that you get an expected error.

        public const int MIN_EXCLUDE_NTERM_AA = 0;
        public const int MAX_EXCLUDE_NTERM_AA = 10000;
        public const int MIN_MIN_LENGTH = 2;
        public const int MAX_MIN_LENGTH = 100;
        public const int MIN_MAX_LENGTH = 5;
        public const int MAX_MAX_LENGTH = 200;

brendanx67 (approved):

PR #2783, opened 2023-11-21
Change the way developers can enable Prosit on local builds

bspratt:
If a file "Model/Prosit/Config/PrositConfig_production.xml" is available, the Skyline Jamfile will always copy it to PrositConfig.xml for embedding in the exe. When the bjam commandline includes "--official" the build still fails if "Model/Prosit/Config/PrositConfig_production.xml" is not available.

Also, remove a stale reference to Model\Prosit\Config\UpdatePrositConfiguration.bat in the .csproj file.

I'll update the wiki page once this is merged to master.

### DIFF
--- a/pwiz_tools/Skyline/Jamfile.jam
+++ b/pwiz_tools/Skyline/Jamfile.jam
@@ -343,15 +343,15 @@ if [ modules.peek : NT ] && --i-agree-to-the-vendor-licenses in [ modules.peek :
         ) > "$(<)"
     }
 
+    constant SKYLINE_PROSIT_CONFIG : "$(SKYLINE_PATH)/Model/Prosit/Config/PrositConfig_production.xml" ;
     if --official in [ modules.peek : ARGV ]
     {
-        constant SKYLINE_PROSIT_CONFIG : "$(SKYLINE_PATH)/Model/Prosit/Config/PrositConfig_production.xml" ;
         if ! [ path.exists $(SKYLINE_PROSIT_CONFIG) ]
         {
             errors.user-error "PrositConfig_production.xml must be present for an official build" ;
         }
     }
-    else
+    else if ! [ path.exists $(SKYLINE_PROSIT_CONFIG) ]
     {
         constant SKYLINE_PROSIT_CONFIG : "$(SKYLINE_PATH)/Model/Prosit/Config/PrositConfig_development.xml" ;
     }
--- a/pwiz_tools/Skyline/Skyline.csproj
+++ b/pwiz_tools/Skyline/Skyline.csproj
@@ -3596,7 +3596,6 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
     <None Include="Model\Prosit\Config\PrositConfig_development.xml" />
-    <None Include="Model\Prosit\Config\UpdatePrositConfiguration.bat" />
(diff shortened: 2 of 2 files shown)


brendanx67 (approved):
Okay. I was thinking of a 3rd name like PrositConfig_override.xml, but I guess this works. It does make it so that dev builds in my release projects will always use PrositConfig_production.xml regardless of whether I have --official. I guess I am okay with that until I come up with a reason it would cause me pain. Nothing comes to mind.

nickshulman:
I am not sure, but I don't think the Jamfile.jam change is going to do what you want it to in the case that the "PrositConfig_production.xml" file is not there.

What you have there is supposed to change the value of "SKYLINE_PROSIT_CONFIG" if the file could not be found, but if I recall correctly, that's not how constants work in bjam.

Once a constant has been assigned a value, you cannot change it.
So, "SKYLINE_PROSIT_CONFIG" always ends up being set to "PrositConfig_production.xml", and the 
```
if ! [ path.exists $(SKYLINE_PROSIT_CONFIG) ]
    {
        constant SKYLINE_PROSIT_CONFIG : "$(SKYLINE_PATH)/Model/Prosit/Config/PrositConfig_development.xml" ;
    }
```
does not actually do anything.

brendanx67:
Maybe we could just use a different file name like PrositConfig_override.xml

pwiz_tools/Skyline/Jamfile.jam
```diff
@@ -343,15 +343,15 @@ if [ modules.peek : NT ] && --i-agree-to-the-vendor-licenses in [ modules.peek :
         ) > "$(<)"
     }
 
+    constant SKYLINE_PROSIT_CONFIG : "$(SKYLINE_PATH)/Model/Prosit/Config/PrositConfig_production.xml" ;
```
nickshulman, line 346: I might be misremembering how constants work in .jam files, but I kind of remember that you are only allowed to assign a constant a value once.

If you find that this build fails in "PrositConfig_production.xml" is not there, then you might need to change this to:
```
    constant SKYLINE_PROSIT_CONFIG_PRODUCTION : "$(SKYLINE_PATH)/Model/Prosit/Config/PrositConfig_production.xml" ;
    if [ path.exists $(SKYLINE_PROSIT_CONFIG_PRODUCTION) ] 
    {
        constant SKYLINE_PROSIT_CONFIG : $(SKYLINE_PROSIT_CONFIG_PRODUCTION) ;
    }
    else if --official in [ modules.peek : ARGV ]
    {
        errors.user-error "PrositConfig_production.xml must be present for an official build" ;
    }
    else
    {
        constant SKYLINE_PROSIT_CONFIG : "$(SKYLINE_PATH)/Model/Prosit/Config/PrositConfig_development.xml" ;
    }
```

But, if the build works correctly when "PrositConfig_production.xml" is not there, then it means I was wrong.
(I have tried to test this on my own computer and am just getting myself more confused)

nickshulman (approved):
Looks good to me. As far as I can tell, the thing that I was worried about is not true.

pwiz_tools/Skyline/Jamfile.jam
chambm: I didn't remember myself whether constant works that way but it was easy enough to test:
```
constant foo : "Constants are really constant" ;
constant foo : "Constants can be changed" ;
echo $(foo) ;
```

Echoes "Constants can be changed".

chambm (approved):

PR #2795, opened 2023-11-29
Add option to add annotations through the command line (Issue 973)

henrytsanford:
[As requested by the MacCoss lab,](https://skyline.ms/issues/home/issues/details.view?issueId=973) this change adds the option to add annotations through the command line. It adds an argument group "Document settings".
document_settings
This change also adds two arguments “annotation-file” and “annotation-conflict-resolution” which are functional and tested, but not documented as we do not anticipate demand from users.


pwiz_tools/Skyline/TestData/CommandLineTest.cs
```diff
@@ -853,6 +853,203 @@ private static void CheckRefSpectraAll(IList<DbRefSpectra> refSpectra)
             Assert.IsTrue(!refSpectra.Any());
         }
 
+        [TestMethod]
+
+        public void ConsoleAddAnnotationsTest()
```
nickshulman, line 858: I would recommend moving this test into a separate class (e.g. "ConsoleAnnotationsTest" which inherits from "AbstractUnitTestEx") because CommandLineTest has gotten rather large.

Also, if there are steps in this test that do not depend on other steps in this test, then it would be better if they were separate [TestMethod] functions such as "ConsoleAddInvalidAnnotationTargetListTest" so that there is less code to step through when you are trying to debug a test failure.

pwiz_tools/Skyline/CommandArgs.cs
```diff
@@ -1088,6 +1099,105 @@ private bool ValidateChromatogramArgs()
...
+        /// <summary>
+        /// Parse a comma-separated list of target types to apply an annotation to. Case-insensitive.
+        /// </summary>
+        /// <param name="p">A NameValuePair where Match is the argument and Value is a
+        /// comma-separated list</param>
+        /// <exception cref="ValueInvalidAnnotationTargetListException">Thrown if no targets are
+        /// parsed from the list</exception>
+        private void ParseAnnotationTargets(NameValuePair p)
+        {
+            var userTargets =
```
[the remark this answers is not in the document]
henrytsanford: To handle invalid lists from the user, would you recommend wrapping AnnotationTargetSet.ParseElements in a Try Catch statement? Or would it be better to build error handling into my ParseElements method?

pwiz_tools/Skyline/TestData/ConsoleAddAnnotationsTest.cs
```diff
@@ -0,0 +1,248 @@
...
+                def.AnnotationTargets.Equals(annotationTargets) &&
+                def.Type.Equals(annotationType) &&
+                def.Items.SequenceEqual(annotationValues));
+
+            Assert.IsTrue(annotationInSettings);
+        }
+
+        private static void AssertDocumentAnnotationCount(string documentPath, int annotationCount)
+        {
+            Assert.IsTrue(ResultsUtil.DeserializeDocument(documentPath).Settings.DataSettings.AnnotationDefs.Count == annotationCount);
```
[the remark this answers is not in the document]
henrytsanford: Good point, thanks. This function is used in ConsoleAddAnnotationsFromArgumentsTest and ConsoleAddXmlDefinedAnnotationTest

pwiz_tools/Skyline/TestData/ConsoleAddAnnotationsTest.cs
```diff
@@ -0,0 +1,256 @@
...
+            var output = RunCommand("--new=" + newDocumentPath, // Create a document (without annotations)
+                "--overwrite", // Overwrite, as the file may already exist in the bin
+                "--annotation-file=" + annotationsXml, // Specify file path
+                "save"
+            );
+            CommandLineTest.CheckRunCommandOutputContains(string.Format(
+                    Resources.CommandLine_AddAnnotations_Annotations_successfully_defined_from_file__0__, annotationsXml)
+                , output);
+            // Assert that the document has the correct number of annotations
+            AssertDocumentAnnotationCount(newDocumentPath, 2);
```
nickshulman, line 190: I would recommend that you get rid of the methods "AssertDocumentAnnotationCount" and "AssertAnnotationInDocumentAndSettings" and replace them with the following:
```
            var doc = ResultsUtil.DeserializeDocument(newDocumentPath);
            Assert.AreEqual(2, doc.Settings.DataSettings.AnnotationDefs.Count);
            var peptideQualityAnnotation = new AnnotationDef("Peptide Quality",
                AnnotationDef.AnnotationTargetSet.Singleton(AnnotationDef.AnnotationTarget.peptide),
                AnnotationDef.AnnotationType.value_list, annotationValuesArray);
            CollectionAssert.Contains(doc.Settings.DataSettings.AnnotationDefs.ToList(), peptideQualityAnnotation);
            CollectionAssert.Contains(Settings.Default.AnnotationDefList, peptideQualityAnnotation);

            var bioReplicateAnnotation = new AnnotationDef("BioReplicate",
                AnnotationDef.AnnotationTargetSet.Singleton(AnnotationDef.AnnotationTarget.replicate),
                AnnotationDef.AnnotationType.number, null);
            CollectionAssert.Contains(doc.Settings.DataSettings.AnnotationDefs.ToList(), bioReplicateAnnotation);
            CollectionAssert.Contains(Settings.Default.AnnotationDefList, bioReplicateAnnotation);
```
The easiest way to check whether an annotation is in a list is to create an AnnotationDef with the values you are looking for, and then use CollectionAssert.Contains.

When you are able to do that, "AssertAnnotationInDocumentAndSettings" ends up being a simple enough method that it's probably not worth having it as a method. It's simpler to just deserialize the document once, and do the asserts that you want on it.

Also, once you have the document in a local variable, you might think of other things you might want to assert about it.

henrytsanford: Thanks, that is simpler

PR #2798, opened 2023-11-30
Skyline: Fix Grpc.Core.Channel leaking during tests due to …

brendanx67:
…  excessive Prosit initialization
- Improve ToolsOptionsUI to only initialize and query Prosit when the Prosit tab is activated

### DIFF
--- a/pwiz_tools/Skyline/Model/Prosit/Communication/PrositPredictionClient.cs
+++ b/pwiz_tools/Skyline/Model/Prosit/Communication/PrositPredictionClient.cs
@@ -79,18 +79,10 @@ namespace pwiz.Skyline.Model.Prosit.Communication
         // a set of cached predictions
         public static PrositPredictionClient FakeClient { get; set; }
 
-        void IDisposable.Dispose()
+        public virtual void Dispose()
         {
-            Dispose(true);
-        }
-
-        protected virtual void Dispose(bool disposing)
-        {
-            if (disposing)
-            {
-                _channel?.ShutdownAsync().Wait();
-                _channel = null;
-            }
+            _channel?.ShutdownAsync().Wait();
+            _channel = null;
         }
     }
 }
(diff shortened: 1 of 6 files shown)


nickshulman:
Usually with a leak like this we would try to figure out how to dispose of the Channel that is being created by the PrositPredictionClient, instead of making it so that only one client gets created for the entire test run.

I see that there is a method "GrpcEnvironment.ShutdownChannelsAsync". If we called that, would that fix this leak?

brendanx67:
You are welcome to try that. I found some other posts online about this
leak where channels get put into a static dictionary, and decided to just
avoid creating all the channels which didn't feel like we were testing
anything useful or likely to cause a serious leak in the real world.

The PR is here and I will accept possible improvements. I do prefer not
initializing Prosit every time the ToolsOptionsUI is shown.

brendanx67:
That does seem to work pretty well. I guess the question then is where to put that call. In the SkylineWindow shutdown code?

nickshulman:
> That does seem to work pretty well. I guess the question then is where to put that call. In the SkylineWindow shutdown code?

I think the only channel that we need to shut down is the one in the FakePrositPredictionClient.
When a user is using regular Skyline, only one Channel ever gets created, and it's the one used by the static "_predictionClient" in "PrositPredictionClient". I believe that object only gets created once for the entire test run, so we do not need to shut it down between tests.

If we ever allowed the user to change which server Skyline was talking to, then we would want to dispose that thing.

nickshulman (approved):

brendanx67:
That is not the case. I added the FakePrositPredictionClient in this case.
Before that, it was relying on the standard path, but an exception got
thrown before _predictionClient could get set. It seems that only gets
stored for a successful connection.

brendanx67:
Huh. Seems like jumping through a lot of hoops to prove you know how to dispose a Channel correctly inside of test code. I find the TODO about not actually ever disposing the Channel created in production code sort of humorous. Why not just let all test code use a single channel that never gets disposed of, as I had originally? Sigh.

PR #2799, opened 2023-12-01
Treat wiff1 and wiff2 as separate types

chambm:
- changed Sciex readers to treat WIFF1 and WIFF2 as separate types (e.g. for purposes of file open dialogs)


brendanx67 (approved):
Looks good. Thanks.

PR #2801, opened 2023-12-03
Add --verbose-errors argument

henrytsanford:
Add --verbose-errors argument under the "General input/output" group.
Adding this argument reports exception.ToString() instead of exception.Message, providing more information for troubleshooting.
verbose_errors


brendanx67 (changes requested):
Looks like a good start.

pwiz_tools/Skyline/CommandArgs.cs
```diff
@@ -199,10 +199,12 @@ private static void SetCulture(string cultureName)
         public const string ARG_VALUE_ASCII = "ascii";
         public const string ARG_VALUE_NO_BORDERS = "no-borders";
         public static readonly Argument ARG_VERSION = new Argument(@"version", (c, p) => c.Version());
+        public static readonly Argument ARG_VERBOSE_EXCEPTIONS =
+            new Argument(@"verbose-exceptions", (c, p) => c._out.IsVerboseExceptions = true);
 
         private static readonly ArgumentGroup GROUP_GENERAL_IO = new ArgumentGroup(() => CommandArgUsage.CommandArgs_GROUP_GENERAL_IO_General_input_output, true,
             ARG_IN, ARG_SAVE, ARG_SAVE_SETTINGS, ARG_OUT, ARG_NEW, ARG_OVERWRITE, ARG_SHARE_ZIP, ARG_SHARE_TYPE, ARG_BATCH, ARG_DIR, ARG_TIMESTAMP, ARG_MEMSTAMP,
-            ARG_LOG_FILE, ARG_HELP, ARG_VERSION)
+            ARG_LOG_FILE, ARG_HELP, ARG_VERSION, ARG_VERBOSE_EXCEPTIONS)
```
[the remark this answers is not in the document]
henrytsanford: I see your point. I called it verbose-exceptions because I didn't want users to think it would modify run-of-the-mill errors, but it's probably better to just not use "exception" at all.

pwiz_tools/Skyline/CommandArgUsage.resx
```diff
@@ -991,4 +991,7 @@ greater peak area.</value>
 "Species" excludes any peptide that appears in proteins for more than one species in the background proteome.
 Useful in sample mixtures including multiple species.</value>
   </data>
+  <data name="_verbose_errors" xml:space="preserve">
+    <value>Provide additional information to report to Skyline developers in the case of an error.</value>
```
brendanx67, line 995: "Adds additional information to the program output in the case of an error, that may be helpful to Skyline developers."

pwiz_tools/Skyline/CommandLine.cs
```diff
@@ -186,7 +181,7 @@ private int RunInner(string[] args, bool withoutUsage = false)
             }
             if (commandArgs.ImportingSkyr)
             {
-                if (!ImportSkyr(commandArgs.SkyrPath, commandArgs.ResolveSkyrConflictsBySkipping))
+                if (!ImportSkyr(commandArgs.SkyrPath, commandArgs.ResolveSkyrConflictsBySkipping, commandArgs))
```
brendanx67, line 184: If we are passing commandArgs, maybe we don't need the other parameters.

pwiz_tools/Skyline/CommandLine.cs
```diff
@@ -626,24 +602,11 @@ private bool RefineDocument(CommandArgs commandArgs)
...
-                {
-                    _out.WriteLine(Resources.Error___0_, x.Message);
-                }
-                else
-                {
-                    _out.WriteLine(x.Message);
-                }
-                return false;
-            }
+            }, Resources.Error___0_, true);   // TODO: Not really standard to just report the exception alone
```
brendanx67, line 609: If you don't plan on doing anything about this, change it to CONSIDER:

pwiz_tools/Skyline/CommandLine.cs
```diff
@@ -664,20 +628,7 @@ private bool CreateImsDb(CommandArgs commandArgs)
...
```
brendanx67, line 631: If you don't plan on doing anything about this, change it to CONSIDER:

pwiz_tools/Skyline/CommandLine.cs
```diff
@@ -906,7 +849,7 @@ private bool PerformExportOperations(CommandArgs commandArgs)
                     sharedFileName = FileEx.GetTimeStampedFileName(_skylineFile);
                 }
                 var sharedFilePath = Path.Combine(sharedFileDir, sharedFileName);
-                if (!ShareDocument(_doc, _skylineFile, sharedFilePath, commandArgs.SharedFileType, _out))
+                if (!ShareDocument(_doc, _skylineFile, sharedFilePath, commandArgs.SharedFileType, _out, commandArgs))
```
brendanx67, line 852: Maybe drop commandArgs.SharedFileType as you have done elsewhere when adding commandArgs to the function arguments.

pwiz_tools/Skyline/TestData/ConsoleVerboseExceptionsTest.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System.IO;
```
brendanx67, line 1: Need to add the licensing header comment.

pwiz_tools/Skyline/CommandArgs.cs
```diff
@@ -131,13 +131,16 @@ private static void SetCulture(string cultureName)
             (c, p) => c.HideAllChromatogramsGraph = true) {InternalUse = true};
         public static readonly Argument ARG_TEST_NOACG = new Argument(@"noacg",
             (c, p) => c.NoAllChromatogramsGraph = true) {InternalUse = true};
+        public static readonly Argument ARG_TEST_EXCEPTION = new Argument(@"exception",
```
brendanx67, line 134: I don't see this getting used in a test. I think it should be. I guess you were just using it in your own manual testing?

brendanx67 (changes requested):
Mostly looks pretty good. I had a few suggestions. Too much change for me to verify that it is all totally correct. Fingers crossed that our tests do a good enough job.

pwiz_tools/Skyline/CommandLine.cs
henrytsanford: This is one case where I have to leave that argument because in when publishing to Panorama it can be called with a different share type than the one in CommandArgs, depending on PanoramaPublishUtil.DecideShareTypeVersion.

pwiz_tools/Skyline/CommandArgs.cs
henrytsanford: Good point. I was trying to write my test to provoke an exception without using this argument, but it is more readable to use this argument in the unit test.

brendanx67 (approved):

PR #2805, opened 2023-12-06
Figure out why StackOverflowException on TeamCity is not a failure

nickshulman:
Trying to track down why TeamCity does not notice that a StackOverflowException which causes TestRunner.exe to abruptly exit with exit code 2147943401 should be treated as a test failure.

### DIFF
--- a/pwiz_tools/Skyline/TestFunctional/StackOverflowExceptionTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/StackOverflowExceptionTest.cs
@@ -1,0 +1,31 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using pwiz.SkylineTestUtil;
+
+namespace pwiz.SkylineTestFunctional
+{
+    /// <summary>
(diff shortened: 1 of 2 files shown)


chambm:
I'm pretty sure this has to be either test runner or boost build. So it
should be reproducible locally but I haven't tried yet. Did you?

chambm:
It just hung indefinitely. When I run it from Boost.Build, your version of the stack overflow test hangs on my machine locally as well. When I try to attach and break to see what's happening, I get:
image
and it seems to exits with code 0. When I start it from the Visual Studio debugger, I get the stack overflow immediately and it exists non-zero.

I made my own version before you made this PR. It didn't go through DoTest, the [TestMethod] just called itself directly. Running that via Boost.Build did exit with stack overflow exception and non-zero exit code. But your version is closer to what the functional test actually did.

Perhaps the easiest way forward would be to set a TC failure condition on the string `terminated due to StackOverflowException` and also end the build immediately to avoid the hang.

nickshulman:
Looks like TeamCity now correctly detects the stack overflow message in the build log.

PR #2806, opened 2023-12-07
BiblioSpec fixes for DIA-NN run by FragPipe

chambm:
BiblioSpec:
- fixed recognizing a DIA-NN run by FragPipe: it now checks for either "library.tsv.speclib" or "lib.predicted.speclib" and the report filename in the diann-output subdirectory can be "diann-output.tsv" or "report.tsv"
- fixed N-terminus mods from DIA-NN to be moved/merged to the N-terminal AA


pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
```diff
@@ -191,7 +191,11 @@ inline std::string get_aas(const std::string &name, vector<SeqMod>& mods)
             i += modPrefixLength;
             mod = name.substr(i, end-i);
             CVID unimodCvid = (CVID) (UNIMOD_unimod_root_node + lexical_cast<int>(mod));
-            mods.emplace_back(SeqMod((int) j, pwiz::data::unimod::modification(unimodCvid).deltaMonoisotopicMass()));
+            int position = max(1, static_cast<int>(j));
+            if (!mods.empty() && mods.back().position == position)
+                mods.back().deltaMass += unimod::modification(unimodCvid).deltaMonoisotopicMass();
+            else
+				mods.emplace_back(position, unimod::modification(unimodCvid).deltaMonoisotopicMass());
```
brendanx67, line 198: Strange indenting. Tabs instead of spaces?

pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
```diff
@@ -601,10 +605,18 @@ bool DiaNNSpecLibReader::parseFile()
     auto diannReportFilepath = bal::replace_last_copy(specLibFile, "-lib.tsv.speclib", "-report.tsv");
 
     // special case for FragPipe
-    if (specLibFilePath.filename().string() == "library.tsv.speclib" && bfs::exists(specLibFilePath.parent_path() / "diann-output" / "diann-output.tsv"))
+    if (specLibFilePath.filename().string() == "library.tsv.speclib" || specLibFilePath.filename().string() == "lib.predicted.speclib")
```
brendanx67, line 608: I wonder if we should simply look for a "diann-output" subdirectory containing either report.tsv or diann-output.tsv. Having these explicit .speclib file names seems less future-proof. Can you think of cases where a .speclib file ends up in a Fragpipe folder with a diann-output subfolder where we would not want to use the contents of that subfolder?

brendanx67 (approved):
A couple of minor considerations. The code and test look they cover the current important case, though.

pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
chambm: It's not a stretch to imagine that a DIA-NN GUI user chooses "diann-output" as their output folder, and report.tsv would be the default report name in that case. The default spectral library name is report-lib.tsv. If I change the default report name to reportfoo, then the default spectral library name becomes reportfoo-lib.tsv. On the other hand, it might be stretching that a GUI user picked the diann-output subdirectory for the report.tsv and the parent directory for the speclib. Unless they plan to re-use that speclib file for multiple reports... I don't have experience using DIA-NN that way, but IIRC you said that's pretty typical?

So I thought "library.tsv.speclib" and  "lib.predicted.speclib" were sufficiently specific to (and hardcoded in) FragPipe because it's not anything like the default. Possibly it would be better to look for the FragPipe log or param file or something to be really sure that it's a FragPipe run, and then skip the report filename check and just do the diann-output check.

brendanx67:
I tested the artifacts and the fixes work for me. Please merge.

PR #2807, opened 2023-12-07
Add relative abundance plot

henrytsanford:
Add a plot showing relative abundance of targets with the options:
- Show proteins or peptides
- Show replicates by all, single, or best
- Exclude peptide lists
- Exclude standards
- Show CV
- Log scale
- Add formatting (with the same formatting options as the volcano plot)
DevelopedRelativeAbundance

Add the option to create a match expression using a list of identifiers, instead of a regex.

Add semi-opaque backgrounds to labels on the relative abundance plot and the volcano plot.
SemiOpaque


pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -1713,6 +1717,16 @@ public void ShowProductTransitions()
             SetDisplayTypeChrom(DisplayTypeChrom.products);
         }
 
+        public void ShowRelativeAbundanceFormatting()
+        {
+            foreach (var summary in _listGraphPeakArea.Where(summary => summary.Type == GraphTypeSummary.abundance))
```
brendanx67, line 1722: Can there possibly ever be more than one of these? Maybe just find it using FirstOrDefault()

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -4393,6 +4461,22 @@ public void ShowPeakAreaPeptideGraph()
...
+            ShowGraphPeakArea(true, GraphTypeSummary.abundance);
+            foreach (var summary in _listGraphPeakArea)
+            {
+                if (summary.Type == GraphTypeSummary.abundance)
+                {
+                    summary.Window = this;
+                }
+            }
+            UpdatePeakAreaGraph();
+            SynchronizeSummaryZooming();
```
brendanx67, line 4477: I can't see how this would apply to this graph. I don't think you could synchronize this with any other graph.

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -4541,6 +4625,42 @@ private void peptideCvsContextMenuItem_Click(object sender, EventArgs e)
...
+        }
+
+        private void abundanceTargetsPeptidesMenuItem_Click(object sender, EventArgs e)
+        {
+            abundanceTargetsProteinsMenuItem.Checked = false;
+            abundanceTargetsPeptidesMenuItem.Checked = true;
+            setAreaProteinTargets(false);
+        }
+
+        public void setAreaProteinTargets(bool areaProteinTargets)
```
brendanx67, line 4642: Use title casing, not camel casing. Camel case is for event handlers only.

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -4541,6 +4625,42 @@ private void peptideCvsContextMenuItem_Click(object sender, EventArgs e)
...
+        {
+            Settings.Default.AreaProteinTargets = areaProteinTargets;
+            UpdateSummaryGraphs();
+        }
+        private void excludeTargetsPeptideListMenuItem_Click(object sender, EventArgs e)
+        {
+            setExcludePeptideListsFromAbundanceGraph(!Settings.Default.ExcludePeptideListsFromAbundanceGraph);
+        }
+
+        public void setExcludePeptideListsFromAbundanceGraph(bool excludePeptideLists)
```
brendanx67, line 4652: Use title case

pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -4557,6 +4677,7 @@ private void areaPropsContextMenuItem_Click(object sender, EventArgs e)
         {
             switch (ContextMenuGraphSummary.Type)
             {
+                case GraphTypeSummary.abundance:
```
brendanx67, line 4680: This seems a bit strange that the abundance graph should share a properties form with the comparison forms.

pwiz_tools/Skyline/TestFunctional/PeakAreaRelativeAbundanceGraphTest.cs
```diff
@@ -0,0 +1,68 @@
...
+                SkylineWindow.setAreaProteinTargets(false);
+                curveList = peakAreaGraph.GraphControl.GraphPane.CurveList[1];
+                Assert.AreEqual(curveList.Points.Count, 125);
+
+                // Verify that excluding peptide lists reduces the number of points
+                SkylineWindow.setExcludePeptideListsFromAbundanceGraph(true);
+                curveList = peakAreaGraph.GraphControl.GraphPane.CurveList[1];
+                Assert.AreEqual(curveList.Points.Count, 45);
+
+                //CONSIDER add quantitative checks for relative abundance results
```
brendanx67, line 63: Feels like you should also be adding testing of the formatting forms in this context.

pwiz_tools/Skyline/TestFunctional/VolcanoPlotFormattingTest.cs
```diff
@@ -414,6 +416,27 @@ private void AssertVolcanoPlotCorrect(FoldChangeVolcanoPlot plot, VolcanoPlotPoi
...
+                    exprInfo.ExpectedPointsInfo.Color, exprInfo.ExpectedPointsInfo.PointSymbol,
+                    exprInfo.ExpectedPointsInfo.PointSize));
+                formattingDlg.ClickCreateExpression(bindingList.Count - 1);
+            });
+            var matchExprListDlg = ShowDialog<MatchExpressionListDlg>(createExprDlg.ClickEnterList);
+            RunUI(() =>
+            {
+                // Verify that typing into the list is parsed to a REGEX
+                matchExprListDlg.proteinsTextBox.Text = @"Cas9" + '\n' + @"Hsp70";
+                Assert.AreEqual(createExprDlg.Expression, "^Cas9$|^Hsp70$");
```
brendanx67, line 437: Ideally, you would have some edge/error case testing, like when you add things that don't match, or you add empty lines, etc. I can imagine wanting the matches to be case-insensitive also. Looks like you have only case-sensitive matching?

brendanx67 (changes requested):
Looking good. Still some work to be done on this.

rita-gwen:
It looks like there are still some fixes needed before this can be merged. Properties context menu does nothing. Selection context menu item seems to do nothing. If some formatting criteria are specified it probably should show a legend.

brendanx67:
On the last one, we are just trying to be comparable to the volcano plot.
Let's get this to the level of that plot and make a first checkpoint merge.
Then we can return to ways that both plots can be improved. I would
probably go for the label placement work over the legend in a list of
priorities for both plots.

PR #2808, opened 2023-12-07
Agilent MH12.1 transition export

rita-gwen:


pwiz_tools/Skyline/Model/Export.cs
```diff
@@ -2986,197 +3017,284 @@ private static string GetModuleName(Process proc)
...
+            public PeptideDocNode nodePep;
+            public TransitionGroupDocNode nodeTranGroup;
+            public TransitionGroupDocNode nodeTranGroupPrimary;
+            public TransitionDocNode nodeTran;
+            public int step;
         }
 
-        public AgilentMassListExporter(SrmDocument document, DocNode node)
-            : base(document, node)
+        public class TransitionField
```
nickshulman, line 3035: Instead of defining a list of functions that take a TransitionContext and return a value, and then constructing TransitionField's from those function, I would recommend making it so that TransitionContext has public Properties on it for each of these fields, and then use reflection to list those properties.

Let me know if you'd like me to write some example code for this.

PR #2809, opened 2023-12-08
BiblioSpec pepXML check base_name parent path 

chambm:
- fixed BiblioSpec pepXML reader to check for spectrum files in the base_name's parent path (if present)


chambm:
Ugh. Must fix it to parse Windows paths on Linux and Linux paths on Windows.

brendanx67:
> Ugh. Must fix it to parse Windows paths on Linux and Linux paths on Windows.

Huh. I expect the C# Path class to do this in .NET. I guess you are saying the bsf::path does not support this automatically.

brendanx67:
Works for me. Waiting for tests to pass and then I will merge.

PR #2810, opened 2023-12-09
Skyline/work/20231026 reporter ion improvements

bspratt:
When determining the "best" peak for determining retention time, some fragment types should not be considered - for example, reporter ions like TMT. 

In this example, the relatively abundant TMT signal pulls Skyline's RT calculation away from the actual peptide fragments peak:
image

but with this change only the peptide fragments are used for RT determination
image


nickshulman:
I don't think classes like "IFoundPeak", "IPeakFinder", "ChromKey", or "ChromTransition" should know anything about the "ParticipatesInScoring" flag.
That is, the only things that need to know that a particular transition is non-scoring are things that already have the TransitionDocNode, such as "ChromData"

bspratt:
Thanks for looking at this. I'll have a look at the impact of your suggestion.

bspratt:
@nickshulman That reduces the changes considerably, thanks. This makes it much easier to see the uses of "area" vs "scorable area", I'd appreciate it if you could inspect those.

nickshulman (commented):
Yes, looks much better.

I wonder whether we need to give the user some way of telling Skyline whether a reporter ion should be scoring (participate in library dot product) or not.
When we first conceived of reporter ions, they were supposed to be things that easily fall off of the peptide which can be used to identify the peptide. As such, they probably should participate in the library dot product.

The reason that we don't want TMT reporter ions to participate in the dot product is not because they are reporter ions, per se, but that they typically come from samples in different vials which have been pooled together.

I know we are planning on making Group Comparison work with TMT, which means that users are probably going to need to tell us how the TMT ions relate to sample cohorts, etc.
So, maybe the way to determine that an ion is non-scoring is that it has some cohort information.

I don't think TransitionGroupChromInfo needs to have separate "Area" and "AreaScoring". If we didn't need to worry about backward compatibility, there would only be "Area" and it should exclude non-scoring peak areas.

bspratt:
You raise good questions - @brendanx67 was hoping to avoid UI complexity, initially at least.

>> I don't think TransitionGroupChromInfo needs to have separate "Area" and "AreaScoring". If we didn't need to worry about backward compatibility, there would only be "Area" and it should exclude non-scoring peak areas.

I was nervous about changing the meaning of Area.

PR #2811, opened 2023-12-10
small molecule chromatogram extraction wasn't using spectral library retention times …

bspratt:
… (didn't initially have that capability, and the limitation was never lifted)


brendanx67 (approved):
Looks simple enough. Surprised the Baker lab or Chris Ashwood did not report this while using libraries and iRTs. I guess if they got correct peaks picked they didn't worry about having full gradient chromatograms?

PR #2822, opened 2024-01-06
Speed up "BiblioSpecLiteLibrary.FindSource" by caching results in a Dictionary

nickshulman:
@chambm noticed that, when doing "Associate Proteins" on a large document with many replicates, most of the time is spent in BiblioSpecLiteLibrary.FindSource.

This fix reduces the time spent in that function to nearly zero.
We might want to encapsulate this "list along with a synchronized dictionary of things we've looked for before" into a single class that is used in multiple places.

We currently have three different types of libraries which can contain multiple files: BiblioSpecLiteLibrary, ChromatogramLibrary and EncyclopeDiaLibrary. EncyclopeDiaLibrary has slightly different logic for deciding which file is the best match for a given file path, but their logic is all pretty similar.
I am hoping that if we solve this problem then we can get rid of the "ref int? fileIndex" parameters in methods like "Library.GetRetentionTimesWithSequences".

In theory, it also should be possible to avoid calling PeptideDocNode.ChangeSettings on every peptide in the document. Unfortunately, "SrmDocument.UpdateResultsSummaries" really wants to call that for every peptide that cannot be found in the original tree.

I am hoping this pull request inspires creative ideas.

### DIFF
--- a/pwiz_tools/Skyline/Model/Lib/BiblioSpecLite.cs
+++ b/pwiz_tools/Skyline/Model/Lib/BiblioSpecLite.cs
@@ -142,6 +142,7 @@ namespace pwiz.Skyline.Model.Lib
 
         private BiblioLiteSourceInfo[] _librarySourceFiles;
         private bool _anyExplicitPeakBounds;
+        private Dictionary<MsDataFileUri, int> msDataFileUriLookup;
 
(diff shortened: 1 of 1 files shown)


nickshulman:
These changes got incorporated into PR #2764

PR #2825, opened 2024-01-09
Fix for NRE in the Detections plot

rita-gwen:


pwiz_tools/Skyline/Controls/Graphs/DetectionsPlotPane.cs
```diff
@@ -46,12 +46,12 @@ public abstract class DetectionsPlotPane : SummaryReplicateGraphPane
 
         protected DetectionsPlotPane(GraphSummary graphSummary) : base(graphSummary)
         {
-            MaxRepCount = graphSummary.DocumentUIContainer.DocumentUI.MeasuredResults.Chromatograms.Count;
+            MaxRepCount = graphSummary.DocumentUIContainer.DocumentUI?.MeasuredResults?.Chromatograms?.Count ?? 0;
```
nickshulman, line 49: I think the only "?" that is necessary is the one between "DocumentUI" and "MeasuredResults".
The other "?" characters should probably be removed.

nickshulman (approved):
Looks good, but I'd recommend removing the unnecessary "?" characters since they imply that some of these things that can never be null might be null.
But, yes, all code needs to realize "MeasuredResults" property on a "SrmDocument" might be null.

pwiz_tools/Skyline/Controls/Graphs/DetectionsPlotPane.cs
rita-gwen: Sure. I wasn't sure if ?. is short-circuiting.

PR #2828, opened 2024-01-11
Speed up display of protein tooltips when lots of losses

nickshulman:
Sped up display of protein tooltips (and Library Match window) when Max Neutral Losses is high (reported by Juan)


brendanx67 (changes requested):
Seems like a good case for a quick unit test.

PR #2832, opened 2024-01-17
Fix ReSharper warnings related to Dispose'ing enumerators

nickshulman:
The newest version of ReSharper complains if we don't call Dispose on our enumerators.
Most of the cases where we were not doing this were straightforward to fix, such as using "foreach" or LINQ.

The methods "ChromDataSet.MergePeaks" and "PeptideChromDataSets.MergePeakGroups" were both doing more complicated stuff with their enumerators so I created a new class "DisposableCollection" which makes sure that Dispose gets called on each of the items that get put into it.
It's not clear to me that "DisposableCollection" was a good idea. Maybe I should have just disabled the warning with:
`// ReSharper disable once NotDisposedResourceIsReturned `
The enumerators returned from ordinary collections such as "List" or "Array" definitely do not need to have Dispose called on them, because they never have to do any cleanup.


pwiz_tools/Shared/Lib/JetBrains.Annotations.dll
bspratt, line 1: Does this change every time there's a major ReSharper version change?

bspratt:
>> It's not clear to me that "DisposableCollection" was a good idea.

If there's no performance hit, it's nice that this indicates that we really thought about as opposed to just blowing off the warning, even if that was a legitimate choice in the immediate case.

pwiz_tools/Shared/Lib/JetBrains.Annotations.dll
nickshulman: "JetBrains.Annotations.dll" does not usually change very much with new versions of ReSharper. I needed to get a new one this time because I wanted to use the new "MustDisposeResourceAnnotation" class.
It's not really a problem if we have an old version of this dll-- it just means we cannot annotate our code as richly as we might want.
Note that JetBrains.Annotations.dll is only used at compile time. It is not included in the set of dll's that ship with Skyline.

bspratt:
Doesn't build for me:

C:\Dev\BuildTrouble\pwiz_tools\Skyline\Model\Results\PeptideChromData.cs(764,44): error CS0246: The type or namespace name 'DisposingCollection<>' could not be found (are you missing a using directive or an assembly reference?) [C:\Dev\BuildTrouble\pwiz_tools\Skyline\Skyline.csproj]
C:\Dev\BuildTrouble\pwiz_tools\Skyline\Model\Results\PeptideChromData.cs(772,26): error CS0019: Operator '-' cannot be applied to operands of type 'method group' and 'int' [C:\Dev\BuildTrouble\pwiz_tools\Skyline\Skyline.csproj]
C:\Dev\BuildTrouble\pwiz_tools\Skyline\Model\Results\PeptideChromData.cs(774,21): error CS0023: Operator '!' cannot be applied to operand of type 'Task<bool>' [C:\Dev\BuildTrouble\pwiz_tools\Skyline\Skyline.csproj]
C:\Dev\BuildTrouble\pwiz_tools\Skyline\Model\Results\PeptideChromData.cs(781,20): error CS0019: Operator '>' cannot be applied to operands of type 'method group' and 'int' [C:\Dev\BuildTrouble\pwiz_tools\Skyline\Skyline.csproj]
C:\Dev\BuildTrouble\pwiz_tools\Skyline\Model\Results\PeptideChromData.cs(788,33): error CS0019: Operator '<' cannot be applied to operands of type 'int' and 'method group' [C:\Dev\BuildTrouble\pwiz_tools\Skyline\Skyline.csproj]
C:\Dev\BuildTrouble\pwiz_tools\Skyline\Model\Results\PeptideChromData.cs(811,21): error CS0023: Operator '!' cannot be applied to operand of type 'Task<bool>' [C:\Dev\BuildTrouble\pwiz_tools\Skyline\Skyline.csproj]

bspratt:
And still a couple of warnings: 
image

bspratt:
Can confirm that this now gives a clean inspection on my local machine with the current ReSharper. Haven't actually run the tests though.

PR #2836, opened 2024-01-19
Display a message in the Immediate Window if Skyline is discarding chromatograms because of the Explicit Retention Time.

nickshulman:
Add class "ImmediateWindowWarningListener" which listens for anything that might have been logged with "Trace.TraceWarning" (or TraceError or TraceCritical), and shows the Immediate Window. Also, send these warnings to the console when running Skyline from the command line.

Also, remove or change to TraceInformation a couple of places where we were calling TraceWarning and probably didn't want it to start showing up in the Immediate Window because it was already being shown to the user as an error message.


bspratt (commented):
I worry about large data sets with repeated errors - say, one of those 1,500,00 transition Skyline docs where some goofy setting makes every chromatogram bark.  Should we have some kind of mechanism for Skyline effectively saying "I just told you about 25 of these, there are about 1.5 million more" rather than crapping up the immediate window?

PR #2839, opened 2024-01-19
Improve failure messages when "TestPanoramaClient" or "PanoramaClientDownloadTest" fails.

nickshulman:
I made it so that PanoramaFilePicker.ClickFile and PanoramaFolderBrowser.SelectNode return a boolean so that the test can assert that the item to be clicked was actually there.

This should make it easier to figure out what is going wrong if this test ever fails in the future.


vagisha (approved):
Thanks!

PR #2844, opened 2024-01-24
- fixed --tran-product-*-special-ion commands

chambm:
- fixed --tran-product-*-special-ion commands not being processed if they are the only filter arguments passed


brendanx67 (approved):

chambm:
@brendanx67 Want me to cherry pick to release as well?

brendanx67:
It feels like we have fixed enough about command-line settings since 23.1
that the answer to a support request about this is likely to be to use
Skyline-daily or the hack you gave Philip. Philip is certainly not
expecting to use 23.1

PR #2845, opened 2024-01-24
Improved error handling when attempting to parse formulas with errors in mass offsets, e.g. "C12H5H3[+3.2/3,3]":…

bspratt:
… use double.TryParse instead of double.Parse so we can throw the more generally handled ArgumentException


bspratt:
@nickshulman nothing controversial here, I think, but can I get a quick blessing for form's sake? Thanks.

nickshulman:
I would recommend moving "SplitFormulaAndMasses" into the class "ParsedMolecule" since that's the only place it gets used.
That way, it would be in the Skyline project instead of CommonUtil which would mean that you would have access to the Skyline .resx files.

It's usually easier to understand code if it throws as soon as it knows it cannot succeed. You would usually only have a variable like "success" if the algorithm needed to try a few different things before finally giving up.

The way I probably would have done this would be to change "SplitFormulaAndMasses" into a private function called "SplitFormulaAndMassesOrThrow" and have it throw whatever exception it wants, and then the public "SplitFormulaAndMasses" catches that and turns it into a pretty message that is ready to be shown to the user:
```
        public static void SplitFormulaAndMasses(string formulaAndMasses, out string formula, out double? modMassMono,
            out double? modMassAverage)
        {
            try
            {
                SplitFormulaAndMassesOrThrow(formulaAndMasses, out formula, out modMassMono, out modMassAverage);
            }
            catch (Exception ex)
            {
                throw new InvalidDataException(
                    string.Format(
                        Resources.BioMassCalc_CalculateMass_The_expression__0__is_not_a_valid_chemical_formula,
                        formulaAndMasses), ex);
            }
        }
        private static void SplitFormulaAndMassesOrThrow(string formulaAndMasses, out string formula, out double? modMassMono, out double? modMassAverage)
        {
        // Keep the current implementation with "Double.Parse" calls that will throw ArgumentException etc.
```

But your way of doing things is fine too.

bspratt:
Good call on the move refactor, thanks. Personally I like error handling code to have a single place where I can set a breakpoint to figure out why things are going sideways, so I'm going to leave that one alone - but  I appreciate the attention to detail, thanks.

pwiz_tools/Skyline/Properties/Resources.resx
```diff
@@ -4387,4 +4387,7 @@ information, there is no spectrum to show</value>
   <data name="PeptideMod_SetTerminus_A_peptide_modification_must_be_added_before_giving_it_a_terminal_or_amino_acid_specificity_" xml:space="preserve">
     <value>A peptide modification must be added before giving it a terminal or amino acid specificity.</value>
   </data>
+  <data name="ParsedMolecule_SplitFormulaAndMasses_Cannot_parse_formula___0__" xml:space="preserve">
+    <value>Cannot parse formula "{0}"</value>
```
brendanx67, line 4391: We don't enforce this as a style, but I definitely prefer not to drop articles when it it is really not necessary. This should be "Cannot parse the formula "{0}""

pwiz_tools/Skyline/Util/ParsedMolecule.cs
```diff
@@ -204,6 +206,90 @@ public static bool TryParseFormula(string formula, out ParsedMolecule resultMole
...
+        /// <param name="modMassMono">effect of any mono mass modifiers e.g. "C12H5[-1.2/1.21]-C2H[-1.1]" => 0.1 </param>
+        /// <param name="modMassAverage">effect of any avg mass modifiers e.g. "C12H5[-1.2/1.21]-C2H[-1.1]" => 0.11</param>
+        private static void SplitFormulaAndMasses(string formulaAndMasses, out string formula, out double? modMassMono, out double? modMassAverage)
+        {
+            modMassMono = null;
+            modMassAverage = null;
+            formula = formulaAndMasses?.Trim();
+            if (string.IsNullOrEmpty(formula))
+            {
+                return;
```
brendanx67, line 224: Should you normalize this to either "null" or String.Empty? As the code stands it will return either null or "".

I guess I would also mildly prefer this function to return a struct or class FormulaAndMasses object. Maybe Nick also said this.

brendanx67: Then it would not have out parameters, and return a single object that you new at the top of the function.

pwiz_tools/Skyline/Util/ParsedMolecule.cs
```diff
@@ -204,6 +206,90 @@ public static bool TryParseFormula(string formula, out ParsedMolecule resultMole
...
+            }
+            // A few different possibilities here, e.g. "C12H5", "C12H5[-1.2/1.21]", "[+1.3/1.31]", "1.3/1.31"
+            // Also possibly  "C12H5[-1.2/1.21]-C2H[-1.1]"
```
brendanx67, line 227: Really feels like this code could benefit from some sub-functions, like something that just parses the inside of [<offset>], and something that parses <formula>[<offset>]. This looks like challenging code to maintain.

pwiz_tools/Skyline/Util/ParsedMolecule.cs
```diff
@@ -204,6 +206,90 @@ public static bool TryParseFormula(string formula, out ParsedMolecule resultMole
...
+                {
+                    formula = formula.Substring(0, position) + formula.Substring(close + 1);
+                }
+            }
+            if (success && formula.Contains(@"/"))
+            {
+                // e.g.  "1.3/1.31"
+                var parts = formula.Split(new[] { '/' });
+                double modMassAverageParsed = 0;
+                success = double.TryParse(parts[0], NumberStyles.Any, CultureInfo.InvariantCulture, out var modMassMonoParsed) &&
```
brendanx67, line 278: This appears to be the only place where we set success to false. Why not just reverse the test and throw as soon as the failure is detected? This should greatly simplify the code, because you can assume otherwise that you are parsing successfully.

brendanx67 (changes requested):
Not a fan of the complexity of this function. Seems like it could be clarified with subfunctions and throwing at the point of failure.

pwiz_tools/Skyline/Util/ParsedMolecule.cs
bspratt: It's a local supporting function for something that does actually return a class, it feels strange to beat up on the heap for a thing like that.

pwiz_tools/Skyline/Util/ParsedMolecule.cs
bspratt: Agreed, done.

pwiz_tools/Skyline/Util/ParsedMolecule.cs
brendanx67: Yeah. Just a little more like recursive descent parsing, since it is allowed to be:
{expression} = {formula}[{offset}]
or
{expression}-{expression}

I may have set a bad example elsewhere when I originally added {formula}-{formula} for peptide modifications. I am not sure of that. But, the addition of the multi-format offset:

{offset} = {double}
{offset} = {double}/{double}

Reaches the point where I would really like to see it in recursive descent and not a loop.

bspratt: I encapsulated the recursion in a regular expression. It reads pretty cleanly now, I think.

pwiz_tools/Skyline/Util/ParsedMolecule.cs
```diff
@@ -204,6 +206,106 @@ public static bool TryParseFormula(string formula, out ParsedMolecule resultMole
...
+                    {
+                        formulaOut += @"-";
+                    }
+                    formulaOut += thisFormula;
+                    var brackets = match.Groups[3].Value;
+                    if (!string.IsNullOrEmpty(brackets))
+                    {
+                        if (!TryParseMassOffsetValues(brackets, sign, out var monoMassParsed, out var averageMassParsed))
+                        {
+                            throw new ArgumentException(string.Format(UtilResources.ParsedMolecule_SplitFormulaAndMasses_Cannot_parse___0___as_a_formula_and_mass_offset, formulaAndMasses));
```
brendanx67, line 262: I actually wanted something like:

string.Format(UtilResources.ParsedMolecule_SplitFormulaAndMasses_Cannot_parse___0___as_a_mass_offset, brackets)

But maybe it should provide more detail like:

string.Format(UtilResources.ParsedMolecule_SplitFormulaAndMasses_Cannot_parse___0___as_a_mass_offset_in_the_text__1__, brackets, formulaAndMasses)

The code is failing on the "brackets" text. So, it is a good idea to call that out clearly, but I can see how the extra information about the full starting text might be useful. With either of your more general errors, I worry that the use may be confused looking at the formula itself for errors, when there may be none.

bspratt: Good point.  Now it throws this:
image

PR #2846, opened 2024-01-26
Reduce memory footprint of the "TransitionChromInfo" objects

nickshulman:
This change reduces the amount of memory that the TransitionChromInfo objects are taking up in memory by sometimes representing them as "columns" of data. Many of the columns have a lot of repeated values such as "OptimizationStep" that can be represented by a smaller data structure. Also, many of the columns such as "ChromFileInfoId" are the same across all "Results" objects and can be shared between DocNode's.

Before:
image

After:
image


nickshulman:
Maybe sometime in the future

PR #2850, opened 2024-01-29
Speed up "ModEquivalentTest".

nickshulman:
Also, fix incorrect implementation of "ReferenceValue.Equals" (unrelated to anything else).


brendanx67:
How much faster does this make it?

nickshulman:
> How much faster does this make it?

20 times faster.

brendanx67:
So, 120 seconds to 6 seconds. Seems okay. Thanks.

PR #2851, opened 2024-01-30
Skyline: Add support for precursor isotopes and reporter ions to transition list import

brendanx67:
- reporter ions was requested by Thermo for TMT support
- others have requested support for the MS1 precursor isotopes supported by the Targets list


nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/Model/Import.cs
```diff
@@ -1314,6 +1320,67 @@ private ExTransitionInfo CalcTransitionExplanations(ExTransitionInfo info, strin
...
+
+            /// <summary>
+            /// Finds the closest custom product ion if any to a specified product m/z value.
+            /// </summary>
+            private ProductExp GetCustomIonProductExp(double productMz, ExTransitionInfo info)
+            {
+                double minDeltaMz = double.MaxValue;
+                ProductExp productExp = null;
+                var massType = Settings.TransitionSettings.Prediction.FragmentMassType;
+
```
bspratt, line 1367: Is it always the case that we aren't provided at least a charge to go with the m/z? That could help disambiguate, when/if available.

brendanx67: Right. In proteomics, we are not guaranteed the user supplied a charge. I think the minimum information is sequence, precursor m/z, and product m/z.

bspratt: That's what I figured, but it seems like we ought to consult it when it is provided?

brendanx67: Good point. I did that for the precursor charge, but not the custom ion charge. Good catch.

brendanx67:
@nickshulman @bspratt When I started messing with the first transition in my test list, I realized things were a bit more complicated than I had thought. The old code failed if you started the list with a non-monoisotopic precursor or a custom ion. That the monoisotopic precursor worked was a bit of a fluke, just because it is valid for both MS1 and MS/MS spectra. I added code to fix these cases and tests to make sure they work.

PR #2852, opened 2024-01-30
* copied DiaUmpire's InstrumentParameter config struct from C++ to C#

chambm:
* copied DiaUmpire's InstrumentParameter config struct from C++ to C# to avoid using CLI bindings to access it; hopefully this is the source of crashes after DiaUmpireTutorial tests on BSPRATT-UW2

Brian I guess you'll want to point BSPRATT-UW2's at this branch for the integration nightlies?


bspratt:
I confess I don't know what's going on with this change, but I can certainly set up BSPRATT-UW2 to run this branch.

chambm:
This change is simply porting the InstrumentParameter code, that DiaUmpire uses for its config variables, from C++ to C#. So it doesn't have to be accessed through the CLI bindings. I also removed dead code previously used to run DiaUmpire in-process in Skyline.

bspratt:
Sounds good. Tomorrow's test reports from BSPRATT-UW2 will reflect these changes.

chambm:
I guess we'll let integration run a few more days, but it seems to have fixed the access violation. I don't understand why not disposing the MSData from IsValidFile() would lead to any kind of corruption though. :(

bspratt (approved):
Excellent, once you've done the merge I'll move on with other uses for the Integration build, thanks!

PR #2855, opened 2024-02-05
PanoramaClient updates

vagisha:
- Look for the CSRF token in both the response and request cookies. When a request is redirected, the CSRF token will be in the request cookies instead of the response cookies.  CSRF tokens are required for POST requests to LabKey Server. 
- Request JSON instead of HTML response from LabKey by setting the HttpRequestHeader.Accept header.  The response will contain an error message if the request failed on the server even if the HTTP status was 200.  Include the LabKey error in the error message dialog. Example: when file upload fails because file name is not accepted by LabKey)

    image
In this screenshot the "Response status" is parsed from the JSON response.  The actual HTTP status was 200. 
JSON response: ```{"exception":"Filename may not contain space followed by dash.","success":false,"status":400}```

- Look for LabKey-specific error message in the response when the request results in a WebException (anything other than 200 HTTP status). 
image
JSON response in the WebException was: ```{"exception" : "You must log in to view this content.","success" : false}```
The error message includes both the WebException message and the LabKey error. The LabKey error here is not helpful to the user so maybe we should ignore it? 

- Move some methods from PanoramaPublishUtil.cs to the PanoramaClient project. Not all methods could be moved because they rely on classes that are not accessible in the PanoramaClient project (e.g. LongWaitDlg).

### Changes:
- Code was refactored to make testing easier.  An IRequestHelper interface was added

```
IDisposable
   ^
   |
IRequestHelper
   ^
   |
AbstractRequestHelper
   ^                  ^
   |                  |
PanoramaRequestHelper  TestRequestHelper
```

`IRequestHelper` provides methods for HTTP GET, POST etc. methods. 
`AbstractRequestHelper` handles common processing logic.
`PanoramaRequestHelper` focuses on the specifics of making HTTP requests with credentials and CSRF token handling.  It uses a subclass of C# `WebClient` to perform the actual operations. 
`TestRequestHelper` allows for controlled testing of methods that depend on HTTP requests without making actual network calls, simulating HTTP interactions based on the provided URI.

`AbstractRequestHelper` checks for:
    - A valid JSON response from a GET request. We expect all GET requests to LabKey server to return a JSON response.  Throws `PanoramaServerException` if response could not be parsed as JSON.  
    - A LabKey error in the response.  Throws a `PanoramaServerException` if response contains an error. 
    - If the GET request throws a WebException, the response in the exception is parsed and a PanoramaServerException is thrown.
  
- ` IPanoramaClient` and implementing classes were moved to a new file PanoramaClient.cs
- Added extension methods to return error messages based on the enum state for ServerStateEnum, UserStateEnum and FolderState.  CONSIDER: remove these enums entirely.
- Added `ErrorMessageBuilder` to build the error message for `PanoramaServerException`.  Removed constructors from `PanoramaServerException` that took various permutations of ServerStateEnum, UserStateEnum, error string and requestUri. 
- Moved some methods from `IPanoramaPublishClient` to `IPanoramaClient`.  Some methods could not be moved because they use classes not accessible in the PanoramaClient project (e.g. LongWaitDlg, ShareType).


pwiz_tools/Shared/PanoramaClient/PanoramaUtil.cs
```diff
@@ -675,37 +349,128 @@ private static string GetErrorMessage(FolderState state, string folderPath, stri
...
+            }
+            if (_uri != null)
+            {
+                sb.AppendLine(string.Format(Resources.GenericState_AppendErrorAndUri_URL___0_, _uri.AbsoluteUri));
+            }
+
+            if (!string.IsNullOrEmpty(_responseString))
+            {
+                // TODO: Should we truncate the response string, or display it at all?
+                //       CommonAlertDlg truncates the displayed string to 50000.
```
nickshulman, line 451: FWIW, CommonAlertDlg hangs onto the entire message in case the user presses the copy button. It does truncate the message to display it.

pwiz_tools/Shared/PanoramaClient/RequestHelper.cs
```diff
@@ -0,0 +1,317 @@
...
+
+        public abstract string DoPost(Uri uri, string postData); // Used only in AuditLogTutorialTest
+
+        public abstract void DoAsyncFileUpload(Uri address, string method, string fileName);
+        
+        public abstract void AddHeader(HttpRequestHeader header, string value);
+
+        public abstract string GetResponse(HttpWebRequest request);
+
+        public abstract Func<WebException, LabKeyError> GetWebExceptionResponseReader();
```
nickshulman, line 59: It's unusual to have an abstract method which returns a Func.
Usually this would be:
`public LabKeyError GetErrorFromWebException(WebException webException)`
Then, if you needed a Func to pass to "GetErrorFromWebException" itself, because if you refer to a method name without any parentheses you can use it the way you would a Func or an Action.

pwiz_tools/Shared/PanoramaClient/PanoramaUtil.cs
```diff
@@ -675,37 +349,128 @@ private static string GetErrorMessage(FolderState state, string folderPath, stri
...
-            var message = mainMessage;
+            Exception(ex, null);
+            return this;
+        }
 
-            if (error != null || uri != null)
+        // If the given exception is a WebException, this method tries to parse the response as JSON and
+        // get a LabKey error if it was set. The method accepts a function to parse the response in the 
+        // WebException so that tests can pass in a custom function.
+        public ErrorMessageBuilder Exception(Exception ex, Func<WebException, LabKeyError> labkeyErrorGetter)
```
nickshulman, line 401: It might be better if this method took an Exception and a LabKeyError.
It's unusual to pass a Func into a method when the caller could so easily have called the function themselves and passed in the result instead.

pwiz_tools/Shared/PanoramaClient/PanoramaUtil.cs
```diff
@@ -675,37 +349,128 @@ private static string GetErrorMessage(FolderState state, string folderPath, stri
...
+            return this;
+        }
+        public ErrorMessageBuilder Response(string response)
+        {
+            _responseString = response;
+            return this;
+        }
 
-            return message;
+        public string Build()
```
nickshulman, line 433: I would probably call this method "ToString" because that's what the method on "StringBuilder" is called, but "Build" is a fine name too.

nickshulman (commented):
Looks very nice. Here are some comments.

pwiz_tools/Shared/PanoramaClient/PanoramaClient.cs
```diff
@@ -0,0 +1,762 @@
...
+                        // If progress is > 1% half the wait time
+                        _waitTime = Math.Max(1, _waitTime / 2);
+                    }
+                    else if (delta < 1)
+                    {
+                        // If progress is < 1% double the wait time, up to a max of 10 seconds.
+                        _waitTime = Math.Min(10, _waitTime * 2);
+                    }
+
+                    Thread.Sleep(_waitTime * 1000);
```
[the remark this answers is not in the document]
nickshulman: It looks like this gets called from "LongWaitDlg.RunWork".
If the user hits the Cancel button in the LongWaitDlg, the dialog stays up until the thread it launched finishes.
For this reason, instead of doing Thread.Sleep(_waitTime * 1000), it would be better to do:
```
for (int i = 0; i < _waitTime * 10; i++) {
    if (progressMonitor.IsCancelled)
         return;
    Thread.Sleep(100);
}
```
so that the user never has to wait more than 100msec for the LongWaitDlg to disappear.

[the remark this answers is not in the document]
vagisha: Thanks!

Issue #2755, opened 2023-10-20
Small molecules added from NIST-type library are missing ExplicitRetentionTime


meowcat:
Steps to reproduce: 
1) create blank document in molecule mode
2) View -> spectral libraries. Dropdown -> Add, choose a name, add the MSP file from this gist: 
https://gist.github.com/meowcat/4f0a08cd803d941cd1ac53829931ee5f
3) Select the new library in the dropdown, select the "Metamitron" molecule, "Add"
4) Confirm "adding the library to your document"
5) Right-click on the added entry "Metamitron...", modify
Explicit Retention Time is missing.

Investigations:
* RetentionTime is parsed correctly from the MSP.
* When performing `AddPeptide`: https://github.com/ProteoWizard/pwiz/blob/fd07aa9749745d79c21fa38ae68e4d55cff9ccdc/pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs#L1734 At L1784 breakpoint, in `pepMatcher.PeptideMatches[0].Value`, the retention time is present in `LibInfo.RT` but absent in `NodePep.ExplicitRetentionTime`.
* I haven't figured out yet where the `PeptideDocNode` entries are created from (supposedly) the library info.

meowcat:
In https://github.com/ProteoWizard/pwiz/blob/1f52dda5cab072f40ce0f641ed36a4a62c1d0726/pwiz_tools/Skyline/Model/LibKeyModificationMatcher.cs#L339, the nodePep created from CreateDocNodeFromSettings is missing the ExplicitRetentionTime

meowcat:
I have a fix in https://github.com/ProteoWizard/pwiz/pull/2860 but I don't know if that's the right way to do it.

PR #2860, opened 2024-02-07
Skyline/work/20240207 nist library rt

meowcat:
Addresses https://github.com/ProteoWizard/pwiz/issues/2755: When using NIST (MSP) spectral libraries for small molecules, the retention time is stored in the library but not added to the document.


brendanx67:
Explicit RT is too strong a statement to take from a library where we have
no idea if it comes from exactly the same chromatography. Without extra UI
to specify this choice it is too strong an assumption to make by default.
This is a misunderstanding of what Explicit RT means and its intended
purpose.

meowcat:
OK, I guess I need to understand the current model better then. 
* If the retention time from a library is not used as Explicit Retention Time, why is it imported in the first place / what is it currently used for? 
* When adding a library in "Edit Library", there is the checkbox "Use explicit peak bounds". I don't actually know what it does, but as a user I would have expected this to govern whether or not RT are imported. Currently, I see no change in the import whether or not I activate the checkbox. Does this do something else I am not aware of?
* Would it be an acceptable UI choice to have a checkbox (deactivated by default) "Add explicit retention time"?

As a background, our motivation: We maintain in-house MS2 libraries with precise RT annotated. I'm aware that I can `Modify` the explicit RT by hand, compound by compound, but this is a lot of work and error-prone if the user is handling 100+ compounds.  On the other hand, when we start via Transition List (where the RT can be specified), I cannot add the library spectrum to a compound. 
* Is there an existing workflow to add an explicit retention time (e.g. from a CSV file) *in batch* to molecules imported from a Spectral Library? 
* Is there an existing workflow to add spectra from spectral libraries to molecules added with a Transition List (with explicit RT specified)?

This would be enough for us, but I haven't found this yet...

bspratt:
> * If the retention time from a library is not used as Explicit Retention Time, why is it imported in the first place / what is it currently used for?

Yikes, I haven't looked at this PR in a while. Perhaps your question has already been answered elsewhere, but here's where that library RT information gets used:
image

PR #2863, opened 2024-02-07
Quiet the TeamCity Release Branch ReSharper code inspection …

bspratt:
… with a few changes of inspections from WARNING to SUGGESTION, and backporting a handful of "Resharper disable once" comments from master.

Only MsDataFileImpl.cs changes semantically, but that did seem like an important change as it involved a return value being disposed before the caller could use it.

### DIFF
--- a/pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
+++ b/pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
@@ -26,7 +26,6 @@ using pwiz.CLI.cv;
 using pwiz.CLI.data;
 using pwiz.CLI.msdata;
 using pwiz.CLI.analysis;
-using pwiz.CLI.util;
 using pwiz.Common.Chemistry;
 using pwiz.Common.Collections;
 using pwiz.Common.Spectra;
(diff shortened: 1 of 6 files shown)


nickshulman (approved):
Looks good to me.

bspratt:
@chambm the MsDataFileImpl.cs change is the only thing that makes me feel a bit wiggly, though all the tests seem happy enough. Any thoughts on that?

brendanx67:
Can you explain why the storage() is okay to hang onto after the object gets disposed? That is certainly not obvious. The tests, of course, were fine before the change. So, the dispose doesn't seem to be tearing down this object completely, but it does seem like a bad idea to hang onto owned objects of a disposed object without any clear ownership release.

nickshulman:
> Can you explain why the storage() is okay to hang onto after the object gets disposed? That is certainly not obvious. The tests, of course, were fine before the change. So, the dispose doesn't seem to be tearing down this object completely, but it does seem like a bad idea to hang onto owned objects of a disposed object without any clear ownership release.

We made this same change in the master branch at the end of last summer right after we forked for 23.1:
https://github.com/ProteoWizard/pwiz/pull/2700/files/64433e5315012705bb3c9fcf41de5ccbf7d96cd4

At the time, the old code was seen as a "why did that ever work?" thing.

chambm:
> Can you explain why the storage() is okay to hang onto after the object gets disposed? That is certainly not obvious. The tests, of course, were fine before the change. So, the dispose doesn't seem to be tearing down this object completely, but it does seem like a bad idea to hang onto owned objects of a disposed object without any clear ownership release.

The value returned from Storage() is a managed array (even if it has to be copied from native storage), so it's not subject to the Dispose/Finalize, which only affects unmanaged resources.

brendanx67 (approved):

PR #2868, opened 2024-02-13
Add a null check to deal with exception in transition list reader …

bspratt:
…reported anonymously at https://skyline.ms/announcements/home/issues/exceptions/thread.view?entityId=4fb740bf-abe7-103c-a8e3-22f5355681f8&_anchor=63438 . I don't have a repro but this seems like reasonable prophylaxis.


pwiz_tools/Skyline/Model/SmallMoleculeTransitionListReader.cs
```diff
@@ -201,8 +201,14 @@ public SrmDocument CreateTargets(SrmDocument document, IdentityPath to, out Iden
...
         {
-            var adduct = precursor.Adduct;
+            var adduct = precursor?.Adduct;
+            if (Adduct.IsNullOrEmpty(adduct))
+            {
+                return true; // Can't process
+            }
+            // ReSharper disable PossibleNullReferenceException
             var precursorMonoMz = adduct.MzFromNeutralMass(precursor.MonoMass);
+            // ReSharper restore PossibleNullReferenceException
```
nickshulman, line 211: An alternative way to suppress the ReSharper warning would be to put an exclamation mark before the ".":
`adduct!.MzFromNeutralMass`

nickshulman (approved):

PR #2884, opened 2024-02-28
* changed MSAmanda temporary directory to use ~SK temp prefix

chambm:
* changed MSAmanda temporary directory to use standard ~SK temp prefix instead of ~SK_MSAmanda which should bring the path length issues under control on KAIPO-DEV

### DIFF
--- a/pwiz_tools/Skyline/Model/DdaSearch/MSAmandaSearchWrapper.cs
+++ b/pwiz_tools/Skyline/Model/DdaSearch/MSAmandaSearchWrapper.cs
@@ -68,13 +68,14 @@ namespace pwiz.Skyline.Model.DdaSearch
         private const string MAX_LOADED_SPECTRA_AT_ONCE = "MaxLoadedSpectraAtOnce";
         private const string CONSIDERED_CHARGES = "ConsideredCharges";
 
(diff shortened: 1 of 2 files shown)


bspratt:
I like it, but I don't see why we're only doing it in test situations. There are lots of users out there with similar installations, and even longer usernames.

chambm:
There is sense in having all of MSAmanda's junk stored in a single place in the temp directory. As for needing to use a shorter path for real user situations, I think there's plenty of room after even a very long username's temp directory:
```
Realistic temp paths (IMO):
C:\Windows\Temp
C:\Users\slartibartfast.rumpelstiltskin@somereallylongdomain.com\AppData\Local\Temp

Not realistic:
C:\Users\kaipo\Documents\Nightly\SkylineTesterForNightly_integration\pwiz_tools\Skyline\TestResults\TestRunner results\~&TMP ^\TDS1279试试
```

bspratt (approved):
I don't love it since (as acknowledged) it just kicks the can down the road, but for test purposes it gets us moving again on the unrelated original issue of mysterious memory spikes, so OK.

bspratt:
Thanks. I've merged this into the integration branch.

PR #2885, opened 2024-02-28
Library Explorer improvements inspired by larger small molecules libraries:

bspratt:
* Peptides and/or molecules are shown in natural sort order (i.e. aware of numerical values embedded in strings) - this matters more for molecules with constructed names like "mass123.456", "mass123.0456" etc. For peptides it generally amounts to the same thing as a plain alphanumeric sort.
* We now show the number of things that match the filter rather than always just giving the total count.
* We remember the user's search string for each filter type as they choose and return to different filters.
* Fixed a bottleneck that was preventing display of ion mobility values in the library explorer.

This required a new natural sort class to complement the existing one, which turns out to be filename specific. The original doesn't understand that "z123.4056" comes before "z123.456" because it doesn't see "123.4056" as a decimal value - it sees "123" and "4056" as two different numbers and is comparing 4056 to 456. No real surprise since it's just a wrapper for a windows call for sorting filenames.

This was all done in the service of the Feature Finding work but stands alone nicely so I've given it a separate PR.

### DIFF
--- a/pwiz_tools/Shared/CommonUtil/Collections/NaturalComparer.cs
+++ b/pwiz_tools/Shared/CommonUtil/Collections/NaturalComparer.cs
@@ -18,7 +18,6 @@
  */
 
 using System;
-using System.Collections.Generic;
 using System.Globalization;
 using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
@@ -50,12 +49,12 @@ namespace pwiz.Common.Collections
     // Similar to the filename comparer, except that decimals are understood
     // e.g. for the filename comparer, "123.4056" comes after "123.456" because 4056 > 456 -
     // it doesn't seem them as two parts of a decimal value
-    public class NaturalStringComparer : IComparer<string>
+    public class NaturalStringComparer
     {
         // Regular expression pattern to match decimal parts - note it accepts both . and , decimal separator
         private static readonly Regex REGEX = new Regex(@"(\d+(?:[.,]\d+)?|\D+)", RegexOptions.Compiled);
 
-        public int Compare(string x, string y)
+        public static int Compare(string x, string y)
         {
             if (x == null)
             {
--- a/pwiz_tools/Shared/CommonUtil/Collections/NaturalComparer.cs
+++ b/pwiz_tools/Shared/CommonUtil/Collections/NaturalComparer.cs
@@ -17,7 +17,11 @@
  * limitations under the License.
  */
 
+using System;
+using System.Collections.Generic;
+using System.Globalization;
 using System.Runtime.InteropServices;
+using System.Text.RegularExpressions;
 
 namespace pwiz.Common.Collections
 {
@@ -31,7 +35,7 @@ namespace pwiz.Common.Collections
     /// e.g. (1AC6, 1AC66, 1AC7, 4C47 --> 1AC6, 1AC7, 1AC66, 4C47)
     /// https://www.pinvoke.net/default.aspx/shlwapi.strcmplogicalw
     /// </summary>
-    public class NaturalComparer
+    public class NaturalFilenameComparer
     {
         [DllImport("shlwapi.dll", CharSet = CharSet.Unicode, ExactSpelling = true)]
         private static extern int StrCmpLogicalW(string x, string y);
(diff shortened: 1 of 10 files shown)


pwiz_tools/Shared/CommonUtil/Collections/NaturalComparer.cs
```diff
@@ -42,4 +46,75 @@ public static int Compare(string x, string y)
             return StrCmpLogicalW(x, y);
         }
     }
+
+    // Similar to the filename comparer, except that decimals are understood
+    // e.g. for the filename comparer, "123.4056" comes after "123.456" because 4056 > 456 -
+    // it doesn't seem them as two parts of a decimal value
+    public class NaturalStringComparer : IComparer<string>
```
chambm, line 53: Why can't this be static? Something to do with the Regex?

chambm: Oh nevermind. Can't be IComparer if it's static., duh.

chambm: Well, maybe not nevermind. The original code called Sort with NaturalComparer.Compare without needing to `new` the Comparer class. Why does the new code need `new`?

bspratt: You're probably right, could probably have a static. I'll try that.

bspratt: Yeah, we don't really need it to implement IComparer, may as well keep it simple. I've changed it, thanks.

pwiz_tools/Skyline/SettingsUI/ViewLibraryPepInfoList.cs
```diff
@@ -31,7 +31,8 @@ namespace pwiz.Skyline.SettingsUI
 {
     public class ViewLibraryPepInfoList : AbstractReadOnlyList<ViewLibraryPepInfo>
     {
-        private readonly ImmutableList<ViewLibraryPepInfo> _allEntries;
+        private readonly ImmutableList<ViewLibraryPepInfo> _allEntries; // Alpha sorted
+        private readonly ImmutableList<int> _naturalSortIndex; // Natural sort order of the members of _allEntries
```
nickshulman, line 35: Is there a reason that we don't want to just start off with having everything sorted naturally, instead of maintaining one ordering for the way things are initially displayed, and different one for when they apply a filter?

pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs
```diff
@@ -1361,7 +1367,24 @@ private void comboFilterCategory_SelectedIndexChanged(object sender, EventArgs e
...
+                    textPeptide.Text = newText;
+                    needsUpdate = false; // Update will file automatically
+                }
+            }
+            if (needsUpdate)
+            {
+                FilterAndUpdate();
+            }
+
+            textPeptide.Focus(); // Assume that the next thing the user wants to do is work with the filter value
```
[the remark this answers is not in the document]
bspratt: Can do. There are a number of similar calls in that file, are they of similar concern?

pwiz_tools/Skyline/SettingsUI/ViewLibraryPepInfoList.cs
bspratt: It's about searching for partial matches while the user is typing in a filter string. We don't need or want anything fancier than alphanumeric matching for that, so the sort order is just straight up string comparison. And it's faster too.

pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs
```diff
@@ -2962,11 +2988,9 @@ private ViewLibraryPepInfo SetIonMobilityCCSValues(ViewLibraryPepInfo entry)
         /// </summary>
         public IonMobilityAndCCS GetIonMobility(ViewLibraryPepInfo pepInfo)
         {
-            var bestSpectrum = _selectedLibrary.GetSpectra(pepInfo.Key,
-                IsotopeLabelType.light, LibraryRedundancy.best).FirstOrDefault();
-            if (bestSpectrum != null)
+            if (_selectedLibrary.TryGetIonMobilityInfos(new[] { pepInfo.Key }, out var ionMobilities))
             {
-                return bestSpectrum.IonMobilityInfo;
+                return ionMobilities.GetIonMobilityDict().Values.First().FirstOrDefault();
```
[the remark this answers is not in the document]
bspratt: Nice, thanks.

PR #2890, opened 2024-03-05
Use Nick's AssortResources logic in our custom code inspection to remind devs to move resources from main Skyline resource file to appropriate smaller resources files.

bspratt:
Developers will see messages like this:

!!! CodeInspection FAILED
Assert.Fail failed. 6 code inspection failures found:

4 non-shared resource(s) should be moved from pwiz_tools\Skyline\Properties\Resources.resx into \pwiz_tools\Skyline\Controls\Graphs\GraphsResources.resx: 
SummaryIntensityGraphPane_SummaryIntensityGraphPane_Protein_Rank 
Extensions_CustomToString_Relative_Abundance
AreaPeptideGraphPane_UpdateAxes_Peptide_Rank
SummaryRelativeAbundanceGraphPane_UpdateAxes_Molecule_Rank 
1 non-shared resource(s) should be moved from pwiz_tools\Skyline\Properties\Resources.resx into pwiz_tools\Skyline\SettingsUI\SettingsUIResources.resx: 
SpectrumGraphItem_library_entry_provides_only_precursor_values 
1 non-shared resource(s) should be moved from pwiz_tools\Skyline\Properties\Resources.resx into pwiz_tools\Skyline\Controls\GroupComparison\GroupComparisonResources.resx: 
VolcanoPlotFormattingDlg_VolcanoPlotFormattingDlg_Protein_Expression_Formatting

This can be done with command:
"pwiz_tools\Skyline\bin\x64\Release\AssortResources.exe" --resourcefile "pwiz_tools\Skyline\Properties\Resources.resx" --projectfile "pwiz_tools\Skyline\Skyline.csproj"


bspratt:
Note that this WILL fail the code inspection test for the moment, since there are in fact strings that need to be relocated. I'll make that change before actually merging to master.

nickshulman:
This pull request seems to be missing some files in "Test/AssortResources" which are referenced by Test.csproj

I think it probably is a good idea to have a separate copy of the source files from "AssortResources" and, in the future, the things that this test wants to do with that code will diverge from what the command line tool "AssortResources.exe" does.

I think the method "CodeInspection" is getting really big now and it would be good if it were split into multiple [TestMethod]'s, and maybe even multiple class files with a base class that knows how to read all the source files in the source tree.
It would be nice if each of the things that is currently a "AddTextInspection" was itself a different [TestMethod] so you would be able to tell just from the name of the test that failed which rule was violated.

I realize that if the inspection were split up into different [TestMethod]'s it would mean that the source files would need to be read multiple times instead of being able to do everything in one pass, but I think the reading of source files could be parallelized and made fast enough that it wouldn't matter.

bspratt:
Oops, those were not meant to be copies (I misunderstood the IDE's use of folder as organizational concept - it's a literal folder and they quietly copy stuff into it). It's actually important that the code be the same for the test and the  AssortResources
 executable, since the test tells the developer how to use that same code in the executable to fix the detected problem. The two should never diverge. That is, if it's worth AssortResources worrying about, it's worth our code inspection worrrying about.

You could be right about busting up the inspection test but I don't think it's a very high priority. It already tells you exactly what the problems are, if any, and under normal use doesn't have to tell you anything at all.

bspratt:
N.B. this is largely in response to the issue addressed in the Jan 3 2024 email thread 'Weird behavior of ReSharper's "Move to Resource" dialog'. 

It's just easier to move new strings to the main Skyline resx, then use the AssortResources tool to distribute them properly. It is, in fact, the only way to since VS2022 and/or ReSharper doesn't want to add to the new resx files once created.

The only problem is that we don't currently enforce the use of AssortResources, so when I go to use it on my own work I end up changing others work too. This new automation solves that problem - no one will commit without using AssortResources as needed.

nickshulman:
> N.B. this is largely in response to the issue addressed in the Jan 3 2024 email thread 'Weird behavior of ReSharper's "Move to Resource" dialog'.

I think that was caused by ReSharper being confused by where the .Designer.cs files were in the .csproj file.
You can fix this by going into "Extensions > ReSharper > Options > Environment > General" and uncheck the checkbox "Read solution model directly from project/solution files"
I think I have managed to make that step unnecessary with PR#2891 where I moved the .designer.cs files around in the .csproj file. I committed that to master this morning.
If you sync to that, ReSharper might still have trouble on an existing project, in which case you might need to jiggle (uncheck, save, recheck) the "Read solution model directly" checkbox, but I think it will work perfectly for a new checkout.

bspratt:
Excellent, though I still like the automation that AssortResources provides for creating entirely new .resx files consistently and conveniently, and for conveniently and consistently directing strings to existing files as well.

brendanx67:
Should we be running AssortedResources as a build step? Or at least in the
test the way we now auto-fix our XML schema files for a release build?

bspratt:
A test that alters the code feels a bit creepy, and running AssortedResources as a build step breaks the IDE's Undo ability. My gut says to make it a manual step before commit. It's an easy manual step since the test gives you the needed commandline.

chambm:
I wonder if there's a way to apply code changes that keeps undo ability? Probably with a VS extension. It would definitely be more useful than this one!
https://marketplace.visualstudio.com/items?itemName=lettucemode.raauu

What about this?
https://marketplace.visualstudio.com/items?itemName=SrirachaSource.CliOnSave
The AssortedResources command could run when the csproj file is saved. I /think/ that would only happen when ALL files are saved, right? That might still clear undo history though, which is less than ideal.

brendanx67:
> A test that alters the code feels a bit creepy, and running AssortedResources as a build step breaks the IDE's Undo ability. My gut says to make it a manual step before commit. It's an easy manual step since the test gives you the needed commandline.

I guess it is simpler than the xsd issue, but we do have a test that makes the necessary changes when it fails. Sounds like you are saying that code should be moved to a separate EXE?

bspratt:
There are some interesting ideas here, but I think we risk letting perfect be the enemy of good. We have before us a reasonably convenient solution that was very inexpensive to implement. If somebody wants to create a TODO to make this even more slick, go for it, but for now I really think we should take the W and move on.

bspratt:
> > A test that alters the code feels a bit creepy, and running AssortedResources as a build step breaks the IDE's Undo ability. My gut says to make it a manual step before commit. It's an easy manual step since the test gives you the needed commandline.
> 
> I guess it is simpler than the xsd issue, but we do have a test that makes the necessary changes when it fails. Sounds like you are saying that code should be moved to a separate EXE?

I'm only saying that specifically in the context of "inspection" I'm really not expecting the code to change - especially not in a way that has no IDE Undo.

chambm:
> There are some interesting ideas here, but I think we risk letting perfect be the enemy of good. We have before us a reasonably convenient solution that was very inexpensive to implement. If somebody wants to create a TODO to make this even more slick, go for it, but for now I really think we should take the W and move on.

Agree. Maybe add a warning "Before running this command, save all your changes in the IDE."

bspratt:
Oh yeah, good point.

PR #2892, opened 2024-03-07
Add a catch clause in ModifyDocument for InvalidOperationException…

bspratt:
… (alongside IdentityNotFoundException, InvalidDataException, and IOException). In this instance, to catch "Adduct [M88H2+2H] calls for labeling more H atoms than are found in the molecule C69H'109N7O42" as reported in exception web https://skyline.ms/announcements/home/issues/exceptions/thread.view?entityId=25cffaa0-b487-103c-a8e3-22f5355681f8&_anchor=63569


nickshulman:
I think for that particular exception, the catch clause should go somewhere inside of "EditMenu.ModifyPeptide".
I don't think we want a general catch clause in "ModifyDocument" for "InvalidOperationException" because that exception is so often indicative of an actual bug in our code (it's really similar to "NullReferenceException" but for Nullables).

bspratt:
I suppose you can make the case that it's the wrong kind of exception to be throwing in the first place.

bspratt:
I think it really is just the wrong exception to be throwing. I've revised it to use InvalidDataException.

bspratt:
Fair point - are you requesting a further change to the code?

brendanx67:
You can show me how it works tomorrow.

bspratt:
Can do. Looking at the call stack in https://skyline.ms/announcements/home/issues/exceptions/thread.view?entityId=25cffaa0-b487-103c-a8e3-22f5355681f8&_anchor=63569#row:63569 you can see that it's using ModifyPeptide to try to construct a new node, if the user has found any way at all to screw that up then they get a nice exception message - this was already how we handled other possible weirdness. I'm not really modifying much in the end, just throwing a more reasonable exception type for this particular kind of user error,

brendanx67:
This is an improvement over "completely broken", i.e. user editing causes an unhandled error to get posted to our website. Even better would be to catch the problem in the form and set focus to the adduct field as the cause of the issue.

I have now reproduced the problem in Skyline with a form like this:
image
If I OK this form I get the unhandled exception. Getting the message in a MessageDlg is better. Maybe that is enough. Anyway, I would like you to just show me the new behavior and we can discuss it a bit further.

bspratt:
The error is now handled thusly
image

The title of this PR is inaccurate now, the error handling is done in the UI code.

PR #2898, opened 2024-03-11
Only show "Do you want to add decoys" message if the document has at least 25 peptides

nickshulman:
Changed so that the "Do you want to add decoys" message only gets shown when importing results if the document has at least 25 peptides in it.


nickshulman:
Still have things to work out.

PR #2899, opened 2024-03-12
missed a couple of tweaks per Nicks last code review on commit b5b0d400be0a90f3bdda72c07a5772dfae3655ff…

bspratt:
… "Library Explorer improvements inspired by larger small molecules libraries"

- don't change focus when running as automated test
- more caution against possible missing entries when looking up ion mobility values


pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs
```diff
@@ -1384,7 +1384,10 @@ private void comboFilterCategory_SelectedIndexChanged(object sender, EventArgs e
                 FilterAndUpdate();
             }
 
-            textPeptide.Focus(); // Assume that the next thing the user wants to do is work with the filter value
+            if (!Program.FunctionalTest)
```
brendanx67, line 1387: Why is this necessary or better? We don't usually have code like this. It should at least be really well commented as to why it is there to avoid others writing their UI like this without justification.

pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs
```diff
@@ -2990,7 +2993,7 @@ public IonMobilityAndCCS GetIonMobility(ViewLibraryPepInfo pepInfo)
         {
             if (_selectedLibrary.TryGetIonMobilityInfos(new[] { pepInfo.Key }, out var ionMobilities))
             {
-                return ionMobilities.GetIonMobilityDict().Values.First().FirstOrDefault();
+                return ionMobilities.GetIonMobilityDict().Values.SelectMany(x => x).FirstOrDefault();
```
brendanx67, line 2996: Should this have ?? IonMobilityAndCCS.EMPTY; to maintain parity with the other return case where nothing is found? Is returning null hear meaningful? When you use FirstOrDefault() it seems to imply that you are going to sometimes return null.

brendanx67 (changes requested):
Might just be better commenting, but these changes are not clear to me.

pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs
bspratt: That would be better, yes.

pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs
bspratt: It's a @nickshulman suggestion.

nickshulman: Oops. My suggestion should have probably been:
`if (!Program.SkylineOffscreen)`
We don't want Skyline stealing the focus when it's supposed to be running offscreen. But, it's fine to steal the focus when not running off screen.

I am not sure that it makes sense to have the "textPeptide.Focus()" call here at all. We normally never have to set the focus to controls on a form when they are first shown.

bspratt: The context here is that the user has just operated the "what kind of thing do I want to search for" control, it's just a convenience to swap focus to the "what thing do I want to search for". I was annoyed at the extra click, there's really no where else I'd want to click next.

bspratt: But yes, that makes sense to avoid doing when running offscreen.

nickshulman: Oh, yes, I see. That Focus() makes sense.

I might be wrong about the "if (!Program.SkylineOffscreen)" being necessary. I'm not 100% sure that calling "Focus()" on a control actually takes the focus if the form is not already active...

bspratt: We certainly see that in the code, especially around graphs, so I expect it's probably a good thing, certainly harmless anyway.

nickshulman: I'm pretty sure I'm 100% wrong about that being necessary.
I tried writing a test which keeps setting focus to text boxes, and it never seemed to steal the focus from whatever other app I was using.
The really important thing for preventing focus stealing is the `override bool ShowWithoutActivation` in the class "FormEx".

bspratt: Excellent, I'll back it out.

brendanx67 (approved):

PR #2900, opened 2024-03-12
* switched Skyline from legacy Prosit server to new Koina framework

chambm:
* changed PrositConfig_development.xml to koina.proteomicsdb.org since Koina is freely accessible
* added new models supported by Koina (ms2pip_2021_HCD, Prosit_2020_intensity_CID, Prosit_2020_intensity_HCD, Prosit_2023_intensity_timsTOF, Deeplc_hela_hf, AlphaPept_rt_generic) except for TMT, crosslinking, and CCS
* switched from Tensorflow to Triton
* centralized HasPrositServer() function
* fixed PrositSkylineIntegrationTest to explicitly encode C+57 to make test validation code simpler
* added BiblioSpecLite.WriteDebugMgf() function to make it easy to view library peaks in an external program
* added new ModifiedSequence constructor that constructs from a modified sequence instead of unmodified sequence and list of mods
* added AssertEx.LibraryEquals (didn't end up using it though)


pwiz_tools/Skyline/Model/ModifiedSequence.cs
```diff
@@ -141,6 +141,61 @@ public ModifiedSequence(string unmodifiedSequence, IEnumerable<Modification> exp
             _defaultMassType = defaultMassType;
         }
 
+        public ModifiedSequence(string modifiedSequence, MassType defaultMassType)
```
nickshulman, line 144: It looks like this only gets used by test code.
Could this functionality of constructing a ModifiedSequence from a string be moved into the TestUtil project?

pwiz_tools/Skyline/Model/ModificationMatcher.cs
```diff
@@ -353,6 +353,32 @@ public static StaticMod GetStaticMod(string unimodNameOrId, ModTerminus? modTerm
                 };
                 if (UniMod.DictUniModIds.TryGetValue(key, out var mod))
                     return mod;
+
+                // if both AA and terminus are provided, look up keys with one or the other set instead of both
```
nickshulman, line 357: Was there a particular unimod modification that wasn't getting matched correctly that necessitated this change?
If so, then I imagine it was probably also a problem if the user tried to add the modified peptide to the document using "Edit > Insert > Peptides".

chambm: This code was originally added for the command-line modification parameters, and this change will only help that case by making it more tolerant of the user being TOO specific. For example, if they do: "--pep-unimod-add 4 --pep-unimod-aa C --pep-unimod-term N" it would add "Carbamidomethyl (C)". I had to add this for the new ModifiedSequence constructor so it would parse "C[UNIMOD:4]PEPTIDEK" correctly (because that C is both N-terminal and C).

pwiz_tools/Skyline/Model/ModifiedSequence.cs
chambm: I suppose it could, but it's pretty useful functionality and I was surprised it wasn't already in the Skyline code. Most of the code was copied from ModificiationMatcher, even.

brendanx67: Is there any way to share code with ModificationMatcher, then? Or at least add a comment, in case one gets changed/improved.

pwiz_tools/Skyline/Model/ModificationMatcher.cs
brendanx67: Maybe we should add Nick's new code to prefer "common" modifications over not common modifications.

https://github.com/ProteoWizard/pwiz/pull/2859

I worry here that we are preferring AA mods over terminal mods, which proved to be the wrong choice for "Acetyl" which has many AA versions and also an N-terminal version which is common. In a tutorial, we ended up choosing lots of AA mods instead of Acetyl (N-term) which covered everything with a single modification.

nickshulman: Oops. I didn't realize that "GetStaticMod" never actually gets used by the class "ModificationMatcher"-- it's only used by CommandLine.

"ModificationMatcher" is a nightmare of a class which tries to guess what modifications a user intended based on whatever they typed in. It was written by brilliant interns decades ago and we are terrified of changing anything about it.

This "GetStaticMod" method probably should be moved into the class "UniMod" and should maybe be renamed to something like "FindById".

pwiz_tools/Skyline/Model/ModifiedSequence.cs
chambm: I'll see about factoring out the mutual enumeration of modification blocks in the string.

chambm:
Should I rename all the Prosit stuff in the code to Koina? Or Predict[ion]?

brendanx67:
Depends on how specific it is to Koina now, or it was to Prosit. If you feel you now have generalized interfaces that could serve other predictors, those can be named "Predict[ion]". Anything that seems highly specific to Koina should probably be named for Koina. I guess the ultimate test will be in Bo implementing SkylineAI so that we have multiple prediction options and not just multiple models and settings on a single predictor.

pwiz_tools/Skyline/Model/Prosit/Models/PrositModel.cs
```diff
@@ -359,35 +406,48 @@ private TPrositOut Predict(PredictionService.PredictionServiceClient predictionC
...
+                    localProgressStatus = localProgressStatus.ChangeMessage(TextUtil.SpaceSeparate(
+                            PrositResources.PrositModel_BatchPredict_Requesting_predictions_from_Prosit,
+                            processed.ToString(), @"/", totalCount.ToString()))
+                        .ChangePercentComplete(REQUESTING_INPUTS_FRACTION * processed / totalCount);
+                    progressMonitor.UpdateProgress(localProgressStatus);
+                    Console.WriteLine(localProgressStatus.Message);
+                }
+            }, maxThreads: 4);
+
+            var prositOutputAll = prositOutput[0];
```
nickshulman, line 434: I get ArgumentOutOfRangeException on this line when I have an empty Skyline document and I try to build a spectral library from Prosit.

chambm:
@brendanx67 did you get a chance to try out this PR?

brendanx67:
I have completed buddy-testing UI review of the new Koina interface. It looks good and works nicely in the cases I usually demo during courses/workshops. Full code review coming soon.

pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
```diff
@@ -255,6 +256,15 @@ public static string EscapePathForXML(string path)
             return path;
         }
 
+        /// <summary>
+        /// Replaces invalid filename characters (\/*:?&lt;>|" and non-printable chars) with a replacement character (default '_').
+        /// </summary>
+        public static string ReplaceInvalidFilenameCharacters(string filename, char replacementChar = '_')
+        {
+            const string illegalFilename = "\\/*:?<>|\"";
+            return new string(filename.Select(t=>t < 0x20 || t == 0x7f || illegalFilename.Contains(t) ? '_' : t).ToArray());
```
brendanx67, line 265: Maybe use "c" instead of "t" since it is a character. Also, I don't love code that always creates a new array of characters even if nothing is changing. It's a memory thing, and not a speed thing. So, you could use Helpers.AssignIfEquals().

Or you could create a named static function PathEx.IsValidChar(char c) and then do return filename.All(IsValidChar) ? filename : new string(filename.Select(c => IsValidChar(c) ? c : '_').ToArray())

I think the second option would likely result in cleaner code.

pwiz_tools/Skyline/Model/ModificationMatcher.cs
```diff
@@ -146,6 +146,31 @@ private IEnumerable<AAModInfo> EnumerateSequenceInfos(string seq, bool includeUn
             bool isSpecificHeavy = FastaSequence.OPEN_MOD.All(paren => aas.Length > seq.Count(c => c == paren));
             var lossOnlyMods = Settings.PeptideSettings.Modifications.StaticModifications
                 .Where(m =>m.HasLoss && !m.HasMod).ToArray();
+            foreach (var modInfo in EnumerateMods(seq, false))
```
brendanx67, line 149: I think I need you to walk me through the need for such big changes to ModificationMatcher.

pwiz_tools/Skyline/Properties/Resources.resx
```diff
@@ -3839,372 +3836,6 @@ Go to the Peptide Settings Prediction tab to fix.</value>
   <data name="PanoramaDownload" type="System.Resources.ResXFileRef, System.Windows.Forms">
     <value>..\Resources\PanoramaDownload.bmp;System.Drawing.Bitmap, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value>
   </data>
-  <data name="RefineDlg_RefineDlg_Precursors" xml:space="preserve">
```
brendanx67, line 3842: Seems like a lot of deleted strings with no clear relation to Prosit/Koina. Can you explain this? Is Nick's new tool somehow responsible?

pwiz_tools/Skyline/TestPerf/EncyclopeDiaSearchTutorialTest.cs
```diff
@@ -71,12 +70,12 @@ public void TestEncyclopeDiaSearchTutorial()
                     //"23aug2017_hela_serum_timecourse_wide_1f.mzML",
                 },
 
-                FinalTargetCounts = new[] { 317, 552, 552, 3922 },
+                FinalTargetCounts = new[] { 370, 725, 725, 5104 },
```
brendanx67, line 73: Do we understand why these numbers get so much bigger?

brendanx67 (commented):
Mostly looks okay. Lots of search and replace. A few places I would like more of an explanation. We can talk those through in person. But since it passed all my manual testing and passes all our automated testing, it is likely very close to being ready.

pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
chambm: Although it's probably a bit of premature optimization, I think your second idea is spot on.

pwiz_tools/Skyline/Model/ModificationMatcher.cs
chambm: These changes are in order to share the modification enumeration code between ModificationMatcher and the new ModifiedSequence ctor (that takes a modified sequence instead of an unmodified one and a list of mods). Happy to have a meeting to go over it though.

pwiz_tools/Skyline/Properties/Resources.resx
chambm: Probably a merge mistake. I'll try rebasing.

pwiz_tools/Skyline/TestPerf/EncyclopeDiaSearchTutorialTest.cs
chambm: Not entirely. I had been thinking it was due to slight differences in the Prosit peaks and retention times predict with Koina, but a deeper comparison of the old prosit blib and the new koina one reveals some pretty extreme RT differences:
image
The peak count differences are not nearly as extreme. Do you think the RT differences would explain EncyclopeDIA results? I need to dig into that to see what's going on.

brendanx67: It would probably be good, then, to look at the RT regression plots in residuals mode to get a sense of how far from the predictions the detected peaks end up. It would definitely make a big difference if Koina improved its iRT prediction a lot. I found that Prosit's iRT predictions were generally twice as far away from the detected peaks as empirical values.

https://skyline.ms/_webdav/home/software/Skyline/%40files/2020-ASMS-Skyline-Prosit.pdf#page=11

It might be a good idea for us/me to rerun these datasets with Koina predictions. You can see on page 11 of the presentation how detections went way down with Prosit unless the chromatogram extraction window was doubled from 10 minutes to 20 minutes.

chambm: Turns out the old PredictBatches function I implemented for EncyclopeDIA was never getting the iRTs right (the same peptide was getting different iRTs!). So we can't really compare the new results with the old. All I can do is say that the peptides are getting the correct iRTs now (into the blib) and it matches what Koina predicts when I use their python example code. I think maybe that EncyclopeDIA tutorial needs a critical and knowledgeable eye to go over it to see if any of the settings are unreasonable. You mentioned used a really big RT window for matching when using Prosit and it seems to be using quite a small window (based on the MacCoss lab EncyclopeDIA guide I followed, IIRC).

chambm:
Shall I go ahead and merge this @brendanx67 ?

brendanx67:
Yes!

rita-gwen:
Matt, it looks like you did not include the Settings.settings file into this PR. I see new Koina properties in the Settings.designer.cs file, but when I add a setting of my own the Koina ones disappear.

nickshulman:
> Matt, it looks like you did not include the Settings.settings file into this PR. I see new Koina properties in the Settings.designer.cs file, but when I add a setting of my own the Koina ones disappear.

I'll make a PR to fix this.

PR #2902, opened 2024-03-18
Fix ArgumentOutOfRangeException in EditPepModsDlg.LoadLists

nickshulman:
Fixed error using "Edit Modifications" dialog (reported anonymously on exception web)

Here's an example of the exception:
https://skyline.ms/announcements/home/issues/exceptions/thread.view?rowId=63793

This fix makes it so that "SelectedIndex" has the value "-1" if it would otherwise be beyond the end of the list of items. (It's better to make it be -1 instead of "Items.Count - 1" because the last item in the list is typically "&lt;Edit List...>")


brendanx67:
This seems relatively easy to reproduce using <Edit list...> in one of these combos. Did you poke around at this a bit? I would like to see a test added to cover this case.

brendanx67:
Looks like comments can't deal the &lt; and removed the "Edit list..." in my comment.

nickshulman:
> Looks like comments can't deal the < and removed the "Edit list..." in my comment.

If you want to include a "`<`" in your comment you can either encode it as "`&lt;`" or you can mark it as "code" by enclosing it in backquote characters.

brendanx67:
That's how I got the &lt; into my follow-up comment. Either didn't know before, or I just forgot. Anyway, you got the point about using "Edit list..." to reproduce this unhandled exception, and the need for a test. In this form, certainly, I think we would want to see any change to the list that removes the selected modification set the modification back to none.

brendanx67:
This form seems fairly broken to me in dealing with list changes and selection. Maybe I should be the one to fix it?

PR #2903, opened 2024-03-20
Skyline: Fix LiteDropDownList use to set SelectedIndex to -1

brendanx67:
after list is cleared

- Remove Nick's prior attempt at a fix
- Expand LiteDropdownListTest to cover the problem case


nickshulman (approved):

bspratt (approved):
Pleasantly straightforward

PR #2904, opened 2024-03-20
Skyline: Add string extension TextUtil.ToLines for line separation

brendanx67:
- Improve diagnostics to help understand why KAIPOT-PC1 fails MassListSpecialIonsTest


nickshulman (approved):
Looks good to me.
I probably would have made it so that "ToLines" returned `IEnumerable<string>` and make the caller call "ToList()" if they needed that, but this way is fine too.

brendanx67:
> Looks good to me. I probably would have made it so that "ToLines" returned `IEnumerable<string>` and make the caller call "ToList()" if they needed that, but this way is fine too.

I had been thinking of string[] File.ReadAllLines() and was feeling I should rename it when I noticed that File.ReadLines() does what you suggest. So, I renamed it to string.ReadLines() and did what you suggested. Looks better to me.

PR #2907, opened 2024-03-23
Proper exception handling for case in "Modify Molecule" form where user tries to force charge state change in a way that makes no sense…

bspratt:
…  with the current adduct


brendanx67 (commented):
I played around with this form a bit for the last unhandled exception, and I found editing the adduct somewhat squirly in the way it kept trying to update while I typed. Generally, I would expect most validation to happen in OkDialog() instead of while the user types.

I guess the fact that it short-circuits when it can't validate as a signed number probably helps to avoid showing messages.

But, generally, our UI design has favored letting the user work and then validating their work when they click the OK button.

brendanx67:
Otherwise, I don't really see a problem with this code.

PR #2909, opened 2024-03-27
Allow for adducts like [M+H-H2O] on molecules described as mass only.…

bspratt:
… Formerly would complain that there's no H2O to remove from an empty formula, now it just subtracts the mass.

### DIFF
--- a/pwiz_tools/Skyline/Util/Adduct.cs
+++ b/pwiz_tools/Skyline/Util/Adduct.cs
@@ -1354,6 +1354,9 @@ namespace pwiz.Skyline.Util
                 return Unlabeled.ApplyToMolecule(molecule);
             }
 
+            var monoMassOffset = molecule.MonoMassOffset;
+            var averageMassOffset = molecule.AverageMassOffset;
+
             var resultDict = new Dictionary<string, int>(molecule.Molecule);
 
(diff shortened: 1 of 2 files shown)


pwiz_tools/Skyline/Util/Adduct.cs
```diff
@@ -1363,33 +1366,43 @@ public ParsedMolecule ApplyToMolecule(ParsedMolecule molecule)
                 {
                     resultDict[element] *= MassMultiplier;
                 }
+                monoMassOffset *= MassMultiplier;
```
brendanx67, line 1369: This seems like it would impact molecules that are not mass-only, but have both a formula and a mass offset. Am I right about that? If so, I am not seeing a test for that case.

brendanx67 (approved):
Looks straightforward enough, assuming all the changes at the end are caused by indentation change due to nesting in the else clause. Nice to see a unit test for it.

I suggested that you may want to consider testing the mass multiplier on a chemical formula with offset, if that is appropriate.

pwiz_tools/Skyline/Util/Adduct.cs
bspratt: Ooh, excellent point. That wasn't properly implemented, nor tested. Set now.

PR #2910, opened 2024-03-27
Skyline (23.1): Fix signing code in Skyline/Jamfile.jam and ...

brendanx67:
Installer/Jamfile.jam
- Going forward we are using DigiCert KeyLocker and not .pfx files with passwords

### DIFF
--- a/pwiz_tools/Skyline/Executables/Installer/Jamfile.jam
+++ b/pwiz_tools/Skyline/Executables/Installer/Jamfile.jam
@@ -72,14 +72,14 @@ rule msbuild_deploy ( targets + : sources * : properties * )
     
     local .artifact-name = [ on $(targets) return $(ARTIFACT_NAME) ] ;
 
-    if [ path.exists $(PFX_PATH) ]
+    if [ path.exists $(CRT_PATH) ]
     {
-        echo "NOTE: signing Skyline MSI with key at: $(PFX_PATH)" ;
-        SIGNTOOL_COMMAND on $(targets) = "if %ERRORLEVEL% EQU 0 signtool sign /csp \"DigiCert Signing Manager KSP\" /kc key_637015839 /f \"$(PFX_PATH)\" /tr http://timestamp.digicert.com /td SHA256 /fd SHA256 /v \"$(.artifact-name).msi\"" ;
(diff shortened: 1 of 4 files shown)


pwiz_tools/Skyline/Executables/Installer/Jamfile.jam
```diff
@@ -71,12 +71,11 @@ rule msbuild_deploy ( targets + : sources * : properties * )
...
     
     local .artifact-name = [ on $(targets) return $(ARTIFACT_NAME) ] ;
-    local .pfx-password = [ MATCH --pfx-password=(.*) : [ modules.peek : ARGV ] ] ;
 
-    if [ path.exists $(PFX_PATH) ] && $(.pfx-password)
+    if [ path.exists $(PFX_PATH) ]
     {
         echo "NOTE: signing Skyline MSI with key at: $(PFX_PATH)" ;
-        SIGNTOOL_COMMAND on $(targets) = "if %ERRORLEVEL% EQU 0 signtool sign /f \"$(PFX_PATH)\" /p $(.pfx-password) /tr http://timestamp.digicert.com /v \"$(.artifact-name).msi\"" ;
+        SIGNTOOL_COMMAND on $(targets) = "if %ERRORLEVEL% EQU 0 signtool sign /csp \"DigiCert Signing Manager KSP\" /kc key_637015839 /f \"$(PFX_PATH)\" /tr http://timestamp.digicert.com /td SHA256 /fd SHA256 /v \"$(.artifact-name).msi\"" ;
```
chambm, line 78: Does key_637015839 need to change occasionally?

brendanx67: We don't expect it to change for 3 years, and then who knows if other aspects of this signing method will change as well. It took Brian C and me a lot of time to work out this exact incantation for DigiCert One. I could imagine putting the kc, csp, and tr parameters together in some single location where they would be easier to change together, but then what about SHA256? and just the command-line itself which changed unrecognizably from just a year ago. It is not possible to imagine a way that we could have written the signing code last year that would have easily adapted to this year.

Still, I have decided to clean up the code a bit more, parameterizing the key with a local in Skyline/Jamfile.jam, and keeping the ability to generate a wrapper batch file for SignAfterPublish.bat.

PR #2911, opened 2024-03-28
Skyline: Attempt to fix intermittent failures in PanoramaClientDownloadTest

brendanx67:
- Add code to avoid deleting downloaded .sky.zip files before SkylineWindow is done with them


nickshulman (approved):
Looks good to me.

PR #2912, opened 2024-03-29
Fix using shared downloads folder on macs2

nickshulman:
On Macs2, the value of the system environment variable "SKYLINE_DOWNLOAD_PATH" is D:\Shared\SkylineDownloadPath.
Users can override that by setting their own value for that environment variable.
The cleanup code at the end of tests tries to rename some of the folders in there which fails because the folders are owned by someone else.

This change makes it so that this renaming does not happen if the folder path has been set in a system environment variable instead of a user level environment variable.

A different way that this could have been implemented would be to check whether the folder was owned by a different user, but the system/user environment variable seems more likely to work in all scenarios.


brendanx67 (approved):
Looks good

PR #2913, opened 2024-03-29
Persistent files dir modification check and related fixes to some tests

chambm:
* added DirectoryEx.IsWritable()
* changed DiaUmpireDdaConverter to use document directory instead of source file directory if source directory isn't writable or running as functional test
* changed EncyclopeDiaSearchTest to not use persistent dir for mzML files (they're pretty small)
* added CheckForModifiedPersistentFilesDir() in TestFilesDir.Cleanup()


pwiz_tools/Skyline/Util/UtilIO.cs
```diff
@@ -1058,6 +1058,38 @@ public static bool IsTempZipFolder(string path, out string zipFileName)
...
+                // create the file with write permission
+                using (new FileStream(randomFilepath, FileMode.Create, FileAccess.ReadWrite))
+                {
+                }
+
+                // cleanup the file
+                File.Delete(randomFilepath);
+                return true;
+            }
+            catch (IOException)
```
nickshulman, line 1088: I think you also need to catch other exception types here such as "UnauthorizedAccessException" which is not an IOException.
Might as well change this to "catch (Exception)".

nickshulman (commented):
I wonder if, instead of checking that you have write permission on the folder by creating and deleting a file with a random name, you should create a file called something like "Skyline.pid" and hold a lock on that file for the entire amount of time that you are doing your work.

If you are unable to create the file called "Skyline.pid" then it means either that you don't have write permissions on the folder or that there is another instance of Skyline running somewhere that is already doing work in that directory, and, either way, you would want to fall back and use the document folder instead.

chambm:
> I wonder if, instead of checking that you have write permission on the folder by creating and deleting a file with a random name, you should create a file called something like "Skyline.pid" and hold a lock on that file for the entire amount of time that you are doing your work.
I think that's a solution to a somewhat different "problem": multiple Skyline instances working in the same directory. But I think locking down an entire directory is too restrictive. As long as they're working on different files, multiple instances should be able to work in one folder without a problem: suppose someone stores lots of experiments in the same folder but selects subsets of the files to import into different Skyline instances. Solving that narrower problem is messier than a single .pid file: one way I can think of is to reserve the output filenames we wanted up front, or a .pid file for each output file. Then another instance will know there's a conflict when it tries to reserve the same pid files. I think we can wait for someone to complain about this problem before trying to solve it though.

pwiz_tools/Skyline/Util/UtilIO.cs
chambm: Good catch, I didn't know unauthorized access on I/O wouldn't be derived from IOException!

pwiz_tools/Skyline/Executables/AssortResources/CsProjFile.cs
```diff
@@ -3,6 +3,7 @@
 using System.IO;
 using System.Linq;
 using System.Xml.Linq;
+#nullable enable
```
chambm, line 7: @nickshulman Any problem with adding these?

nickshulman: Looks good to me.

PR #2914, opened 2024-03-30
Skyline (23.1): Fix intermittent failures in HeavyNicknamesTest

brendanx67:


brendanx67:
Testing the release branch for a patch, TestHeavyNicknames failed my parallel testing. It took 2 separate fixes to get this test to where I could run it in parallel for all 5 languages and not fail in 1000 iterations.

In this test, I changed DismissAutoManageDialog() to have a deterministic variant that expects the auto-manage MultiButtonMessageBox and waits indefinitely for it and fails if it doesn't appear.

I did not change the function to be completely deterministic. It could, in theory, still just do nothing, if the MultiButtonMessageBox() has the wrong text in it.

I would prefer to have this function be completely deterministic always. We should know whether we expect this message box or not. The original code depends on a timeout, the document changing, or an iRT form appearing. This seems overly complicated and error prone to me. And it definitely was not dependable in TestHeavyNicknames.

bspratt:
Fair enough, will this be ported to master?

PR #2916, opened 2024-04-02
Skyline: Remove MultiButtonMsgDlg introduced inside a ModifyDocument method

brendanx67:
- ModifyDocument is a loop and should never contain UI code

It seems that this was the cause of the failure in the HeavyNickNames test. My prior attempt at a fix just moved the problem.

### DIFF
--- a/pwiz_tools/Skyline/SkylineFiles.cs
+++ b/pwiz_tools/Skyline/SkylineFiles.cs
@@ -2240,6 +2240,42 @@ namespace pwiz.Skyline
                     return;
             }
 
+            RefinementSettings refineAutoPick = null;
+            if (importer.InputType == SrmDocument.DOCUMENT_TYPE.small_molecules)
+            {
+                // We create new nodes with automanage turned off, but it might be interesting to user to have that on for isotopes etc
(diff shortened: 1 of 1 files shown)


nickshulman (approved):
Looks good to me.

By the way, a different way to solve this would be by surrounding everything with `lock (GetDocumentChangeLock())`. The usual time when we have had to resort to that is when modifying the document takes a really long time and we need to show a LongWaitDlg and there's no way to move that work outside of the ModifyDocument but it could also be used when you just want to ask the user a question.

By the way, it looks like this function ends up turning on "Auto Manage" for all Molecules and Precursors in the document when the user chooses "Enable". It would make more sense to me if it only turned it on for the newly added items. That is, if I import a transition list into a document that already had a bunch of stuff in it, I wouldn't expect that my answer to this question would affect the things that were already in my document to begin with.

brendanx67:
I agree with your observation, Nick. That would be better. I am going to go ahead and fix the problem at hand. I have been thinking about adding something like this for peptides also to allow Thermo to add just fragment ions by transition list and have code like this add both MS1 precursors and TMT reporter ions. So, I will try to sort out doing it all for only the added precursors when I do that.

PR #2922, opened 2024-04-09
Avoid some potential test race conditions by being explicit about which tests need DismissAutoManageDialog()…

bspratt:
… and which do not.


pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -2204,32 +2204,25 @@ public static void PasteSmallMoleculeListNoAutoManage()
...
-            if (expectAutoManage ?? false)
+            var autoManageDlg = WaitForOpenForm<MultiButtonMsgDlg>();
+            if (autoManageDlg == null)
             {
-                wantAutoManageDlg = WaitForOpenForm<MultiButtonMsgDlg>();
+                AssertEx.Fail($@"Expected dialog with message like ""${Resources.SkylineWindow_ImportMassList_Do_you_want_to_use_the_document_settings_to_automanage_these_new_transitions}""");
             }
-            else
+            // Make sure we haven't intercepted some other dialog
+            else if (IsFormattedMessage(Resources.SkylineWindow_ImportMassList_Do_you_want_to_use_the_document_settings_to_automanage_these_new_transitions, autoManageDlg.Message))
```
brendanx67, line 2215: This feels like it should be replaced with a simple assert, getting rid of the if clause entirely and the need to call AssertEx.Fail() yourself.

brendanx67 (changes requested):
I think the use of conditions versus straight assertions could be improved in DismissAutoManageDialog()

brendanx67 (approved):

brendanx67:
I simplified the DismissAutoManageDialog() function a bit more and removed IsFormattedMessage() in favor of using AssertEx.AreComparableStrings(). I also ended up improving diagnostics to help understand failures I saw running parallel tests. Sorry for adding unrelated changes to your PR.

bspratt:
I like my implementation better as it tells you exactly what the test was waiting for (the AutoManage dialog) rather than just "I was expecting a generic message dialog", but not enough to make a fuss about it.

brendanx67:
Yeah. You can learn that from the call stack pretty easily. Not worth adding code complexity to put that in the message.

brendanx67:
I set the timeout to 1 millisecond to see what a failure report looks like. Funny that by the time it starts reporting the failure the MultiButtonMsgDlg is present, but you can see that DismissAutoManageDialog is in the callstack.

Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException: Assert.Fail failed. Timeout 0 seconds exceeded in WaitForOpenForm(MultiButtonMsgDlg). Open forms: SequenceTreeForm (), SkylineWindow (Skyline), MultiButtonMsgDlg ()
   at pwiz.SkylineTestUtil.AbstractFunctionalTest.WaitForOpenForm[TDlg](Int32 millis) in C:\proj\pwiz\pwiz_tools\Skyline\TestUtil\TestFunctional.cs:line 783
   at pwiz.SkylineTestUtil.AbstractFunctionalTest.DismissAutoManageDialog() in C:\proj\pwiz\pwiz_tools\Skyline\TestUtil\TestFunctional.cs:line 2198

PR #2924, opened 2024-04-10
- added support for Ardia remote accounts through Open Data Source dialog

chambm:
- added support for Ardia remote accounts through Open Data Source dialog: you can login, browse Ardia, and select RAW files which will be downloaded to the document directory when importing (if they aren't already there)
* added downloadPath parameter to OpenMsDataFile for Ardia support
* added WebView2 NuGet package for Ardia login support

Things left to do:
- [x] Add progress monitoring for RAW file downloading.
- [x] Add a perftest with the bigger files in `Skyline/Large Files` on the Ardia server. Need Thermo to tell us what's in those files so we can make a .sky for it.
- [ ] Add and test handling of Ardia accounts for command-line access.
- [x] Test that opening a document with Ardia results allows viewing a spectrum (clicking on a chromatogram point) even if locally cached RAW files have been deleted (i.e. it redownloads if necessary). Current functional test is importing results into a fresh document.
- [x] Consider allowing user to leave Skyline's Ardia username/password (or at least password) blank if the user only wants to enter them in the Ardia web dialog (e.g. they don't want Skyline to store the credentials, so they will have to be re-entered if Skyline needs to redownload a RAW).
- [x] Fix leaks: `# TestArdia deltas (25): managed = 321.6 KB, heap = 108.1 KB, memory = 326.3 KB, user-gdi = -0.3, total = 6.6`
- [x] Add user option to delete local RAW files after importing. (requested by Thermo)


pwiz_tools/Skyline/Model/Results/RemoteApi/Ardia/ArdiaUrl.cs
```diff
@@ -0,0 +1,267 @@
...
+
+            var response = client.GetAsync(presignedUrl, HttpCompletionOption.ResponseHeadersRead, openMsDataFileParams.CancellationToken).Result;
+            response.EnsureSuccessStatusCode();
+            var responseStream = response.Content.ReadAsStreamAsync().Result;
+            using (var progressStream = new ProgressStream(responseStream, RawSize ?? 1))
+            using (var fileStream = new FileStream(rawFilepath, FileMode.CreateNew))
+            {
+                void ProgressStream_OnUpdateProgress(object sender, ProgressEventArgs e)
+                {
+                    var newPercentComplete = (int) Math.Round(e.Progress * 100);
```
nickshulman, line 243: ProgressStatus objects do something a little weird when you set their progress to 100, so you might want to change this to:
`var newPercentComplete = Math.Min(99, (int) Math.Round(e.Progress * 100));`

The specific weird thing that they do is that if they have multiple segments, the first time you set the progress to 100, "ProgressStatus.ChangePercentComplete" does the following:
`// Turn off progress zooming, if the end has been reached`
and the second time you call ChangePercentComplete(100) it sets "State" to ProgressState.complete

Once the progress state has been set to "Complete", "MultiProgressStatus.ChangeStatus" will stop noticing changes on it because:
`// Avoid overwriting a final status with a non-final status`

chambm: Is there a good way to indicate that the importing stage is actually downloading? There doesn't seem to be an individual text status for replicates in the AllChromatogramsGraph dialog.

nickshulman: I can't think of a good way to do something like that.

A different approach that we could take could be to make it so that Skyline does not even start trying to extract chromatograms until all of the Ardia files have been downloaded.
That is, we would implement a BackgroundLoader ("ArdiaDownloadManager") that looks at the MsDataFileUri's in the document and figures out whether any of them are ArdiaUrl objects, and, if so, starts downloading them.
We would change "ChromatogramManager.IsReadyToLoad" so that it returns false if the ArdiaDownloadManager says it's not finished.
In this way, the experience would be similar to what the user gets when their are retention time alignments that need to be performed by RetentionTimeManager, and DocumentRetentionTimes.IsNotLoadedExplained returns something indicating that there is still work to be done.
The ArdiaDownloadManager could provide progress that would show up in the Status Bar at the bottom of the Skyline window.

Doing it that way would probably be a worse experience because no chromatogram extraction would happen until all of the files had been downloaded. Also, it's hard to see how to make it work with the "delete after chromatogram extraction" checkbox. But, it's something to keep in mind as a possibility.

pwiz_tools/Skyline/Properties/Settings.cs
```diff
@@ -824,6 +824,25 @@ public OptimizationLibraryList OptimizationLibraryList
             }
         }
 
+        [UserScopedSettingAttribute]
+        public SerializableDictionary<string, string> LastArdiaLoginCookieByUsername
```
chambm, line 828: @nickshulman I don't remember why I stored the cookie directly in settings instead of as an Account property that would get stored in the RemoteAccountList, but I'm questioning the wisdom of my decision now. Maybe it had something to do with the immutability of RemoteAccount. What do you think? The cookie is used in lieu of the username/password for the duration of the session and should probably be regenerated every time Skyline is started (or when it stops working).

nickshulman: It sounds like this dictionary should just be a static variable somewhere instead of being a property on the Settings object. (If you wanted you could probably just remove the "[UserScopedSettingAttribute]" if you wanted to prevent it from being persisted, but making it a static variable somewhere else is probably cleaner)

nickshulman: I don't think there's anything inherently wrong with having transient properties in an object that do not get serialized to XML, but, when you find yourself doing that it's often a good idea to make two separate classes so you don't get confused.
That is, the thing that is stored in the Settings list has a username and password. And then, there's another type of object that has a username, a password and a cookie. And some methods want just the username and password, and other things want a username, a password and a cookie, and the easiest way to indicate which methods are which is which type is to have a different class for these two things.

PR #2926, opened 2024-04-12
Skyline: Reset standards in File > New to decrease confusion

brendanx67:
- Users often confused by "light" standard type so make it not sticky


nickshulman (approved):
Looks good to me.
I wonder if there other settings we might also want to forget when you create a new document.

brendanx67:
That’s why I put it in a function to allow room for it to grow. But for
this PR, I just wanted to solve this one issue.

I am also mulling just making it easier to create a new document with
default settings as I teach that all the time, and it requires several
clicks to get one with an empty audit log.

PR #2929, opened 2024-04-16
Honor explicit adduct charge declarations (e.g. the trailing "+" in "[M+2CH3+Cl]+")

bspratt:
Skyline may occasionally improperly calculate the ion charge of complicated adduct descriptions (e.g. "[M+2CH3+Cl]" where we did not pick up on the contributed charge +2 of 2CH3, and only used the -1 of Cl). Adding an explicit charge declaration (e.g. "[M+2CH3+Cl]**+**") would formerly just cause an error, with Skyline insisting on its own calculated charge. Now Skyline respects the declared charge.

Skyline will still complain about obviously wrong declarations like "[M+2H]-".

(Also, "CH3" is now understood as imparting a charge of +1, so  "[M+2CH3+Cl]" is now handled properly.)

Reported by user Alex


nickshulman (approved):
Looks good to me.
I see that the original unhandled exception was already fixed by the try/catch in PR #2907

PR #2930, opened 2024-04-16
Fix TestTargetedMSMSTutorial and TestRedundantComboBox

nickshulman:
The contents of the ListBox in ViewLibraryDlg sometimes gets reset inside "ViewLibraryDlg_Shown" so it's best to continually set the SelectedIndex while waiting for the graph to update.

### DIFF
--- a/pwiz_tools/Skyline/TestFunctional/RedundantComboBoxTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/RedundantComboBoxTest.cs
@@ -93,11 +93,7 @@ namespace pwiz.SkylineTestFunctional
             RunUI(() => dlg.SelectedIndex = i);
             // The peptide at index one does not have redundant spectra
             int waitMs = 1000;
-            if (!TryWaitForConditionUI(waitMs, () =>
-                {
-                    dlg.SelectedIndex = i; // keep setting the selected index in case it has been changed by "ViewLibraryDlg_Shown"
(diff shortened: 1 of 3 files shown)


pwiz_tools/Skyline/TestFunctional/RedundantComboBoxTest.cs
```diff
@@ -92,16 +93,24 @@ private static void VerifyRedundant(ViewLibraryDlg dlg, int i, bool visible, int
             RunUI(() => dlg.SelectedIndex = i);
             // The peptide at index one does not have redundant spectra
             int waitMs = 1000;
-            if (!TryWaitForConditionUI(waitMs, () => IsViewLibraryDlgState(dlg, i, visible, peakCount)))
+            if (!TryWaitForConditionUI(waitMs, () =>
```
brendanx67, line 96: Should we put this code somewhere both tests can use it, like TestFunctionalEx? Maybe start a set of methods focused on driving the ViewLibraryDlg? Hate to see code that is this complex copied to multiple tests.

nickshulman: I was planning on removing "TestRedundantComboBox" from "TargetedMsmsTutorialTest" because that's not actually part of the tutorial.

But, actually, I've thought of a better way to fix this: the "UpdateListPeptide" call in "ViewLibraryDlg_Shown" should pass in "listPeptide.SelectedIndex" to preserve the current selection in that big list box.

PR #2931, opened 2024-04-18
Fix Agilent CE optimization scenario

nickshulman:
Fixed optimization step incorrect for some Agilent collision optimization data (reported by Marilyn)

"ChromatogramDataProvider.SetOptStepsForGroup" now checks to see whether the chromatograms have the same Q3 value and, if so, tries to use the CE value to figure out which chromatogram corresponds to which optimization step.


pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -724,6 +724,12 @@ public string GetChromatogramId(int index, out int indexId)
 
         private static readonly string[] msLevelOrFunctionArrayNames = { "ms level", "function" };
 
+        public double? GetChromatogramCollisionEnergy(int chromIndex)
+        {
+            using var chrom = ChromatogramList.chromatogram(chromIndex, DetailLevel.FullMetadata);
+            return chrom.precursor?.activation?.cvParam(CVID.MS_collision_energy)?.value;
+        }
+
```
chambm, line 732: Which part of this are concerned about? It's looking in the right place for the collision energy, but of course it may not be there or it may be 0.

nickshulman: I guess I was most worried about the "DetailLevel.FullMetadata".
There are other places where there's complicated code which first tries a lower detail level and if that doesn't work tries again with escalating levels of detail.
But, it's probably not a concern here because reading chromatograms from raw files is always incredibly fast and we are going to be reading the binary data eventually anyway.
Just wanted to make sure...

PR #2938, opened 2024-04-23
Improved tolerance for ion mobility data with bad CCS calibrations - …

bspratt:
…Skyline was throwing an exception when CCS->IM conversion on transition-list-provided CCS values failed (conversion happens on the vendor DLL side, nothing we can do about that), now Skyline just proceeds without ion mobility filtering for the precursor in question.

Reported by Lauren Royer at Mobilion


bspratt:
N.B. waiting on permission to use the provoking data for a test

nickshulman (approved):

PR #2942, opened 2024-04-25
Add a static code inspection to identify tutorial links presented in the Skyline start page that do not correspond to the links found in our tests

bspratt:
Note this test will fail until these are cleaned up, at which time we can merge this to master.


bspratt:
ref https://skyline.ms/issues/home/issues/details.view?issueId=953 "add a code inspection test to check that the start page Tutorials tab uses the same links as the Tutorials tests"

PR #2945, opened 2024-04-29
Fix spectral library explorer error when 'J' amino acid has a modification on it

bspratt:
Fixed spectral library explorer error when amino acid 'J' had a modification on it (reported on exception web)
This had been fixed for 'X' in PR #1708 but the same logic needed to be used for all letters for which AminoAcid.IsAA returns false.


bspratt:
I provoked an error with a multi-line message by temporarily putting a throw in the code that normally populates the tip text:

image

pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs
```diff
@@ -2654,48 +2654,57 @@ public class PeptideTipProvider : ITipProvider
...
+                        IsotopeLabelType.light, LibraryRedundancy.best).FirstOrDefault();
+                    if (bestSpectrum != null)
+                    {
+                        _ionMobility = bestSpectrum.IonMobilityInfo;
+                    }
+                }
+                catch (Exception e)
                 {
-                    _ionMobility = bestSpectrum.IonMobilityInfo;
+                    _errorMessage = e.Message; // Very annoying to encounter a blocking window in a mouseover, just show error in the tip
```
brendanx67, line 2707: Should check whether the exception is a programming error. I really prefer much more targeted exception handling than this kind of bulk handling over the Exception type. Also, I would prefer to see you introduce a wrapper function with the try-catch in it rather than just indenting the entire contents of a function into a try-catch block. Using a new wrapper will produce easier-to-understand diffs.

pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs
```diff
@@ -2713,44 +2722,58 @@ public Size RenderTip(Graphics g, Size sizeMax, bool draw)
 
                 using (RenderTools rt = new RenderTools())
                 {
-                    // Draw sequence
-                    sizeSeq = DrawTextParts(g, 0, 0, _seqPartsToDraw, rt);
-
-                    // Draw small mol info
-                    if (_smallMoleculePartsToDraw != null)
+                    if (!string.IsNullOrEmpty(_errorMessage))
```
brendanx67, line 2725: Is there a way to make this change with a small diff? Or maybe introduce sub-functions?

brendanx67 (changes requested):
I really don't like just blindly try-catch(Exception) on everything. That will decrease the likelihood of us getting information about actual programming errors, which is the entire point of the exception report form. Now, if there is a NRE in this code, we likely will not hear about it. Nick created a function to check for programming error exceptions in catch(Exception) clauses. You should at least use that here.

pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs
bspratt: Yes, should have used IsProgrammingDefect, that's in there now.

I'm not sure what you're asking for re wrappers though. The diff is already pretty simple, especially if your diff tool can ignore indent changes. If you want to wrap the ctor where it's invoked, that's in more than one place so the diff gets even bigger. If you want to move the guts of the ctor to a sub function that doesn't seem any less complicated to understand in a diff than just indenting. But I'm happy to do whatever you want once I understand it.

nickshulman:
I think we also need to handle this exception in ViewLibraryPepInfoList.CalculateEveryMz because that calls "ViewLibraryPepInfo.CalcMz" which calls "ViewLibraryPepInfo.GetPeptideInfo".

nickshulman:
Here's a Skyline document which causes this exception. The library has a modification on a "J" amino acid.
[InvalidAminoAcidJ.sky.zip](https://github.com/ProteoWizard/pwiz/files/15154449/InvalidAminoAcidJ.sky.zip)

Most errors in peptide sequences do not cause exceptions in the Library Explorer because "CachedLibrary.FilterInvalidLibraryEntries" would prevent it from being seen by any other code.

I think the right fix for this bug is either to make "FilterInvalidLibraryEntries" more strict so that it doesn't allow modifications on the "J" amino acid, or we should change "StaticMod.Validate" so that it does not forbid modifications on "J".

bspratt:
> I think we also need to handle this exception in ViewLibraryPepInfoList.CalculateEveryMz because that calls "ViewLibraryPepInfo.CalcMz" which calls "ViewLibraryPepInfo.GetPeptideInfo".

Good point - I suppose we just post the report window and break the loop.  In the event that an exception is thrown that is not a programming error, do we pop up a message and break the loop, or just break?

pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs
brendanx67: By "wrapper", I mean that you would create a new method called something like "void Init(...)" that gets called inside the constructor, which becomes just:
try
{
Init();
}
Catch (Exception e)
{
}

A diff tool will do much better with this kind of change, and it avoids simply indenting everything because you want to add exception handling.

bspratt: Still seems like a more violent change to me - when you move the guts of the ctor try block elsewhere you get this:

image

This strikes me as even more distracting. But again, happy to do so if that's your preference.

If this ctor had been written with exception handling in the first place, and with the form it has now, would you have felt moving most of the ctor code to a subfunction necessary or even desirable?

brendanx67: Strict believer in non-violence that I am, I will concede. I didn't originally notice it was a constructor when I made the suggestion. That does make things more complicated. I have used the pattern I described with functions needing exception handling quite successfully in the past, but a constructor with readonly variables makes life a bit more complicated.

I suppose you could create another constructor with different variables, e.g. make assignments that can't throw, and then call an inner constructor with fewer arguments wrapped in a try-catch. probably not worth it.

I just don't love seeing this kind of change in the code. Oh well. At least it now can avoid catching programming errors somewhat.

PR #2948, opened 2024-05-02
Re-enable Prosit tests

nickshulman:


chambm:
Was this just for avoiding the error on TeamCity or also on the nightly builds? If it was just the former (so code inspection would still run for example) then I think muting the test on TeamCity would have been more expedient/convenient. Those mutes can be set to expire either when the test starts passing or on a certain date.

nickshulman:
> Was this just for avoiding the error on TeamCity or also on the nightly builds? If it was just the former (so code inspection would still run for example) then I think muting the test on TeamCity would have been more expedient/convenient. Those mutes can be set to expire either when the test starts passing or on a certain date.

I didn't know about muting tests. Yes, I should have probably done that instead. Thanks!

PR #2951, opened 2024-05-06
Display document import failure message returned by LabKey

vagisha:
Job status JSON returned by LabKey contains the error message for failed Skyline document imports.  Display this message in Skyline, and write to log for command-line usage. 

An example message would be:
image


brendanx67 (approved):
Looks good to me.

PR #2952, opened 2024-05-07
Allow setting the replicate name using result file rules

nickshulman:
Added "Replicate Name" to the set of things that can be set using Result File Rules (requested by Brendan)


pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -2439,8 +2439,11 @@ public static void SetStaticModifications(Func<IList<string>, IList<string>> cha
             if (expectedErrorMessage != null)
             {
                 var dlg = WaitForOpenForm<MessageDlg>();
-                Assert.IsTrue(dlg.DetailMessage.Contains(expectedErrorMessage));
-                dlg.CancelButton.PerformClick();
+                RunUI(() =>
```
brendanx67, line 2442: I think it would be good to make this a RunDlg() or use OkDialog() to close the form just in case the next instruction tries to do something that shows a MessageDlg. This code doesn't ensure that the MessageDlg is gone before continuing on.

pwiz_tools/Skyline/TestFunctional/ReplicateNameMetadataRuleTest.cs
```diff
@@ -0,0 +1,133 @@
...
+using pwiz.Skyline.SettingsUI;
+using pwiz.SkylineTestUtil;
+
+namespace pwiz.SkylineTestFunctional
+{
+    [TestClass]
+    public class ReplicateNameMetadataRuleTest : AbstractFunctionalTest
+    {
+        [TestMethod]
+        public void TestReplicateNameMetadataRule()
```
brendanx67, line 35: Did you run this new test in parallel in all languages at least a few hundred times?

brendanx67 (approved):
Looks good. A few minor questions.

PR #2954, opened 2024-05-07
Method export for ThermoFisher Linear ion trap instrument

rita-gwen:


pwiz_tools/Skyline/Executables/BuildMethod/BuildThermoMethod/Program.cs
```diff
@@ -169,6 +172,12 @@ public static ListItem FromLine(int lineNum, string[] fields, Dictionary<int, st
                 if (curHeader.Equals("compound", StringComparison.InvariantCultureIgnoreCase))
                 {
                     item.Compound = curValue;
+                    var match = Regex.Match(curValue, @".+\(([\+|-])\)?(\d+)\).*");
```
brendanx67, line 175: Consider whether code would be simplified by using Stellar isolation list. Otherwise, comment this better to explain what is being done. And probably explain that this EXE uses the Fusion isolation list for the Stellar. (confusing)

brendanx67 (changes requested):
Almost there. Some changes suggested.

brendanx67 (approved):

PR #2957, opened 2024-05-09
Fix Agilent MS1 centroided spectra erroneously treated as SIM

nickshulman:
Fixed incorrect MS1 chromatograms from some Agilent centroided data (reported by Darshak)


chambm (approved):

nickshulman:
Actually, this fix breaks "AgilentGCEIChromatogramTest" because the isolation window is used for something related to MSe. 
I will keep investigating.

PR #2958, opened 2024-05-09
--associate-proteins-fasta argument

chambm:
- added --associate-proteins-fasta argument for associating existing …document peptides with a FASTA (i.e. without having to use --import-fasta); --import-fasta will still be used if --associate-proteins-fasta is not provided
* fixed valid value range for --associate-proteins-min-peptides (1-1000) but added error handling in AssociateProteinsDlg's MinPeptidesPerProtein property to convert a '0' value to '1'


brendanx67 (approved):
Looks good. Thanks for doing it so quickly.

PR #2959, opened 2024-05-09
Skyline: Fix loss annotation in MS/MS spectra to work for mass-only losses

brendanx67:
- Code to annotate MS/MS spectra based hide/show status on the formula keeping mass-only losses hidden always (fix requested by Dan - FragPipe team)


nickshulman (approved):
Looks good to me.

PR #2960, opened 2024-05-10
Fix outputting retention time in Sciex scheduled transition lists

nickshulman:
Fixed "retention time is always zero" in exported scheduled Sciex transition list (reported by Jeff)

Based on this support request:
https://skyline.ms/announcements/home/support/thread.view?rowId=63103

This was my best guess about how to fix this one.


rita-gwen:
I think this will mess up the method export logic. I would rather do 
```
            else   // CSV
            {
                return string.Format(@",{0},{1},{2}",
                    dp,
                    ce,
                    rt);
            }

```
in GetOptionalColumns method. Do you know if RT has to be the second column in the csv file?

nickshulman:
> in GetOptionalColumns method. Do you know if RT has to be the second column in the csv file?

That's a good question. I must confess I don't know anything about how this works (or even how to test this). Can you fix this?

brendanx67:
> I think this will mess up the method export logic. I would rather do
> 
> ```
>             else   // CSV
>             {
>                 return string.Format(@",{0},{1},{2}",
>                     dp,
>                     ce,
>                     rt);
>             }
> ```
> 
> in GetOptionalColumns method. Do you know if RT has to be the second column in the csv file?

Rita, this definitely fixes the old transition list export support. I don't see that it would break method export. Though, I think you need to fix the comment where you switched the code from using the "dwell" column for for retention time when the method is scheduled to using a new RT column.

                // If the ScheduledMethod flag was set assume that the Dwell time column in the 
                // transition list file has retention time.
                if (ScheduledMethod)
-                    retentionTime = transition.Dwell;
+                    retentionTime = transition.Rt ?? 0;

Are you positive that this is now always correct? It seems like this fix would again make it possible to use transition.Dwell. It is also never correct to use zero for RT when it is a scheduled method.

So, I think more attention is needed here. When exactly would transition.Rt be null?

brendanx67:
So maybe:

            // If the ScheduledMethod flag was set assume that the Dwell time column in the 
            // transition list file has retention time.
            if (ScheduledMethod)
                retentionTime = transition.Rt ?? transition.Dwell;

PR #2970, opened 2024-05-17
* fixed leak with EncyclopeDIA MultiProgressControl

chambm:
With the cancelability aspect of the test enabled, some DataGridViewProgressCells were not being disposed.

### DIFF
--- a/pwiz_tools/Shared/Common/Controls/DataGridViewProgressBar.cs
+++ b/pwiz_tools/Shared/Common/Controls/DataGridViewProgressBar.cs
@@ -38,6 +38,12 @@ namespace CustomProgressCell
         {
             CellTemplate = new DataGridViewProgressCell();
         }
+
+        protected override void Dispose(bool disposing)
+        {
+            CellTemplate.Dispose();
+            base.Dispose(disposing);
+        }
     }
 }
 
@@ -133,6 +139,14 @@ namespace CustomProgressCell
             Warning = 3 // yellow
         }
 
+        protected override void Dispose(bool disposing)
+        {
+            _progressBar.Dispose();
+            _animationStepTimer.Dispose();
+            _animationStopTimer.Dispose();
+            base.Dispose(disposing);
+        }
+
         private void SetProgressBarState(ProgressBarState state)
         {
             SendMessage(_progressBar.Handle, PBM_SETSTATE, (IntPtr)state, IntPtr.Zero);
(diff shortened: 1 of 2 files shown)


pwiz_tools/Shared/Common/Controls/DataGridViewProgressBar.cs
```diff
@@ -38,6 +38,12 @@ public DataGridViewProgressColumn()
         {
             CellTemplate = new DataGridViewProgressCell();
         }
+
+        protected override void Dispose(bool disposing)
+        {
+            CellTemplate.Dispose();
```
nickshulman, line 44: I think this technically needs to be:
`if (dispose) { CellTemplate.Dispose();}`
The reason is that "Dispose(false)" can be called from the finalizer, and, in that case you are not supposed interact with any other managed objects.

nickshulman (approved):
Looks good to me but I think the code that you added inside of the Dispose block needs to be surrounded by "if (disposing)"

pwiz_tools/Shared/Common/Controls/DataGridViewProgressBar.cs
chambm: That confuses me. Dispose(false) should dispose unmanaged resources. That's not supposed to include unmanaged resources held by member objects?

chambm: OK after reading about it a bit I think I understand.

nickshulman: Right. Any other managed object which might hold an unmanaged resource will have its own finalizer to handle the case where it got garbage collected before being Dispose'd.

Here's what ChatGPT said to do when "disposing" is false:
_You should only free unmanaged resources.
Do not interact with other managed objects. These objects may have already been finalized, leading to potential ObjectDisposedException or other unexpected behaviors.
You should not reference other managed objects because the finalizer does not guarantee the order in which objects are finalized, and those objects may no longer be in a valid state._

chambm: A suppose an exception would be a managed object which (improperly) doesn't have a finalizer?

nickshulman: Yes, my understanding is that if another object does not have a finalizer then it is completely safe to interact with inside of your own finalize method.

PR #2971, opened 2024-05-17
* update Shimadzu IOModule to 4.1.0.7489

chambm:
* update Shimadzu IOModule to 4.1.0.7489 (adds support for recent instruments)


nickshulman (approved):
Looks good to me.

It would be nice if there was some way to run a test which is currently being skipped. Perhaps TestRunner should always run tests which have been explicitly mentioned by name with the "test=" commandline parameter. But, that's not something to worry about now.

PR #2975, opened 2024-05-21
* fixed errors building Koina library with modified sequences

chambm:
* fixed errors building Koina library with modified sequences (reported by Phil)


chambm:
@brendanx67 Need a new daily with this fix. Koina's broken for regular library building because it'll fail for any modifications (or maybe only C+57, I'm not sure). The bug didn't affect EncyclopeDIA's use of Koina because it comes in from an alternate PredictBatches pathway. Unfortunately I assumed the older pathway was already being tested with at least some modifications but apparently this was not the case!

brendanx67:
Looking at the test change I was left wondering if C[+57] is the only
modification expected to work or should we have tests for more
modifications? e.g. M[+16]

PR #2981, opened 2024-05-23
Update tutorial audit log file for TestDiaQeDiaUmpireTutorial

nickshulman:
My computer is still generating the audit log files for other languages, but here's what the change is for English.
The following line no longer appears in the audit logs.
`Removed suffix = ".mzML"`

The reason for this change is that the Import Peptide Search wizard no longer includes the filename extension as part of the suggested replicate name as a result of part of the feature finding commit on May 6th.
In "ShowRemovePrefixDialog" it was changed to use "Path.GetFileNameWithoutExtension".


chambm:
Looks like that line was actually removed in the new logs. I'm not sure what caused that, either. :/

nickshulman:
> Looks like that line was actually removed in the new logs. I'm not sure what caused that, either. :/

Oops. You're right.
I think this failure began with the Hardklör commit on May 6th.

brendanx67:
The problem with lots of other tests failing is that you often find less
noisy test were also failing. This is a perfect example and why getting
things fixed as quickly as possible is so important.

PR #2984, opened 2024-05-25
Show warning when training mProphet model if explicit peak bounds from libraries are being used

nickshulman:
Changed to show warning message when adding decoys or training mProphet model if document uses explicit peak bounds from spectral libraries.
Changed to show warning tooltip on Transition Settings Full Scan tab if retention time filtering is used with PRM acquisition method.

Also speed up "Library.FindSource" so that it remembers its answers because EncyclopeDiaLibrary was really slow sometimes. The logic for matching MsDataFileUri's to library source file names is now in class "LibraryFiles" instead of duplicated in EncyclopeDiaLibrary, BiblioSpecLiteLibrary and ChromatogramLibrary.

Warning message for "Refine > Add Decoys":
image

Warning message using "<Add...>" or "<Edit Current...>" at "Refine > Reintegrate":
image

Warning tooltip if using retention time filtering with PRM:
image
("SureQuant" or "Targeted" may be displayed instead of "PRM").


brendanx67:
I don't love the use of "OK | Cancel" here. These are pretty broken things to do, and "OK" is the default. Also "OK" is not a valid response to "Are you sure... ?" I suppose it might be better to use "Yes" and "No", with no default or "No" as default.

Or use "Add Decoys" and "Cancel" ("Train Model" and "Cancel") with no default.

pwiz_tools/Skyline/Menus/MenusResources.resx
```diff
@@ -319,4 +319,14 @@
...
     <value>Unexpected character '{0}' found on line {1}.</value>
   </data>
+  <data name="RefineMenu_ShowGenerateDecoysDlg_Are_you_sure_you_want_to_add_decoys_to_this_document_" xml:space="preserve">
+    <value>Are you sure you want to add decoys to this document?
+    
+This document uses peak boundaries from spectral libraries. When these boundaries are used, Skyline does not perform its own peak detection.
+
+To be ready to train a peak scoring model effectively, follow these steps:
+
+1. Navigate to the "Libraries" tab at "Settings > Peptide Settings" and edit the library list.
```
brendanx67, line 329: Navigate to the "Settings > Peptide Settings - Libraries" tab, and click "Edit List".

pwiz_tools/Skyline/Menus/MenusResources.resx
```diff
@@ -319,4 +319,14 @@
...
+2. Edit the library items and uncheck the "Use explicit peak bounds" checkbox.</value>
```
brendanx67, line 330: Select a library item, click "Edit", and uncheck the "Use explicit peak bounds" checkbox.

pwiz_tools/Skyline/SettingsUI/SettingsUIResources.resx
```diff
@@ -1164,4 +1164,24 @@ information, there is no spectrum to show</value>
...
+Use this option only if your acquisition method includes multiple targets with the same m/z.</value>
+  </data>
+  <data name="EditPeakScoringModel_ExplictPeakBoundsWarning" xml:space="preserve">
+    <value>Are you sure you want to train a peak scoring model for this document?
+      
+This document uses peak boundaries from spectral libraries. When these boundaries are used, Skyline does not perform its own peak detection.
+
+Before training a peak scoring model, it is recommended that you follow these steps:
+
+1. Navigate to the "Libraries" tab at "Settings > Peptide Settings" and edit the library list.
```
brendanx67, line 1181: Make consistent with other instance of these instructions per review.

pwiz_tools/Skyline/SettingsUI/SettingsUIResources.resx
```diff
@@ -1164,4 +1164,24 @@ information, there is no spectrum to show</value>
...
+2. Edit the library items and uncheck the "Use explicit peak bounds" checkbox.
+3. Navigate to 'Edit > Manage Results' and click the 'Re-score' button.
```
brendanx67, line 1183: Use double quotes for consistency.

brendanx67 (changes requested):
Surprising that this requires no test changes. Please add testing of the new UI.

nickshulman:
> I don't love the use of "OK | Cancel" here. These are pretty broken things to do, and "OK" is the default. Also "OK" is not a valid response to "Are you sure... ?" I suppose it might be better to use "Yes" and "No", with no default or "No" as default.
> 
> Or use "Add Decoys" and "Cancel" ("Train Model" and "Cancel") with no default.

I updated the message text and the button text.

I actually can't figure out how to make it so the MultiButtonMsgDlg has no default (accept) button.

Since the form has only two buttons on it, one of those two buttons is going to have the focus, and is therefore going to respond to the Enter key.
However, I cannot even figure out how to make it so the Cancel button is the default button.
That is, even if I set "dlg.AcceptButton = dlg.CancelButton", the other button still ends up being the one that is selected when the dialog comes up ("dlg.AcceptButton = null" does not work either).

I even tried this with a simpler form ("PasteFilteredPeptidesDlg") and I could not get it to have either no default button or have it so the Cancel button is the default button.

PR #2988, opened 2024-05-28
Back out the changes to suppress these tests that were bogging down due to issues on the Uniprot server backend

bspratt:
internal test issue only


bspratt:
This commit adds a timeout mechanism to TestRunner that uses the same attribute as VSTest.

brendanx67:
Can you post text showing what a failure looks like?

bspratt:
Here I've set an absurdly short timeout of 3 seconds:
image

brendanx67:
Looks okay to me then, if it passes all of the tests in serial and parallel mode. I think this would be better than passing, but taking over 1 hour per test, as that required us to figure out which tests were having that problem. Thanks, Brian.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
```diff
@@ -271,6 +275,53 @@ private static string GetAssemblyPath(string assembly)
             return Path.Combine(runnerExeDirectory, assembly);
         }
 
+        private void RunTestWithTimeout(TestInfo test, object testObject)
+        {
+            // Signals completion of the test
+            var taskCompletedEvent = new ManualResetEvent(false);
+
+            // Run the test on a thread
+            Thread testThread = new Thread(() =>
```
chambm, line 284: I wonder whether this would cause problems with other tests that do stuff with threads. If you run every test with a timeout but make it default to a large number (so they all run in a thread like this), do they all work?

pwiz_tools/Skyline/TestPerf/PerfUniquePeptidesTest.cs
```diff
@@ -78,43 +78,43 @@ void LogToConsole(string message, [CallerLineNumber] int lineNumber = 0)
...
-        [TestMethod, NoParallelTesting(TestExclusionReason.RESOURCE_INTENSIVE), SkipTestUntil(2024, 5, 29, "UniProtKB REST API slowness issue")]
+        [TestMethod, NoParallelTesting(TestExclusionReason.RESOURCE_INTENSIVE), Timeout(TEST_TIMEOUT_MSEC)]
         public void UniquePeptides4PerfTest()
         {
             // 4)  Current background proteome not same as in new settings
             scenario(PeptideFilter.PeptideUniquenessConstraint.species, "human_and_yeast.protdb", "human_and_yeast_no_metadata.protdb");
         }
 
-        [TestMethod, NoParallelTesting(TestExclusionReason.RESOURCE_INTENSIVE)]
+        [TestMethod, NoParallelTesting(TestExclusionReason.RESOURCE_INTENSIVE), Timeout(TEST_TIMEOUT_MSEC)]
```
chambm, line 117: It would be nice if a timeout in one of these tests would cause the others get skipped (timeouts in Prosit/Koina tests would be another place that would be useful): would that be excessively complicated to implement? Perhaps some optional static flag variable in the Timeout attribute (if that's even possible)?

bspratt: Interesting idea, but probably a different attribute - I'm trying to match the standard timeout behavior in Test Explorer.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
bspratt: Yeah, something in there is unhappy:
image
I'll try identifying the existing test thread rather than nesting it.

bspratt: (That was with 8 parallel workers, FWIW)

chambm: I considered that idea of killing the existing test thread but I'm skeptical that you can kill the existing test thread without breaking the RunTests flow. It's worth a try though. I think that DragDrop error probably comes from no longer running on the UI thread and something didn't InvokeRequired/Invoke when it perhaps should be.

chambm: Obviously MSTest/VSTest is able to do UI tests with a timeout of course, but perhaps it does quite a bit of extra thread voodoo when a timeout is used.

PR #2989, opened 2024-05-28
Bug fix in PanoramaClient - catch errors when async file upload completes

vagisha:
If there is an error during asynchronous file upload, trying to access the [Result ](https://learn.microsoft.com/en-us/dotnet/api/system.net.uploadfilecompletedeventargs.result?view=net-9.0#system-net-uploadfilecompletedeventargs-result)property on [UploadFileCompletedEventArgs ](https://learn.microsoft.com/en-us/dotnet/api/system.net.uploadfilecompletedeventargs?view=net-9.0)throws an exception.  Check the Error and Cancelled properties first to determine if the upload request completed correctly.

I found this problem when I tried uploading a Skyline document to a folder on my dev machine.  I had changed the permissions on the @files directory to be read-only earlier when I was testing something else.  The upload caused Skyline to crash because PanoramaClient tried to access the Result property to parse the server response when an Error (502 Bad Gateway) was set on UploadFileCompletedEventArgs. 

This fix addresses that problem.  But I was surprised to see the 502 error.  I had expected a successful (200) response from the server with the actual LabKey error ("File could not be created on the server") in the JSON response.  This is what I had seen when I had tested this scenario about a month ago for this PR (https://github.com/ProteoWizard/pwiz/pull/2855). 

What I found was that if my test file is small (~4MB) the Response property contains the LabKey response.  But larger files cause a 502 error on the server, and the Response property is not valid.  I will try to figure out why.


brendanx67 (approved):
Looks okay to me.

PR #3000, opened 2024-06-05
Fixed the negative RT export for all Agilent instruments

rita-gwen:


pwiz_tools/Skyline/Model/Export.cs
```diff
@@ -3504,6 +3508,7 @@ public static string GetTargetedHeader(char fieldSeparator)
                     SchedulingReplicateIndex, SchedulingAlgorithm, false, out var windowRT);
                 if (predictedRT.HasValue)
                 {
+                    predictedRT = Math.Max(predictedRT.Value, windowRT + 0.1);
                     retentionTime = (RetentionTimeRegression.GetRetentionTimeDisplay(predictedRT) ?? 0).ToString(CultureInfo);  // Ret. Time (min)
                     deltaRetentionTime = Math.Round(windowRT, 1).ToString(CultureInfo); // Delta Ret. Time (min)
```
brendanx67, line 3513: Yeah. This "Delta Ret. Time (min)" does sound like it should be 1/2 windowRT, and not windowRT, which is a true "window" extending from predictedRT - windowRT/2 to predictedRT + windowRT/2.

pwiz_tools/Skyline/TestFunctional/ExportIsolationListTest.cs
```diff
@@ -467,6 +473,60 @@ private string FieldSeparate(params object[] valuesIn)
             }
             return sb.ToString();
         }
+
```
brendanx67, line 476: Thanks for adding a test. Let's make sure we get the Delta RT field right, and I see another opportunity to use your constant instead of 0.1 as a magic value below.

brendanx67 (changes requested):
Not ready to go yet. Most importantly, we need to understand whether we have ever correctly output the retention time window value to the Agilent field that is supposed to be getting window/2.

brendanx67 (approved):
Looks good to me. Thanks.

PR #3001, opened 2024-06-07
* added dotCover support to parallel testing mode

chambm:
* added dotCover support to parallel testing mode (coverage=on/off and dotcoverexe path relative to pwiz root)
* changed host worker to run as a child process instead of an extra thread in the main TestRunner process
* changed ReportExceptionForRetry() to not do a delay on the first retry
* added TestRunner checkdocker parameter to allow skipping the check that the current always_up_runner image can actually be run

With 6 workers, I can run coverage on `test=Test.dll,TestData.dll,TestTutorial.dll,TestFunctional.dll,TestConnected.dll` in about 11 minutes, compared to about 60 minutes for running dotCover in Visual Studio, and 9 minutes for running those same tests without coverage!). The coverage results are very similar. There's also more that could be done with comparing current coverage to the last successful snapshot (kind of like what TeamCity does), and maybe opening the snapshot in Visual Studio to provide a more interactive experience.

Here's a trimmed down zip for the dotCover. You can unzip it to the pwiz root and add `coverage=1` to the TestRunner command. There may be a better way to include the command-line tools, like downloading this zip on the fly.
[jetbrains.dotcover.commandlinetools.zip](https://github.com/user-attachments/files/15743331/jetbrains.dotcover.commandlinetools.zip)


pwiz_tools/Skyline/TestRunner/Program.cs
```diff
@@ -439,6 +444,9 @@ static int Main(string[] args)
 
             if (commandLineArgs.ArgAsBool("wait"))
             {
+                GC.Collect();
```
nickshulman, line 447: What exactly is this "GC.Collect" for?
If doing a garbage collection improves things, it might mean we're doing something wrong somewhere else.
Was there a particular test that you added this for? Was there a particular file on disk that was being locked?

The last time I had to debug something like this, the "pending finalizers" list in dotMemory was very helpful.

chambm: This is for the "wait until user presses a button" step at the end of TestRunner execution. I used it for leak finding in a test in a loop and collecting a snapshot at the end before the process exits. It doesn't have anything to do with the coverage feature though.

nickshulman (approved):
Looks good to me (but I don't actually understand this code)

chambm:
Did you try running it? I had some trouble with non-deterministic failures where Docker workers would spontaneously quit due to an I/O error, but it seemed to go away after rebooting. Still not sure what was up with that.

chambm:
@brendanx67 do you mind if I merge this before feature freeze? It's got some good improvements for parallel testing and the coverage support shouldn't have an effect if you don't configure it.

brendanx67:
Sure. Go ahead. Fingers crossed 🤞

PR #3007, opened 2024-06-11
Speed up TestRescoreInPlace

nickshulman:
Make "TestRescoreInPlace" run a little bit faster.


pwiz_tools/Skyline/Program.cs
```diff
@@ -652,6 +652,7 @@ private static void ReportExceptionUI(Exception exception, StackTrace stackTrace
 
         public static void AddTestException(Exception exception)
         {
+            Trace.TraceError(@"AddTestException: {0}", exception);
```
brendanx67, line 655: How does this look in the TestRunner output? It seems to me that it would cause double output of exceptions. Let's have a quick Zoom where you demo it to me with a forced failure.

nickshulman: Yes, it probably will result in double output.
I was planning on worrying about that after we see what it looks like in a nightly run and figure out whether an exception is the cause of the hang we are investigating.

brendanx67: Before you came up with this, I was thinking the TestFunctional class could just silently maintain its own list of the exceptions that I could easily look at in the debugger. That solution seems like it may be as valuable as this one without impacting anything visible to anyone not running a debugger.

brendanx67: The problem in the debugger was not being able to get to the Program class global variable.

.gitignore
```diff
@@ -380,6 +380,7 @@ Version.cpp
 /pwiz_tools/Bumbershoot/**/AssemblyInfo.cs
 /pwiz_tools/Bumbershoot/idpicker/Resources/Resources.*
 /pwiz_tools/Skyline/Executables/SkylineBatch/SkylineBatch/Properties/AssemblyInfo.cs
+/pwiz_tools/Skyline/Model/Koina/Config/KoinaConfig.xml
```
brendanx67, line 383: This is being fixed in PR#3004 by changing the Koina/Config/.gitignore which was referencing only PrositConfig files.

pwiz_tools/Skyline/Program.cs
nickshulman: I was hoping that if the hang was caused by an exception that we would see that exception in the log file without having to attach a debugger.
If we don't mind attaching a debugger for this sort of hang then I'll revert this change.

brendanx67 (approved):
Looks good. Thanks.

PR #3009, opened 2024-06-12
Display values for "Normalized Area" etc. in Document Grid even when some transitions are truncated

nickshulman:
This change makes it so that values like "Normalized Area", "Protein Abundance", etc. have sub-properties "Raw", "Strict" and "Message" which have different values depending on whether there was a problem calculating the value.
image

If there are missing or truncated transitions, the "Raw" value will have the best value that Skyline was able to calculate, whereas "Strict" will be null (which is displayed as "#N/A" in the Document Grid).
Whenever "Strict" is null, the "Message" property will display the reason.

When the parent property ("Normalized Area", etc) is displayed on the Document Grid, it will have "*" prepended to its value and there will be a red exclamation mark that you can hover over to see the message.
image

If there is a problem calculating a value on the Calibration Curve, it will still be displayed but it will always be excluded and drawn without being filled in. If the selected replicate has a problem the message will be displayed.
image

Also, this change adds "Protein Abundance Transition Summed" value which is the regular protein abundance multiplied by the number of Transitions, and "Protein Abundance Transition Averaged" value which is calculated in the old way.
image


brendanx67:
How about All, Strict, and Annotated as the modifier names?

nickshulman:
> How about All, Strict, and Annotated as the modifier names?

The "Annotated" value is actually the parent column-- it displays the value along with a tooltip.
The "Message" sub-property is just the text of the tooltip or blank if there is no message. The reason it is there is because you can't see the tooltip when you export to CSV-- the tooltip is just represented by a "*".

I understand how "All" makes sense as a column header, since if you look in that column you will see all the values, whereas if you look in the "Strict" column you will only see a subset of the values, but I wish we could come up with a word that makes more sense as a description of a single value, since the column headers are always singular nouns.

brendanx67:
Unfiltered?

PR #3020, opened 2024-06-19
Fix relative abundance formatting

nickshulman:
Changed to remember peak area relative abundance formatting settings in the Skyline document.
Changed protein abundance value on relative abundance plot to be sum of transitions instead of average (requested by Mike)


brendanx67 (approved):
Looks good. Thanks, Nick.

PR #3030, opened 2024-06-21
Fix hang in ProteinMetadataManager

nickshulman:
Fixed potential hang extracting chromatograms when some protein groups have proteins with nonstandard accession numbers (reported by Mike)

The problem is that when some protein groups have invented accession numbers, a partially-searched ProteinGroupMetadata might be put into ProteinMetadataManager._processedNodes. When that happens, then incompletely searched protein groups in the document never get searched to completion, and ProteinMetadataManager keeps saying it has more work to do but just calls the very expensive method "SrmDocument.ChangeChildrenChecked" with nothing new.

This change makes it so it checks whether the thing in "_processedNodes" is only partially completed, and, if so, makes sure to queue up requests for the unsearched parts of the group.

(This diff is more simple if you hide whitespace changes)

### DIFF
--- a/pwiz_tools/Skyline/Model/Proteome/ProteinMetadataManager.cs
+++ b/pwiz_tools/Skyline/Model/Proteome/ProteinMetadataManager.cs
@@ -211,45 +211,68 @@ namespace pwiz.Skyline.Model.Proteome
                         var proteinsToSearch = new List<ProteinSearchInfo>(); 
                         foreach (PeptideGroupDocNode node in docOrig.PeptideGroups)
                         {
-                            if (node.ProteinMetadata.NeedsSearch() && !_processedNodes.ContainsKey(node.Id.GlobalIndex)) // Did we already process this?
+                            if (!node.ProteinMetadata.NeedsSearch())
                             {
-                                void ProcessNode(FastaSequence seq, ProteinMetadata proteinMetadataOrGroup, int i, int nResolved2, ref IProgressStatus progressStatus2)
+                                continue;
+                            }
+
+                            var completedNames = new HashSet<string>();
+                            if (_processedNodes.TryGetValue(node.Id.GlobalIndex, out var processedNode))
+                            {
+                                if (!processedNode.NeedsSearch())
                                 {
-                                    var proteinMetadata = proteinMetadataOrGroup.ProteinMetadataList[i];
-                                    if (proteinMetadata.WebSearchInfo.IsEmpty()) // Never even been hit with regex
+                                    continue;
+                                }
+
+                                foreach (var part in processedNode.ProteinMetadataList)
+                                {
+                                    if (!part.NeedsSearch())
                                     {
-                                        // Use Regexes to get some metadata, and a search term
-                                        var parsedProteinMetaData = FastaImporter.ParseProteinMetaData(proteinMetadata);
-                                        if ((parsedProteinMetaData == null) || Equals(parsedProteinMetaData.Merge(proteinMetadata), proteinMetadata.SetWebSearchCompleted()))
-                                        {
-                                            // That didn't parse well enough to make a search term, or didn't add any new info - just set it as searched so we don't keep trying
-                                            if (_processedNodes.TryGetValue(node.Id.GlobalIndex, out var processedProteinGroupMetadata))
-                                                _processedNodes[node.Id.GlobalIndex] = processedProteinGroupMetadata.ChangeSingleProteinMetadata(proteinMetadata.SetWebSearchCompleted());
-                                            else
-                                                _processedNodes.Add(node.Id.GlobalIndex, proteinMetadataOrGroup.ChangeSingleProteinMetadata(proteinMetadata.SetWebSearchCompleted()));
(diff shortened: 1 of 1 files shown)


pwiz_tools/Skyline/Model/Proteome/ProteinMetadataManager.cs
```diff
@@ -211,46 +211,69 @@ void CheckBackgroundProteome(FastaSequence seq, ProteinMetadata proteinMetadataO
...
                         foreach (PeptideGroupDocNode node in docOrig.PeptideGroups)
                         {
-                            if (node.ProteinMetadata.NeedsSearch() && !_processedNodes.ContainsKey(node.Id.GlobalIndex)) // Did we already process this?
+                            if (!node.ProteinMetadata.NeedsSearch())
                             {
-                                void ProcessNode(FastaSequence seq, ProteinMetadata proteinMetadataOrGroup, int i, int nResolved2, ref IProgressStatus progressStatus2)
+                                continue;
+                            }
+
+                            var completedNames = new HashSet<string>();
```
bspratt, line 219: Relying on just the name makes me nervous - conceivably ambiguous. We really ought to be hashing all available attributes even if many are empty pre-search.

nickshulman: These names are only used to decide which parts of a ProteinGroupMetadata are not finished.
The name only needs to be unique in the ProteinGroupMetadata.

The problem was that ProteinGroupMetadata objects were being put into "_processedNodes" where some but not all of the individual ProteinMetadata's inside were finished retrieving their metadata.
The code was assuming that anything that was in "_processedNodes" was finished, and so some PeptideGroupDocNode objects would be forever in a state of thinking they needed metadata fetched, but not metadata fetching would ever happen.

bspratt: >> The name only needs to be unique in the ProteinGroupMetadata.

OK, but is that guaranteed? Or is it the case that Description, Species etc are never populated in this context? (If so, a comment to that effect would be good)

nickshulman: That is an interesting question.
"ProteinGroupMetadata.ChangeSingleProteinMetadata" only looks at the Name when it is told to update part of itself:
```
        public override ProteinMetadata ChangeSingleProteinMetadata(ProteinMetadata singleProteinMetadata)
        {
            Assume.IsTrue(singleProteinMetadata?.ProteinMetadataList?.Count == 1);
            return new ProteinGroupMetadata(ProteinMetadataList.ReplaceElement(singleProteinMetadata, m => m.Name == singleProteinMetadata!.Name));
        }
```
If it were possible that a `ProteinGroupMetadata `might contain two `ProteinMetadata`'s with the same Name, then we might be in a state where the ProteinGroupMetadata could never be marked as completed. I'll see if I can get that to happen.

bspratt: Seems like one could just compare hashes based on all the settable attributes, for best safety.

chambm:
IMO it should be an error to have multiple proteins with the same name in the same document, much less in the same protein group. How the heck could users tell what node a name was referring to if it's not unique? RenameProteinsDlg.UseFastaFile() for example will throw an error if multiple proteins have the same name. I don't know how strictly we enforce that in other places that proteins are imported though.

bspratt:
> IMO it should be an error to have multiple proteins with the same name in the same document, much less in the same protein group.

Same name in same group seems bad, yes, but is there no use case for grouping proteins in various ways?

pwiz_tools/Skyline/Model/Proteome/ProteinMetadataManager.cs
nickshulman: Brian, it sounds like you are worried about the scenario of what happens if the user changes the protein metadata in between the time that the metadata is requested and the response comes back. The code that protects against that is still intact down on line 324.

The fix that I am proposing here is that, if a protein group has multiple parts that Skyline has to make sure to mark all of the parts of the protein metadata as so ProteinMetadataManager always thinks it still has work to do.

If we want to go with the minimal code change, the other part of this pull request:
```
               if (ArrayUtil.ReferencesEqual(listProteins, docOrig.MoleculeGroups.ToList()))
                {
                    return null;
                }
```
would be sufficient to fix the hang.

ProteinMetadataManager would still think it has work to do, but, since it does not keep calling "SrmDocument.ChangeChildren" it won't appear to the user like a hang.

bspratt: My concern was just that Name could possibly be non-unique across proteins in a group, even though as Matt points out that would be weird and maybe even wrong. But I'd expect that if there was a Name conflict there would be disambiguation in some of the other attributes, even before full metadata resolution.

 I'm not familiar enough with the Protein Groups functionality to comment further.

PR #3032, opened 2024-06-24
Revert to English in .ja and .zh-CHS resx files when English text has changed

nickshulman:
This change makes it so that entries in the .ja.resx and .zh-CHS.resx files have the English text for all of the cases where the English text has changed between 23.1 and 24.1.
It also ensures that the order of elements in the .resx files are identical.


nickshulman:
This pull request has been abandoned in favor or PR #3040

PR #3033, opened 2024-06-24
Skyline: UI re-organization of BuildLibraryDlg to support integration with Carafe

brendanx67:
This is a new design for the BuildLibraryDlg with more flexible panel layout that is easier to design using the WizardPages class, and which creates a new final page for Carafe that allows a user to specify transfer learning data sources.

I believe it still passes all of the existing tests and may be a useful starting point for Jason later as he integrates Carafe.


brendanx67:
This work seems ready. It now provides comparable UI and passes all tests, and it would be necessary for adding either or both AlphaPeptDeep or/and Carafe which we know we want to do.

PR #3035, opened 2024-06-24
Normalize whitespace in .resx files

nickshulman:
Change encoding to UTF-8 no byte order mark
Replace tabs in comments with two spaces
Remove text nodes from top-level elements and reformat XML

The goal is to make the diff in PR #3032 smaller


brendanx67 (approved):
Looks okay to me.

PR #3037, opened 2024-06-24
Replace "<value />" with "<value></value>" in resx files

nickshulman:
Use "`<value></value>`" instead of "`<value />`" for empty values in .resx files


chambm:
Will Visual Studio change these? How did they get to be `<value/>` in the first place?

nickshulman:
> Will Visual Studio change these? How did they get to be `<value/>` in the first place?

They got changed to `<value/>` by a tool that I wrote in PR #3035. We are pretty sure that Visual Studio always writes them out as  `<value></value>` so I changed them back to the way they were before.

PR #3039, opened 2024-06-26
Tool for managing localized resources

nickshulman:
Command line tool that reads all .resx files into a SQLite database, can figure out which strings have changed between two databases and output .resx files.


chambm:
Impressive! Would it be possible to use nuget to get the binaries so we don't need them in the repo?

PR #3041, opened 2024-06-27
Parallelize feature finding

bspratt:
Refactor our Hardklor-based feature finding to operate in parallel


pwiz_tools/Skyline/FileUI/PeptideSearch/HardklorSearchControl.cs
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        public override UpdateProgressResponse UpdateProgress(IProgressStatus status)
+        {
+            if (IsCanceled)
+            {
+                return UpdateProgressResponse.cancel;
+            }
+
+            lock (_status)
```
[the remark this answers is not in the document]
bspratt: Thanks for eyes on this. I'll try lock (this.progressBar) instead.

pwiz_tools/Skyline/FileUI/PeptideSearch/HardklorSearchControl.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            if (IsCanceled)
+            {
+                return UpdateProgressResponse.cancel;
+            }
+
+            lock (progressBar)
+            {
+                _status = _status.ChangeMessage(status.Message).ChangePercentComplete((100 * _currentStep++) / _totalSteps);
+                Invoke(new MethodInvoker(() => UpdateSearchEngineProgress(_status)));
```
brendanx67, line 97: I don't love seeing Invoke() inside a lock {}. Invoke is inherently single-threaded because it blocks until the code can be run on the UI thread, which is a single thread. Maybe the lock should go only around the update to _status and it should be on "this" because it is inherently about keeping _currentStep, _totalSteps, and _status in synch. If those are in synch, I don't think you need any further synchronization on UpdateSearchEngineProgress().

In fact, I would generally recommend against forcing an update on the UI thread for every change to _status. Instead, you want updates to happen based on a timer that fires on the UI thread and checks the current _status whatever it is. Status updates are about letting the user know about progress, which the user only needs every so often, e.g. 200 ms. I have seen real performance issues that came down to code like this where every status update forces a UI update.

Even if it is not a problem here, writing status update code like this forces the programmer to consider how often they need to update status to balance user need to see some progress and the performance consideration that UI updates can be much slower than the processing they are reporting on.

bspratt: Thanks, I'll have to think about that. I was following a pattern from elsewhere in our code.

FWIW these status updates don't happen very often - just at the end of each completed process (msconvert, hardklor etc)

chambm: Can't it just use BeginInvoke instead?

bspratt: > Can't it just use BeginInvoke instead?

I'll give that a try, thanks.

brendanx67: If you are confident about the timing issue, then I think you can just move the Invoke outside of the lock. I don't see why it needs to be inside the lock.

The problem with putting it in the lock is that if the UI thread somehow triggers this code and stops at the lock, and while a background thread is blocked on the Invoke() then you get a deadlock.

If you feel everything else after your Invoke call can happen without waiting for it to complete, you can use BeginInvoke() here, but that also seems to be to indicate the Invoke() doesn't need to be in the lock {}. i.e. it means that the what happens in BeginInvoke() doesn't need to happen in the lock. Does the queueing of this work need to happen in the lock? I don't see why.

brendanx67: I would use lock(this) and have it only contain the update to _status, which clearly involved multiple member variables you are trying to keep in sync. The update to the UI could happen at any time as long as the _status is coherent.

And, I would still say that in general status UI should be implemented as a timer checking for status updates rather than pushing the UI in real-time from a background thread. But, point taken that this may be good enough for the processing in question.

bspratt: OK, I'll switch the lock to (this)

bspratt:
Thanks for the reviews - this is working well with the exception if the RT alignment, which has an N-squared problem as currently implemented - it bogs down as more replicates are added. Nick is having a look at that.

bspratt:
I spent some time with a profiler, RT alignment is now working well enough that it's not currently the bottleneck in the CoralEggs 12 replicate data set.

PR #3054, opened 2024-07-02
Skyline/work/20240701 rc updates 24 1

brendanx67:
Skyline (24.1): Format version updates to prepare for release


nickshulman (approved):
Looks good to me.

It is not possible to see it in this pull request, but git is smart enough to realize that "Skyline_24.1.xsd" is identical to a file which is currently in the repository, so, if you look at its history and set "Walk behavior" to "follow renames" you will see the long history of that file.

PR #3056, opened 2024-07-03
Label drag box implementation

rita-gwen:


rita-gwen:
Brendan, could you have a look at the functionality ? I tested it as much as I could, but another pair of eyes would be nice.

pwiz_tools/Shared/zedgraph/ZedGraph/GraphPane.cs
```diff
@@ -1549,39 +1559,64 @@ public void AdjustLabelSpacings(List<LabeledPoint> labPoints, Control parentCont
...
                 if (visiblePoints.Any())
                 {
-                    GraphObjList.RemoveAll(obj => obj is BoxObj || obj is LineObj);
+                    var savedLayout = existingLayout ?? new List<LabeledPoint.PointLayout>();
+					var newPoints = new List<LabeledPoint>();
                     foreach (var point in visiblePoints)
                     {
-                        if (_labelLayout.PlaceLabel(point, g))
+                        // if we already have a saved point layout for this peptide we do not need to re-adjust it
+						// CONSIDER: There is no stable persistent protein ID, so we cannot match the points using identityPath
```
nickshulman, line 1571: In theory, you can get a persistent ID from an IdentityPath by doing:
`new ElementRefs(srmDocument).GetNodeRef(identityPath)`

brendanx67:
I am wondering if this really belongs in the .sky file where small changes
need to go into the audit log, and will get put on the undo stack. Maybe we
should be saving it to the .sky.view file, which just gets snapshotted at
save time and stores other view state like the selection and scroll
position of the Targets view.

nickshulman:
> Maybe we should be saving it to the .sky.view file, which just gets snapshotted at save time and stores other view state like the selection and scroll position of the Targets view.

The positions of the labels on the graph are being saved in the .sky.view file.
In this pull request, the method "GraphSummary.GetPersistentString()" has been augmented to ask the GraphPane if it has any more information to save in the .view file.

There's also the "RelativeAbundanceFormatting" which does get saved in the document, but that's the things like the list of regular expressions controlling colors.

brendanx67:
Okay. Great. That is what I was hoping for. The formatting controlled by
the formatting editor saved in the .sky file, and the more nuanced label
position adjustments saved in the .sky.view. Thanks for confirming that.

rita-gwen:
Well,  it is saved into the .view file. Initially,  I planned to save it
into the document,  but then a change in the layout causes the document
change and a series of unwanted UI refreshes that are hard to suppress.

chambm:
Failing with:
Z:\pwiz\pwiz_tools\Skyline\SkylineTester\SkylineTesterWindow.Designer.cs(3596,17): error CS0104: 'ToolTip' is an ambiguous reference between 'System.Windows.Forms.ToolTip' and 'ZedGraph.ToolTip' [Z:\pwiz\pwiz_tools\Skyline\SkylineTester\SkylineTester.csproj]

PR #3062, opened 2024-07-09
Read CCS values from .msp formatted spectral library files

bspratt:
Adding support for reading CCS values in .msp files, as keyword "CCS" or "CCS_SqA"

requested by user Ryan


bspratt:
The user has been provided with a custom build, reports this is working well

pwiz_tools/Skyline/Model/Lib/NistLibSpec.cs
```diff
@@ -1789,6 +1823,91 @@ public override bool TryGetIrts(out LibraryRetentionTimes retentionTimes)
...
+                ionMobilities = new[]{ _libraryEntries[i].IonMobility};
+                return true;
+            }
+
+            return base.TryGetIonMobilityInfos(key, filePath, out ionMobilities);
+        }
+
+        public override bool TryGetIonMobilityInfos(LibKey[] targetIons, MsDataFileUri filePath, out LibraryIonMobilityInfo ionMobilities)
+        {
+            ILookup<LibKey, IonMobilityAndCCS[]> ionMobilitiesLookup;
```
nickshulman, line 1840: I think this should just be:
`ILookup<LibKey, IonMobilityAndCCS>`
without the "[]" after "IonMobilityAndCCS".

Lookups are always a map between a single key and a list of values.
I don't think you want a list of list of values.

This code seems to have been copied from BiblioSpecLite.cs.

nickshulman (commented):
If "IonMobility" were a member of "ICachedSpectrumInfo", would you be able to move the implementation of "TryGetIonMobilityInfos" into "CachedLibrary" so that "BiblioSpecLiteLibrary" and "NistLibraryBase" would be able to use the same implementation?

bspratt:
> If "IonMobility" were a member of "ICachedSpectrumInfo", would you be able to move the implementation of "TryGetIonMobilityInfos" into "CachedLibrary" so that "BiblioSpecLiteLibrary" and "NistLibraryBase" would be able to use the same implementation?

Maybe? The big difference between NIST and BiblioSpec is that NIST doesn't refer to more than one input source, and so has no need for per-file RT and IM values and no sense of "best".

brendanx67:
In other words, it stores no redundant information. Though it may come from
many, many more input sources they are not referenced. And it consists
entirely of non-redundant canonical (or best) values.

In BiblioSpec the non-redundant library refers only to the non-redundancy
of spectra. I NIST equivalent of any BiblioSpec library could be produced
by discarding all redundant information and keeping only the values marked
as “best”, which would also remove the need for specifying “best” on the
remaining values.

bspratt:
> In other words, it stores no redundant information. Though it may come from many, many more input sources they are not referenced. And it consists entirely of non-redundant canonical (or best) values. In BiblioSpec the non-redundant library refers only to the non-redundancy of spectra. I NIST equivalent of any BiblioSpec library could be produced by discarding all redundant information and keeping only the values marked as “best”, which would also remove the need for specifying “best” on the remaining values.


Exactly so. It doesn't look to me like ICachedSpectrumInfo wants to deal with redundancy.

PR #3063, opened 2024-07-10
- fixed search errors from Search control to be more specific

chambm:
- fixed search errors from Search control to be more specific (the main message from the Exception is put in the message box instead of just "Error")
* added TestDdaSearchMsFraggerBadFasta to check for specific error message
* added special handling for "process cannot access the file 'X' because it is being used by another process" in functional tests: now the locking process will be reported

### DIFF
--- a/pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs
@@ -21,6 +21,8 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Reflection;
+using System.Text;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using pwiz.Common.Chemistry;
 using pwiz.Skyline.Alerts;
@@ -45,8 +47,9 @@ namespace pwiz.SkylineTestFunctional
     [TestClass]
     public class DdaSearchTest : AbstractFunctionalTest
     {
-        public struct ExpectedResults
+        public class ExpectedResults
         {
+            public Exception expectedException;
             public int proteinCount;
             public int peptideCount;
             public int precursorCount;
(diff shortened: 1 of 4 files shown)


pwiz_tools/Skyline/TestUtil/AbstractUnitTestEx.cs
```diff
@@ -199,5 +203,25 @@ public static bool HasKoinaServer()
...
             return !string.IsNullOrEmpty(KoinaConfig.GetKoinaConfig().Server);
         }
+
+        /// <summary>
+        /// Get a system resource string. Useful when checking that a test throws an expected error message but the text comes from the OS or .NET.
+        /// Use an ID from https://github.com/ng256/Tools/blob/main/MscorlibMessage.md
+        /// </summary>
+        /// <example>GetSystemResourceString("IO.FileNotFound_FileName", "SomeFilepath")</example>
+        public string GetSystemResourceString(string resourceId, params object[] args)
+        {
```
chambm, line 213: Did I miss an existing function in Skyline to do this?

brendanx67: Probably not. This is a great addition, but I doubt we have been testing this specifically for messages thrown by .NET

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -1820,6 +1823,27 @@ private void WaitForSkyline()
             }
             catch (Exception x)
             {
+                // if it's a file locking issue, wrap the exception to report the locking process
```
brendanx67, line 1826: Nice. Indeed, seems amazing that we've never implemented this.

pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs
```diff
@@ -45,8 +45,9 @@ namespace pwiz.SkylineTestFunctional
     [TestClass]
     public class DdaSearchTest : AbstractFunctionalTest
     {
-        public struct ExpectedResults
+        public class ExpectedResults
         {
+            public Exception expectedException;
```
brendanx67, line 50: I guess we can leave this for consistency, but it is non-standard naming to have camel-case public member variables. Our standard style is title-case for all of the public interface. We use camel-case preceded by an underscore for private member variables:

private string _privateMember;

brendanx67 (approved):
Looks good to me. Thanks for the thorough testing.

pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs
chambm: Yes, might change all those to title case at some point. Probably not this PR since it'd just add noise.

PR #3065, opened 2024-07-11
Fix AlertDlg small molecule mode UI translation

nickshulman:
Fix bug where message boxes in small molecule mode would say "Peptide" instead of "Molecule" (reported by Will)


bspratt (approved):
Looks good, thanks!

brendanx67 (approved):

PR #3066, opened 2024-07-11
fix unexpected behavior with ion mobility values in spectral libraries (and no ion mobility library)

bspratt:
SrmDocument.Settings.GetIonMobilities(LibKey[] targetIons, MsDataFileUri filePath) wasn't retrieving ion mobility values from spectral libraries unless filePath was specified - in the unspecified case it should be taking the median value for all files referenced in the library


bspratt:
image

brendanx67 (changes requested):
Can you add a test to prove what was broken now works?

bspratt:
> Can you add a test to prove what was broken now works?

Done.

bspratt:
I'd like to backport this to release.

brendanx67 (approved):
Looks good. Okay to cherry pick to release.

PR #3072, opened 2024-07-12
Fix missing small molecule chromatograms

nickshulman:
Fixed missing chromatograms for small molecules in some SRM documents (reported by Jess, recently introduced in Skyline-daily)

I will write a unit test for this if I get permission to use the data files.


brendanx67 (approved):
Can we add a test that would have failed when we broke this recently?

PR #3083, opened 2024-07-17
Make "Auto Zoom > Best Peak" fall back to "Auto Zoom > None" if there is no best peak

nickshulman:
Changed so "Auto Zoom > Best Peak" zooms out to the extracted chromatogram if there is no best peak (requested by Mike)

Mike has been frustrated by the fact that if "Auto Zoom X-Axis > Best Peak" is selected, and there is no chosen peak, the chromatogram graph just remains zoomed to wherever it was when the previous peptide was selected. If the chromatogram was extracted for only a certain number of minutes around the MS/MS IDs then the chromatogram is often zoomed to a time range where there is no data.

This change makes it so that if there is no chosen peak, and "Auto Zoom > Best Peak" was selected, then it behaves exactly as if "Auto Zoom > None" was chosen, which means that the graph is zoomed to the time range of the extracted chromatogram.

This also makes it so that if the peptide is missing a chosen peak and/or a predicted retention time, and "Auto Zoom > Both" is chosen, then it ends up behaving like "Auto Zoom > Best Peak", or "Auto Zoom > Predicted" or "Auto Zoom > None" depending on what values are available.

This fix is different from what Mike has been asking for.
Mike would prefer that Auto Zoom > Best Peak would zoom to the average retention time across all of the replicates if a particular replicate is missing its peak. We could certainly implement something like that, but it does seem too complicated.


brendanx67:
Agreed. We should just teach Mike how to use Synchronize Zooming better, which would likely make him desire what he is asking for less.

PR #3087, opened 2024-07-18
Read CID/EAD mode from wiff2 files

david-cox-sciex:
Use DataAPI for wiff2 to get detailed experiment information which includes fragmentation mode for spectra.


pwiz_aux/msrc/utility/vendor_api/ABI/WiffFile2.ipp
```diff
@@ -831,8 +840,7 @@ void WiffFile2Impl::setSample(int sample) const
             IList<ISample^>^ unwrappedAllSamples = allSamples;
             this->msSample = (ISample^)unwrappedAllSamples[sample - 1];
 
-            auto experimentRequest = DataReader()->RequestFactory->CreateExperimentsReadRequest();
-            experimentRequest->SampleId = msSample->Id;
+            auto experimentRequest = DataReader()->RequestFactory->CreateExperimentsReadRequest(msSample->Id, true);
```
chambm, line 843: What does the true parameter do?

chambm (changes requested):
Looks good except tabs need to be converted to spaces. Can this wait for the EAD energy term to be approved and the CV updated in pwiz? I will do the latter although it may be more involved than usual because it looks like they've been messing around with embedding other ontologies inside the OBO.

pwiz_aux/msrc/utility/vendor_api/ABI/WiffFile2.ipp
david-cox-sciex: It retrieves detailed experiment info, such as fragmentation type and electron kinetic energy.
This line should be called only once per sample, so I think the impact to performance will be very small.

IExperimentsReadRequest CreateExperimentsReadRequest(
	string sampleId,
	bool readDetailedParameters
)

david-cox-sciex:
I'll take care of the tabs to spaces now. I think the EAD term is now in psi-ms.

david-cox-sciex:
@chambm 
I converted tabs to spaces.

Should I run the update_cv.bat, and then include the file changes it generates? I see it bringing in the electron beam energy term, it also makes some other changes to obo file that I think are okay but I'm not certain.

david-cox-sciex:
Yes. I run this and get a functioning MSConvertGUI.exe
quickbuild.bat --i-agree-to-the-vendor-licenses address-model=64 debug

chambm (approved):

chambm:
One thing that's missing here is a new test data set. Do you think you can get a tiny WIFF/WIFF2 set with a few spectra? One neat feature of Thermo QualBrowser I discovered is it can export a single spectrum as a RAW file. The spectrum metadata is preserved although file-level metadata is lost. But that's still useful for testing that different spectrum types are converted properly! I'm not sure if Sciex software has any analogous capability.

david-cox-sciex:
I have a ~15 MB file I've been using as a manual test. 
Can you point me at where the tests are defined?

chambm:
Reader_ABI_Test.cpp . All WIFF and WIFF2 files in the .data directory are tested. 15MB is rather big though. Any way to get it down to just a few spectra?

david-cox-sciex:
that is the smallest one I can find. 

Should I put these files in 
\pwiz\data\vendor_readers\ABI\Reader_ABI_Test.data

[7600ZenoTOFMSMS_EAD_TestData.zip](https://github.com/user-attachments/files/16396606/7600ZenoTOFMSMS_EAD_TestData.zip)

chambm:
OK, if you've reviewed the mzML and it's converting ok, go ahead and add it. Thanks!

david-cox-sciex:
OK. the example files are added. If everything is okay, will you (@chambm) be doing the merge. For me it is blocked.

david-cox-sciex:
Did the Skyline PR Perf check fail because of this one extra file?
https://teamcity.labkey.org/buildConfiguration/ProteoWizard_SkylinePrPerfAndTutorialTestsWindowsX8664/3127062?buildTab=log&linesState=82&logView=flowAware

chambm:
Sorry I was out sick yesterday. Is this a DIA file? I'm wondering why it's not getting isolation width.

david-cox-sciex:
Not DIA, just a looped MSMS. It's possible we only show isolation width on DIA experiments.

chambm:
It turns out that only all WIFF files in that directory are tested; WIFF2 files have to be added by name. I also added some index subsetting to keep the size of the reference mzML down.

chambm:
Thanks for adding this @david-cox-sciex !

PR #3088, opened 2024-07-19
Move all grid items into "View > Live Reports" sub-menu

nickshulman:
Move spectrum-related stuff onto "View > Library" sub-menu

image
image


brendanx67 (changes requested):
Let's make View > Spectra 
View > Libraries
Instead and Spectral Libraries 
View > Libraries > Library Explorer
Instead. For consistency with Library Match and the naming we started for chromatogram libraries, and we often talk about View > Stectral Libraries as the Spectral Library Explorer.

So we finally get "Live Reports" and "Library Explorer" into the UI.

Looks good otherwise.

brendanx67 (approved):

PR #3090, opened 2024-07-22
* parallelized CreateLibraryFromSpectra

chambm:
* parallelized CreateLibraryFromSpectra as it is a big time sink when predicting full proteome with Koina (and added corresponding perf test)


chambm:
Happy to hear if you've got ideas to remove remaining bottlenecks Nick. The test I added is really good for profiling and can easily adjust how many peptides.

nickshulman:
I don't remember what I said about removing bottlenecks.

This looks good to me.

I suspect that the actual execution of the insert statements ends up being single threaded, but it looks like parallelizing setting all the Parameters on the SQLiteCommand objects does makes things go faster.

If you used a QueueWorker instead of a "ParallelEx.ForEach" you would have access to the "threadIndex" parameter that gets passed into the Consume action. That might be cleaner than using "Thread.CurrentThread.ManagedThreadId", or it might not make any difference.

I would recommend creating an inner class which has _insertSpectraCmd, _insertAnnotationsCmd, _insertPeaksCmd, _insertRetentionTimesCmd, _insertModificationsCmd.
Then "SpectrumInserter" would only need one ConcurrentDictionary instead of having one ConcurrentDictionary for each of those.

chambm:
One big change I thought would be nice would be to start writing the blib while getting predictions from Koina. But that seems like it'd be a big restructure. I agree about consolidating the insert commands, I'll do that.

nickshulman:
You might be able to speed things up by inserting more than one row with each SQLiteCommand.
This seems to have sped things up in my "replacing .skyd with SQLite" branch.

The SQL for inserting four peaks at once would be something like:
```
INSERT INTO RefSpectraPeaks (RefSpectraId,PeakMZ,PeakIntensity) 
SELECT (? AS RefSpectraId,? AS PeakMZ,? AS PeakIntensity)
UNION ALL (?,?,?)
UNION ALL (?,?,?)
UNION ALL (?,?,?)
```
The maximum number of parameters that a SQLiteCommand can have seems to be 1024.
I suspect performance will be best if we keep the number of parameters in the SQLiteCommands less than 256, but I'm not sure.

chambm:
I was indeed able to speed things up 20-30% more by buffering spectrum and peak inserts. Squeezed a bit more by limiting the parallelism in the ParallelEx.For, and I have no idea why the benefit maxes out at 4-6 threads when there is still a huge percentage of time taken up by Deflate. I think this is good enough to be getting on with though.

nickshulman:
I have a branch Skyline/work/20240727_SessionQueue where I have one thread devoted to just executing IDbCommand's and other threads that are enqueuing the IDbCommand objects.
That other branch has a whole "CommonDatabase" project which uses the NHibernate mapping to create a SQLiteCommand's which insert multiple rows at a time. (Maybe this will be the year that we figure out how to make things fast enough that we can replace the .skyd file with a SQLite database)

You can speed things up a little if you do "PRAGMA synchronous = OFF" and "PRAGMA journal_mode = MEMORY"

I think things can be sped up a few percent even more if you do ""PRAGMA defer_foreign_keys = ON", but if you actually temporarily violate a foreign key constraint such as inserting a RefSpectraPeaks before its RefSpectra, things slow down a lot.

chambm:
How much speedup did you get with the producer/consumer setup? Do you see any drawback to turning off journaling here?

PR #3092, opened 2024-07-24
Fix a problem in Settings>Transitions>IonMobility where Skyline would hang if user tried to "Edit current" when there is no current IMSDB…

bspratt:
…, or throw a null pointer exception when user tried to add an IMSDB via "Edit List > Add" instead of just using "Add"


pwiz_tools/Skyline/Properties/Settings.cs
```diff
@@ -2359,8 +2359,9 @@ public override string GetDisplayName(IonMobilityLibrary item)
             IEnumerable<IonMobilityLibrary> existing, object tag)
         {
             var ionMobilityFilteringUserControl = (owner as IonMobilityFilteringUserControl) ??
-                                                  (owner as TransitionSettingsUI)!.IonMobilityControl;
-            var ionMobilityWindowWidthCalculator = ionMobilityFilteringUserControl!.IonMobilityWindowWidthCalculator;
+                                                  ((owner as TransitionSettingsUI)?.IonMobilityControl) ??  // Accessed via Settings>TransitionSettings>IonMobility>Add
+                                                  ((owner as Form)?.Owner as TransitionSettingsUI)?.IonMobilityControl; // Accessed via Settings>TransitionSettings>IonMobility>EditList>Add|EditCurrent
+                var ionMobilityWindowWidthCalculator = ionMobilityFilteringUserControl!.IonMobilityWindowWidthCalculator;
```
nickshulman, line 2364: This line is indented wrong.

nickshulman (approved):
Looks good to me.

It's unfortunate that we don't have a good mechanism for making sure that the "EditItem" method gets passed what it needs.

There are other SettingsList subclasses which use the "object tag" argument to get their data.
The classes which use the tag are:
GroupComparisonDefList
MetadataRuleSetList
PeakScoringModelList
ServerList

These classes have all been written so that if the tag is null, they behave in a document-agnostic way.
The tag is always null when invoked from the `<Edit List...>` item in a SettingsListComboDriver.

But your way of looking at the Form Owner works well too.

Is there any concept of a generic "IonMobilityWindowWidthCalculator" that you could fall back to using if you can't find a "IonMobilityFilteringUserControl"?

bspratt:
> Is there any concept of a generic "IonMobilityWindowWidthCalculator" that you could fall back to using if you can't find a "IonMobilityFilteringUserControl"?

Unfortunately not - that's a document setting we need for "Use Results".

pwiz_tools/Skyline/Properties/Settings.cs
bspratt: I always appreciate your eye for detail, thanks.

bspratt:
> > Is there any concept of a generic "IonMobilityWindowWidthCalculator" that you could fall back to using if you can't find a "IonMobilityFilteringUserControl"?
> 
> Unfortunately not - that's a document setting we need for "Use Results".

(But it might not be saved to the current document yet, which is why we grab it from the control)

PR #3095, opened 2024-07-26
Figures of merit using bilinear fit and bootstrap calibration curves

nickshulman:
This pull request takes just the calibration curve part of PR #2572 (optimize transitions dialog)

On the "Quantification" tab of Peptide Settings, there is a new option in the "Calculate LOD By" dropdown which available if the "Regression Fit" is "Bilinear":
image

The option is currently called "Bilinear turning point standard error" but definitely needs a better name.

When that option is chosen, the "Max LOQ bias" textbox becomes disabled but you can still put a number in the "Max LOQ CV" box:
image

When those options are chosen, the Limit of Detection is one standard deviation above the mean of the baseline points:
image

The Lower Limit of Quantification is calculated by generating one hundred other calibration curves ("bootstrap curves) with random subsets of points and choosing the concentration level where the standard deviation of the curves is below the "Max LOQ CV".

The user can choose to display the bootstrap curves:
image

image

There is also a new "Calibration Curve Display Options" form which lets you set the line width and font size:
image


brendanx67:
It would be interesting to see what this does with the data from the Absolute Quantification tutorial. The curve you are showing has multiple points per concentration. This is unusual for quantitative experiments that are not entirely focused on figures of merit.

brendanx67:
I think we should have a Zoom meeting with Will Thompson and Monika from Genentech to understand their intended use of our existing calibration features so that in adding bilinear fit we do not make that use case worse, and ideally we make that use case work better also, as it is clearly quite cumbersome for Monika.

brendanx67:
I was not able to see the new option when I tried it with the Absolute Quantification tutorial. Though, I could see other changes you made in this PR like moving "Qualitative ion ration threshold" outside the "Figures of merit" group box, and the new "More Options..." menu item and resulting form.

nickshulman:
> I was not able to see the new option when I tried it with the Absolute Quantification tutorial.

The "Bilinear turning point standard error" and "Bilinear turning point (obsolete)" options only show up if you have chosen "Bilinear" in the Regression Fit dropdown.
Updating the Absolute Quantification tutorial sounds like a great test of this functionality as well as Eduardo's converting tutorials to HTML work.

brendanx67:
I think you could call them "Bilinear fit (legacy)"

and

"Bilinear fit"

I think that gets across the one being replaced by the other as the best option, and that the "(legacy)" one is no longer recommended but there because it may have been used in the past.

brendanx67:
I mean "Bilinear turning point" of course

PR #3096, opened 2024-07-29
fix https://skyline.ms/issues/home/issues/details.view?issueId=985 Target tree's Find function results window says "Peptide" where it should say "Molecule"

bspratt:
The contents of the "Type" column in the Find Results window would not get translated from "Protein" and "Peptide" as appropriate


nickshulman (approved):
Looks good to me.

If you wanted, you could look at the IdentityPath in the Bookmark in order to decide whether to display "Molecule" or "Peptide" in a mixed document, but that would probably be weird.

bspratt:
> If you wanted, you could look at the IdentityPath in the Bookmark in order to decide whether to display "Molecule" or "Peptide" in a mixed document, but that would probably be weird.

Yeah, there are plenty of places in the Mixed UI where we just say "molecule" even if it is actually a peptide.

PR #3098, opened 2024-07-30
Update the HTML version of the MS1 DDA Search tutorial

nickshulman:
Replace "Import Peptide Search" with "Run Peptide Search" and replace one image with a new screenshot.

I had to edit the HTML page by hand because when I used SeaMonkey Composer it reformatted the entire HTML document to be 80 characters wide instead of the 150 that was checked in.


brendanx67:
This seems to imply a bug in the DOCX file, which would mean a bug in the currently posted PDF. While the DOCX and PDF files are primary, we need to make fixes in them, and then ideally propagate those through automated re-translation.

nickshulman:
Moved to https://github.com/nickshulman/pwiz/pull/3
This was just a pull request to play around with HTML tutorial

PR #3100, opened 2024-07-31
Fix an issue where editing a molecule's formula or mass in the Targets tree would not update the m/z values for non-auto-managed precursor transitions

bspratt:
Reported by user Nicholas D.

see 
https://skyline.ms/announcements/home/support/thread.view?rowId=62755
and
https://skyline.ms/issues/home/issues/details.view?issueId=983


nickshulman:
I think all of the complicated logic about updating the children of the PeptideDocNode should be moved inside of "PeptideDocNode.ChangeCustomIonValues".
That method has enough information to update all of its children, so it doesn't make sense for the method to return a PeptideDocNode that still needs work done on it in order to be valid.

Do you feel like moving that logic from "EditMenu.ModifyPeptide" into PeptideDocNode.ChangeCustomIonValues, or would you like me to do it?

bspratt:
> I think all of the complicated logic about updating the children of the PeptideDocNode should be moved inside of "PeptideDocNode.ChangeCustomIonValues". That method has enough information to update all of its children, so it doesn't make sense for the method to return a PeptideDocNode that still needs work done on it in order to be valid.
> 
> Do you feel like moving that logic from "EditMenu.ModifyPeptide" into PeptideDocNode.ChangeCustomIonValues, or would you like me to do it?

Good suggestion, I'll implement - thanks

bspratt:
Thanks for yet another thoughtful code review. In the end it wasn't about moving the logic to PeptideDocNode.ChangeCustomIonValues but rather tweaking the logic that was already in PeptideDocNode.ChangeCustomIonValues.

nickshulman (approved):

PR #3101, opened 2024-08-01
Make Precursor Collision Energy nullable in the Document Grid

nickshulman:
Changed to display "#N/A" for Collision Energy in Document Grid when no collision energy formula provided (not reported by anyone but inspired by import transition list Issue #1006)


bspratt (approved):
Looks proper. Once this is merged I'll look at improved round-tripping with our transition list readers

PR #3097, opened 2024-07-29
Migrate all tutorials to Html

eduardo-proteinms:
This PR migrates all docx files inside of the tutorial(pwiz_tools\Skyline\Documentation\Tutorials) directory. Careful with opening the full diff for this PR as it is too large to process. To take a look at the result the best way would be to checkout the branch and look through the documents that are generated in comparison to the existing document. 

### Html Conversion
Converting docx documents to html has alot of complications. Not all word document elements or styles can be matched perfectly while keeping a clean html output. I researched many conversion libraries and only found [mammoth](https://github.com/mwilliamson/mammoth.js/tree/master) to generate clean html and was easily extensible/configurable. Even though this library has different language implementations, javascript is the most maintained and simplest to fork.
 
We leverage mammoth-js for the following key features:
- Merge viable docx elements - The docx format is extremely verbose. By merging elements which do not need to be 'fresh' it is able to condense elements with matching styles that do not need to be separated.
- Configurable image reader - Makes it simple to write images to files and do transformations as needed. 
- Simple style mappings - Recognizes and warns user of styles which could require additional formatting. We use this to map certain styles to css classes

The implementation of mammoth-js is not overly complex and for certain use cases I found the need to fork the library. This fork currently lives in my space but I will move it to the pwiz one.  [mammoth.js/tree/skyline-document-generation](https://github.com/eduardo-proteinms/mammoth.js/tree/skyline-document-generation)
I implemented the following changes in the library:
- Table Indentation - Tables in our tutorials at times have indentations to match the text they correspond to. I added a change to find the "w:ind" property in the table element and apply the indentation to the corresponding html. 
- KeepNext - Many paragraphs in our documents are configured to keep the text with the next element. Usually this is used to avoid splitting image and text. I added a change to find the "w:keepNext" property and assign these elements a class in the html. In the browser this wont make a difference but if you print the document PDF you will see this is respected. 
-  Added [skyline.js](https://github.com/eduardo-proteinms/mammoth.js/blob/skyline-document-generation/lib/skyline.js) - Does the actual conversion of the specified folders. This could be implemented outside of mammoth but kept this in the project for simplicity. 

### Image Conversion
Images are written to files by simply leveraging mammoth's image reader. The html document has a relative link to this image. 

[...]

PR #3103, opened 2024-08-02
Update the "ResourcesOrganizer" tool to work with HTML files

nickshulman:
I have started the work of making the "ResourcesOrganizer" read localized strings from HTML files.

I copied just the "Data Independent Acquisition" tutorial from PR #3097.
I changed the folder layout a little so that there are three sibling folders for the languages: "en", "ja", and "zh-CHS".
Also, the name of the HTML file in each folder is "index.html".

In the localized folders, there is also a new HTML file called "invariant.html" which is the HTML file that the localized index.html was translated from.

In theory, the ResourcesOrganizer tool will be able to compare "ja/invariant.html" with "en/index.html" and figure out which paragraphs need to be sent to the localizers for re-translation.


brendanx67:
Note that I have a strong desire to a have the HTML file(s) sort tot the top of the file list by default in these folders. Sadly, index.html still comes after image-##.* We can always name the images z-image-##.* Just keep this in mind as we design these folders.

Also, I want the folder names to match the shorter names we are using on the website which we combine with a version number for the PDF files, and also we use them for the covershot images. Probably we should add this name as an attribute to the test function and add a mode in SkylineTester that allows you to view the tutorials by these names to make the link between tests and tutorial documents clearer.

nickshulman:
I'll create a different pull request for this.

PR #3104, opened 2024-08-04
Prompt to save the document when .skyd file has extra results

nickshulman:
If the user tries to import results and Skyline notices that there are replicates in the .skyd file that are not in the document, Skyline asks the user if they want to save the document.
This makes better the confusing behavior where you might remove some replicates in Manage Results and then import again and get the same results as before because they were already in the .skyd file.

Changed to prompt to save document before importing new results if the .skyd file contains results which are not in the loaded document.


brendanx67:
Thanks, Nick. I should have done this when I added the SpectrumChromDataProvider. The idea of the SKYD file as merely a "cache" of what was in the raw data files stopped being valid at that point.

PR #3109, opened 2024-08-08
backport improvements to feature detection's RT alignment, for improved detection of features appearing in multiple replicates

bspratt:


brendanx67:
So, you think this is enough of an improvement and safe enough that it should go into our next patch release?

bspratt:
> So, you think this is enough of an improvement and safe enough that it should go into our next patch release?

Improvement, yes - RT alignment simply wasn't happening in some cases when the KDEAligner got distracted by noise at the start and end of a run. With these changes we ignore the first and last 20% of the run.

Safe - that's a call for @nickshulman since there are changes to KDEAligner that might affect other uses. It does not cause any test failures though.

PR #3110, opened 2024-08-08
Improve RT alignment in feature finding by using only the middle 60% of each replicate…

bspratt:
…, avoiding the noise typically found at start and end of a mass spec run.


pwiz_tools/Skyline/Model/RetentionTimes/KdeAligner.cs
```diff
@@ -235,77 +303,69 @@ private void TraceSouthWest(float[,] histogram, int bestXi, int bestYi, LinkedLi
...
-
         private void Stamp(float[,] histogram, float[,] stamp, int x, int y)
         {
-            var stampLength0 = stamp.GetLength(0); // GetLength is more expensive than you'd think
-            var stampLength1 = stamp.GetLength(1);
-            var histogramLength0 = histogram.GetLength(0);
-            var histogramLength1 = histogram.GetLength(1);
-
-            for (int i = x - stampLength0/2; i <= x + stampLength0 / 2; i++)
+            for (int i = x - stamp.GetLength(0) / 2; i <= x + stamp.GetLength(0) / 2; i++)
```
nickshulman, line 332: Does the "GetLength is more expensive than you'd think" comment still apply, or has more recent profiling disproven it?

pwiz_tools/Skyline/Model/RetentionTimes/KdeAligner.cs
```diff
@@ -73,56 +76,95 @@ public KdeAligner(int origXFileIndex, int origYFileIndex, int resolution = 1000)
...
         }
 
-        public KdeAligner(int resolution = 1000) 
+        /// <summary>
+        /// Constructs a KdeAligner with a specific resolution and stretchFactor
+        /// </summary>
+        /// <param name="resolution">The number of points in the X and Y axes </param>
+        /// <param name="eccentricity">Experimental parameter controlling the amount that the CosineGaussian stamp is stretched.
+        /// This value is interpreted as the eccentricity of an ellipse whose major axis is along the Y=X axis.</param>
+        public KdeAligner(int resolution = 1000, double eccentricity = 0)
```
nickshulman, line 85: This "eccentricity" parameter should be removed. It was a misguided thing that I thought might help some scenario but never improved things.

pwiz_tools/Skyline/Model/RetentionTimes/KdeAligner.cs
bspratt: I haven't reprofiled it. But I can remove the comment if it's distracting.

pwiz_tools/Skyline/Model/RetentionTimes/KdeAligner.cs
bspratt: Will do, thanks

pwiz_tools/Skyline/Model/RetentionTimes/KdeAligner.cs
bspratt: Oh right - uninitentional undoing of an optimization, thanks. Dang, I love your reviews.

PR #3113, opened 2024-08-09
Skyline MSFragger DIA search and tutorial

chambm:
- fixed MSFragger search to set default data_type correctly depending on workflow type (PRM -> data_type=1, DDA -> data_type=0; and user can override the data_type in Additional Settings)
- added TestDiaSearchStellarTutorial (Stellar GPF) and TestDiaSearchQeTutorial (QE wide window DIA, same dataset as EncyclopeDIA tutorial but with less aggressively subsetted files)
- increased overlap tolerance in Edit Isolation Scheme dialog from 0.0001 to 0.001
- added logging of commands used to launch MSFragger and Crux

Possible todos:
- [ ] Add workflow option for non-DIA Umpire DIA that sets isolation scheme properly
- [ ] Add workflow option for GPF (or checkbox when DIA workflow is selected) that sets data_type=2 automatically and imports files as multi-injection replicate runs
- [ ] When above workflows are selected, limit search engine selection to MSFragger
- [ ] Add tolerance when visualizing overlaps in DiaIsolationSchemeGraphForm
- [ ] Add on-demand demultiplexing when above workflows are used, like EncyclopeDIA search mode?
- [ ] Add a way to filter out alignment DIA scans from search and extraction: e.g. allow user to delete rows from isolation scheme, and if rows were deleted, automatically run files through msconvert's new "isolationWidth" filter (or perhaps another new one that requires fully specifying the absolute m/z isolation windows to include/exclude)

@brendanx67 I've added the todo list here in lieu of detailing these considerations in an issue. I will let Mike test what's already done while we discuss how best to accomplish the workflow options.


chambm:
@brendanx67 This PR currently doesn't have any visible UI changes. I think it'd be a good one to cherry pick to 24.1 for the next patch. Then I can do a follow-up PR to change the workflow options to make DIA-Umpire optional. Sound good?

brendanx67 (changes requested):
You are going to need to add the full set of tutorial audit logs for 5 languages. You currently only have English. Also, just a good idea to run any new test in all 5 languages.

brendanx67 (approved):

brendanx67:
Sure.

PR #3116, opened 2024-08-10
Fix a problem in Settings>Transitions>IonMobility where Skyline would hang if user tried to "Edit current" when there is no current IMSDB… (#3092)

nickshulman:
Fix a problem in Settings>Transitions>IonMobility where Skyline would hang if user tried to "Edit current" when there is no current IMSDB, or crash when user tried to add an IMSDB via "Edit List > Add" instead of just using "Add"

Reported anonymously via exceptionweb


brendanx67:
Maybe we should add DropDownOpenning handling to SettingsListComboDriver that adds or removes this element based on whether SelectedItem == default. We could also put the ComboHelper.AutoSize(Combo) in there.

Great to fix this of course. Just saying it would be even better, if the user didn't see the option.

PR #3117, opened 2024-08-12
Set the option to use UTF8 to encode filenames in sky.zip files

vagisha:
Skyline document imports can fail on Panorama if filenames in the sky.zip contain special characters characters such as ç ([cedilla](https://en.wikipedia.org/wiki/Cedilla))
Java zip utility classes assume UTF-8 encoding unless specified otherwise.  Here is the support board request where this problem was reported - https://skyline.ms/announcements/home/support/thread.view?rowId=65292


nickshulman (approved):
Looks good to me.

brendanx67 (changes requested):
This covers what is necessary for Panorama, but now for Skyline to read its own shared ZIP files, there are a couple of cases of ZipFile.Read() that will need to become:

ZipFile.Read(SharedPath, new ReadOptions { Encoding = Encoding.UTF8 })

vagisha:
> This covers what is necessary for Panorama, but now for Skyline to read its own shared ZIP files, there are a couple of cases of ZipFile.Read() that will need to become:
> 
> ZipFile.Read(SharedPath, new ReadOptions { Encoding = Encoding.UTF8 })

I think this will cause a problem with sky.zip files where the filename encoding is not UTF-8. When I try to read 
`special-ç-char.sky.zip`, created with default encoding but read with `ReadOptions { Encoding = Encoding.UTF8 }`, the names of the files in the exploded directory have ç replaced with �. 
`special-ç-char.sky.zip (default encoding) -> special-�-char.sky (read as UTF-8)`

But when I read a sky.zip file created with UTF-8 filename encoding, without specifying an encoding option while reading, then the file name is read correctly.
`special-ç-char.sky.zip (UTF-8 encoding) -> special-ç-char.sky (read without encoding option)`

I think this second case works correctly because the ZIP format has a flag that is set if the encoding of the file name field is UTF-8. 
https://en.wikipedia.org/wiki/ZIP_(file_format)#Internationalization_issues
The ZIP format also has a Unicode Path Extra Field that is supposed to have the UTF-8 encoded filename.  But that is either not getting set when the sky.zip file is created, or not being read when the sky.zip is opened.   Otherwise the first case - create with default encoding, read with UTF-8, would also have worked.

brendanx67:
> I think this will cause a problem with sky.zip files where the filename encoding is not UTF-8. When I try to read `special-ç-char.sky.zip`, created with default encoding but read with `ReadOptions { Encoding = Encoding.UTF8 }`, the names of the files in the exploded directory have ç replaced with �. `special-ç-char.sky.zip (default encoding) -> special-�-char.sky (read as UTF-8)`

Good point for backward compatibility. Glad that a file written with UTF-8 automatically works without any special ReadOptions. Thanks for clarifying. Looks good to me!

brendanx67 (approved):

PR #3118, opened 2024-08-12
Replace uses of Trace.Trace* with new wrapper methods

bspratt:
The thinking is that a bare call to TraceWarning looks too much like debug code. 

In instances where Trace.TraceWarning is called with a string literal (rather than a string resource) I have left it alone as this does in fact seem more like debug code. Whether or not these should remain in the code is a matter for review.


bspratt:
@nickshulman @brendanx67 any thoughts on this?

nickshulman:
I was hoping that we would eventually get to a place where the Skyline code base had lots of logging statement and, when Skyline is behaving mysteriously, users would be able to go to "Tools > Options" and increase the verbosity level and maybe see some helpful information.
Currently, we effectively have that verbosity level set at "Warning" and therefore all Warning and Fatal messages get written to the Immediate Window.
How can we get there from here?

brendanx67:
Having a UserMessage class seems like a good start so that we are clearer
about what we want the user to see. The problem with the old situation was
that it looked like debug code and was more likely to escape notice that
the messages were not localizable, and also increased the likelihood of
leaving debug code in place thinking it was meant for the user to see.

bspratt:
So are we satisfied with this as a step in the right direction? And with leaving the instances where Trace.TraceWarning is called with a string literal (rather than a string resource) as-is, being debug code? I'd like to get it merged and off my PR list.

bspratt:
@brendanx67 are we satisfied with this as a step in the right direction? And with leaving the instances where Trace.TraceWarning is called with a string literal (rather than a string resource) as-is, being debug code? I'd like to get it merged and off my PR list.

brendanx67:
Not very satisfied, since all Trace.TraceWarning() output goes to the Immediate Window. I would like to see a solution that now uses a different mechanism, if we are going to leave these in unlocalized for debugging purposes. I don't mind adding another class called DebugMessage which we use to wrap Trace.TraceWarning() for debug use, and put the call to Trace.TraceWarning() in a conditional for being under a debugger, which prevents a user from ever seeing the message.

Either that, or we need to create a different mechanism for UserMessage adding messages to the Immediate Window and disconnect this "clever" overloading of Trace.TraceWarning().

brendanx67:
Sorry you have ended up saddled with this for being the person I finally noticed using this feature that Nick added without my realizing the long-term impact.

chambm:
What is the Skyline "Immediate Window" (I think of Visual Studio when I read that)?

brendanx67:
Yeah. It is a bit like the VS Immediate Window. I had Dany Broudy create it for external tools as a place to write external tool console output that could be copied and pasted, and we also added a bit of ability to run commands to the window. I can't remember exactly what, but definitely the ability to run a tool from the Immediate Window.

bspratt:
I think I'd have more clarity on this if I knew what @nickshulman thinks should happen with those untranslated messages.

nickshulman:
> I think I'd have more clarity on this if I knew what @nickshulman thinks should happen with those untranslated messages.

I think any TraceWarning's that we wouldn't want the user to see should be changed to TraceInformation.

bspratt:
> I think any TraceWarning's that we wouldn't want the user to see should be changed to TraceInformation.

Done, but my gut (and Brendan's too, I think) is that we want to wrap that in something that makes the intent clear e.g. "AnsyncDebugMessage" while hiding the actual implementation (TraceInformation).

brendanx67:
Let’s create a parallel class to UserMessage called DebugMessage and wrap
this use in it and then add to our static analysis that these are the only
two cases of these functions in the code. So, that misuse doesn’t creep
back.

brendanx67:
Sure. That also is fine with me.

bspratt:
So, Messages.WriteAsyncUserMessage and Messages.WriteAsyncDebugMessage?

nickshulman (approved):

PR #3119, opened 2024-08-12
Edit > Integration > Peak Imputation menu item

nickshulman:
image

image


brendanx67:
I think I prefer this on the Refine menu with Refine > Reintegrate. Something like:
Impute Integration
Impute Peaks
It doesn't really feel like an Edit operation.

nickshulman:
> I think I prefer this on the Refine menu with Refine > Reintegrate.

I originally had this on the "Refine" menu but I moved it to "Edit > Integration" because what it does is so similar to what "Apply peak to all" does.
You use "Apply peak to all" when you know that the correct peak is chosen in the currently selected replicate. You use "Peak Imputation" when you want Skyline to figure out which replicate has the best peak.
They became more similar once "Apply peak to all" was changed to work on multiple peptides selected in the Targets tree.

But, yes, "Refine" is also a place that it would make sense to put this menu item.

nickshulman:
I am going to open a new pull request with a different implementation of this feature.
The version associated with this closed pull request can be installed from here:
https://proteome.gs.washington.edu/~nicksh/SpecialSkylines/PeakImputation/Version445.application

PR #3120, opened 2024-08-13
Add a convenience function GetUIValue() for getting a value on the UI thread.

bspratt:
Fixes a problem with exceptions being thrown under the debugger in TestLibraryBuild

### DIFF
--- a/pwiz_tools/Skyline/TestUtil/TestFunctional.cs
+++ b/pwiz_tools/Skyline/TestUtil/TestFunctional.cs
@@ -682,6 +682,16 @@ namespace pwiz.SkylineTestUtil
             return  (millis * waitMultiplier) / SLEEP_INTERVAL; // Return the wait cycle count
         }
 
+        /// <summary>
+        /// Convenience function for getting a value on the UI thread
+        /// </summary>
+        public static T GetUIValue<T>(Func<T> act)
+        {
+            T result = default;
+            RunUI(() => result = act() );
+            return result;
+        }
+
         public static TDlg TryWaitForOpenForm<TDlg>(int millis = WAIT_TIME, Func<bool> stopCondition = null) where TDlg : Form
         {
             int waitCycles = GetWaitCycles(millis);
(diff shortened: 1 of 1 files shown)


brendanx67 (approved):
Looks good. Thanks. I can imagine we may end up using this elsewhere.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -682,6 +682,16 @@ private static int GetWaitCycles(int millis = WAIT_TIME)
             return  (millis * waitMultiplier) / SLEEP_INTERVAL; // Return the wait cycle count
         }
 
+        /// <summary>
+        /// Convenience function for getting a value on the UI thread
+        /// </summary>
+        public static T GetUIValue<T>(Func<T> act)
```
nickshulman, line 688: I would have probably named this "RunUI" (or maybe "CallUI" if "RunUI" caused too much compiler confusion)
This method does the same thing as "RunUI", but has a Func parameter instead Action.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -696,7 +706,7 @@ private static int GetWaitCycles(int millis = WAIT_TIME)
                     var multipleViewProvider = tForm as IMultipleViewProvider;
                     if (multipleViewProvider != null)
                     {
-                        formType += "." + multipleViewProvider.ShowingFormView.GetType().Name;
+                        formType += "." + GetUIValue(() => multipleViewProvider.ShowingFormView.GetType().Name);
```
nickshulman, line 709: A different way to solve this would be to move this line of code down two lines so it's inside of the "RunUI" that is already there.

nickshulman (commented):
Looks good to me too.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
bspratt: Hah, good point

PR #3121, opened 2024-08-15
Add test for "File > Share" with international characters in file name

nickshulman:
Add "InternationalFilenamesTest" which tests doing "File > Share" still works even if the document file path has international characters in it.
Also, verifies that the .sky.zip file can be opened, which was only recently fixed in PR #3117.

Fixed a problem where, when opening a document whose name was the same as the previous document, sometimes the document library would never get opened. This was happening because "_loadedLibraries" and "_loadingLibraries" only paid attention to the Name of the library.


brendanx67 (approved):
Looks good. Thanks, Nick.

PR #3122, opened 2024-08-16
Skip over zoom scans when extracting chromatograms

nickshulman:
Changed to skip over "zoom" spectra when extracting chromatograms (requested by Mike)


chambm:
I'm not sure this is the ideal path but haven't gotten a chance to reply to the email chain yet. In the future those scans may not be annotated as zoom scans (like Mike said, they're not really zoom scans). My idea is based on what Mike asked for originally: a way to delete isolation windows in the grid after importing from file. If the user deleted windows, then that will prevent them from being extracted due to the isolation window not matching, right? And for DIA Search mode, I could add logic in that case so that the MsconvertDdaConverter would get called to make subset mzMLs without those alignment scans (using the new isolationWindows filter to exclude exactly the isolation windows that the user deleted from the grid).

nickshulman:
> If the user deleted windows, then that will prevent them from being extracted due to the isolation window not matching, right? 

Unfortunately, no. Specifying an isolation scheme is a way for the user to tell Skyline what the width of the isolation window is given what was isolated. There used to be some mass spectrometers which only told us the center of the isolation window, and there also were cases where people told the mass spectrometer to isolate a window of one width, but wanted Skyline to only use part of the window.
Mike's data has one isolation window [500-550] which he would like to exclude, but there is another window [524.48832-526.48922] which he wants to keep.
If "[500-550]" were not in the isolation scheme then Skyline would consider treat those [500-550] scans as if their window was  [524.48832-526.48922] because that's the window that matches the isolated precursor.

The way to work around not being able to delete rows from the isolation scheme grid is to copy everything to Notepad, delete the rows there, copy, close the isolation scheme editor, open it up again and paste from the clipboard.

chambm:
So it would need a "Skip spectra with isolation windows that do not match isolation scheme" checkbox to be added somewhere as well (and the logic inside the extractor to do that). Definitely more complicated than the zoom scan filter, but I do think it's more future proof. Also do we know for sure that there aren't any users depending on being able to extract from zoom scans? It seems unlikely, but I probably would have said that about the so-called SIM scans too (a long time ago).

nickshulman:
This has been made obsolete by PR #3125

PR #3125, opened 2024-08-16
Global spectrum filters

nickshulman:
Users can now specify a "Global Spectrum Filter" which gets used for all extracted ion chromatograms in the document.

The place to edit it is on the "Instrument" tab of Transition Settings. It might make more sense to put it on the "Full Scan" tab, but "Instrument" has a lot more room.

It looks like this:
image

When you push the "Edit..." button it brings up the spectrum filter editor:
image

You could add an "Isolation Window Width" filter (which is also a new feature in this pull request):
image

When you have a filter, the read-only textbox next to the "Edit..." button shows the current value of the filter:
image

The "Ignore sim scans" checkbox has been removed from the "Full scan" tab:
image

If you open an older document which has "Ignore sim scans" set then that gets translated into a filter:
image


pwiz_tools/Skyline/Model/Results/SpectraChromDataProvider.cs
```diff
@@ -292,10 +294,13 @@ private void ExtractChromatogramsLocked()
                 }
 
                 UpdatePercentComplete();
-
```
chambm, line 295: Is this CurrentSpectrum call slow and that's why it was in the if blocks below? If so, it could be changed to only be called when _globalSpectrumClassFilter != null.

pwiz_tools/Skyline/Model/Results/SpectraChromDataProvider.cs
```diff
@@ -51,6 +51,7 @@ internal sealed class SpectraChromDataProvider : ChromDataProvider
         private bool _isSingleMzMatch;
         private bool _sourceHasPositivePolarityData;
         private bool _sourceHasNegativePolarityData;
+        private Predicate<SpectrumMetadata> _globalSpectrumClassFilter;
```
chambm, line 54: Should it be _globalSpectrumFilter? Why the Class bit?

pwiz_tools/Skyline/Model/TransitionGroupDocNode.cs
```diff
@@ -247,6 +248,12 @@ public IList<DocNode> GetPrecursorChoices(SrmSettings settings, ExplicitMods mod
 
         public SpectrumClassFilter SpectrumClassFilter { get; private set; }
```
chambm, line 249: SpectrumClassFilter here but SpectrumFilter in TransitionSettingsUI.

chambm (approved):
Looks great! Can't wait to convert those to msconvert filters and see how it maps. :)

pwiz_tools/Skyline/Model/TransitionGroupDocNode.cs
chambm: So I'm still curious what is the reason for the "Class" bit in the name?

nickshulman: > So I'm still curious what is the reason for the "Class" bit in the name?

Skyline already had a class called "SpectrumFilter" which is a pretty important class which does all of the work of extracting chromatograms from spectra. In that class name, the word "filter" basically means "transformer".
The feature that I implemented last year is called "Spectrum Filter" in the user interface, but it had to be called something else in the code because the name "SpectrumFilter" was already being used.

chambm: Did you consider "SpectrumMetadataFilter"? Not a big deal obviously, just something to ponder.

PR #3128, opened 2024-08-19
Prompt user to configure Koina before EncyclopeDIA search

nickshulman:
Added prompt to configure Koina before running EncyclopeDIA search (reported by Eric)

If the user does "File > Search > EncyclopeDIA Search..." and they have not configured Koina yet they will see this message:
image

The wizard will not start if they choose "No" or if they don't actually configure Koina.

### DIFF
--- a/pwiz_tools/Skyline/SkylineFiles.cs
+++ b/pwiz_tools/Skyline/SkylineFiles.cs
@@ -50,6 +50,7 @@ using pwiz.Skyline.Model.ElementLocators.ExportAnnotations;
 using pwiz.Skyline.Model.Esp;
 using pwiz.Skyline.Model.IonMobility;
 using pwiz.Skyline.Model.Irt;
+using pwiz.Skyline.Model.Koina.Models;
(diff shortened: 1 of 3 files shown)


chambm (changes requested):
Looks good. To fix tests I think:
```
Settings.Default.KoinaIntensityModel = KoinaIntensityModel.Models.First();
Settings.Default.KoinaRetentionTimeModel = KoinaRetentionTimeModel.Models.First();
```
will have to be moved before:
```
var searchDlg = ShowDialog<EncyclopeDiaSearchDlg>(SkylineWindow.ShowEncyclopeDiaSearchDlg);
```

nickshulman:
Thanks! I have fixed the failing tests.

I wonder if we might want to change the steps of the tutorial to tell the user to go to the "Tools > Options" dialog and make sure that the Koina settings are set.

chambm:
Definitely. I will put that on my TODOs. Right after I actually write the proper tutorial. I think it's just a brief PPTX right now. :P

PR #3129, opened 2024-08-19
Extend the transition list reader from importing just per-transition Notes to also reading per-precursor and per-molecule notes. 

bspratt:
This adds new supported column types "Precursor Note" and "Molecule Note", and adds "Transition Note" as a synonym for the existing "Note" column.  

Requested by user Chris A.


bspratt:
@nickshulman  can you give this the blessing? Thx

nickshulman (approved):

PR #3131, opened 2024-08-21
AutoQC Loader updates

vagisha:
- Added a text field for entering the path to an annotations file.
image
- Added a FileWatcher that watches for changes in the annotations file.
- Added a SCIEX WIFF2 option to the "Instrument type" combo box for importing .wiff2 files
- Ping software version to Panorama server.
- Deleted SignAfterPublish.bat

### Bug fixes:
- Fix UI freeze. This happens when: 1) Start a config 2) Switch to the log tab 3) Stop the config.
- Fix log message for selected Skyline settings.
- Fix OutOfBoundsException in Logger.cs
- Fix switching log file in the Log tab when a config is selected.
- Change width of "Folder on Panorama" textbox in the "Panorama" tab
- Renamed AutoQcConfigManager.State property to AutoQcState to get rid of VisualStudio warning about hiding an inherited member.
- Do not force user to select "...a root folder for the configurations" if the .qcfg file being imported is in the "Downloads" folder.
- Save the available Skyline options to user.config after initializing. Otherwise, Skyline and Skyline-daily options are disabled in the "Skyline" setting tab when creating a new configuration.   
- Catch "Exception" instead of an "ArgumentException" when validating a configuration with PanoramaSettings.ValidateSettings(true). This can throw both an ArgumentException and a PanoramaServerException which, in PanoramaClient project, does not inherit from ArgumentException.
- Fixed bug: Config did not show "Error" status if there was an error uploading to Panorama, and the config was stopped manually. This happened because AutoQcConfigManager.SetConfigEnabled() returns a new State with new config runners that do not copy the RunnerStatus from the current State.
- Fixed bug: On app startup, PanormaServer validation was happening in one background thread per configuration.  This caused a warning "The state of the configuration list has changed...".  Configurations that failed Panorama server validation were not being displayed in red.
- Added "fatal" Panorama errors that will cause a config to stop with an error status.  Retrying the import in these cases is never going to succeed unless user takes an action. Example import error on Panorama - "QC folders allow new imports to add or remove peptides...".  
- Upload document to Panorama after the configuration is started if:
  - One or more existing raw files were imported
  - OR none of the existing raw files were imported but the document contains replicates


vagisha:
Functional tests will be added in a separate PR.

PR #3132, opened 2024-08-22
Fixes to the Relative Abundance graph formatting

rita-gwen:
This PR implements following changes go the Relative Abundance graph:
* make sure there is always a little white space between the first (leftmost) point marker in the graph and the Y axis
* avoid resetting the graph zoom due to selection change or other changes that do not modify the document
* decouple Log Scale setting for the Relative Abundance graph from other peak area graphs and set its default to true

These are based on Mike's suggestions in the e-mail thread from Jul 29.


pwiz_tools/Skyline/Controls/Graphs/SummaryRelativeAbundanceGraphPane.cs
```diff
@@ -219,16 +219,21 @@ public override void UpdateGraph(bool selectionChanged)
...
             // is not yet created
+            var dataChanged = false;
             if (_graphData?.GraphPointList == null ||
                 !ReferenceEquals(document, _graphData.Document) ||
-                !Equals(graphSettings, _graphData.GraphSettings))
+                !Equals(graphSettings, _graphData.GraphSettings) ||
+                !Equals(_graphData.AggregateOp, aggregateOp))
             {
                 _graphData = CreateGraphData(document, graphSettings);
+                dataChanged = true;
```
nickshulman, line 234: "dataChanged" is used to decide whether to reset the graph by calling "UpdateAxes".
I wonder if we could do something smarter than just looking at whether the Document is different.
Maybe we could compare the new graph data to the old graph data and only update the axes if there are more points or the maximum Y value of all of the points has changed.

As it stands, the graph will reset itself when the document changes for reasons that do not affect the layout, such as editing the Note on something in the Targets tree or when the document changes because of a background loader.

PR #3134, opened 2024-08-26
DIA search workflow UI changes

chambm:
- reworked DIA workflow radio button to work with and without DIA-Umpire
- added checkboxes next to DIA workflow for DIA-Umpire and Gas Phase Fractionation
* added faster TestDiaSearch test (sharing TestDiaSearch DiaUmpire code)
* updated audit logs

GPF checkbox in Extract Chromatograms:
image

Conversion settings always shown for DIA workflow, with checkbox for DIA-Umpire:
image


chambm:
@brendanx67 I added a screenshot of the main UI change here.

brendanx67:
Should DIA-Umpire be on by default? I wonder if we should get rid of "with MS1 filtering" from the DDA option. That is our own terminology and somewhat fading into the past. We may also want to improve the wizard to allow the DDA extraction option for MS/MS now that we have that. Quant is still based on MS1, but I am not sure "with MS1 filtering" adds anything but term that never really caught on.

brendanx67:
Maybe also a tip on "DIA-Umpire" explaining that it provides deconvolution of chimeric DIA MS/MS spectra to single-precursor, pseudo-DDA MS/MS spectra.

brendanx67:
Also a tip about what "Gas phase fractionation" is. It should not be in title case despite the fact that it is given the acronym GPF. DIA-Umpire is a proper name, which is why both words get capitalized.

chambm:
> Maybe also a tip on "DIA-Umpire" explaining that it provides deconvolution of chimeric DIA MS/MS spectra to single-precursor, pseudo-DDA MS/MS spectra.
Also a tip about what "Gas phase fractionation" is. It should not be in title case despite the fact that it is given the acronym GPF. DIA-Umpire is a proper name, which is why both words get capitalized.

Done. Without going into any UI reorganization, the last thing to do here is to make the GPF checkbox import the given files as a multi-run replicate. I'm not sure offhand how to do that but I will look at the import results dialog to figure it out.

brendanx67:
I do think it would be better to have the user just choose "DIA" and then make these choices later on pages with more space to explain better what they are choosing.

brendanx67:
image

It is a little weird that we offer the choice of DIA Umpire on the first page, but don't really get into explaining it until this page.

brendanx67:
They are 5 pages apart.

It may have seemed to make sense to add the checkbox on the first page, because we used to say "DIA with DIA-Umpire" but that was just meant to inform people of an important aspect of the search over which they had no control. If we are going to give them control of that, then all they really need to care about at this stage is that they are telling us they are supplying DIA data.

The point of the original text was to tell them early before they went through those 5 other pages and then suddenly realized we were going to be doing DIA-Umpire doconvolution.

PR #3135, opened 2024-08-26
Fixed an unexpected error loading a spectral library with IMS columns…

brendanx67:
… in RefSpectra and none in the RetentionTimes table. (reported by Bo)


nickshulman:
Skyline is actually looking at the "minorVersion" column in the "LibInfo" table to decide whether to expect to be able to find the ion mobility columns in the RetentionTimes table.
`bool hasGeneralIonMobility = (SchemaVersion >= 6);`
("SchemaVersion" is the value from the "minorVersion" column of the LibInfo table)

I wonder if we should make it so that Skyline looks at the schema version and makes sure that all of the columns that it expects to be present are present and, if not, refuses to open the library at all.
Alternatively, we could make it so that the schema version gets ignored and Skyline figures out what is there based on which columns are present.

PR #3137, opened 2024-08-27
Make sorting by NaturalStringComparer faster

nickshulman:
Fixed slow to display Library Explorer when library has hundreds of thousands of peptides (reported by Bo)

Bo had a .blib file with 600,000 peptides in it.
It took 15 seconds for the Library Explorer to come up because the code for natural sorting was spending a lot of time parsing decimals.
In this pull request I have changed it so that there's a method "NaturalStringComparer.MakeCompareKey" which can be used with methods like "OrderBy" so that the decimal parsing only has to happen once per string, instead of for each comparison.

Also "NaturalStringComparer.Comparer" was not well behaved in the case that part of the string looked like a number but could not be parsed as a Decimal.
That is, there was not a strict ordering of the strings:
```
"01234",
"1234",
"1112222222222222222222222222222222222"
```
There was code that said that if either of the segments of the string could not be converted to a Decimal that they should be compared as strings. I changed that so that if one but not both of the strings could not be converted to a Decimal then the one that could be parsed would sort first.


pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs
```diff
@@ -426,7 +426,7 @@ private void LoadLibrary()
                     {
                         var message = TextUtil.LineSeparate(string.Format(Resources.ViewLibraryDlg_LoadLibrary_An_error_occurred_attempting_to_import_the__0__library, selectedLibrarySpec.Name),
                                         x.Message);
-                        MessageDlg.Show(this, message);
+                        MessageDlg.ShowWithException(this, message, x);
```
bspratt, line 429: Will this show the exception message twice, as it's already part of the value of message?

bspratt (commented):
Seems good, especially as it passes existing unit tests

pwiz_tools/Skyline/SettingsUI/ViewLibraryDlg.cs
nickshulman: > Will this show the exception message twice, as it's already part of the value of message?

When the MessageDlg is displayed, the only thing that the user will see is "message".
The exception would only be shown if the user pressed the "More Info" button.

bspratt (approved):

PR #3141, opened 2024-08-29
Remove "SaveSettings" call from CommandLine "Associate Proteins" and "Import Reports"

nickshulman:
Fixed error using command line to import reports or associate proteins when user's settings folder cannot be written to (reported by Aaron). Importing reports with "--report-add" now requires additional parameter "--save-settings" for reports to persist beyond current Skyline instance.

I am not sure whether we should add anything to the documentation for "--report-add" to mention that "--save-settings" now needs to be specified to get the old behavior.
The documentation for "--save-settings" already strongly implied that "--report-add" did not save settings, because it said:

Argument|Description
---|---
--save-settings | Causes settings changes to the global environment, e.g. added reports, libraries, enzymes, etc. to be saved to the program configuration file.
--report-add=path/to/file.skyr|Add reports from a saved report template file to the running program.


pwiz_tools/Skyline/CommandLine.cs
```diff
@@ -2284,7 +2284,6 @@ private bool AssociateProteins(CommandArgs commandArgs)
                     commandArgs.AssociateProteinsMinPeptidesPerProtein.GetValueOrDefault(),
                     progressMonitor);
                 Settings.Default.LastProteinAssociationFastaFilepath = fastaPath;
-                Settings.Default.Save();
```
chambm, line 2287: Shouldn't this also be conditional on commandArgs.SaveSettings?

nickshulman: I was assuming that it would be taken care of by the "SaveSettings" call in "CommandLine.RunInner" but I will double check and maybe also write a unit test to verify that.

pwiz_tools/Skyline/Test/CmdLineAssociateProteinsTest.cs
```diff
@@ -0,0 +1,93 @@
...
+            var processStartInfo = new ProcessStartInfo(batchFileName, skylineCmdExePath)
+            {
+                WorkingDirectory = workingDirectory,
+                UseShellExecute = false
+            };
+            var outputWriter = new StringWriter();
+            var processRunner = new ProcessRunner();
+            IProgressStatus status = new ProgressStatus(string.Empty);
+            processRunner.Run(processStartInfo, null, null, ref status, outputWriter);
+
```
chambm, line 61: This is quite different from the existing CommandLine tests (which all (or almost all?) start with "Console" in the name). Why call SkylineCmd with ProcessRunner? This approach also means the batch file you're using isn't easy to track changes for. Is there some caveat about saving settings I'm not considering?

chambm: If we keep this form, then perhaps prefixing the test with "TestSkylineCmd..." would make it clear this test actually runs SkylineCmd.

nickshulman: I figured we should have a test which runs SkylineCmd.exe using a .bat file.
I was also thinking that if we wanted to test whether settings were being saved, the best way to test that would be to run SkylineCmd.exe multiple times out of process so that it actually has to read the settings from disk.
The usual way of testing the command line involves calling the command line functions in the same process as the unit tests where it would be trickier to figure out whether the "user.config" settings file was getting saved.

chambm: I agree about running SkylineCmd.exe, I just don't think we need to hide the commands inside a batch file inside a zip file. The commands are simple enough we can keep them in the code.

brendanx67:
Agree. I have just been writing command-line tests where I generate quite a
few of the files on the fly and have less in the ZIP.

pwiz_tools/Skyline/Test/CmdLineAssociateProteinsTest.cs
chambm: So Nick, are you proposing you'll keep the batch file in a future commit to this PR, it'll just be using the .data directory (and thus plaintext in git)? If so, I still wonder why that approach is better than just putting the commands in the code, but there's no question it's better than putting it in a zip.

nickshulman: I wanted the experience of writing a .bat file which called SkylineCmd.exe.
It certainly would be possible to accomplish the same thing with multiple calls to ProcessRunner.Run, but I was thinking that we should have at least one test which uses a .bat, because that is how some users write their scripts.
Is there a reason that you think we should avoid calling SkylineCmd.exe from .bat files during our suite?

chambm: As long as the batch script is in plain text I think it's fine, it's just a little more cumbersome to maintain than having it inline in the code. But as you say, it's useful to have at least one test which runs Skyline this way. :shipit:

PR #3142, opened 2024-08-30
Skyline: Fixed --import-pep-list to turn of auto-manage when any modi…

brendanx67:
…fications are present. (reported by Dan)

- Also added support for multi-peptide lists to Edit > Paste to make this easier to test and make support comparable to --import-pep-list

### DIFF
--- a/pwiz_tools/Skyline/Model/ModificationMatcher.cs
+++ b/pwiz_tools/Skyline/Model/ModificationMatcher.cs
@@ -40,6 +40,8 @@ namespace pwiz.Skyline.Model
         private IProgressStatus _status;
         private const int DEFAULT_ROUNDING_DIGITS = 6;
 
+        public bool HasSeenMods { get; private set; }
+
         public void CreateMatches(SrmSettings settings, IEnumerable<string> sequences,
             MappedList<string, StaticMod> defSetStatic, MappedList<string, StaticMod> defSetHeavy, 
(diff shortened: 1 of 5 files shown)


pwiz_tools/Skyline/Model/ModificationMatcher.cs
```diff
@@ -75,7 +77,7 @@ public override bool MoveNextSequence()
                 return false;
             // Skip sequences that can be created from the current settings.
             // Check first if the sequence has any modifications, because creating doc nodes is expensive
-            while (!HasMods(_sequences.Current) ||
+            while (!HasModsChecked(_sequences.Current) ||
```
nickshulman, line 80: This function could have been simpler if the condition for the "while" were "MoveNextSingleSequence".

I would probably rewrite this function as:
```
        public override bool MoveNextSequence()
        {
            while (MoveNextSingleSequence())
            {
                bool hasMods = HasMods(_sequences.Current);
                HasSeenMods |= hasMods;
                // Skip sequences that can be created from the current settings.
                // Check first if the sequence has any modifications, because creating doc nodes is expensive
                if (hasMods)
                {
                    if (CreateDocNodeFromSettings(new Target(_sequences.Current), null, DIFF_GROUPS, out _) == null)
                    {
                        return true;
                    }
                }
            }

            return false;
        }
```

nickshulman (approved):
Looks good to me.

PR #3144, opened 2024-08-30
Allow test zip files to be folders

nickshulman:
This change makes it so that instead of having your test use a ".zip" file, it can use a folder whose name ends in ".zip" instead.
One thing this would be useful for would be if the .zip file contained some big files which do not change often, but also contains some smaller files that do change.
It also would be useful if the .zip file was small and contained text files and we wanted to make it easier to observe changes in the files inside it.

This pull request contains a couple of small .zip files which have been changed into directories.
I don't know what sort of best practices we would want to use, such as whether the files in the .zip folder should be added individually to the .csproj or added with wildcards.


chambm:
Can the directory versions be named .data, .testdata, or something other than .zip?

nickshulman:
> Can the directory versions be named .data, .testdata, or something other than .zip?

I have not tested that, but it would probably work to name the folder something which ends in something other than ".zip".
I think the test .zip files that we currently use could also have a different filename extension if you really wanted (although that would be a weird thing to do).

The one thing that will not currently work if its name does not end in ".zip" is when SkylineTester creates a .zip file containing Skyline and all of the test data. The method `CreateZipInstallerWindow.FindZipFiles` only enumerates files and folders whose names end in ".zip". If you wanted to use another extension it would be good to add support for it to `CreateZipInstallerWindow.FindZipFiles`.

PR #3151, opened 2024-09-10
Fix a bug in m/z calculation where the mass of the adduct was not being added to molecules described only as a mass with no chemical formula

bspratt:
Reported by user Trevor


nickshulman (approved):
Looks good to me. This should probably be cherry picked back to 24.1

PR #3153, opened 2024-09-10
Improvements to tutorial screenshot taking

eduardo-proteinms:
Work in progress...

In current state a new button is added when a file name is provided to `PauseAndContinueForm`.

image

Currently testing with `AbsoluteQuantTutorialTest` creating images in the new tutorial folder location:
image


pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -1514,7 +1838,12 @@ protected void RunFunctionalTestOrThrow(string defaultUiMode)
 
             UnzipTestFiles();
 
-            _shotManager = new ScreenshotManager(TestContext, SkylineWindow);
+            _shotManager = new ScreenshotManager(SkylineWindow);
+
+            // Run PauseAndContinue ui thread
+            pauseAndContinueThread = new Thread(StartPauseAndContinueForm);
```
brendanx67, line 1844: We need to not create a PauseAndContinueForm until it is first required by a test. Otherwise, the test creates this extra thread and form for all functional tests. In most of our testing, it will never be used.

pwiz_tools/Skyline/Util/UtilIO.cs
```diff
@@ -843,6 +843,25 @@ public static bool IsFile(string path)
             return !IsDirectory(path);
         }
 
+        public static bool IsWriteLocked(string path)
```
nickshulman, line 846: Usually "write locked" would mean that no one can read the file.
This method seems to be checking whether it's possible to write to the file, so a more appropriate name would be "IsReadLocked" or maybe "IsLocked" or "IsUnwritable"

pwiz_tools/Shared/CommonUtil/SystemUtil/FormUtil.cs
```diff
@@ -20,11 +20,14 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Windows.Forms;
 using pwiz.Common.Collections;
 
 namespace pwiz.Common.SystemUtil
 {
+    public enum ScrollDirection { horz = 0, vert = 1 }
```
nickshulman, line 29: The pinvoke website actually uses the existing enum `System.Windows.Forms.Orientation` for this instead of defining a new enum.
https://www.pinvoke.net/default.aspx/user32.setscrollpos

nickshulman (commented):
Looks good to me. Here are a couple of comments.

pwiz_tools/Skyline/Util/UtilIO.cs
brendanx67: I guess I see your point. I prefer a name that doesn't require a ! For use for probably IsWritable is better than IsUnwritable or IsNotWitable. And correct use of IsReadLocked seems confusing.

PR #3154, opened 2024-09-10
Fix unescaped ampersands in msfragger pepXML output

bspratt:


brendanx67:
This seems overly simplistic. What if there is an entity in the XML? This would double escape the ampersand in the entity. Ideally, you would be looking for an ampersand in a specific attribute and even checking for a semicolon within 5 characters (e.g. &quot;)

brendanx67:
You don't want to create an instance of &amp;amp;quot;

bspratt:
Agreed, the Regex is looking only for unescaped ampersands: "&(?!amp;|lt;|gt;|quot;|apos;)"

bspratt:
image

brendanx67:
> Agreed, the Regex is looking only for unescaped ampersands: "&(?!amp;|lt;|gt;|quot;|apos;)"

I  missed that subtlety. Thought it was a string.replace().

That covers the named entities. Maybe that is good enough. For completeness you would have to include:

&amp;#\d\d;
&amp;#x\x\x;

bspratt:
OK, added those and made the whole thing less subtle, hopefully

brendanx67 (approved):

chambm (approved):
Looks good.

PR #3163, opened 2024-09-17
Change tutorial images to count from 1

eduardo-proteinms:
Changed tutorial image counter to start with 1 to align with alt text we are adding


eduardo-proteinms:
closing as we don't see this as important anymore since the figure alt text would not match with this numbering due to shared images and eventual image naming.

PR #3164, opened 2024-09-18
Missing rdotp line (issue 1033)

rita-gwen:


nickshulman (commented):
If the peptide has transitions with multiple charge states, there could be multiple TransitionGroupDocNode's that have a dot product. 

It might make the most sense to show the average dot product across those TransitionDocNode's instead of just showing the dot product from the first TransitionGroupDocNode.

rita-gwen:
Thank you for the review. I re-wrote it to show the average, as suggested.

brendanx67:
Please don't merge yet. I wan't completely satisfied with my review, but go
pulled away to other things.

rita-gwen:
OK

PR #3166, opened 2024-09-19
Target to generate "MergedTutorials.zip"

nickshulman:
I made a tool "TutorialLocalizer" in "pwiz_tools\Skyline\Executables\DevTools".
It looks at the "index.html" and "invariant.html" files in the pwiz_tools\Skyline\Documentation\Tutorials folders and produces a file called "MergedTutorials.zip" which has all of the English tutorials, and then has localized "index.html" files which is the latest version of the English index.html with as much localized text substituted in.
The .zip file also contains "localization_ja.csv" and "localization_zh-CHS.csv" which contain the strings that we would want to send off to the localizers.


chambm:
Have you looked at Executables/Installer/Jamfile.jam? It'll be like that, without the testing stuff. Basically copy your changes into a new Jamfile.jam in that new location, and copy over the build-properties rule as well. Then decide whether you want that to be built every time or only explicitly.

If you want me to do it, no problem, but I won't be able to do it today. I can do it next week.

chambm (approved):

PR #3168, opened 2024-09-24
Deal with ambiguity in TargetResolver where a mass-only molecule had very small differences in iRT library and document…

bspratt:
… (mass serialized as "344.300548579909" vs "344.300548580")


nickshulman:
In the TargetResolver constuctor, there's one code path that does this:
```
                _targetsByName = targets.Select(t => t.ToSerializableString())
                    .Distinct()
                    .Select(Target.FromSerializableString).ToLookup(GetTargetDisplayName);
```
This part of it:
```
Select(t => t.ToSerializableString())
                    .Distinct()
                    .Select(Target.FromSerializableString)
```
round-trips the target to a string, and I think it is important for removing duplicates.

The other code path does not do that:
`_targetsByName = accessions.ToLookup(a => a.Item1, a => a.Item2);`

I'm thinking a better fix might be to change:
`foreach (var target in targets)`
to:
`foreach (var target in targets.Select(t=>t.ToSerializableString()).Distinct().Select(Target.FromSerializableString))`
on line 68 in TargetResolver.cs

bspratt:
I'll have a look, thanks!

bspratt:
@nickshulman OK, a slightly different take on your idea, normalizes everything at the start of the ctor.

nickshulman (approved):

PR #3180, opened 2024-10-03
Skyline: Fixes to translated HTML to match the fixes made to English

brendanx67:
- Fixed all links to other tutorials to go to translated versions when available

PR #3181, opened 2024-10-03
Restructure Tutorial Image naming

eduardo-proteinms:
This restructuring is automated so we can rerun if if we need to change the source before we merge this.

This PR changes the following:
- Moves icon/button images to shared folder
- Names all images which are not coming from Skyline using the format "s-name"
- All other images are renamed / renumbered to the format "s-#"

This will make it simpler to output images from tutorial tests with a counter for now. Also we can use the s- prefix to add alt text / anchors in html programmatically

Example:
image


eduardo-proteinms:
Will rerun script once #3180 is merged

brendanx67:
Looks great. Definitely an improvement on several levels. Please start the s-# with s-1 instead of s-0. These are not indices into an array. I know the # from s-# will not line up with "Figure #", but I still think it is better to have the s-# and Figure # both start with 1.

brendanx67:
I have merged PR #3180

eduardo-proteinms:
> Looks great. Definitely an improvement on several levels. Please start the s-# with s-1 instead of s-0. These are not indices into an array. I know the # from s-# will not line up with "Figure #", but I still think it is better to have the s-# and Figure # both start with 1.

Numbering starts at 1 now. This should be ready for merge.

PR #3183, opened 2024-10-06
Skyline: Wait to import peptide list until after importing peptide search

brendanx67:
- This avoids adding a lot of transitions that will get re-picked after the library is built


brendanx67:
This really needs a test added, but I may ask the FragPipe team to test it for their uses first.

PR #3185, opened 2024-10-07
Transition settings TransitionFilter.Equals was not considering its _smallMoleculeFragmentAdducts member…

bspratt:
…, which caused UI problems when that was the only value being edited (would not save)


bspratt:
@brendanx67 - shall I backport to release branch?

brendanx67:
Sure. Seems pretty safe and only of benefit.

brendanx67 (approved):

PR #3191, opened 2024-10-10
Extension of bruker precursor information (2nd attempt)

xjasg:
Extension of the analysis of Bruker files (precursor information) based on the timsconvert Python code
MS_inverse_reduced_ion_mobility, MS_collisional_cross_sectional_area, MS_peak_intensity.
lowest observed mz and highest observed mz included


pwiz/data/vendor_readers/Bruker/SpectrumList_Bruker.cpp
```diff
@@ -485,6 +490,26 @@ PWIZ_API_DECL SpectrumPtr SpectrumList_Bruker::spectrum(size_t index, DetailLeve
...
+                it = std::max_element(mz.begin(), mz.end());
+                std::size_t index2 = std::distance(mz.begin(), it);
+
+                if (mz.size() > index1)
+                    result->set(MS_lowest_observed_m_z, mz[index1], MS_m_z);
+
+                if (mz.size() > index2)
+                    result->set(MS_highest_observed_m_z, mz[index2], MS_m_z);
+            }
+        }
```
chambm, line 512: The FullMetadata detail level should provide the same metadata as the FullData detail level, and this change would violate that (only FullData would provide lowestMz). This is why we have the spectrum list filter MetadataFixer, if the user really wants these values even though they're not actually provided by the API. So revert this please.

pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.cpp
```diff
@@ -854,16 +862,21 @@ void TimsSpectrum::getIsolationData(std::vector<IsolationInfo>& isolationInfo) c
     if (HasPasefPrecursorInfo())
     {
         const auto& info = GetPasefPrecursorInfo();
-        isolationInfo.resize(1, IsolationInfo{ info.isolationMz, IsolationMode_On, info.collisionEnergy });
+        isolationInfo.resize(1, IsolationInfo{
```
chambm, line 865: Just keep this on one line, thanks.

pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.cpp
```diff
@@ -601,6 +604,11 @@ double TimsDataImpl::oneOverK0ToCCS(double oneOverK0, double mz, int charge) con
     return tims_oneoverk0_to_ccs_for_mz(oneOverK0, charge, mz);
 }
 
+double TimsDataImpl::oneOverK0ToCCS(double oneOverK0, int charge) const
```
chambm, line 607: This does not seem to be used anymore? So this whole overload should be removed along with largestPeakMz.

pwiz/data/msdata/Serializer_MGF.cpp
```diff
@@ -52,13 +52,13 @@ class Serializer_MGF::Impl
...
 };
 
-template <typename T>
-struct nosci10_policy : boost::spirit::karma::real_policies<T>   
-{
-    //  we want to generate up to 10 fractional digits
-    static unsigned int precision(T) { return 10; }
-    //  we want the numbers always to be in fixed format
-    static int floatfield(T) { return boost::spirit::karma::real_policies<T>::fmtflags::fixed; }
+template <typename T>
```
chambm, line 55: Revert whitespace only change.

pwiz/data/msdata/Serializer_MGF.cpp
```diff
@@ -78,22 +78,22 @@ void Serializer_MGF::Impl::write(ostream& os, const MSData& msd,
...
-        {
-            // s = sl->spectrum(i, true);
-            s = spectrumWorkers.processBatch(i);
-        }
-        catch (std::exception& e)
-        {
-            if (continueOnError)
-            {
+    {
+        SpectrumPtr s;
```
chambm, line 82: Revert whitespace only change.

pwiz/data/msdata/Serializer_mzXML.cpp
```diff
@@ -535,7 +535,7 @@ void write_precursors(XMLWriter& xmlWriter, const vector<PrecursorInfo>& precurs
             attributes.add("activationMethod", it->activation);
         if (it->windowWideness != 0)
             attributes.add("windowWideness", it->windowWideness);
-
+
```
chambm, line 538: Revert whitespace only changes.

pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.cpp
```diff
@@ -538,22 +541,22 @@ TimsDataImpl::TimsDataImpl(const string& rawpath, bool combineIonMobilitySpectra
                     {
                         const auto& precursor = frame->pasef_precursor_info_[p];
 
-                            // add MS2s that don't have PASEF info (between last precursor's scanEnd and this precursor's scanBegin)
-                            if (p > 0)
+                        // add MS2s that don't have PASEF info (between last precursor's scanEnd and this precursor's scanBegin)
```
chambm, line 544: This indentation fix is OK.

pwiz/data/msdata/Serializer_MGF.cpp
```diff
@@ -132,6 +132,16 @@ void Serializer_MGF::Impl::write(ostream& os, const MSData& msd,
                 os << " " << intensityParam.valueFixedNotation();
             os << '\n';
 
+            CVParam inverseReduceIonMobility = scan->cvParam(MS_inverse_reduced_ion_mobility);
```
chambm, line 135: The SpectrumListCache crash is due to a null pointer dereference here. Scan may be null so you need to handle that. It might be better to change the way Scan is defined though, like this:
`const Scan& scan = !s->scanList.empty() ? s->scanList.scans[0] : Scan();`

Then change:
`CVParam scanTimeParam = scan ? scan->cvParam(MS_scan_start_time) : CVParam();`
to
`CVParam scanTimeParam = scan.cvParam(MS_scan_start_time);`

And your new code from `scan->` to `scan.`

xjasg:
Okay, thank you very much. I really appreciate that. I would try to fix the things you mentioned.

pwiz/data/msdata/Serializer_MGF.cpp
```diff
@@ -132,6 +132,16 @@ void Serializer_MGF::Impl::write(ostream& os, const MSData& msd,
...
+            if (!inverseReduceIonMobility.empty())
+                os << "INVREION=" << inverseReduceIonMobility.valueFixedNotation();
```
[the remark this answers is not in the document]
xjasg: okay, I´ll check this.

pwiz/data/msdata/Serializer_MGF.cpp
```diff
@@ -132,6 +132,16 @@ void Serializer_MGF::Impl::write(ostream& os, const MSData& msd,
...
+            os << '\n';
+
+            CVParam collisionalCrossSectionalArea = scan->cvParam(MS_collisional_cross_sectional_area);
+            if (!collisionalCrossSectionalArea.empty())
+                os << "COLLCROSSSA=" << collisionalCrossSectionalArea.valueFixedNotation();
```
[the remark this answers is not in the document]
xjasg: okay, Is there an example of adding some parameters to the title string?

pwiz/data/vendor_readers/Bruker/SpectrumList_Bruker.cpp
```diff
@@ -24,6 +24,8 @@
 
 
 #include "SpectrumList_Bruker.hpp"
+
+#include "pwiz/data/vendor_readers/Thermo/ChromatogramList_Thermo.hpp"
```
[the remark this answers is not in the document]
xjasg: don´t know, how this happend :-) I will fix this.

pwiz/data/vendor_readers/Bruker/SpectrumList_Bruker.cpp
xjasg: Sorry, but I was not aware of the Metadatafixer. But that's a good thing, then you don't need my code here. How is the class used? We use msconvert.exe with the appropriate parameters. What would you have to do to get the two parameters (lowest / highest mz) into the result file?

pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.cpp
xjasg: okay.

pwiz/data/msdata/Serializer_MGF.cpp
xjasg: sure, I´ll fix this.

pwiz/data/msdata/Serializer_MGF.cpp
xjasg: okay.

pwiz/data/msdata/Serializer_mzXML.cpp
xjasg: okay.

pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.cpp
xjasg: okay.

pwiz/data/msdata/Serializer_MGF.cpp
xjasg: Oh, I tried to find the reason. Thank you very much. I changed a lot of things, but the build still crashed. That saves me a lot of time.

pwiz/data/msdata/Serializer_MGF.cpp
chambm: I'm still seeing whitespace changes here. You know how to review the diff in GitHub? You can also review it locally if you set your diff engine to show whitespace changes.

pwiz/data/msdata/Serializer_MGF.cpp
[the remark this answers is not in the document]
chambm: There's a titleMaker filter that allows for easy customization of the MGF title string with a user-defined format. Although it doesn't support CCS currently, only instrument-level ion mobility metrics. If you don't want to use that, you can add "ccs=<value>" to the title string above this, before the line:
            if (!scanTimeParam.empty())
                os << "RTINSECONDS=" << scanTimeParam.timeInSeconds() << '\n';

pwiz/data/msdata/SpectrumList_MGF.cpp
```diff
@@ -235,6 +235,16 @@ class SpectrumList_MGFImpl : public SpectrumList_MGF
                                 else
                                     selectedIon.set(MS_selected_ion_m_z, value, MS_m_z);
                             }
+                            else if (name == "INVREION")
+                            {
+                                bal::trim(value);
+                                scan.cvParams.push_back(CVParam(MS_inverse_reduced_ion_mobility, value, MS_volt_second_per_square_centimeter));
+                            }
+                            else if (name == "COLLCROSSSA")
```
chambm, line 243: When moving CCS to the title, it can be parsed back in the `name == TITLE` block above.

pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.cpp
```diff
@@ -862,21 +854,16 @@ void TimsSpectrum::getIsolationData(std::vector<IsolationInfo>& isolationInfo) c
     if (HasPasefPrecursorInfo())
     {
         const auto& info = GetPasefPrecursorInfo();
-        isolationInfo.resize(1, IsolationInfo{
-	                             info.isolationMz, IsolationMode_On, info.collisionEnergy,
-	                             info.intensity
```
chambm: You reverted your intensity population here.

pwiz/data/msdata/Serializer_MGF.cpp
xjasg: yes, sorry, but unfortunately it's not finished yet. Yes, It's easy to see whitespace changes, but undoing them without removing other important things is a bit difficult for me. But I´m on it. Thank you.

pwiz/data/msdata/Serializer_MGF.cpp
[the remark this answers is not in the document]
xjasg: Okay, that sounds good, I'll implement it that way.

pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.cpp
xjasg: Thank you for finding this. I will fix it.

pwiz/data/msdata/Serializer_MGF.cpp
xjasg: Whitespaces should now be removed.

pwiz/data/msdata/Serializer_mzXML.cpp
xjasg: Done.

pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.cpp
xjasg: Done.

pwiz/data/msdata/Serializer_MGF.cpp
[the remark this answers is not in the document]
xjasg: I hope it´s done now. :-)

pwiz/data/msdata/Serializer_MGF.cpp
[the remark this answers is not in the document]
xjasg: Should be finished.

pwiz/data/msdata/Serializer_MGF.cpp
xjasg: Should be removed.

pwiz/data/msdata/SpectrumList_MGF.cpp
xjasg: Done.

chambm (changes requested):
Please add ion mobility and CCS cvParams in Serializer_MGF_Test so this new code gets tested. Maybe for just one of the spectra there so it gets tested with an without.

xjasg:
@chambm Thank´s for your review and comment. yes, okay, I´ll write some tests.

xjasg:
@chambm Thank´s for your review and comment. yes, okay, I will write some tests.

chambm:
You don't need to add any new test, just update the existing one in Serializer_MGF_Test by adding the cvParams to one of the spectra there.

xjasg:
@chambm okay.

pwiz/data/msdata/Diff.hpp
```diff
@@ -228,14 +228,17 @@ struct PWIZ_API_DECL DiffConfig : public pwiz::data::BaseDiffConfig
 
     bool ignoreDataProcessing;
 
+    bool ignoreSpectrumTitle;
+
```
chambm, line 232: Why is this necessary? I would expect the new CCS-in-title code to round trip the CCS so the title will be preserved and not different. What diffs were you getting without this?

pwiz/data/msdata/SpectrumList_MGF.cpp
```diff
@@ -379,6 +379,9 @@ class SpectrumList_MGFImpl : public SpectrumList_MGF
             }
         }
 
+        if(ccsStr.empty())
+            ccsStr = title.substr(start,  title.length() - start);
```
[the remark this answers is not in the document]
xjasg: okay, looks good, I change this.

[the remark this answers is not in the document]
xjasg: Done.

xjasg:
@chambm: Hi Matt, have you had a chance to look at the latest changes again? Should I change anything else? Thank you very much. Best regards.

pwiz/data/msdata/Diff.hpp
chambm: Sorry I dropped the ball here. What did you mean "the reference does not contain the text"?

chambm: Did you see this reply @xjasg ?

xjasg: I replied, but my answers were marked as "pending".

Here is what I wrote:

"I have to apologise. I had a lot to do in another project. As far as I understand it, the following happens: testWriteRead(const MSData& msd) uses msd as a reference. msd contains previously generated test data with a specific title. ‘serializer.write(oss, msd);’ extends the title by “ccs=...” and saves the new extended data with the new title. Then "serializer.read(iss, msd2);" reads back the written data with the new title. But after that the comparison with msd fails because of the different title. That is why I have introduced a flag that disables the comparison of the titles. Maybe its better to changed the test than the main program. I`ll try this.

I think I have now understood the problem. The additional text must be removed from the title after loading. Then everything should work. Otherwise, the title could be expanded more than once if we save and load the data twice or more. I think we would have to do the same with other title parameters."

My last commit should have resolved this issue?

chambm:
Thanks @xjasg and @Andre99999999 !

xjasg:
@chambm : I have to thank you for your kind support.

PR #3193, opened 2024-10-14
Dotp indication in Prosit mirror plot (issue 831)

rita-gwen:
https://skyline.ms/issues/home/issues/details.view?issueId=831


pwiz_tools/Skyline/Model/Koina/Models/KoinaModel.cs
```diff
@@ -705,26 +705,79 @@ public static double CalculateSpectrumDotpMzMatch(LibraryRankedSpectrumInfo spec
...
+        /// Calculates the normalized contrast angle of the square rooted
+        /// intensities of the two given spectra using all peaks and possibly ignoring
+        /// mismatched peaks
+        /// </summary>
+        /// <param name="spectrum1">First spectrum</param>
+        /// <param name="spectrum2">Second spectrum</param>
+        /// <param name="mzTolerance">Tolerance for considering two mzs the same</param>
+        /// <param name="ignoreMismatches1">if true, unmatched peaks from the first spectrum are ignored</param>
+        /// <param name="ignoreMismatches2">if true, unmatched peaks from the second spectrum are ignored</param>
+        public static double CalculateSpectrumDotpMzFull(LibraryRankedSpectrumInfo spectrum1,
```
nickshulman, line 718: I think `CalculateSpectrumDotpMzMatch` should be changed to call `CalculateSpectrumDotpMzFull` passing in `false` for `ignoreMismatches1` and `ignoreMismatches2` so that there is less duplicated code.

I think the implementation in "CalculateSpectrumDotpMzMatch" is a lot cleaner in terms of iterating over the "RankedMI" lists in the two spectra. I think it would be better if "CalculateSpectrumDotpMzFull" looked more like the way "CalculateSpectrumDotpMzMatch" looks like now.

pwiz_tools/Skyline/Model/Koina/Models/KoinaModel.cs
```diff
@@ -705,26 +705,79 @@ public static double CalculateSpectrumDotpMzMatch(LibraryRankedSpectrumInfo spec
...
+                (mz, i) => new SpectrumPeaksInfo.MI() { Mz = mz, Intensity = (float)i }).ToList();
+
+            var intensities1All = new List<double>(vectorLength);
+            var intensities2All = new List<double>(vectorLength);
+            var matchIndex1 = 0;
+            var matchIndex2 = 0;
+            while (matchIndex1 < spectrum1.Intensities.Count() && matchIndex2 < spectrum2.Intensities.Count())
             {
-                var other = matched2.Where(m => m.MatchedIons.Intersect())
+                switch (mzTolerance.CompareWithTolerance(s1[matchIndex1].Mz, s2[matchIndex2].Mz))
```
nickshulman, line 733: I think the way that `CalculateSpectrumDotpMzMatch` does this with:
```
if (mzTolerance.IsWithinTolerance(mz1, mz2))
{
...
}
else if (mz1 < mz2)
{
...
}
else
{
...
}
```
is easier to understand than this way of doing a `switch` on the return value of `CompareWithTolerance`.

pwiz_tools/Shared/CommonUtil/Chemistry/MzTolerance.cs
```diff
@@ -85,6 +85,17 @@ public bool IsWithinTolerance(double a, double b)
             return (a >= b - this) && (a <= b + this);
         }
 
+        public int CompareWithTolerance(double a, double b)
```
nickshulman, line 88: I would recommend removing this method because it's entirely obvious just from the method's name what exactly it's going to do.
If you were going to keep this method, I think its implementation should make use of `IsWithinTolerance`:
```
        public int CompareWithTolerance(double a, double b)
        {
            if (IsWithinTolerance(a, b))
            {
                return 0;
            }

            return a < b ? -1 : 1;
        }
```

pwiz_tools/Skyline/Model/Koina/Models/KoinaModel.cs
```diff
@@ -705,26 +705,79 @@ public static double CalculateSpectrumDotpMzMatch(LibraryRankedSpectrumInfo spec
...
+            LibraryRankedSpectrumInfo spectrum2, MzTolerance mzTolerance, bool ignoreMismatches1, bool ignoreMismatches2)
         {
-            var matched1 = spectrum1.PeaksMatched.ToArray();
-            var matched2 = spectrum2.PeaksMatched.ToArray();
-            var intensities1All = new List<double>(matched1.Length + matched2.Length);
-            var intensities2All = new List<double>(matched1.Length + matched2.Length);
-
-            foreach (var match1 in matched1)
+            var vectorLength = spectrum1.Intensities.Count() + spectrum2.Intensities.Count();
+            var s1 = Enumerable.Zip(spectrum1.MZs, spectrum1.Intensities,
```
nickshulman, line 722: This is almost the same as `s1=spectrum1.Peaks.ToList()`, except that the code here produces a List of MI whereas "Peaks" is a list of RankedMI.

pwiz_tools/Skyline/Model/Koina/Models/KoinaModel.cs
```diff
@@ -662,69 +662,77 @@ public static bool KoinaSettingsValid
...
-                    intensities2All.Add(matched2[matchIndex2].Intensity);
-
-                    ++matchIndex1;
-                    ++matchIndex2;
+                    intensities1All.Add(s1[matchIndex1++].Intensity);
+                    intensities2All.Add(s2[matchIndex2++].Intensity);
                 }
-                else if (mz1 < mz2)
+
+                if (mz1 < mz2)
```
nickshulman, line 708: I am pretty sure this needs to be "else if (mz1 < mz2)" because otherwise "matchIndex1" or "matchIndex2" will potentially get incremented twice.

nickshulman (commented):
Looks almost perfect.

PR #3196, opened 2024-10-16
Update TestDiaXxx tutorial audit logs

nickshulman:


brendanx67 (approved):
Huh. Looks like just the capitalization of "q" presumably in "q value". I thought you might have fixed it already, but realized that perf tests only get run in one language and it looks like they only got broken in Japanese.

PR #3197, opened 2024-10-16
Retry up to 10 times when error requesting Koina predictions

nickshulman:
Here's one way that we could deal with errors that we get from the Koina server.
However, given that the failure always seems to happen on the 1000th request, it should probably be fixed on the server side.

This change makes it retry up to 10 times.

Other changes:
1. Fix inefficient use of "Enumerable.Skip" in "EnumerateBatches". When this code was originally written in 2019 it seems that the developer did not realize that "Skip" requires iterating over all of the skipped elements. This fix might make the "EnumerateBatches" part fast enough that it would not make sense to be doing it in ParallelEx.
2. Speed up "RefreshProgressTextbox". It was doing too many string concatenations
3. I added an application setting to Settings.settings which then required fixing inconsistencies in App.config. Brian Pratt is probably going to fix those inconsistencies in a separate pull request.


pwiz_tools/Skyline/FileUI/PeptideSearch/SearchControl.cs
```diff
@@ -142,8 +143,11 @@ protected void UpdateSearchEngineProgress(IProgressStatus status)
             }
 
             // look at the last 10 lines for the same message and if found do not relog the same message
-            if (_progressTextItems.Skip(Math.Max(0, _progressTextItems.Count - 10)).Any(entry => entry.Message == message))
```
chambm, line 145: It seems this blatant missing optimization has been in .NET since 2016 or so, but alas doesn't seem to have made it into Framework.

chambm (approved):

pwiz_tools/Skyline/app.config
```diff
@@ -690,7 +690,7 @@
             <setting name="MixedMoleculeSettingsTab" serializeAs="String">
                 <value>0</value>
             </setting>
-            <setting name="Prosit" serializeAs="String">
+            <setting name="Koina" serializeAs="String">
```
brendanx67, line 693: Leave these changes for Brian Pratt's static analysis PR. Stick to only the one settings change required for this PR.

PR #3198, opened 2024-10-17
added static code inspection for values that don't agree between Settings.settings and app.config

bspratt:
a little bit of voodoo around settings that are collections of strings - Settings Designer may not create an app.config entry if the collection is empty. On the other hand, if the collection is non-empty then edited to be empty, the app.config entry will exist, but with an empty string as its value.

also had to change "Prosit" to "Koina" in app.config

### DIFF
--- a/pwiz_tools/Skyline/Properties/Settings.settings
+++ b/pwiz_tools/Skyline/Properties/Settings.settings
@@ -869,6 +869,9 @@
     <Setting Name="FullScanPropertySheetVisible" Type="System.Boolean" Scope="User">
       <Value Profile="(Default)">False</Value>
     </Setting>
+    <Setting Name="FeatureFindingMinIntensityPPM" Type="System.Double" Scope="User">
+      <Value Profile="(Default)">100.0</Value>
+    </Setting>
     <Setting Name="FeatureFindingMinIdotP" Type="System.Double" Scope="User">
       <Value Profile="(Default)">0.9</Value>
     </Setting>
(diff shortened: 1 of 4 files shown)


pwiz_tools/Skyline/Properties/Settings.settings
```diff
@@ -886,6 +883,9 @@
     </Setting>
     <Setting Name="RelativeAbundanceLogScale" Type="System.Boolean" Scope="User">
       <Value Profile="(Default)">True</Value>
-    </Setting>	
+    </Setting>
+    <Setting Name="FeatureFindingMinIntensityPPM" Type="System.Double" Scope="User">
```
brendanx67, line 887: Change this value in place rather than letting it jump to the end.

pwiz_tools/Skyline/app.config
```diff
@@ -873,9 +870,15 @@
...
             </setting>
+            <setting name="GroupComparisonSuspendLabelLayout" serializeAs="String">
+                <value>False</value>
+            </setting>
             <setting name="RelativeAbundanceLogScale" serializeAs="String">
-                 <value>True</value>
-            </setting>			
+                <value>True</value>
+            </setting>
+            <setting name="FeatureFindingMinIntensityPPM" serializeAs="String">
```
brendanx67, line 879: Change this value in place rather than letting it jump to the end.

brendanx67 (changes requested):
One minor change to avoid unnecessary diffs in app.config and Settings.settings. Otherwise, assuming the code finds the things you fixed and passes after you fixed them, this looks good.

brendanx67 (approved):

PR #3202, opened 2024-10-18
Add better exception handling in transition list input - in particular make the FileModifiedException more informative

bspratt:
In response to https://skyline.ms/announcements/home/issues/exceptions/thread.view?entityId=96760a55-6efc-103d-a666-22f53556a042&_anchor=66300

I don't have a repro for the problem, unfortunately

### DIFF
--- a/pwiz_tools/Skyline/Model/SrmDocument.cs
+++ b/pwiz_tools/Skyline/Model/SrmDocument.cs
@@ -1509,6 +1509,17 @@ namespace pwiz.Skyline.Model
                         Replace(TextUtil.SEPARATOR_TSV_STR, @" ");
                     errorList.Add(new TransitionImportErrorInfo(x.PlainMessage, x.ColumnIndex, x.LineNumber + 1, line));  // CONSIDER: worth the effort to pull row and column info from error message?
                 }
+                catch (Exception ex)
+                {
+                    if (ExceptionUtil.IsProgrammingDefect(ex))
+                    {
+                        throw; // Let it fall through so we hear about it on ExceptionWeb
+                    }
+                    else
+                    {
+                        errorList.Add(new TransitionImportErrorInfo(ex.Message, null, null, string.Empty));
+                    }
+                }
             }
             else
             {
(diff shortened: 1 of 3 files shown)


pwiz_tools/Skyline/Model/SrmDocument.cs
```diff
@@ -1509,6 +1509,17 @@ public SrmDocument ImportMassList(MassListInputs inputs,
                         Replace(TextUtil.SEPARATOR_TSV_STR, @" ");
                     errorList.Add(new TransitionImportErrorInfo(x.PlainMessage, x.ColumnIndex, x.LineNumber + 1, line));  // CONSIDER: worth the effort to pull row and column info from error message?
                 }
+                catch (Exception ex)
```
brendanx67, line 1512: I don't think this is the right place to put this. This is supposed to be for issues with parsing a line, not for global errors. You should be catching other exceptions higher up the callstack.

pwiz_tools/Skyline/Model/SrmDocument.cs
```diff
@@ -1544,6 +1555,17 @@ public SrmDocument ImportMassList(MassListInputs inputs,
                 {
                     errorList.Add(new TransitionImportErrorInfo(x));
                 }
+                catch (Exception ex)
```
brendanx67, line 1558: I don't think this is the right place to put this. This is supposed to be for issues with parsing a line, not for global errors. You should be catching other exceptions higher up the callstack.

brendanx67 (changes requested):

pwiz_tools/Skyline/Model/SrmDocument.cs
bspratt: Catching it here is advantageous as it ties into the reporting mechanism for both the UI and commandline use cases.

pwiz_tools/Skyline/Model/SrmDocument.cs
bspratt: Catching it here is advantageous as it ties into the reporting mechanism for both the UI and commandline use cases.

bspratt:
@brendanx67 do you still want to change this? I could add two sets of error handling (UI and commandline) if you feel strongly about it.

PR #3203, opened 2024-10-22
Fix a problem with reading small molecule transition lists, where precursors with similar m/z but distinct isotope labels were being combined

bspratt:
The logic to simplify messy redundant transition lists wasn't thinking about labels properly: e.g.

Molecule Name,Molecular Formula,Precursor Adduct,Precursor Charge 
Glutamine,C5H10N2O3,M2C13-H,-1
Glutamine,C5H10N2O3,M1C131N15-H,-1
Glutamine,C5H10N2O3,M1C131H2-H,-1

would result in a single precursor C5H10N2O3[M2C13-H] as all have similar m/z - but of course they have different isotope envelopes. 

Reported by user Jamie


bspratt:
@nickshulman @brendanx67 this is a well-tested code area, I have high confidence that this change doesn't have unwanted side effects for the kinds of transition lists we've already seen

nickshulman (approved):
I don't really understand this but it looks good to me.

PR #3205, opened 2024-10-24
Address translation feedback

eduardo-proteinms:
see [localization.ja_JA-JP_MI_KS.xlsx](https://github.com/user-attachments/files/17513943/localization.ja_JA-JP_MI_KS.xlsx)


pwiz_tools/Skyline/Model/Results/ResultsResources.resx
```diff
@@ -333,7 +333,7 @@ It is recommended that you delete this file so that Skyline can create a new fil
     <value>Discarding chromatograms for '{0}' because the explicit retention time {1} is not between {2} and {3}</value>
   </data>
   <data name="IonMobilityFinder_EvaluateBestIonMobilityValue_need_a_value_for_Transition_Settings___ion_mobility_filtering___Window_type" xml:space="preserve">
-    <value>need a value for  Transition Settings  Ion mobility filtering  Window type</value>
+    <value>need a value for Ion mobility filtering Window type in Transition Settings</value>
```
brendanx67, line 336: We tend to want to direct people from the top down. So, in sentence form, this would be:
In the Transition Settings - Ion Mobility tab, the Ion mobility filtering, Window type field must have a value.

If I were reviewing all these messages from that perspective, there are more issues with them.

brendanx67 (approved):
One comment, take it for what it is worth. I am not sure I would bother with the last one, if we are not going to do more general fixes for these messages. It is not a big enough improvement, I think. You could also just add commas.

eduardo-proteinms:
> need a value for Transition Settings Ion mobility filtering Window type

makes sense, will add commas.

PR #3206, opened 2024-10-28
TestRunner: if more than one leak type is detected, log them all rather than just the first

bspratt:
internal test issue, not reported by any user

### DIFF
--- a/pwiz_tools/Skyline/TestRunner/Program.cs
+++ b/pwiz_tools/Skyline/TestRunner/Program.cs
@@ -192,17 +192,18 @@ namespace TestRunner
 
             public string GetLeakMessage(LeakTracking leakThresholds, string testName)
             {
+                var message = string.Empty;
                 if (ManagedMemory >= leakThresholds.ManagedMemory)
-                    return string.Format("!!! {0} LEAKED {1:0.#} Managed bytes\r\n", testName, ManagedMemory);
+                    message += string.Format("!!! {0} LEAKED {1:0.#} Managed bytes\r\n", testName, ManagedMemory);
                 if (HeapMemory >= leakThresholds.HeapMemory && !MutedHeapMemoryLeakTestNames.Contains(testName))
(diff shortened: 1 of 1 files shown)


chambm (approved):

brendanx67:
Have you tested this with the testresults module to make sure it produces the desired behavior for nightly test runs? I suppose as long as it doesn't break import it will be as good as current for summary reporting and possible to learn more from the logs... Though, you can already learn more from looking at the logs since we log all rates of leaking at the end of leak checking a test whether it passes or fails.

bspratt:
> Have you tested this with the testresults module to make sure it produces the desired behavior for nightly test runs? I suppose as long as it doesn't break import it will be as good as current for summary reporting and possible to learn more from the logs... 

I haven't, as I don't know how to provoke multiple (any, really) leaks. As I read the code, it shouldn't hurt we already can report multiple kinds of leaks (Handle, CRT, "Leak") for a single test. It would be good to break "Leak" into different memory types, but possibly that would break whatever downstream tool is reading the report. Maybe the thing to do is leave that alone, and introduce new categories for backward compatibility.

> Though, you can already learn more from looking at the logs since we log all rates of leaking at the end of leak checking a test whether it passes or fails.

I'm not really sure what's a leak and what isn't just from looking at the log: e.g. this snippet with a heap leak, but maybe also something else ("memory = 23.4 KB")?

```
[06:39]   1.597 TestManageResultsUndo                          (fr)   0 failures, 79.75/400.21/719.8 MB, 141/11598 handles, 1 sec.
[06:39]   1.597 TestManageResultsUndo                          (tr)   0 failures, 79.75/400.24/719.8 MB, 141/11598 handles, 1 sec.
[06:39]   1.597 TestManageResultsUndo                          (ja)   0 failures, 79.75/400.28/719.8 MB, 141/11599 handles, 1 sec.
!!! TestManageResultsUndo LEAKED 28454.9 Heap bytes
# TestManageResultsUndo deltas (25): managed = -6.9 KB, heap = 27.8 KB, memory = 23.4 KB, user-gdi = -0.9, total = -3.4
```

PR #3210, opened 2024-11-01
Library spectrum match graph tooltip implementation

rita-gwen:


pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
```diff
@@ -131,6 +133,47 @@ public MzRange GetMzRange(SpectrumControlType controlType)
...
+            LibraryRankedSpectrumInfo.RankedMI _peakRMI;
+            public ToolTipImplementation(LibraryRankedSpectrumInfo.RankedMI peakRMI)
+            {
+                _peakRMI = peakRMI;
+            }
+            public bool HasTip => _peakRMI != null;
+
+            public Size RenderTip(Graphics g, Size sizeMax, bool draw)
+            {
+                if (HasTip)
```
nickshulman: When the entire body of a function is inside of an if statement, I usually try to arrange things to reduce the amount that the code has to be indented.
That would be:
```
if (!HasTip)
{
   return Size.Empty;
}
// rest of the body of the function
```

pwiz_tools/Shared/zedgraph/ZedGraph/GraphPane.cs
```diff
@@ -2385,8 +2385,126 @@ curve is HiLowBarItem || curve is OHLCBarItem ||
...
+                        if (nearestStick != null && minDist <= Default.NearestTol)
+                        {
+                            return true;
+                        }
+                    }
+                }
+            }
+            return false;
+        }
+        public static int Between<T>(T x, T low, T high) where T : IComparable<T>
```
nickshulman: I would recommend getting rid of this method.
The problem with implementing a method "Between" is that the order of parameters and return value are not something that is universally agreed upon.
The natural ordering of arguments is "low, middle, high", but a case can also be made for the "middle, low, high" that you have here.

pwiz_tools/Shared/zedgraph/ZedGraph/GraphPane.cs
```diff
@@ -2385,8 +2385,126 @@ curve is HiLowBarItem || curve is OHLCBarItem ||
...
+                            // yVal is the user scale Y value of the current point
+                            if (yAxis._scale.IsAnyOrdinal && !stickItem.IsOverrideOrdinal)
+                                yVal = (double)iPt + 1.0;
+                            else
+                                yVal = points[iPt].Y;
+
+                            if (xVal != PointPair.Missing &&
+                                yVal != PointPair.Missing)
+                            {
+                                if (Between(yAct, 0, yVal) != 0)
```
nickshulman: Is this equivalent to `if (yAct < 0 || yAct > yVal)`?

pwiz_tools/Skyline/Controls/Graphs/GraphsResources.resx
```diff
@@ -544,6 +544,18 @@ Here are the last 3 errors:
   <data name="GraphSpectrum_UpdateUI_Failure_loading_spectrum__Library_may_be_corrupted" xml:space="preserve">
     <value>Failure loading spectrum. Library may be corrupted.</value>
   </data>
+  <data name="GraphSpectrum_ToolTip_mz">
+    <value>Observed M/z:</value>
```
nickshulman: The "m" and the "z" must have the same case as each other.

In the Document Grid, "M/Z" is written in upper case, but I'm not sure that's correct.
I see that many menu items have it in lowercase.
I'm not sure whether this should be "Observed M/Z:" or "Observed m/z:"

pwiz_tools/Skyline/Controls/Graphs/GraphsResources.resx
```diff
@@ -751,4 +763,7 @@ Here are the last 3 errors:
   <data name="GraphChromatogram_UpdateUI_No_quantitative_chromatograms_found" xml:space="preserve">
     <value>No quantitative chromatograms found</value>
   </data>
+  <data name="ToolTipImplementation_RenderTip_Calculated_Mass" xml:space="preserve">
+    <value>Calculated M/z</value>
```
nickshulman: However, I wonder this should be "Ion m/z" because the context menu item which causes these numbers to be shown on the graph is "Ion m/z Values"
image

pwiz_tools/Skyline/Controls/Graphs/SpectrumGraphItem.cs
```diff
@@ -343,15 +343,22 @@ private string GetLabel(LibraryRankedSpectrumInfo.RankedMI rmi)
...
         }
 
-        private string GetLabel(MatchedFragmentIon mfi, int rank, bool showMz)
+        public static string GetMassErrorString(LibraryRankedSpectrumInfo.RankedMI rmi, MatchedFragmentIon mfi)
+        {
+            var sb = new StringBuilder();
+            var massError = mfi.PredictedMz - rmi.ObservedMz;
+            massError = SequenceMassCalc.GetPpm(mfi.PredictedMz, massError);
+            massError = Math.Round(massError, 1);
+            sb.Append(string.Format(Resources.GraphSpectrum_MassErrorFormat_ppm, (massError > 0 ? @"+" : string.Empty), massError));
```
nickshulman: The StringBuilder here is unnecessary.
This could just be:
`return string.Format(Resources.GraphSpectrum_MassErrorFormat_ppm, (massError > 0 ? @"+" : string.Empty), massError);`

pwiz_tools/Skyline/Model/SpectrumPropertiesResx.resx
```diff
@@ -157,6 +160,18 @@
   <data name="Description_SpectrumCount" xml:space="preserve">
     <value>Number of measured spectra for this peptide in the library</value>
   </data>
+  <data name="Description_PeakCount">
+    <value>Number of non-zero peaks in the library spectrum</value>
```
nickshulman: I don't think "non-zero peaks" means anything (specifically, there's no such thing as a zero peak).

I might say "Number of m/z values in the spectrum that had non-zero intensity", but I also might be wrong.

pwiz_tools/Shared/zedgraph/ZedGraph/GraphPane.cs
```diff
@@ -2385,8 +2385,126 @@ curve is HiLowBarItem || curve is OHLCBarItem ||
...
+            double[] y;
+            double[] y2;
+            ReverseTransform(mousePt, out x, out x2, out y, out y2);
+
+            if (!AxisRangesValid())
+                return false;
+
+            double yPixPerUnitAct, yAct, yMinAct, yMaxAct, xAct;
+            double minDist = 1e20;
+            double xVal, yVal, dist = 99999, distX, distY;
```
nickshulman: dist, distX and distY are not used.

By default, ReSharper does not perform code analysis on the files in the ZedGraph project but you can right-click on the pause button in the top right corner of the code editor and choose "Resume Analysis"
image

There are quite a few unused local variables in this function.

rita-gwen: The original ZedGraph code has too many code inspection problem, so I usually keep it disabled there. I cleaned that FindNearestStick method from the unused variables, though.

pwiz_tools/Shared/zedgraph/ZedGraph/GraphPane.cs
```diff
@@ -2385,8 +2385,92 @@ curve is HiLowBarItem || curve is OHLCBarItem ||
...
+                        if (nearestStick != null && minDist <= Default.NearestTol)
+                        {
+                            return true;
+                        }
+                    }
+                }
+            }
+            return false;
+        }
+        private double LogScalePixelDistance(double positionA, double positionB, double totalDistance)
```
nickshulman: I know this isn't part of this pull request, but I don't trust the math in this "LogScalePixelDistance".

Typically, the way to figure out the pixel distance between two coordinate values is to call "axis.Scale.Transform" on both of them and subtract the two results. That works for all types of scales.

pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
```diff
@@ -1324,23 +1377,25 @@ private void DoUpdate()
...
                         double? fullDotp = null;
                         SpectrumGraphItem mirrorGraphItem = null;
                         if (mirrorSpectrum != null)
                         {
                             var peaksInfo = spectrum != null
                                 ? RescaleMirrorSpectrum(mirrorSpectrum, spectrum)
                                 : mirrorSpectrum.SpectrumPeaksInfo;
-                            mirrorGraphItem = MakeGraphItem(mirrorSpectrum, selection, settings, peaksInfo);
-                            mirrorGraphItem.Invert = true;
+                            MirrorGraphItem = MakeGraphItem(mirrorSpectrum, selection, settings, peaksInfo);
```
nickshulman, line 1417: The local variable `mirrorGraphItem` is still being used in a few places in this function and causes NullReferenceException's.
Those need to be changed to the capital `MirrorGraphItem`

pwiz_tools/Skyline/Controls/Graphs/SpectrumGraphItem.cs
```diff
@@ -343,15 +343,21 @@ private string GetLabel(LibraryRankedSpectrumInfo.RankedMI rmi)
...
-                sb.AppendLine().Append(string.Format(Resources.GraphSpectrum_MassErrorFormat_ppm, (massError > 0 ? @"+" : string.Empty), massError));
+                sb.AppendLine().Append(GetMassErrorString(rmi, rmi.MatchedIons.First()));
             }
             return sb.ToString();
         }
 
-        private string GetLabel(MatchedFragmentIon mfi, int rank, bool showMz)
+        public static string GetMassErrorString(LibraryRankedSpectrumInfo.RankedMI rmi, MatchedFragmentIon mfi)
+        {
+            var sb = new StringBuilder();
```
nickshulman: This StringBuilder is no longer used.

pwiz_tools/Shared/zedgraph/ZedGraph/GraphPane.cs
rita-gwen: Yes, your intuition was right, as usual. That function returned incorrect values. Removed.

pwiz_tools/Skyline/Skyline.csproj
```diff
@@ -216,15 +216,11 @@
     <Compile Include="Alerts\AlertLinkDlg.Designer.cs">
       <DependentUpon>AlertLinkDlg.cs</DependentUpon>
     </Compile>
-    <Compile Include="Alerts\ArdiaLogoutDlg.cs">
-      <SubType>Form</SubType>
-    </Compile>
+    <Compile Include="Alerts\ArdiaLogoutDlg.cs" />
```
brendanx67: This file looks like all changes that are not relevant to your PR.

pwiz_tools/Skyline/TestTutorial/TargetedMSMSTutorialTest.cs
```diff
@@ -1241,6 +1241,7 @@ private void TestPropertySheet()
 
         private void TestLibraryMatchPropertySheet()
         {
+            //TODO: [RC] Fix tests, add some tooltip testing code
```
brendanx67: Is there any testing of this new code? Would prefer to see this removed before you merge, and ideally clear new testing added. Please run code coverage and make sure your new code gets hit and covered well during some test run.

brendanx67 (changes requested):

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/TestTutorial/TargetedMSMSTutorialTest.cs
rita-gwen: Added the testing code and run the code coverage. Looks like the new tooltip code is exercised well.
image

pwiz_tools/Skyline/TestTutorial/TargetedMSMSTutorialTest.cs
```diff
@@ -1360,5 +1361,47 @@ private static void ValidateScoresMatch(double expectedScore, PropertyGrid pg, V
             Assert.AreEqual(expectedScore.ToString(CultureInfo.CurrentCulture), score?.ToString());
             Assert.AreEqual(expectedScore, dlg.GraphItem.SpectrumInfo.Score);
         }
+
+        private void ValidatePeakTooltips()
+        {
+            // 191, 95; 169, 175
+            var testData = new Dictionary<Point, string>()
+            {
+                {new Point(191, 95), "Observed m/z:\t951.6229\nIntensity:\t3814516\nRank:\t1\nMatched Ions\tIon m/z (calculated)\ny8\t951.4782  -152.1 ppm"},
```
nickshulman: This test seems to fail in French ~~, Chinese and Japanese~~.

rita-gwen: Thank you, Nick. I think it's the numbers formatting.

PR #3218, opened 2024-11-14
Fix a problem with ProteinMetadataManager hanging while trying to resolve protein metadata for peptide lists…

bspratt:
…, which often aren't resolvable because they aren't named as actual proteins (e.g. "Library Peptides")


pwiz_tools/Skyline/Model/Proteome/ProteinMetadataManager.cs
```diff
@@ -164,14 +164,23 @@ private SrmDocument LookupProteinMetadata(SrmDocument docOrig, IProgressMonitor
                                     }
                                     else if (nodePepGroup.ProteinMetadata.NeedsSearch())
                                     {
-                                        void CheckBackgroundProteome(FastaSequence seq, ProteinMetadata proteinMetadataOrGroup, int i)
+                                        for (var i = 0; i < nodePepGroup.ProteinMetadata.ProteinMetadataList.Count; i++)
                                         {
+                                            var name = nodePepGroup.IsPeptideList ?
```
brendanx67, line 169: Ick. Can you possibly use polymorphism? Give this a base-class implementation and then have the FastaSequence version do what it needs to do to get the name you are truly looking for?

brendanx67 (changes requested):
Can you define a virtual function for getting the type of name you are looking for and use polymorphism?

pwiz_tools/Skyline/Model/Proteome/ProteinMetadataManager.cs
bspratt: I think not - it doesn't make any sense to ask for the i'th name in a peptide list, it only has one name because it doesn't have a FastaSequenceList. (For a peptide list that loop count is always 1.)

brendanx67: Not going to let this one go through until it has received more thought. Seems like clearly poor design to me. If the information is in a parallel list of names to the ProteinMetadataList, then you should instead be returning a list of structs with names and protein metadata instances, so that you don't need to rely on these parallel lists that you have to know so much to decipher. Improve the design, don't go digging like this based on knowledge you shouldn't be required to have from outside the PeptideGroupDocNode class.

bspratt: This is all about the protein association work, which I wasn't part of. Sounds like you want to kick this back to the original author?

chambm: One part I'm not sure about Brian is how much of these cases are really necessary if IsPeptideList is true. How likely is it that `proteomeDb.GetProteinByName(nodePepGroup.Name)` return something other than null for a peptide list? If that's basically impossible, IsPeptideList could be handled outside this ProteinMetadataList loop.

bspratt: A peptide list *may* have a name that's actually a protein identifier for which we can get metadata.

bspratt: I'd like to get this off my plate. Can you guys @chambm @brendanx67  reach a conclusion on this?

chambm: How does something get marked a peptide list? Do we actually want it to get metadata if it happens to match a protein id or name, even if it has peptides that aren't related to the protein?

bspratt: Import Transition List and Spectral Library Add, for instance - though Import Transition List is the more likely case for a bunch of peptides sharing a group name which is recognizable by web services like UniProt. 

By adding in plausible metadata we aren't making any assertions about the appropriateness of the group membership that weren't already implied by using an obviously recognizable protein name.

bspratt: (Note that often as not, we aren't even hitting the internet - just parsing the metadata in names that are canonically structured strings)

PR #3222, opened 2024-11-17
iRT UI fixes

nickshulman:
Fix some bugs when defining an iRT standard using peptides in the current document:
1. Prevent duplicate peptides from being added to IrtDb.DocumentXml
2. Correctly use "PeptideDocNode.ModifiedTarget" instead of "PeptideDocNode.Target" so that peptides with modifications (e.g C[+57]) work correctly.

Other iRT fixes:
1. Suppress "Are you sure you want to create a new database file you will lose changes" message in EditIrtCalcDlg if there is no existing database. Fix performance problem in IrtDbManager.cs
2. Fix message text of not enough standards message so that it says how many standards are required.


nickshulman:
I have copied parts of this pull request into PR #2165.

PR #3225, opened 2024-11-18
* updated macos CI image since macos-12 is no longer available

chambm:


chambm:
Unfortunately don't have time to work on this. I welcome contributions from anybody in the community that actually builds pwiz on MacOS. :)

PR #3228, opened 2024-11-19
Fix tracked files being deleted by clean-apps.bat

nickshulman:
Rename "bin" folder in "pwiz_tools\Skyline\Executables\DevTools\Utf16to8" to "binaries" so that it does not get deleted by clean-apps.bat

I could not find anything in the source tree that might care whether that folder was named "binaries" or "bin", but I might have missed something.


chambm:
If there's nothing in the source tree that uses these binaries (to avoid needing to build the project), why are they even checked in?

nickshulman:
> If there's nothing in the source tree that uses these binaries (to avoid needing to build the project), why are they even checked in?

Good question.
It's probably so that people can run "Utf16to8.bat" on .sky.view files if they ever want to do so in the future.
.sky.view files used to be saved in UTF-16, but that was fixed a couple of weeks ago so going forward they will all be UTF-8.
Committing UTF-16 files to the project is problematic because they get treated as binary.
This only became a problem when checking in test files to ".data" folders so this tool was written to convert .sky.view files.

ekoneil:
Correct. Brendan added this tool recently as part of the larger project
upgrading the tutorials. As far as I know, it's not tied into the build -
rather, we run it manually on the .view files contained in
pwiz_tools\Skyline\TestTutorial\*.zip.

chambm:
Huh I totally missed that issue. I would have thought there was a way to have git treat UTF-16 files as text.

Probably too late to mess with, but what about this in .gitattributes?
`*.sky.view text diff working-tree-encoding=UTF-16LE eol=CRLF`
From https://stackoverflow.com/a/54683489/638445

chambm:
Anyway since it's intended to only be used by devs on demand, this fix looks entirely appropriate. Let's keep bin/obj directories out of the .NET project tree.

PR #3231, opened 2024-11-21
Update Boost to 1.86

chambm:
Boost 1.76 is no longer working on newer GCC/Clang versions.


chambm:
Unfortunately `/clr` with `<mutex>` is apparently broken on the TC build agents' version of VS2022. Works on TCA1 and my local machine. I'll have to update the build agent AMI next week. But hey, at least the Linux and MacOS builds are passing... :)

PR #3241, opened 2024-11-26
Update Custom Reports tutorial

ekoneil:


pwiz_tools/Skyline/SkylineGraphs.cs
```diff
@@ -5539,6 +5539,11 @@ public LiveResultsGrid CreateResultsGrid()
             return _resultsGridForm;
         }
 
+        public LiveResultsGrid GetLiveResultsGrid()
```
brendanx67, line 5542: Probably best to do this as a property as we have done with other forms like this. I can't think of another that uses a function.

pwiz_tools/Skyline/TestTutorial/CustomReportsTutorialTest.cs
```diff
@@ -494,14 +495,23 @@ protected bool DoQualityControlSummaryReports()
             RunUI(SkylineWindow.ShowPeakAreaReplicateComparison);
             WaitForGraphs();
 
-            PauseForPeakAreaGraphScreenShot("Peak Areas view");
+            PauseForScreenShot<GraphSummary.AreaGraphView>("Peak Areas View");
```
brendanx67, line 498: Though, the class name is AreaGraphView, we generally do not capitalize "View" in referring to our views. The name "Peak Areas" gets capitalized because it is a "proper name" that appears in the UI. But the term "view" does not appear in the UI, it is just something we use to qualify the proper name.

pwiz_tools/Skyline/TestTutorial/CustomReportsTutorialTest.cs
```diff
@@ -494,14 +495,23 @@ protected bool DoQualityControlSummaryReports()
...
             return true;    // Continue subsequent tests
         }
 
         protected void DoResultsGridView()
         {
             // Results Grid View
             RestoreViewOnScreen(26);
+            RunUI(() =>
+            {
+                var areaGraphFloatingWindow = FindFloatingWindow(SkylineWindow.GraphPeakArea);
```
brendanx67, line 508: I wonder if these values should be relative to the bottom-right of the Skyline window. That is how their placement looked to me in the screenshot. Consider, if we wanted to make the Skyline window a little bigger for some reason, what edges would we want this window placement anchored to?

pwiz_tools/Skyline/SkylineGraphs.cs
ekoneil: Makes sense - good call.

pwiz_tools/Skyline/TestTutorial/CustomReportsTutorialTest.cs
ekoneil: Typo, my bad - good catch.

pwiz_tools/Skyline/TestTutorial/CustomReportsTutorialTest.cs
ekoneil: Yes, will fix.

pwiz_tools/Skyline/TestTutorial/CustomReportsTutorialTest.cs
```diff
@@ -482,7 +484,28 @@ protected bool DoQualityControlSummaryReports()
             PauseForScreenShot<ViewEditor.FilterView>("Customize View - Filter tab", 23);
             OkDialog(viewEditor, viewEditor.OkDialog);
 
-            RunUI(() => documentGridForm.DataGridView.Columns[6].Width = 105);
+            RunUI(() =>
```
brendanx67, line 487: Note that something like this can also go inside a RunUIForScreenShot() call to keep from having to do such minor irrelevant tweaks during normal test runs.

ekoneil: Updated. And used in a few other places. Thx for the pointer.

brendanx67 (approved):
Did a bit of cleanup. Looks ready to go to me. Thanks.

ekoneil:
> Did a bit of cleanup. Looks ready to go to me. Thanks.

Cleanup makes sense. Good changes. Thanks.

PR #3243, opened 2024-11-26
Recent changes (commit 0a7a9390c) to peptide search UI broke the feature detection UI

bspratt:
TestFeatureDetectionTutorialFuture caught the problem.


bspratt:
It's pretty slow, I think a smoke test that might have caught this particular problem would be a whole different animal and wouldn't cover enough to be a useful addition. I think it's fair to say the system worked - though I do personally try to run a complete set of tests locally before a merge of more than moderate complexity, given that we do have that skip list for TC.

chambm:
Yeah I definitely ran a full suite of tests during development but I must have let it get stale, i.e. made some changes after it and only ran selected tests then.

chambm:
Is the problem that Hardklor won't give usable (for testing) output for a very small mzML file?

bspratt:
Really it's that the test does a lot of start/stop/cancel stuff, and does it all more than once to work with different initial conditions. Just not worth the effort to mess with that test any further.

PR #3246, opened 2024-11-27
Fix opposite ticks on dotp axis

eduardo-proteinms:


brendanx67 (changes requested):
While I hugely appreciate this fix, I found it easiest to test simply by adding it to the PR I was working on where I had already moved this particular block to its own function. I worked great! At this point merging this PR would just cost me a slight amount more work dealing with the conflict it would create with my larger PR.

chambm:
Just FYI you can change the target branch for this pr to your branch. That
might reduce the conflict (but maybe not).

brendanx67:
Still. It just seems like more work that simply closing this PR. It is just two lines and I had to put them in the right place in my branch to test them where I was seeing the problem most visibly. I will look into the kind of linked PR you are suggesting another time.

brendanx67:
This fix has been merged in PR#3247

PR #3249, opened 2024-11-28
Skyline: Attempt a fix for the deadlock in TestMs1Tutorial

brendanx67:
Make BackgroundScanProvider communicate with GraphFullScan through BeginInvoke instead of Invoke to avoid the case where GraphFullScan tries to Join the background thread just as it is trying to Invoke to the UI thread.


nickshulman (approved):
Looks good to me.

PR #3253, opened 2024-12-02
Mobilion reader uses HDF5 library, which isn't thread safe. Put a processwide lock in the ctor to ensure only one thread is in use at a time

bspratt:
Looks a bit funny when parallel processing multiple .mbi files in Skyline and MSConvertGUI, as only one proceeds at a time, but it does complete


bspratt:
For the record, this is a design call made by MattC and myself, in anticipation of Mobilion resolving the issue eventually. Not worth the effort to build a "not suitable for parallel use" sense into the UI of either application.

chambm:
Eventually I should add an IterationListener parameter for Reader::read() so the user can be informed why it's paused. Could be useful for other readers as well if/when the `fillInMetadata()` function is a little slower than we'd like. Skyline isn't set up to listen and report those messages, but MSConvertGUI is.

PR #3257, opened 2024-12-03
Update PRM Tutorial

ekoneil:
* Automate screenshot 20 which was manually created.
* Add a new .zip file containing .sky/d files with data used to generate screenshot 20's 5-point calibration curve.
* Minor fixes in other screenshots explicitly set active replicate and use labels to show dot products


brendanx67 (approved):
I decided the dopt inconsistency would be confusing in the tutorial. So, I used the library information from the 2-run file to repopulate the 5-run file and get consistent dotp values and no warning message opening the file.

I think it is ready to merge.

ekoneil:
I see - and copying the .blib file into Extras.zip. Should have tried that myself. :) Thanks for fixing the dotp problem and doing the commit.

PR #3266, opened 2024-12-06
Skyline: Cleanup persistent folder before and after test runs

brendanx67:
- Helps avoid the case where a prior failure causes an echo failure in the next run because cleanup deletes left over files in the persistent folder

### DIFF
--- a/pwiz_tools/Skyline/SkylineTester/SkylineTesterWindow.cs
+++ b/pwiz_tools/Skyline/SkylineTester/SkylineTesterWindow.cs
@@ -1959,6 +1959,7 @@ namespace SkylineTester
             else
             {
                 coverageCheckbox.Enabled = true;
+                runMode.SelectedItem = "Test";  // Only Test mode supported in parallel testing
             }
         }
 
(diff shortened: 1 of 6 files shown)


chambm (approved):
It's great to make it more robust to failures. There are probably more tests that need CleanupPersistentDir() overloads added though, right? (Do you want me to take that on?)

brendanx67:
Yes. Feel free to add them. I am focusing on the 24 officially published
tutorials. Feeling like we need an indicator for whether a tutorial is
officially published as a lot has crept into the list that is not yet
officially published.

Though maybe that is already ending in Tutorial for officially published
and I just need to downgrade the tutorials that are not official to ending
in TutorialDraft

pwiz_tools/Skyline/SkylineTester/SkylineTesterWindow.cs
```diff
@@ -1991,6 +1991,8 @@ private void runMode_SelectedIndexChanged(object sender, EventArgs e)
                 {
                     runLoops.Checked = true;
                     runLoopsCount.Text = 1.ToString();
+                    Offscreen.Checked = false;  // Can't do screenshots offscreen
```
bspratt, line 1994: This doesn't appear to be re-enabled when the control is switched back to Test

bspratt: Probably want a behavior more like what you get when parallel mode is turned on/off

brendanx67: Good catch. I think I fixed the potential inconsistencies the code was allowing before.

PR #3269, opened 2024-12-09
-update Sciex Data API to 1.5.1.391 (May 2024)

david-cox-sciex:
Update Sciex Data API to 1.5.1.391 (May 2024).


david-cox-sciex:
Hi Matt,

Yes, I see the issue with closing and deleting/renaming files. I'm working with the Data API team to see if it can be addressed quickly. I'll try some example dlls in this PR for a bit. Once I have something working, I'll wait for the Data API release, and then use that version.

david-cox-sciex:
I have the new API working. It adds support for recognizing Echo TOF instruments, and probably more improvements since the 2023 version. 
The issue with locking files was solved by calling CloseFile on the API. 
I need to wait for an official release of the API, then I will update this PR with it. 
If you see changes I can make while I'm waiting for that updated dll, let me know.

brendanx67:
The release consideration makes me wonder if we have a license for this version?

brendanx67:
Was meaning to cancel my consideration, but now that it is part of the thread, Tim Blacker has raised the need to update our license in the past. Do you have an estimate for when an "official release" might happen?

david-cox-sciex:
I was hoping to use the current release, but the issue with file locking stopped me. I'd rather wait for the next release so it handles these locked files in a better way. 
I'll confirm with Tim Blacker before putting the final API dll in for this review. I think the release of this next API is going to be very soon, but releases can get held up for reasons that happen in meetings I don't get invited to (thankfully).

pwiz/data/vendor_readers/ABI/SpectrumList_ABI.cpp
```diff
@@ -363,6 +363,9 @@ namespace {const SpectrumIdentity emptyIdentity;}
 size_t SpectrumList_ABI::size() const {return 0;}
 const SpectrumIdentity& SpectrumList_ABI::spectrumIdentity(size_t index) const {return emptyIdentity;}
 size_t SpectrumList_ABI::find(const std::string& id) const {return 0;}
+
+void SpectrumList_ABI::close() {}
```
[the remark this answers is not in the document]
david-cox-sciex: I think so. I'll remove the close and handle it by destructor.

pwiz/data/vendor_readers/ABI/ChromatogramList_ABI.cpp
```diff
@@ -52,6 +52,11 @@ PWIZ_API_DECL ChromatogramList_ABI::ChromatogramList_ABI(const MSData& msd, Wiff
 {
 }
 
+PWIZ_API_DECL ChromatogramList_ABI::~ChromatogramList_ABI()
+{
+    wifffile_->close();
```
chambm, line 57: The *List_ABIs have shared_ptr to a WiffFileImpl (or WiffFile2Impl). It's WiffFile2Impl's destructor that should call close, because it'll be called when both *Lists are destroyed. Calling close manually in the *List destructors is unsafe as the other List may still be using the WiffFile.

david-cox-sciex: WiffFile2Impl destructor was calling close, but then this part of VendorReaderTestHarness was preventing the file from being renamed later. I'm not sure how to handle it.

            if (msd_reverse.run.spectrumListPtr.get())
                for (size_t j = 0, end = msd_reverse.run.spectrumListPtr->size(); j < end; ++j)
                    msd_reverse.run.spectrumListPtr->spectrum(end - j - 1);

chambm: Are you saying that there's a WiffFile2Impl that hasn't been destroyed by the time the rename part of the test happens? Or there's some other way it's preventing the rename? Because I'm not sure how the WiffFile2Impl would survive after msd_reverse goes out of scope (that should destroy both its spectrumListPtr and chromatogramListPtr and thus destroy the shared_ptr<WiffFile2Impl> they hold).

david-cox-sciex: I can't figure out why Wiff2Impl isn't being destroyed in that part of the test. There is no issue on the similar line:
 bfs::remove_all(newRawPath); // remove the copy of the RAW file with non-ASCII characters

but without close in the SpectrumList destructor, the line:
 bfs::rename(rawpath, rawpath + ".renamed");

fails. 
There should be no issue with calling close multiple times. It is only closing a SQL connection. I think keeping the close in SpectrumList and ChromatogramList is okay. Another option would be to make a special case in VendorReaderTestHarness similar to:
if (bfs::exists(bfs::path(rawpath) / "Analysis.yep")

Let me know which option you prefer or if you can think of something else.

brendanx67:
Do you have a virtual destructor? Always define your destructors as
virtual. Otherwise, they can be missed when deleting a pointer to a base
class.

pwiz/data/vendor_readers/ABI/ChromatogramList_ABI.cpp
chambm: I'll take a look at this.

david-cox-sciex:
It does have a virtual destructor. Do we need to call managed garbage collector directly during test?

brendanx67:
Or if the managed class implements IDisposable and gets put in a using
clause. That is the intent of the using clause, to provide deterministic
clean-up during scope clean-up that C++ programmers became so dependent on
with smart pointers and destructors that releases resources automatically
during scope clean-up.

pwiz/data/vendor_readers/ABI/ChromatogramList_ABI.cpp
chambm: The destructor with its CloseFile call was always getting run as expected, but the CloseFile() call wasn't effective every time for some reason. I was able to fix that by running GC::Collect() before CloseFile(), like this:
```
~WiffFile2Impl()
{
    auto dataReader = DataReader();
    if (dataReader != nullptr)
    {
        System::GC::Collect();
        dataReader->CloseFile(((IList<ISample^>^) allSamples)[0]->Sources[0]);
    }
    System::GC::Collect();
}
```

Can you test whether that works for you as well (and you don't need the close() function anymore)?

Other than that, every time Reader_ABI_Test runs, the WIFF2 API is adding 160MB of debug text to a log file at C:\ProgramData\SCIEX\logs\defaultlog_Reader_ABI_Test.log

Can we disable that? If not I can delete the log file at the beginning of every test to prevent it building up, but it won't let me do it after the test (even in an atexit() function) because the log file is locked.

david-cox-sciex:
Reader_Agilent_Test is failing, but I don't see how any of my changes could impact that test. Any ideas?

pwiz/data/vendor_readers/ABI/ChromatogramList_ABI.cpp
david-cox-sciex: The extra GC::Collect works, thanks for figuring that out. I'll reverse the explicit calling of Close() from the other places.

I'll look into the log file size. I think there is a way to configure logging in an app.config file.

david-cox-sciex: Adding a config section for the logging will stop DataAPI from excessive logging. However, it requires including log4net.dll. I'm not sure how to include this for the Reader_ABI_Test.exe program.

chambm: In vendor_api/ABI/Jamfile.jam's vendor-api-requirements:
`        result += <assembly-dependency>$(PWIZ_ROOT_PATH)/pwiz_tools/Shared/Lib/log4net.dll ;`

in install-requirements:
`        result += <source>$(PWIZ_ROOT_PATH)/pwiz_tools/Shared/Lib/log4net.dll ;`

See if that works.

david-cox-sciex: I've found a way to embed the OFX.Logging.dll and log4net.dll in the Data API assembly; this enables configuration of logging. Works on the Reader test, I'll run TC to confirm other tests are passing.

david-cox-sciex: Well, that seems to break some Skyline tests that use logging. I've asked the Sciex API group to look into it, but it will likely be a fix for the next version. 
The current release of MSConvert with the very old Data API also has the large log files. The December 2024 Data API also has the large log files, but still fixes the Echo MS and file locking issues.
Can we proceed with the Dec 2024 Data API?

david-cox-sciex:
The new Data API reports SWATH isolation windows correctly. Previously it would show a Q1 value offset from the centre of the window. Now it shows a correct centre value with symmetric window around it.
I'll see if I can update the test comparison files.

brendanx67:
FYI. There is now a related PR #3992 

I think Matt was intending this new PR to be what we merge. He just told me his internet is down, but you two should work together on this. Thanks for the active work today!

chambm:
> FYI. There is now a related PR #3992
> 
> I think Matt was intending this new PR to be what we merge. He just told me his internet is down, but you two should work together on this. Thanks for the active work today!

Yes I took your changes from this PR, squashed them, updated to the newer API Tim sent, and updated the reference mzML with the improved DIA boundaries. The DIA change seems pretty significant. I wouldn't be surprised if it changed Skyline results pretty substantially when importing WIFF2 DIA (WIFF1 wouldn't be affected).

PR #3271, opened 2024-12-10
Quameter/peak symmetry

keesh0:
Motivation: 
Helps ascertain if there are disproportionately more MS2 triggers which in turn inflates the spectral counts for that protein.  
This alters the ability to assess peak area and contributes to more variability in the assays.  
We would hope for more peaks to be centered near the 1-line (line slope=1).

A convenient Jupyter notebook for example plotting is available from [keesh@ieee.org](mailto:keesh@ieee.org) .
Less changes would have been necessary if not for the near duplicate data structures (i.e. SlimCrawPeak, 
CrawPeak, Peak, CrawdadPeak).


keesh0:
Anything I can do to move this pull request forward?

chambm:
Sorry for the big delay reviewing this. Please revert the whitespace changes to make the changes easier to review and while you're at it remove the // BEG and //END comments (if I want to know where some code came from I'll just use `git blame`). I'm not a big fan of the extra output file. If it's important that the existing output columns stay the same, you can enable the new columns with a command-line option. Other than that it looks fine, thanks!

PR #3272, opened 2024-12-10
Tweaks to deal with new warnings from latest ReSharper.

bspratt:
This one is very popular:
[https://www.jetbrains.com/help/resharper/UsageOfDefaultStructEquality.html](https://www.jetbrains.com/help/resharper/UsageOfDefaultStructEquality.html) 
It suggests that default generated runtime comparisons for structs are usually slow, and we should use our own code for equality checks. I've done so, but I worry a bit that we now have more code to maintain in the event that a member is added to a struct. (Interestingly in the act of doing this, I noticed that Model.TransitionGroup.LossId already has its own equality comparer, but doesn't include all its members in that - I've left a CONSIDER comment there)

### DIFF
--- a/pwiz_tools/Shared/Common/DataAnalysis/Matrices/QrFactorizationCache.cs
+++ b/pwiz_tools/Shared/Common/DataAnalysis/Matrices/QrFactorizationCache.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using MathNet.Numerics.LinearAlgebra;
 using MathNet.Numerics.LinearAlgebra.Double;
 
@@ -42,7 +43,7 @@ namespace pwiz.Common.DataAnalysis.Matrices
 
 
 
-        public struct Key
+        public struct Key : IEquatable<Key>
         {
             public Key(Matrix<double> matrix, double tolerance) : this()
             {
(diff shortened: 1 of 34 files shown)


bspratt:
Brendan isn't convinced that all these new equality methods are worth the churn, and I'm not either. Backing those out and busting that Warning down to a Suggestion is certainly a reasonable option.

bspratt:
BTW @brendanx67 that curious equality method I mentioned in Model.TransitionGroup.LossId is yours - it looks like we're forbidding its use between precursor and fragment, but I wonder if we want to see if fragment types are equal (and if we don't, maybe we can have a comment saying why we don't)

dshteyn:
I guess the way to verify is to time the execution with and without the change and see the actual difference in runtimes. This might entail a greater effort than it is worth.

bspratt:
>> This might entail a greater effort than it is worth.
I think that's the feeling, yeah.

brendanx67:
Our general rubric is 1) make the code as readable and maintainable as
possible, 2) focus on performance in bottleneck situations and prove the
value of more complex code with profiling statistics.

brendanx67:
Especially in UI, you could waste a lot of time and complexity optimizing
code that has no visible impact on the user experience.

bspratt:
I'm going to issue a new PR with a clean history, and that warning busted down to a suggestion.

PR #3273, opened 2024-12-11
Fix for the label clipping in the label layout algorithm

rita-gwen:


pwiz_tools/Skyline/Controls/Graphs/SummaryRelativeAbundanceGraphPane.cs
```diff
@@ -158,7 +158,13 @@ private void UpdateFormatting(IEnumerable<MatchRgbHexColor> colorRows)
         public void OnLabelOverlapPropertyChange(object sender, PropertyChangedEventArgs e)
         {
             if (e.PropertyName == @"GroupComparisonAvoidLabelOverlap")
```
nickshulman, line 160: This should probably have been:
```
if (e.PropertyName == nameof(Settings.Default.GroupComparisonAvoidLabelOverlap))
```
so that "Find References" and other sorts of refactoring would work a little bit better.

pwiz_tools/Skyline/Controls/Graphs/SummaryRelativeAbundanceGraphPane.cs
```diff
@@ -158,7 +158,13 @@ private void UpdateFormatting(IEnumerable<MatchRgbHexColor> colorRows)
...
+            {
+                Settings.Default.PropertyChanged -= OnLabelOverlapPropertyChange;
+                Settings.Default.GroupComparisonSuspendLabelLayout = false;
+                _labelsLayout = null;
                 GraphSummary.UpdateUI();
+                Settings.Default.PropertyChanged += OnLabelOverlapPropertyChange;
```
nickshulman, line 166: If I have to write code which removes a listener and adds it back I usually put it inside of a try/finally:
```
try 
{
    Settings.Default.PropertyChanged -= OnLabelOverlapPropertyChange;
// do other stuff
}
finally
{
    Settings.Default.PropertyChanged += OnLabelOverlapPropertyChange;
}
```
That way, even if some sort of exception happens, things still get back into an expected state.

For something like this, I would probably just have a field:
`private bool _inPropertyChange;
`
and then in the property change handler:
```
if (_inPropertyChange) 
{
   return;
}
try 
{
    _inPropertyChange = true;
    if (e.PorpertyName == nameof(Settings.Default.GroupComparisonAvoidLabelOverlap)) 
    {
        // do some stuff
    }
    else if (e.PropertyName == nameof(Settings.Default.GroupComparisonSuspendLabelLayout)) 
    {
        // do other stuff
    }
}
finally 
{
    _inPropertyChange = false;
}
```
But the way you have this is fine too.

pwiz_tools/Shared/zedgraph/ZedGraph/LabelLayout.cs
```diff
@@ -319,10 +326,13 @@ public bool PlaceLabel(LabeledPoint labPoint, Graphics g)
             var roughGoal = goal;
             // Search the cell neighborhood for a better position
             var goalPoint = _densityGrid[goalCell.Y][goalCell.X]._location;
-            for (var count = 15; count > 0; count--)
+            var chartRect = _graph.Chart.Rect;
+            var allowedRect = new RectangleF(chartRect.X + labelRect.Width / 2, chartRect.Y,
```
nickshulman, line 330: I am not sure I understand why `labelRect.Width / 2` gets added to the X-coordinate, but a similar sort of thing does not get added to the Y-coordinate.
This code might be correct, but I wanted to point out that it looks suspicious.

nickshulman (commented):
This looks good to me.
It definitely seems to lay out better than before.
Here are some comments which may be helpful. If they're not helpful then I would say don't worry about it and go ahead and merge to master.

pwiz_tools/Shared/zedgraph/ZedGraph/LabelLayout.cs
rita-gwen: This is because label coordinates are top-middle, so Y coordinate is not adjusted.

rita-gwen:
The most recent commit should fix the label clipping in small-size plots in the DIA tuturials. This is how it looks: 
DIATtof tutorial
image
DIAPacef tutorial:
image

PR #3275, opened 2024-12-11
Tweaks to quiet the latest ReSharper version's code inspection.

bspratt:
Note in particular the new "UsageOfDefaultStructEquality" warning has been reduced to a suggestion - it wants equality members for structs, because the ones generated automatically are probably inefficient. Worth thinking about when looking at performance issues but not worth a general change to many dozens of files.


nickshulman (approved):
Looks good to me.

PR #3276, opened 2024-12-11
Fix a handful of tests which were accessing UI elements on the test thread, as detected in a debug build.

bspratt:
Note in particular the new convenience function in TestUtil.AbtractFunctionalTest

        /// <summary>
        /// Convenience function for getting a value on the UI thread
        /// e.g. var value = CallUI(() => control.Value);
        /// </summary>
        public T CallUI<T>([InstantHandle] Func<T> func)


pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -334,6 +334,27 @@ public static void RunUI([InstantHandle] Action act)
             });
         }
 
+        /// <summary>
+        /// Convenience function for getting a value on the UI thread
+        /// e.g. var value = GetValueUI(() => control.Value);
+        /// </summary>
+        public T GetValueUI<T>([InstantHandle] Func<T> act)
```
nickshulman, line 341: I would recommend naming this function "CallUI".
It does the same thing as "RunUI", except that it takes a Func instead of an Action.

You might as well make use of "RunUI" in the implementation:
```
        public static T CallUI<T>([InstantHandle] Func<T> func)
        {
            T result = default;
            RunUI(()=>result = func());
            return result;
        }
```

nickshulman (approved):

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
bspratt: I'll try that, thanks.

bspratt: Yes, much tidier, thanks.

bspratt:
@nickshulman good to go?

bspratt:
@nickshulman oh, wait, I see it now, thanks

PR #3278, opened 2024-12-13
Add support for pepxml files with search_engine = Percolator

poshul:
OpenMS' Percolator adapter saves the search_engine as Percolator directly. As a result it's not possible to load OpenMS generated pepxmls that have Percolator postprocessing into Skyline. This patch fixes that.

### DIFF
--- a/pwiz_tools/BiblioSpec/src/PepXMLreader.cpp
+++ b/pwiz_tools/BiblioSpec/src/PepXMLreader.cpp
@@ -144,7 +144,8 @@ void PepXMLreader::startElement(const XML_Char* name, const XML_Char** attr)
    else if (state == STATE_PROPHET_SUMMARY && isElement("inputfile", name)) {
       // Count files for use in reporting percent complete
       numFiles++;
-   } else if(isElement("msms_run_summary",name)) {
+   }
+   else if(isElement("msms_run_summary",name)) {
       fileroot_ = getRequiredAttrValue("base_name",attr);
       Verbosity::comment(V_DEBUG, "PepXML base_name is %s", fileroot_.c_str());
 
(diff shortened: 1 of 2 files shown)


pwiz_tools/BiblioSpec/src/PepXMLreader.cpp
```diff
@@ -419,7 +426,8 @@ void PepXMLreader::startElement(const XML_Char* name, const XML_Char** attr)
            (analysisType_ == PROTEOME_DISCOVERER_ANALYSIS && scoreType_ == MASCOT_IONS_SCORE && score_name == "exp-value") ||
            (analysisType_ == MORPHEUS_ANALYSIS && score_name == "psm q-value") ||
            (analysisType_ == MSGF_ANALYSIS && score_name == "qvalue") ||
-           (analysisType_ == CRUX_ANALYSIS && score_name == "percolator_qvalue")) {
+           (analysisType_ == CRUX_ANALYSIS && score_name == "percolator_qvalue")
+           (analysisType_ == PERCOLATOR_ANALYSIS)) {
```
chambm, line 430: Looks like a missing `||` operator? How did this work?

chambm (changes requested):
Thanks for the PR. It needs a test added though. Easiest way would be taking one of the existing Percolated pepXMLs and changing the search_engine (and removing post-processor parameter if OpenMS doesn't set that). Better would be getting a real OpenMS pepXML but cutting it down to be <1MB. Is it clear enough how to add a test when looking at the tests/Jamfile.jam? (when you run the test the first time it will create a reference/xxx.observed file; just change that .check after reviewing it to be sure it's correct and then the test should pass).

pwiz_tools/BiblioSpec/src/PepXMLreader.cpp
poshul: And this is why I shouldnt push changes on a friday afternoon. Fixed.

poshul:
> Thanks for the PR. It needs a test added though. Easiest way would be taking one of the existing Percolated pepXMLs and changing the search_engine (and removing post-processor parameter if OpenMS doesn't set that). Better would be getting a real OpenMS pepXML but cutting it down to be <1MB. Is it clear enough how to add a test when looking at the tests/Jamfile.jam? (when you run the test the first time it will create a reference/xxx.observed file; just change that .check after reviewing it to be sure it's correct and then the test should pass).

I've been working on the test for this, a meaningful search with decoys, in a small mzml file, that produces better results than "nothing found" is proving to be a pain. I'm not going to have a chance to circle back to this before the holidays, so I'm putting a note in my calendar to come back in early Jan.

PR #3279, opened 2024-12-13
Enable running perftests via Test Explorer in Visual Studio

nickshulman:
This change makes it so the default value of "RunPerfTests" is true. This means that if you run the test in Visual Studio using either Test Explorer or ReSharper's Unit Test Explorer, the test will actually get run. This change does not affect TestRunner (or SkylineTester) because TestRunner always explicitly sets RunPerfTests to false if it has not been specified on the command line.

Also, this change makes the default for "RunSmallMoleculeTestVersions" true as well.

What this does mean is that if you tell Visual Studio to run all of the tests, the perf tests will get run in addition to all of the other tests. My understanding is that nobody really ever tells Visual Studio to run all the tests anymore (probably because it's faster to run all the tests in parallel via SkylineTester).

This change makes it much easier to debug a single perf test, or even to run a single perf test with code coverage via ReSharper.


chambm:
I usually have TestRunner as my startup project and just change its debug commandline (which always has perftests=on) to test=whatever. Then I can F5 or Ctrl-F5 for easy running/debugging, which you can't do with Visual Studio testing AFAIK unless your editor focus is in the test method.

With the old way of doing code coverage on TC using vstest, this would have caused perftests to run under code coverage which would have been bad. But the current way uses TestRunner with perftests=off, so it won't affect that.

I wonder how this would affect the "continuous coverage" mode which was accidentally enabled on one of my VS sessions recently. It was pretty unusable even without perf tests though. :)

bspratt:
I'd be in favor of this, but as one who uses TestRunner/SkylineTester pretty much exclusively I'll defer to others.

pwiz_tools/Skyline/TestUtil/AbstractUnitTest.cs
```diff
@@ -117,7 +117,7 @@ protected bool RecordAuditLogs
         private static string[] SmallMoleculeDevelopers = {"BSPRATT"}; 
         protected bool RunSmallMoleculeTestVersions
         {
-            get { return TestContext.GetBoolValue("RunSmallMoleculeTestVersions", false) || SmallMoleculeDevelopers.Any(smd => Environment.MachineName.Contains(smd)); }
+            get { return TestContext.GetBoolValue("RunSmallMoleculeTestVersions", true) || SmallMoleculeDevelopers.Any(smd => Environment.MachineName.Contains(smd)); }
```
bspratt, line 120: If we're defaulting this to true then we can drop the special treatment for BSPRATT machines.

PR #3284, opened 2024-12-18
Improvements to LogScale in bar graphs

nickshulman:
1. Correct the Y-axis range when displaying values that are less than 1.
image
"CurveList.GetRange" used to think that all bar graphs should include the number 0. That was change to not do that if the axis is a log scale.
Values which would be zero are replaced with PointPair.Missing
Removed the code which did not allow you have a log scale with normalization methods such as "ratio to global standards" or "ratio to heavy" which would often have values that were less than 1.
When normalized to "Total" or "Maximum", log scale is still forbidden.

2. Always switch to BarType.Cluster if the Y-axis is log scale
image
This is because stacked bar graphs with a log scale give a very misleading idea of which bar in the stack is the biggest.
Also, it would be tricky to figure out what the range of the Y-axis should be.

3. This also works for the Peak Area Peptide Comparison
image

I will try to write a unit test which verifies that the selection box is drawn in the right place and the Y-axis range is what we would expect.


brendanx67:
Looks much better. Though, I wonder in the first 2 cases why it 10^-1. Seems like it is adding a lot of unnecessary white space to the graph.

nickshulman:
> Though, I wonder in the first 2 cases why it 10^-1. Seems like it is adding a lot of unnecessary white space to the graph.

Good point. "GraphHelper.ReformatYAxis" was doing the wrong thing when the Y-axis scale was log.
It was incorrectly ensuring that the number 1 (i.e. 10^0) was included because that was the value of Scale.MajorStep.

image

PR #3286, opened 2024-12-18
Simplified handling of bogus lockmass correction requests

bspratt:
While chasing down memory issues, I noticed that there seemed to be some proximity to tests that involved asking for lockmass correction on data that isn't actual Waters raw. I'm not 100%% convinced that this is the issue, but it does seem just as well to simplify this code so that the data processing chain is left completely untouched when spurious requests are made. If nothing else it makes for a more accurate set of CVParams.


chambm (approved):
Looks good assuming all builds and tests pass.

bspratt:
> Looks good assuming all builds and tests pass.

Thanks. It's been running in the Integration branch nightlies for several days now.

PR #3288, opened 2024-12-19
Switch to using "effectivePermissions" property for checking LabKey folder permissions

vagisha:
- "userPermissions" is no longer written to the JSON returned by getContainers API (LK 24.11 release notes - https://www.labkey.org/Documentation/wiki-page.view?name=releasenotes2411#dev).  Get permissions from the "effectivePermissions" property.
- Updated Skyline, AutoQC, and SkylineBatch tests
- Added PanoramaFolder class to generate LabKey folder JSON used for testing
- Instead of requiring folder admin permissions when validating a QC folder in AutoQC Loader, check for permissions associated with the "Editor" LabKey role.


nickshulman:
I think this looks great.
The class "FolderPermission" represents a set of permissions that might be required for an operation.
It has a method "MatchesPermissions" which takes as an argument an array of strings which represents the actual permissions that a user has on a folder.
I would recommend using the same class (maybe renamed to "PermissionSet") for both a required set of permissions and the set of permissions a user has.
Then, "MatchesPermissions" returns true if "this" is a superset of the passed in argument.

Encapsulating things in a class like this would make it easier if, in the future, we wanted to optimize things by, for instance, storing the permission list in a HashSet or replacing the strings with enum values, etc.

nickshulman (approved):

PR #3289, opened 2024-12-19
TransitionGroupDocNode.ChangeSettings wasn't paying attention to the precursor mz exclusion window setting

bspratt:
Noticed this while working on improved filtering for small molecules, but it applies to peptides as well

I don't completely understand the changes I had to make to the test, would appreciate a thorough review there. Note there's already some mystery in there ("TODO: Figure out what is up with the 6")

### DIFF
--- a/pwiz_tools/Skyline/Model/TransitionGroupDocNode.cs
+++ b/pwiz_tools/Skyline/Model/TransitionGroupDocNode.cs
@@ -1125,7 +1125,9 @@ namespace pwiz.Skyline.Model
                         }
 
                         Helpers.AssignIfEquals(ref nodeNew, nodeTransition);
-                        if (settingsNew.TransitionSettings.Instrument.IsMeasurable(nodeNew.Mz, precursorMz))
+                        if (settingsNew.TransitionSettings.Instrument.IsMeasurable(nodeNew.Mz, precursorMz) &&
+                            (nodeNew.Transition.IonType == IonType.precursor || 
+                             !settingsNew.TransitionSettings.Filter.IsExcluded(nodeNew.Mz, precursorMz))) // Watch for unfragmented precursors in MS2
                             childrenNew.Add(nodeNew);
                     }
 
--- a/pwiz_tools/Skyline/TestFunctional/SrmCourseScenarioTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/SrmCourseScenarioTest.cs
@@ -86,7 +86,8 @@ namespace pwiz.SkylineTestFunctional
             // Check changed result state
             const int pepStandardCount = 13, tranStandardCount = 34;
             document = WaitForDocumentChange(document);
-            AssertEx.IsDocumentState(document, null, protCount, pepCount, pepCount*2-pepStandardCount, tranCount*2-tranStandardCount);
+            const int nPrecursorWindowExclusions = 1; // There's a single precursor mz exclusion window hit
+            AssertEx.IsDocumentState(document, null, protCount, pepCount, pepCount*2-pepStandardCount, (tranCount*2-tranStandardCount)- nPrecursorWindowExclusions); 
             int[] missingHeavy = {1, 1, 2, 2};
             for (int i = 0; i < EXPECTED_REPLICATES.Length; i++)
             {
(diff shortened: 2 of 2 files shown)


pwiz_tools/Skyline/Model/TransitionGroupDocNode.cs
```diff
@@ -1125,7 +1125,9 @@ public TransitionGroupDocNode ChangeSettings(SrmSettings settingsNew, PeptideDoc
                         }
 
                         Helpers.AssignIfEquals(ref nodeNew, nodeTransition);
-                        if (settingsNew.TransitionSettings.Instrument.IsMeasurable(nodeNew.Mz, precursorMz))
+                        if (settingsNew.TransitionSettings.Instrument.IsMeasurable(nodeNew.Mz, precursorMz) &&
+                            (nodeNew.Transition.IonType == IonType.precursor ||
```
nickshulman, line 1129: I think this code is in the "else" block of "if (autoSelectTransitions) { ... } else...".

I don't think this code is supposed to be deciding whether transitions that are already in the document pass the filter criteria.

The user is allowed to add manually add transitions that do not pass the filter criteria (although Skyline never allows the user to pick a transition whose m/z is outside of the Instrument min/max mz, hence the "IsMeasurable") and we would not want to remove any transition that they had manually added.

bspratt: Seems kind of arbitrary to trigger on Instrument mz range settings but not on precursor exclusion settings? Why omit one check and not the other?

nickshulman: It sounds like the correct place to put the fix might be in "TransitionGroup.GetTransitions".

Currently there's a block of code that looks like this:
```
            // For small molecules we can't generate new nodes, so just mz filter those we have
            foreach (var nodeTran in groupDocNode.Transitions.Where(tran => tran.Transition.IsNonPrecursorNonReporterCustomIon()))
            {
                if (minMz <= nodeTran.Mz && nodeTran.Mz <= maxMz)
                    yield return nodeTran;
            }
```
I think that's the place to put the fix.
```
            // For small molecules we can't generate new nodes, so just mz filter those we have
            foreach (var nodeTran in groupDocNode.Transitions.Where(tran => tran.Transition.IsNonPrecursorNonReporterCustomIon()))
            {
                if (useFilter)
                {
                    if (!settings.TransitionSettings.Filter.IsExcluded(nodeTran.Mz, precursorMz))
                    {
                        continue;
                    }
                }
                if (minMz <= nodeTran.Mz && nodeTran.Mz <= maxMz)
                    yield return nodeTran;
            }
```

brendanx67: > Seems kind of arbitrary to trigger on Instrument mz range settings but not on precursor exclusion settings? Why omit one check and not the other?

It may seem that way, but there is a meaning behind the distinction. IsMeasurable as the name implies is supposed to be about whether the mass spec can even measure the transition. If it cannot, then including it could cause an error when we export a method. So, if the instrument cannot measure a Q3 greater than 1200, exporting one essentially breaks the file. As Nick has pointed out, that means we will remove the transition even if a user has added it manually, because it is not a viable transition for the instrument configuration.

This is not the case for the precursor exclusion. It may be desirable in most cases to reduce the chance of noisy peaks in DIA. It was originally added for this case by Hannes Roest, but I can still imagine a user adding back excluded transitions in this range to see if they behave well enough, or because the only viable transition to measure falls in this range. You just can't do that when the transition is not even measurable. You have no option to add it back because the mass spec truly is not capable of measuring it.

bspratt: > It sounds like the correct place to put the fix might be in "TransitionGroup.GetTransitions".

Agreed, I have a whole branch dealing with that part of the code, thanks.

bspratt: > It may seem that way, but there is a meaning behind the distinction.

Thanks, I'll make sure this info ends up in a comment there.

PR #3295, opened 2024-12-21
"Back" button in the import peptide search dialog isn't properly disabled if you navigate back to the first page

bspratt:
It tries to take you to page -1.

Noticed this after using PauseTest() in DiaUmpireVendorFormatTest while working on something else. Not reported by any user.


bspratt:
@chambm can you give this the blessing, please?

chambm (approved):

PR #3313, opened 2025-01-03
Fix inconsistent formatting of MS vs MS/MS resolution values in Transition Settings > Full Scan dialog

bspratt:
It's actually not possible to enter this state (inconsistent use of thousands separator) as a user, but automated testing and tutorial screen capture could do so.

image
vs
image


brendanx67 (approved):
Looks good. Glad it was already wrapping everywhere else but the test setting. Thanks for renaming.

PR #3314, opened 2025-01-03
be a little more tolerant of data variants in BlibBuild with SSL and mzML

bspratt:
In SSL, accept "scan=<number>" as meaning just "<number>"
In mzML reader,  accept MS2 scans that lack precursor values

per https://skyline.ms/announcements/home/support/thread.view?rowId=65625

reported by user Morgan


chambm:
What mzml is this that has ms2 without any precursor? That doesn't sound valid: why not just call it an ms1 in that case...

bspratt:
Produced by "mzR", apparently. Purely for the purposes of spectral library generation, no scan times either. Funky, but useful.

chambm:
I think we should handle this as an error (rather than the crash it is now). It doesn't make sense to have a library MS2 without a precursor. MzML isn't meant to be a spectral library format but it can sort of be used as a primitive one, but MS2s still need a precursor. Let's go back to the user and ask them to create their MS2s with precursors, or to use MS1 instead.

Something like this?
```
if (specInfo->msLevel > 1)
{
    if (specInfo->precursors.empty())
        Verbosity::error("MS2 spectrum %s does not have a precursor", specInfo->scanNumber);
    returnData.mz = specInfo->precursors[0].mz;
}
```

bspratt:
Seems sensible. The user was in fact able to remake the mzML to include precursor information, so if we'd barked about that they'd have been able to deal with it themselves.

bspratt:
@chambm are you ok with this now?

PR #3322, opened 2025-01-06
Prevent properties like "SelectedResultsIndex" from showing up in the property sheet in "ViewMenu" form.

nickshulman:
In the form designer for classes that are derived from "SkylineControl", such as "ViewMenu" or "EditMenu", some properties from the base class were showing up in the property sheet.


ekoneil (approved):
Does ModeUiHandler need the same attributes since it has a getter? Otherwise, lgtm. Thanks for the quick fix (and explanation in email).

nickshulman:
> Does ModeUiHandler need the same attributes since it has a getter? Otherwise, lgtm. Thanks for the quick fix (and explanation in email).

Good point. We would prefer not to have that property show up in the Property Sheet either.

PR #3333, opened 2025-01-14
Fix path to the IMS filtering tutorial's data files

ekoneil:
Current path is HTTP, which gets blocked by Chrome. Change to HTTPS to avoid the unnecessary error.


ekoneil:
@eduardo-proteinms, @brendanx67 - ran into a minor issue when testing the Files Tree. Is there any reason not to make this change?

eduardo-proteinms (approved):
lgtm, cant think of a reason to not make this change.

brendanx67 (approved):
Yep. Seems good. The former should automatically forward to the new value, but better to just point to the right location. Thanks.

PR #3337, opened 2025-01-16
* changed Run Peptide Search to keep wizard form close button visible

chambm:
* changed Run Peptide Search to keep wizard form close button visible, but intercept it while search is running and display an explanation message to click cancel search button

image

### DIFF
--- a/pwiz_tools/Skyline/Controls/WizardPages.cs
+++ b/pwiz_tools/Skyline/Controls/WizardPages.cs
@@ -50,6 +50,17 @@ namespace pwiz.Skyline.Controls
             return base.ProcessCmdKey(ref msg, keyData);
         }
     }
+
+    public class WizardPageControl : UserControl
+    {
+        /// <summary>
+        /// Allow controls in a wizard to preempt close requests by the wizard's form. Returns true iff the close request should continue, false if it should be canceled.
+        /// </summary>
+        public virtual bool CanWizardClose()
+        {
+            return true;
+        }
+    }
 }
 
 
--- a/pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
+++ b/pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
@@ -229,7 +229,7 @@ namespace pwiz.Skyline.FileUI.PeptideSearch
         private UserControl GetPageControl(TabPage tabPage)
         {
             // Assume each tabPage only has a single UserControl; if that changes this function will need to be updated
-            return tabPage.Controls.OfType<UserControl>().Single();
+            return tabPage.Controls.OfType<UserControl>().SingleOrDefault();
         }
 
         public SrmDocument Document
@@ -1491,13 +1491,13 @@ namespace pwiz.Skyline.FileUI.PeptideSearch
         private bool CanWizardClose()
         {
             var wizardPageControl = GetPageControl(wizardPagesImportPeptideSearch.SelectedTab) as WizardPageControl;
-            return wizardPageControl?.CanWizardClose() == false;
+            return wizardPageControl == null || wizardPageControl.CanWizardClose();
         }
 
         protected override void OnFormClosing(FormClosingEventArgs e)
         {
             // Ask current WizardPageControl if wizard is in a good state to close
-            if (CanWizardClose())
+            if (!CanWizardClose())
             {
                 e.Cancel = true;
                 return;
--- a/pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
+++ b/pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
@@ -226,6 +226,13 @@ namespace pwiz.Skyline.FileUI.PeptideSearch
             tabPage.Controls.Add(pageControl);
         }
 
+        private UserControl GetPageControl(TabPage tabPage)
+        {
+            var wizardPageControls = tabPage.Controls.OfType<UserControl>().ToArray();
+            Assume.AreEqual(1, wizardPageControls.Length);
+            return wizardPageControls[0];
+        }
+
         public SrmDocument Document
         {
             get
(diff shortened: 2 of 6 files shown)


chambm:
I thought about avoiding the `as WizardPageControl` complexity by making ALL the controls (for each TabPage) inherit from WizardPageControl instead of UserControl, but some of those controls are used elsewhere (FullScanSettomgsControl and TransitionsEttingsControl) so that seemed unwise. I could have wrapped those controls in a WizardPageControl, but that seemed more complex than the WizardPageControl cast. But perhaps if more logic is added into WizardPageControl that applies to all the controls, then that latter approach would make sense.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -1486,6 +1491,13 @@ public void CloseWizard(DialogResult result)
 
         protected override void OnFormClosing(FormClosingEventArgs e)
         {
+            // Ask current WizardPageControl if wizard is in a good state to close
+            if ((GetPageControl(wizardPagesImportPeptideSearch.SelectedTab) as WizardPageControl)?.CanWizardClose() == false)
```
brendanx67, line 1495: Please encapsulate this logic in a function. Looks ugly to me in the if clause.

pwiz_tools/Skyline/FileUI/PeptideSearch/PeptideSearchResources.resx
```diff
@@ -362,4 +362,7 @@ Click 'Retry' to try building again after placing the original spectrum files in
   <data name="ImportPeptideSearchDlg_NextPage_Adding_detected_features_to_document" xml:space="preserve">
     <value>Adding detected features to document</value>
   </data>
+  <data name="SearchControl_CanWizardClose_Cannot_close_wizard_while_the_search_is_running_" xml:space="preserve">
+      <value>Cannot close wizard while the search is running. Press the Cancel Search button first.</value>
```
brendanx67, line 366: Should go the extra step of asking the user if they want to cancel rather than just telling them they need to press the button themselves. You can still deny the close until the cancel is completed, but it is more polite to help the user do what you are telling them is necessary.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -226,6 +226,13 @@ private static void AddPageControl<TControl>(TControl pageControl, TabPage tabPa
             tabPage.Controls.Add(pageControl);
         }
 
+        private UserControl GetPageControl(TabPage tabPage)
+        {
+            var wizardPageControls = tabPage.Controls.OfType<UserControl>().ToArray();
```
brendanx67, line 231: Could also be First() if asserting there are not 2 is not important. First will throw if their are none, just like your Assume.

pwiz_tools/Skyline/FileUI/PeptideSearch/SearchControl.cs
```diff
@@ -266,6 +267,16 @@ public void Cancel()
             btnCancel.Enabled = false;
         }
 
+        public override bool CanWizardClose()
+        {
+            if (btnCancel.Enabled)
+            {
+                MessageDlg.Show(Parent, PeptideSearchResources.SearchControl_CanWizardClose_Cannot_close_wizard_while_the_search_is_running_);
```
brendanx67, line 274: Make this Yes/No. Do you want to cancel now?

pwiz_tools/Skyline/Controls/WizardPages.cs
```diff
@@ -50,6 +50,17 @@ protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
             return base.ProcessCmdKey(ref msg, keyData);
         }
     }
+
+    public class WizardPageControl : UserControl
```
brendanx67, line 54: Time will tell if this is better than an interface, since you currently only use this in one case and assume that not-implemented and base class have the same meaning of "true" the wizard can be closed. An interface would not need a base implementation and the single implementation would not need to call base.CanWizardClose().

brendanx67 (changes requested):
A bit of feedback and at least one change requested.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
nickshulman: BTW. If you want to throw an error if the number of elements is anything other than "1" you can use "Single()":
`return tabPage.Controls.OfType<UserControl>().Single();`

chambm: I was aware of the Linq options here but I thought an explicit Assume was more self-documenting (obviously not enough). I'm making an assumption here that the tabPage only has one UserControl and it should be an error if there isn't.

chambm: I went with `Single()` and a comment in the code about the assumption.

chambm:
Updated screenshot and addressed the review points.

brendanx67 (approved):
One comment where I am not even positive "== false" is correct, except that you seem to be testing it. Easy to change to what I am suggesting. So, approving these changes.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -1484,8 +1488,21 @@ public void CloseWizard(DialogResult result)
             DialogResult = result;
         }
 
+        private bool CanWizardClose()
+        {
+            var wizardPageControl = GetPageControl(wizardPagesImportPeptideSearch.SelectedTab) as WizardPageControl;
+            return wizardPageControl?.CanWizardClose() == false;
```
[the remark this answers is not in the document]
chambm: Ah, I forgot to invert the logic when I made it a function. As written, it's really functioning as a "CannotCloseWizard()" function. I'll invert here and where it's used because CanWizardClose() is more natural.

PR #3339, opened 2025-01-21
* added preliminary support for Timeout attribute

chambm:
* added preliminary support for Timeout attribute when running tests under TestRunner; doesn't always work right now, it may need additional thread cleanup other than just the one thread launched to run the TestMethod

I'm making a draft PR for this to get it out there. It would be really nice to have Timeout support like the Visual Studio test runner has.


chambm:
I'll close this but leave the branch around if we want to get this working properly in the future.

PR #3341, opened 2025-01-21
Add Comet search engine

chambm:
@bspratt I changed mass tolerance combobox population so it happens after search engine initialization. Since Hardklor presumably needs those too, I had to remove one of those `if (ImportPeptideSearch.IsFeatureDetection) return;` statements. The FeatureDetectionTutorial still seemed to run ok, but I didn't do much extra testing.


bspratt:
> @bspratt I changed mass tolerance combobox population so it happens after search engine initialization. Since Hardklor presumably needs those too, I had to remove one of those `if (ImportPeptideSearch.IsFeatureDetection) return;` statements. The FeatureDetectionTutorial still seemed to run ok, but I didn't do much extra testing.

I'll build and test manually.

bspratt:
@chambm yeah, it's good, thanks

PR #3345, opened 2025-01-23
Sanity check for UI state in pwiz.SkylineTestUtil.AbstractFunctionalTestEx.OpenDocument()

bspratt:
It's technically possible in a test to simulate the user opening a document while forms like PeptideSettingsUI or TransitionSettingsUI are open, even though this isn't possible in actual UI use and will doubtless lead to confusing test behavior. So, I've added a check to pwiz.SkylineTestUtil.AbstractFunctionalTestEx.OpenDocument() to make sure that never happens.


pwiz_tools/Skyline/TestUtil/AbstractFunctionalTestEx.cs
```diff
@@ -75,6 +76,12 @@ public void Run(string zipFile = null)
         /// <param name="documentPath">File path of document</param>
         public SrmDocument OpenDocument(string documentPath)
         {
+            // In a test it's possible to programatically open a document while forms like
+            // PeptideSettingsUI or TransitionSettingsUI are open, but this isn't possible
+            // in actual UI use and will doubtless lead to confusing test behavior.
+            var unexpectedOpenForms = FindOpenForms<CommonFormEx>().Where(f => !(f is SkylineWindow)).Select(form => form.Name).ToList();
```
nickshulman, line 82: I'd recommend looking for forms for which the "Modal" property is true, because those are the only forms that would prevent the user from interacting with Skyline's menus:
```
var unexpectedOpenForms = FindOpenForms<Form>().Where(f=>f.Modal).Select(form => form.Name).ToList();
```

bspratt: Makes sense, thanks.

PR #3346, opened 2025-01-24
Use ToLowerInvariant() instead of ToLower() when identifying small molecule column headers…

bspratt:
… (otherwise the "I" in "InChi" does not convert to "i" under language tr-TR)


brendanx67 (approved):
Nice one! Glad we test Turkish.

PR #3352, opened 2025-01-30
Skyline: Improve code for finding Thermo method export DLLs

brendanx67:
- Thermo reported this code as failing on an Ascend at Thermo
- This code was written after a full review of 7 machines in the MacCoss lab and the Ascend at Thermo to do a more complete job of finding the necessary DLLs
- More recent machines have added Thermo.TNG.MethodXMLInterface2.dll to support Adaptive RT so it needs to be copied if it is found
- Added a fairly thorough unit test for this more complicated new code to make sure it works as expected and stays working


brendanx67:
I think this is ready to go. I will do another round of live testing tomorrow in the lab, but I am not expecting any surprises. It is pretty well tested. Would appreciate at least one code review before merging.

nickshulman:
On my desktop computer my registry looks like this:
image
The ProgramPath is the relative path "System\Programs\dependencies\tng" which actually seems to refer to the path "C:\Program Files\Thermo Scientific\Instruments\Astral\1.1\System\Programs\dependencies\tng".

There was a support request from a user last year who also had a relative path in the registry:
https://skyline.ms/announcements/home/support/thread.view?rowId=63895

They apparently were able to fix their problem by changing the relative path in the registry to an absolute path.

The error message I got was:
image

Does Skyline have enough information to figure out how to turn the relative path into an absolute path?
If it were to look in the correct absolute path it would find the dll's it's looking for.

brendanx67:
What is the full path in this case? Pretty explicit error message, though.
If it were posted on the support board, we would know what to tell the user
to do.

brendanx67:
I don't think we are going to fix that problem in this PR. I like that the error message is pretty clear for it. None of the instruments in the MacCoss lab have this issue. If you can reproduce how you ended up with your registry looking like this, we should report it to Thermo, but is may already be fixed on their end.

nickshulman:
> I don't think we are going to fix that problem in this PR.
Sounds good. I could not find anything in my registry that could be used to figure out that "C:\Program Files\Thermo Scientific\Instruments\Astral\1.1\"

brendanx67:
Any other thoughts on this PR before I merge to master?

pwiz_tools/Skyline/Model/DllFinder.cs
```diff
@@ -0,0 +1,322 @@
...
+                    TextUtil.LineSeparate(listMachineNames));
+            }
+
+            return info;
+        }
+
+        private class ProgramPath
+        {
+            public string Path;
+            public bool IsFull;
```
nickshulman, line 288: This probably should be called "IsFullInstall" because otherwise it sure sounds like it means "IsFullPath" given the class it's in.

pwiz_tools/Skyline/FileUI/ExportMethodDlg.cs
```diff
@@ -256,6 +261,13 @@ public ExportMethodDlg(SrmDocument document, ExportFileType fileType)
             UpdateInstrumentControls(MethodType);
         }
 
+        private void SelectInstrumentTypeIfValid(string instrumentType, string[] listTypes)
+        {
+            string selectionType = GetSelectedInstrument(instrumentType);
+            if (listTypes.Contains(selectionType) && GetInstrumentTypeFromSelection(selectionType, true) != null)
+                InstrumentType = listTypes.FirstOrDefault(typeName => typeName.Equals(instrumentType));
```
nickshulman, line 268: This code is a little confusing, because there's "instrumentType", "selectionType" and whatever "GetInstrumentTypeFromSelection(selectionType)" returns.

This particular line is equivalent to:
`InstrumentType = listTypes.Contains(instrumentType) ? instrumentType : null;`

Is that what you intended? (it probably is)
Just wanted to make sure this code is doing what you expect it to.

nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/FileUI/ExportMethodDlg.cs
brendanx67: Okay. Good point. I could barely understand it myself, and I am not sure it worked correctly. I have added a bunch of comments, and several tests of the initial selection of the Instrument type combo box, and also improved the code to work exactly the way I was expecting it to, after debug stepping through it during the tests.

pwiz_tools/Skyline/Model/DllFinder.cs
brendanx67: Okay. Changed to IsFullInstall and added comments.

PR #3357, opened 2025-02-04
Avoid a null reference in graph update in a race condition with an action that removes all chromatograms

bspratt:
bumped into this while working on something else

not reported by any user


rita-gwen:
Thank you! Yes, this looks fine with me. Im going to merge it.

bspratt:
Excellent, thanks. I've deleted the branch.

PR #3358, opened 2025-02-05
Fix for feature finding

bspratt:
The feature finding code uses spectral libraries in an unconventional manner - entries don't have any fragments, just the precursor info. This tripped up the AutoManage code, specifically the use of TransitionSettings.Libraries.MinIonCount. Now we treat libraries with score type ScoreType.HARDKLOR_IDOTP as being immune to that particular filter.

I can't help but wonder if there isn't a more general way to deal with this, though

### DIFF
--- a/pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
+++ b/pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
@@ -909,7 +909,10 @@ namespace pwiz.Skyline.FileUI.PeptideSearch
             // Add the library molecules to the document, in natural sort order
             var status = new ProgressStatus(PeptideSearchResources.ImportPeptideSearchDlg_NextPage_Adding_detected_features_to_document);
             progressMonitor.UpdateProgress(status);
-            Assume.AreNotEqual(Document.Settings.PeptideSettings.Libraries.LibrarySpecs, _existingLibraries.LibrarySpecs);
+            if (Equals(Document.Settings.PeptideSettings.Libraries.LibrarySpecs, _existingLibraries.LibrarySpecs))
+            {
+                MessageDlg.Show(this, PeptideSearchResources.ImportPeptideSearchDlg_AddDetectedFeaturesToDocument_Did_not_detect_any_features___check_your_settings_and_try_again);
+            }
             var docNew = Document;
             foreach (var lib in Document.Settings.PeptideSettings.Libraries.Libraries.Where(l =>
                          !_existingLibraries.Libraries.Contains(l)))
--- a/pwiz_tools/Skyline/FileUI/PeptideSearch/PeptideSearchResources.resx
+++ b/pwiz_tools/Skyline/FileUI/PeptideSearch/PeptideSearchResources.resx
@@ -365,4 +365,7 @@ Click 'Retry' to try building again after placing the original spectrum files in
   <data name="SearchControl_CanWizardClose_Cannot_close_wizard_while_the_search_is_running_" xml:space="preserve">
       <value>Cannot close wizard while the search is running. Do you want to cancel the search?</value>
   </data>
+  <data name="ImportPeptideSearchDlg_AddDetectedFeaturesToDocument_Did_not_detect_any_features___check_your_settings_and_try_again" xml:space="preserve">
+      <value>Did not detect any features - check your settings and try again</value>
+  </data>
 </root>
--- a/pwiz_tools/Skyline/Model/DocSettings/TransitionSettings.cs
+++ b/pwiz_tools/Skyline/Model/DocSettings/TransitionSettings.cs
@@ -25,6 +25,7 @@ using System.IO;
 using System.Xml;
 using System.Xml.Schema;
 using System.Xml.Serialization;
+using pwiz.BiblioSpec;
 using pwiz.Common.Chemistry;
 using pwiz.Common.Collections;
 using pwiz.Common.DataBinding;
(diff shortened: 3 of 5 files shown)


pwiz_tools/Skyline/FileUI/PeptideSearch/PeptideSearchResources.resx
```diff
@@ -365,4 +365,7 @@ Click 'Retry' to try building again after placing the original spectrum files in
   <data name="SearchControl_CanWizardClose_Cannot_close_wizard_while_the_search_is_running_" xml:space="preserve">
       <value>Cannot close wizard while the search is running. Do you want to cancel the search?</value>
   </data>
+  <data name="ImportPeptideSearchDlg_AddDetectedFeaturesToDocument_Did_not_detect_any_features___check_your_settings_and_try_again" xml:space="preserve">
+      <value>Did not detect any features - check your settings and try again</value>
```
brendanx67, line 369: I would prefer "No features were detected. Check your settings and try again." Possibly with a new line between the sentences. This was our standard format for a while in Visual C++. Line 1: state the problem, Line 2: Suggest a possible solution. It is not standard in Skyline, though. But at least the goal is similar and having two separate sentences seems better to me.

pwiz_tools/Skyline/Model/DocSettings/TransitionSettings.cs
```diff
@@ -1686,7 +1687,18 @@ public bool HasMinIonCount(TransitionGroupDocNode nodeGroup)
...
             // seems to be the only time we really know that a precursor ion is being created for MS1 filtering
             // or not.
-            return nodeGroup.Transitions.Count(nodeTran => !nodeTran.IsMs1 || nodeTran.HasLibInfo) >= MinIonCount;
+            if (nodeGroup.Transitions.Count(nodeTran => !nodeTran.IsMs1 || nodeTran.HasLibInfo) >= MinIonCount)
+            {
+                return true;
+            }
+
+            // Special case with feature finding where libraries don't have fragment info, just precursors
+            if (nodeGroup.HasLibInfo && nodeGroup.LibInfo.ScoreType.Equals(ScoreType.HARDKLOR_IDOTP) && nodeGroup.Transitions.All(nodeTran => nodeTran.IsMs1))
```
brendanx67, line 1696: Looks good. Nice effort to minimize the impact. The comment might be improved with a bit more explanation, e.g. "So, having only MS1 transitions is treated as having the minimum ion count."

pwiz_tools/Skyline/Model/SrmDocument.cs
```diff
@@ -2772,6 +2772,11 @@ public override int GetHashCode()
             }
         }
 
+        public override string ToString()
+        {
+            return $@"{MoleculeGroupCount},{MoleculeCount},{MoleculeTransitionGroupCount},{MoleculeTransitionCount}"; // For debugging convenience, not user-facing
```
brendanx67, line 2777: Remove this or make it clearer to a developer who has no idea why these 4 values represent SrmDocument.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -909,7 +909,10 @@ private void AddDetectedFeaturesToDocument(IProgressMonitor progressMonitor)
             // Add the library molecules to the document, in natural sort order
             var status = new ProgressStatus(PeptideSearchResources.ImportPeptideSearchDlg_NextPage_Adding_detected_features_to_document);
             progressMonitor.UpdateProgress(status);
-            Assume.AreNotEqual(Document.Settings.PeptideSettings.Libraries.LibrarySpecs, _existingLibraries.LibrarySpecs);
+            if (Equals(Document.Settings.PeptideSettings.Libraries.LibrarySpecs, _existingLibraries.LibrarySpecs))
+            {
+                MessageDlg.Show(this, PeptideSearchResources.ImportPeptideSearchDlg_AddDetectedFeaturesToDocument_Did_not_detect_any_features___check_your_settings_and_try_again);
```
brendanx67, line 914: This seems strange that there is no return in this if block. The if is quick enough that it seems like you should do it ahead of the progress update and then just return. If there is some reason you need to update the status, that should be noted and then add the line to complete the status and return. It doesn't make sense to go through all of the code below, and it puts a burden on anyone changing it to ensure it still works for this degenerate case.

brendanx67 (changes requested):
Generally good. Little bit of cleanup.

brendanx67:
Also, there should be a test that confirms the new code works as expected.

pwiz_tools/Skyline/FileUI/PeptideSearch/PeptideSearchResources.resx
bspratt: Sounds good.

pwiz_tools/Skyline/Model/DocSettings/TransitionSettings.cs
bspratt: Yeah, could be better:
            // Special case with feature finding where libraries don't have fragment info, just precursors - so a minumum
            // fragment count isn't applicable. Only current scenario for this is the Hardklor-based feature finding.
            // N.B. there is a possibility of the feature finder eventually supporting fragment ions, in which case this needs
            // to be revisited.

bspratt:
> Also, there should be a test that confirms the new code works as expected.

Right - a small modification to the existing test should take care of that, thanks

pwiz_tools/Skyline/Model/SrmDocument.cs
bspratt: They're the same values in the same order shown in the UI, or used in CheckDocumentState calls. And of course they can check the code if they're confused. I'll add a comment to this effect but would prefer to keep this terse for use as value display in the debugger.

bspratt: I've added this comment:
            // For debugging convenience, not user-facing
            // These are the same values in the same order used to summarize the document in the UI (lower right corner of Skyline window).
            // That's also the same values and order as used in CheckDocumentState() calls.
            // This is terse by design, for ease of display in debugger.

brendanx67:
Looks like your new code may cause a NRE in TestSrmTutorialLegacy. Check the TeamCity log for Tutorials and Perf tests.

bspratt:
Yeah, it's hitting a null value for nodeGroup.LibInfo.ScoreType - certainly I can add a null check here, but would it be better to handle that at the LibInfo constructor and set it to ScoreType.UNKNOWN  when null is proposed?

PR #3359, opened 2025-02-06
Move "ProgressStream" into CommonUtil

nickshulman:
Also, speed up "File > Import > Annotations" for large documents.
It was taking an excruciating amount of time to import a large number annotation values into a big document and now it only takes a painful amount of time.

I will try to change "HashingStreamReaderWithProgress" to use "ProgressStream" as well.

Speed up importing annotations by:
1. Avoid calling "CalcIsProteinMetadataPending" when SrmDocument.DeferSettingsChanges is true.
2. Reduce bottleneck in NodeRef.EnumerateSiblingNodeRefs Also, make it so that element locators that cannot be found is treated as a warning shown at the end up importing instead of causing the import to fail.

Fixed performance problems importing annotations into large documents. Changed to a warning elements not found during annotation import.


pwiz_tools/Skyline/Model/Results/RemoteApi/Ardia/ArdiaUrl.cs
```diff
@@ -248,20 +194,13 @@ public override MsDataFileImpl OpenMsDataFile(OpenMsDataFileParams openMsDataFil
             var response = client.GetAsync(presignedUrl, HttpCompletionOption.ResponseHeadersRead, openMsDataFileParams.CancellationToken).Result;
             response.EnsureSuccessStatusCode();
             var responseStream = response.Content.ReadAsStreamAsync().Result;
-            using (var progressStream = new ProgressStream(responseStream, RawSize ?? 1))
+            var progressStream = new ProgressStream(responseStream, RawSize ?? 1);
```
chambm, line 197: Why did you remove automatic dispose?

nickshulman: I wanted to make it clear that calling Dispose on ProgressStream does not do anything (specifically, ProgressStream never closes the Stream that was passed into its constructor).

I could certainly put the word "using" back.
I am not sure whether having the word "using" on that line makes the code more or less clear and correct.

What do you think?

chambm: I think a general rule we could follow is if something has a Dispose() and it has a lifetime that should be limited to the current scope, we should `using` it. You can add a comment like `// does not close stream` ?

chambm: My rationale for applying the rule is that somebody might modify ProgressStream to add some other Dispose()able member, and assume that all callers are already Dispose()ing, so not revisit all the call sites to add `using` back.

PR #3361, opened 2025-02-07
Skyline: Raised the peak truncation threshold

brendanx67:
The TransitionResults.Truncated and PrecursorResults.CountTruncated values have been reported as too sensitive for a while, especially for the Stellar which is often tuned to use very narrow retention time windows for scheduled data acquisition. These values are based on a height proportion threshold currently set to 0.01 or 1%. In the initial commit the value is set to 0.25 of 25%, i.e. quarter-max where 50% is half-max.

### DIFF
--- a/pwiz_tools/Skyline/Documentation/Tutorials/shared/skyline.js
+++ b/pwiz_tools/Skyline/Documentation/Tutorials/shared/skyline.js
@@ -16,30 +16,42 @@ function targetTop()
 });
 }
 
-function addFigureAltText()
-{
+function addFigureAltText() {
+    
+    // Helper function to wrap element with anchor
+    function wrapWithAnchor(element, id) {
+        const anchor = document.createElement('a');
+        anchor.setAttribute('name', id);
+        anchor.setAttribute('id', id);
+        
+        // Replace the element with the anchor
+        const parent = element.parentNode;
+        parent.replaceChild(anchor, element);
+        
+        // Add the element inside the anchor
+        anchor.appendChild(element);
+        
+        return anchor;
+    }
+
     const images = document.querySelectorAll('img');
     var figureCounter = 1;
+
     images.forEach((img) => {
         if (img.src.includes("/s-")) {
             img.title = `Figure ${figureCounter}`;
-
-            var anchor = document.createElement('a');
-            const anchorId = `figure${figureCounter}`
-            anchor.setAttribute('name', anchorId);
-            anchor.setAttribute('id', anchorId);
-            img.parentNode.parentNode.insertBefore(anchor, img.parentNode);
-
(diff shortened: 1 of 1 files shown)


pwiz_tools/Skyline/Model/Serialization/DocumentWriter.cs
```diff
@@ -670,6 +670,9 @@ private static void WriteTransitionGroupChromInfo(XmlWriter writer, TransitionGr
             writer.WriteAttributeNullable(ATTR.height, chromInfo.Height);
             writer.WriteAttributeNullable(ATTR.mass_error_ppm, chromInfo.MassError);
             writer.WriteAttributeNullable(ATTR.truncated, chromInfo.Truncated);
+            // Writing truncated proportion 0 for peaks with a truncated count of zero is redundant
+            if ((chromInfo.Truncated ?? 0) != 0)
```
nickshulman, line 674: This should just be:
`if (chromInfo.Truncated != 0)`

pwiz_tools/Skyline/Model/Databinding/Entities/PrecursorResult.cs
```diff
@@ -116,6 +116,8 @@ public double? RatioDotProduct {
         public double? AverageMassErrorPPM { get { return ChromInfo.MassError; } }
         [Format(NullValue = TextUtil.EXCEL_NA)]
         public int? CountTruncated { get { return ChromInfo.Truncated; } }
+        [Format(Formats.STANDARD_RATIO, NullValue = TextUtil.EXCEL_NA)]
```
nickshulman, line 119: We might want to use `Formats.Percent` here instead of `Formats.STANDARD_RATIO`.
(There is also `Formats.PEAK_AREA_NORMALIZED` which is a percent with up to four decimal places)

nickshulman (commented):
I think "ProportionTruncated" is simple enough that it should just be calculated in the getter for the property in PrecursorResult.
I don't think it's worth storing on the TransitionGroupChromInfo or writing out to the .sky file.

brendanx67:
> I think "ProportionTruncated" is simple enough that it should just be calculated in the getter for the property in PrecursorResult. I don't think it's worth storing on the TransitionGroupChromInfo or writing out to the .sky file.

Should we also do this for PrecursorResults.AverageMassErrorPPM? I was following that as an example. Seems like the two have very similar calculation overhead. Or do we need to have that in Panorama? It is worth remembering that anything not in the .sky file will generally not be available in Panorama by default, unless a user goes through the pain of creating a calculate annotation and forcing it back into the XML.

nickshulman:
This PR fails "TestDiaQeFullSearchTutorialExtra". I will update the expected numbers so that it passes.

PR #3366, opened 2025-02-11
- added checkbox in Associate Proteins dialog for last used FASTA

chambm:
- added checkbox in Associate Proteins dialog to make remembering the last used FASTA an option

Valentine's Day special for Mike:
image

Might tweak the vertical spacing on a bit though. Also it technically doesn't remember the "choice" if it it's using background proteome, but that was the existing behavior (only FASTA was remembered). Should I fix that?


brendanx67 (commented):
If the issue is that it can take a long time to calculate the result of using the same FASTA as last time, this seems to be asking a lot of the user to remember to check this checkbox so they don't pay that price next time. It kind of assumes that users like Mike have gotten annoyed with the remembering. What about just showing a yes/no message box when we bring up this form? You used the FASTA file C:\Users\mchambers\Documents\FASTA\human.fasta last time. Do you want to use this FASTA file again?

chambm:
The checkbox is persistent so they'd only have to endure that once. The calculation should also be cancellable although there may be parts that don't check cancellation frequently enough. I think a yes/no box that showed up _every_ time would be even more annoying than setting this once?

chambm:
@nickshulman what result do you get running TestDiaSearchStellarTutorialDraft ? I can't get 915 locally, only 905.

PR #3375, opened 2025-02-22
Enable using the clipboard when paused in a test

nickshulman:
When tests are running, they use the "internal" clipboard instead of the system clipboard. There used to be code in PauseAndContinueForm to switch to using the system clipboard when the test is paused. That code got removed in PR #3153


brendanx67 (approved):
Thanks, Nick.

PR #3379, opened 2025-02-26
Skyline/work/20250212 diagonal pasef

bspratt:
Support Bruker Diagonal PASEF in msconvert and Skyline


chambm:
I added the singleFrameDiaPASEF switch but when I tried to use it with TestDiaPasefTutorial it fell apart. I made changes to IsolationSchemeReader and SpectrumFilter so it could actually import the files with the singleFrameDiaPASEF switch turned on, but the data that's extracted looks quite different than with it turned off.

bspratt:
> I added the singleFrameDiaPASEF switch but when I tried to use it with TestDiaPasefTutorial it fell apart. I made changes to IsolationSchemeReader and SpectrumFilter so it could actually import the files with the singleFrameDiaPASEF switch turned on, but the data that's extracted looks quite different than with it turned off.

Can we back this out? It messes with the work I've been doing, and if I understand you it doesn't actually work at the moment?

chambm:
It works on the pwiz side from my testing (which admittedly was not exhaustive). DiaPASEF is giving single frames with all the data and the 2 isolation m/z arrays. I'm pretty sure it's the Skyline side that isn't handling it as (not really) "diagonal" data properly yet. 

There's no reason to back it out. All you have to do is turn off the singleFrameDiaPasef switch in MsDataFileImpl. And it would only affect diaPASEF, not diagonalPASEF (unless I messed up).

chambm:
I just saw that I have it default to false so it shouldn't affect you (except for possible code conflicts of course).

Is the GetVirtualSpectra() you're having issues with? Because I think that kind of method or something like it is absolutely needed if we continue with the "splitting raw spectra into virtual subspectra based on isolation m/z". I needed the same thing for the isolation scheme reader code, and from my testing that DOES work with singleFrameDiaPasef (the isolation scheme graph looks the same whether it's turned on or not).

bspratt:
Mostly code conflict problems, I'll make it work.

PR #3383, opened 2025-03-03
Fix for a "polarity mismatch" exception when searching for nearest precursor m/z match in a mixed polarity document...

bspratt:
… when the m/z is negative and has greater absolute value than any negative precursor m/z. 

Test for the fix was added opportunistically to an existing test involving mixed polarity documents.

Reported by user Paul

### DIFF
--- a/pwiz_tools/Skyline/Model/Results/PeptideFinder.cs
+++ b/pwiz_tools/Skyline/Model/Results/PeptideFinder.cs
@@ -58,47 +58,42 @@ namespace pwiz.Skyline.Model.Results
                 return null;
 
             // Find closest precursor Mz match.
-            // Watch out for negative values, which sort before positive values.
             var lookup = new PeptidePrecursorMz(null, precursorMz);
             int i = _precursorMzPeptideList.BinarySearch(lookup, PeptidePrecursorMz.COMPARER);
-            if (i < 0)
+            if (i >= 0)
             {
-                i = ~i;
-                if (i >= _precursorMzPeptideList.Count)
-                    i = _precursorMzPeptideList.Count - 1;
-                else if (i > 0 &&
-                         _precursorMzPeptideList[i - 1].PrecursorMz.IsNegative == precursorMz.IsNegative &&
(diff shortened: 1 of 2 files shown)


nickshulman (approved):
Looks good to me.

You might want to simplify the logic a little. After the binary search, it is either an exact match, or "~i" points to the first item which is greater than the key in which case the closest match is at either "~i-1" or "~i".

Here's what I was thinking this could be:
```
        public PeptideDocNode FindPeptide(SignedMz precursorMz)
        {
            if (_precursorMzPeptideList.Count == 0)
                return null;

            // Find closest precursor Mz match.
            // Watch out for negative values, which sort before positive values.
            var lookup = new PeptidePrecursorMz(null, precursorMz);
            int i = _precursorMzPeptideList.BinarySearch(lookup, PeptidePrecursorMz.COMPARER);
            if (i >= 0)
            {
                return _precursorMzPeptideList[i].NodePeptide;
            }
            PeptidePrecursorMz closestMatch = null;
            double closestDistance = double.MaxValue;
            foreach (var candidateIndex in new[] { ~i - 1, ~i })
            {
                if (candidateIndex < 0 || candidateIndex >= _precursorMzPeptideList.Count)
                {
                    continue;
                }
                var candidate = _precursorMzPeptideList[candidateIndex];
                if (candidate.PrecursorMz.IsNegative != precursorMz.IsNegative)
                {
                    continue;
                }

                var distance = Math.Abs(candidate.PrecursorMz - precursorMz);
                if (closestMatch == null || distance < closestDistance)
                {
                    closestMatch = candidate;
                    closestDistance = distance;
                }
            }

            // Return only if the match is within allowed tolerance.
            if (closestMatch == null || Math.Abs(closestMatch.PrecursorMz - precursorMz) > _mzMatchTolerance)
            {
                return null;
            }
            return closestMatch.NodePeptide;
        }

```

bspratt:
Another thoughtful review, thanks. I have incorporated your suggestions.

PR #3386, opened 2025-03-05
Waters connect implementation

rita-gwen:


rita-gwen:
Current TODO list (not in the priority order):
-  ???There can be only one Quan transition per analyte
-  ???Testing language support
-  Add "Add/Edit" list view items to manage remote accounts
-  Add account alias
-  Add empty template message
-  Test "no accounts defined" situation
-  Fix button text in the Save File dialog
-  Hide local file system icons
-  Fix "Look In" dropbox glitch
-  Add failed upload scenario test
-  Add "multiple methods decided on the server" option
-  Investigate missing folder Id exception
-  Hide run duration for the standard method and set start/end time to 0
-  Add string auto-completion in the file name field of the Template selection dialog

rita-gwen:
These are mostly the changes requested by Stephen earlier this week:
Rita A Chupalov Thanks for that, we done some testing. Some initial findings...
 
1. Remove adduct name from analyte name when passing on the API
 2. Ensure a single transition per adduct has a quan ion set
 3. Do not pass RT if it is unknown
- we are changing this to be nullable on the API, the service will be updated asap
4. Do not pass start and end time if scheduleType = FullGradientTime
- we are changing these to be nullable on the API, the service will be updated asap
5. Display a success dialog with method names and ID's - it may be useful for the user to be able to copy the ID's so they can search with them in w_c.
6. Display a error dialog with the w_c error model info - it may be useful if the user can copy this info.
7. Remove audit reason string - this is geting removed from the API.
8. Sample Set import fails if Sample Set selected (rather than injections inside Sample Set)

re. #5, standard Skyline message dialog allows user to copy the message using Ctrl-C. The error dialog already shows the server response string as part of the exception message. It can be copied using Show More button.
It looks like the start and end times are still not nullable on the API, I am getting an exception with the standard method upload.
I need some clarification regarding the internal standard field since there are multiple standards possible in Skyline. Is this quantification, RTi or QC standard.
This commit does not include the fix on the #8. It is probably result of merging the sample set logic into this branch. Matt's input on this might be helpful.

pwiz_tools/Shared/CommonMsData/RemoteApi/RemoteSession.cs
```diff
@@ -282,4 +298,9 @@ public override int GetHashCode()
             }
         }
     }
+
+    public class ProgressableStreamContents
```
rita-gwen, line 302: Don't need this, remove

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectAccount.cs
```diff
@@ -121,9 +164,42 @@ public string GetFoldersUrl()
...
+
+            var newToken = tokenClient.RequestResourceOwnerPasswordAsync(Username, Password, ClientScope).Result;
+            if (newToken.IsError)
+            {
+                AuthenticationException ex;
+                if (newToken.ErrorType == ResponseErrorType.Exception)
+                    ex = new AuthenticationException(newToken.Error);
+                else
+                    ex = new AuthenticationException(string.Format(CultureInfo.CurrentCulture,
+                        "Failed to authenticate waters_connect account {0} with error: {1}",
```
chambm: Why doesn't this fail code inspection? This string should be moved to a resx and code inspection should have triggered here because it's not `@` literal.

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectAcquisitionMethodUrl.cs
```diff
@@ -0,0 +1,69 @@
...
+            if (!string.IsNullOrEmpty(methodVersionId))
+            {
+                Guid id;
+                if (Guid.TryParse(methodVersionId, out id))
+                {
+                    MethodVersionId = id;
+                }
+                else
+                {
+                    throw new InvalidDataException(string.Format("Invalid method version Id {0}", methodVersionId));
```
chambm: resx

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectSession.cs
```diff
@@ -18,15 +18,24 @@
...
 {
     public class WatersConnectSession : RemoteSession
     {
+        public static readonly string GET_FOLDERS = @"/waters_connect/v1.0/folders";
+        protected HttpClient _httpClient;
+
         public WatersConnectSession(WatersConnectAccount account) : base(account)
         {
+            if (account == null)
+                throw new ArgumentException(@"WatersConnectSession requires a WatersConnectAccount");
```
chambm: Would we expect a user to see this? If so, resx. If not, I think we use Assume for this, right @nickshulman ?

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectSession.cs
```diff
@@ -79,11 +87,57 @@ private IEnumerable<WatersConnectFolderObject> EnumerateChildFolderHierarchy(JOb
...
+            {
+                var folder = folders.FirstOrDefault(f => f.Path.Equals(url.EncodedPath));
+                if (folder != null)
+                {
+                    return url.ChangeFolderOrSampleSetId(folder.Id);
+                }
+            }
+            return url;
+        }
+        public bool TryGetFolderByUrl(WatersConnectUrl url, out WatersConnectFolderObject folder)
```
chambm, line 88: Blank line between methods.

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectSession.cs
```diff
@@ -95,12 +149,13 @@ private ImmutableList<WatersConnectFolderObject> GetFolders(Uri requestUri)
...
+        protected ImmutableList<WatersConnectFolderObject> GetInjections(Uri requestUri)
         {
-            var httpClient = WatersConnectAccount.GetAuthenticatedHttpClient();
-            var response = httpClient.GetAsync(requestUri).Result;
-            response.EnsureSuccessStatusCode();
+            var response = _httpClient.GetAsync(requestUri).Result;
+            EnsureSuccess(response);
             string responseBody = response.Content.ReadAsStringAsync().Result;
+            if (string.IsNullOrEmpty(responseBody))
+                return ImmutableList<WatersConnectFolderObject>.EMPTY;
```
chambm, line 143: This seems like an error rather than a correct "empty folder" result. What prompted this change?

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectSessionAcquisitionMethod.cs
```diff
@@ -0,0 +1,242 @@
+﻿/*
+ * Original author: Matt Chambers <matt.chambers42 .at. gmail.com>
```
chambm: Rita

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectSessionAcquisitionMethod.cs
```diff
@@ -0,0 +1,242 @@
...
+                    var res2 = AsyncFetch(GetAcquisitionMethodsUrl(wcUrl), GetAcquisitionMethods, out ex2);
+                    remoteException = ex1 ?? ex2;
+                    return res1 && res2;
+                }
+
+                remoteException = ex1;
+                return res1;
+            }
+
+            throw new Exception("Url type mismatch: " + wcUrl.Type);
```
chambm: Not sure what this means or whether it could be a user-facing error. If so, resx. If not, Assume the type at the top and the if block wouldn't be necessary.

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectSessionAcquisitionMethod.cs
```diff
@@ -0,0 +1,242 @@
...
+
+        public string UploadMethod(string methodName, string methodJson, IProgressMonitor progressMonitor)
+        {
+            var requestUri = GetAquisitionMethodUploadUrl();
+            var request = new HttpRequestMessage(HttpMethod.Post, requestUri);
+            if (progressMonitor == null)
+                request.Content = new StringContent(methodJson);
+            else
+            {
+                var status = new ProgressStatus(string.Format("Uploading method {0} to {1}", methodName, WatersConnectAccount.ServerUrl));
```
chambm: Resx

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectSessionAcquisitionMethod.cs
```diff
@@ -0,0 +1,242 @@
...
+                var progressStream = new ProgressStream(new StringStream(methodJson));
```
chambm: Can you explain this part, maybe with a comment? I'm not sure what the StringStream is for.

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectUrl.cs
```diff
@@ -37,19 +37,28 @@ public enum ItemType
             folder,
             folder_without_sample_sets,
             sample_set, // a collection of related injections
-            injection   // like a .raw file
+            injection,   // like a .raw file
+            folder_with_methods,
+            method       // a Waters acquisition method
+        }
+
+        public new enum Attr
```
chambm, line 45: Should the FileObject and FolderObject classes also get this Attr treatment?

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectAcquisitionMethodObject.cs
```diff
@@ -0,0 +1,213 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+using Newtonsoft.Json.Linq;
+using System;
+
+namespace pwiz.CommonMsData.RemoteApi.WatersConnect
+{
+    public class WatersConnectAcquisitionMethodObject : WatersConnectObject
```
chambm, line 24: This file's name has redundant WatersConnect in it. Also fix author/date line.

pwiz_tools/Shared/CommonMsData/RemoteApi/RemoteUrl.cs
```diff
@@ -96,6 +97,31 @@ public override MsDataFileUri ChangeLockMassParameters(LockMassParameters lockMa
...
+            {
+                return false;
+            }
+            if (account.AccountType != AccountType)
+            {
+                return false;
+            }
+            return account.ServerUrl == ServerUrl && account.Username == Username;
+        }
+        public bool SameAccountAs(RemoteUrl url)
```
chambm, line 113: Newline

pwiz_tools/Shared/CommonUtil/Mock/MockHttpMessageHandler.cs
```diff
@@ -0,0 +1,163 @@
...
+        }
+        public MockHttpMessageHandler CreateReplaceHandler(string handlerName)
+        {
+            var handler = new MockHttpMessageHandler();
+            _handlers[handlerName] = handler;
+            return handler;
+        }
+
+
+        public HttpMessageHandler getMessageHandler(string handlerName, Func<HttpMessageHandler> defaultHandlerFactory = null)
```
chambm: Fix newlines here. Sometimes too many, sometimes not any.

pwiz_tools/Skyline/Alerts/MessageDlg.cs
```diff
@@ -29,20 +29,30 @@ public static void Show(IWin32Window parent, string message, bool ignoreModeUI =
             new MessageDlg(message, ignoreModeUI).ShowAndDispose(parent);
         }
 
+        public static void Show(IWin32Window parent, string message, bool ignoreModeUI, MessageIcon messageIcon)
```
chambm: Changes to MessageDlg and AlertDlg need review by @nickshulman or @brendanx67 I think.

pwiz_tools/Skyline/Model/Export.cs
```diff
@@ -1016,6 +1035,46 @@ public AbstractMassListExporter ExportWatersQMethod(SrmDocument document, string
 
             return exporter;
         }
+        public AbstractMassListExporter ExportWatersConnectMethod(SrmDocument document, string exportMethodUrlString, string templateMethodUrlString)
```
chambm, line 1039: Newline

pwiz_tools/Skyline/TestConnected/UnifiFunctionalTest.cs
```diff
@@ -80,11 +98,30 @@ protected override void DoTest()
             var importResultsDlg = ShowDialog<ImportResultsDlg>(SkylineWindow.ImportResults);
             var openDataSourceDialog = ShowDialog<OpenDataSourceDialog>(importResultsDlg.OkDialog);
             var editAccountDlg = ShowDialog<EditRemoteAccountDlg>(() => openDataSourceDialog.CurrentDirectory = RemoteUrl.EMPTY);
-            RunUI(()=>editAccountDlg.SetRemoteAccount(_testAccount));
+
+            // Test invalid server URLs
+            RunUI(() => editAccountDlg.SetRemoteAccount(_testAccount.ChangeServerUrl("localhost")));
+            AssertAlertDlgContainsMessage(() => editAccountDlg.TestSettings(), ToolsUIResources.EditRemoteAccountDlg_ValidateValues_Invalid_server_URL_);
+            RunUI(() => editAccountDlg.SetRemoteAccount(_testAccount.ChangeServerUrl("https://localhost:12345"))); // resolves, but no server there
+            AssertAlertDlgContainsMessage(() => editAccountDlg.TestSettings(), "Unable to connect to the remote server");
```
chambm, line 106: I think I tested and these pass on Japanese but I don't understand why. These errors come from the system though rather than Skyline. Perhaps they are only localized if the entire OS is set to that language?

pwiz_tools/Skyline/TestFunctional/ScheduleMethodDlgTest.cs
```diff
@@ -179,7 +182,19 @@ protected override void DoTest()
...
+
+            var csv3 = new DsvStreamReader( File.OpenText(csvPath3), TextUtil.GetCsvSeparator(CultureInfo.CurrentCulture));
+            var csv4 = new DsvStreamReader(File.OpenText(csvPath4), TextUtil.GetCsvSeparator(CultureInfo.CurrentCulture));
+            var startHeader = "protein.name";
+            var endHeader = "cone_voltage";
+            while (!csv3.EndOfStream && !csv4.EndOfStream)
+            {
+                csv3.ReadLine();
+                csv4.ReadLine();
+                Assert.IsTrue(csv3.TryGetHeadersRange(startHeader, endHeader, out var csv3Line) &
```
chambm: Did you really mean bitwise and?

pwiz_tools/Skyline/TestFunctional/WatersConnectMethodExportTest.cs
```diff
@@ -0,0 +1,241 @@
+﻿using System.Collections.Generic;
```
chambm: Author/copyright header

pwiz_tools/Skyline/TestFunctional/WatersConnectMethodExportTest.cs
```diff
@@ -0,0 +1,241 @@
...
+            var methodFileDlg = ShowDialog<WatersConnectSaveMethodFileDialog>(() => schedulingDataDlg.OkDialog());
+            ValidateSkylineFolder(methodFileDlg);
+            RunUI(() =>
+            {
+                var folderToSelect = methodFileDlg.ListViewItems.FirstOrDefault(item => item.Text == @"FolderA");
+                Assert.IsNotNull(folderToSelect, "FolderA not found in the list of folders.");
+                folderToSelect.Selected = true;
+                methodFileDlg.KeyPressHandler(Keys.Enter);
+            });
+            WaitForConditionUI(1000, () => methodFileDlg.ListViewItems.Count == 11, () => "Template selection dialog is not populated within allotted time.");
```
chambm, line 121: Are these error messages from Waters and that's why they're not localized? If so that should be noted in a comment.

pwiz_tools/Skyline/Util/DsvStreamReader.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System.Collections.Generic;
```
chambm: Author/copyright

pwiz_tools/Skyline/Program.cs
```diff
@@ -81,6 +82,10 @@ public static bool FunctionalTest
             }
         }
 
+        public static HttpMessageHandlerFactory HttpMessageHandlerFactory
+        {
+            get { return CommonApplicationSettings.HttpMessageHandlerFactory; }
+        } 
         // TODO(nicksh): Remove this once intermittent failures in these tests are fixed
```
chambm, line 90: Newline

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectAccount.cs
rita-gwen: It looks like code inspection is not enabled for this project somehow. Is it in the project settings?

chambm: I think Brendan recently addressed this. After merging from master you may find these fail code inspection.

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectSession.cs
rita-gwen: The server returns an empty response if there are no injections in the folder. Try to import from "Company/Skyline/Folder with spaces". It might be a glitch due to spaces, but no exception is returned by the server and the request URL seems to be correct. We probably should ask Stephen to create an injection there and check if it is visible.

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectSessionAcquisitionMethod.cs
rita-gwen: ProgressStream is a wrapper around a Stream that adds ability to show progress as the underlying stream is being read. I guess I could add another constructor to the ProgressStream with a string parameter and construct the StringStream there.
I'll add a comment to the code.

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectUrl.cs
rita-gwen: I'm not sure. WatersConnectUrl is built around this name-value list and property names there are used in more than one place. With the file and folder objects the strings are just used once to parse the incoming JSON. I would rather use the approach I used in the MethodModel where class members are mapped to JSON elements using member attribute.

pwiz_tools/Skyline/TestConnected/UnifiFunctionalTest.cs
rita-gwen: They (en and jp) do pass in SkylineTester, but take 0 seconds, so I think they do not really run. When I do it in Visual Studio I get an exception at this line because it actually tries to authenticate against existing w_c server at Waters that is set by the default account. On what branch did you test them?

pwiz_tools/Skyline/TestFunctional/ScheduleMethodDlgTest.cs
rita-gwen: Yes, because I did not want && to short-circuit the second test. I guess I can just do two separate assertions.

pwiz_tools/Skyline/TestFunctional/WatersConnectMethodExportTest.cs
rita-gwen: These messages are only displayed by the test, they are not compared against anything. We do not localize test messages, I think. And this test does not connect to the Waters server.

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectSessionAcquisitionMethod.cs
nickshulman: I'm not sure that "StringStream" needs to exist. I think you could you just do:
`var progressStream = new ProgressStream(new MemoryStream(Encoding.UTF8.GetBytes(methodJson)));`

pwiz_tools/Skyline/TestFunctional/WatersConnectMethodExportTest.cs
chambm: So your mocked tests don't have a "fallback" mode where they do connect to the server? Is there another test that does?

pwiz_tools/Skyline/TestFunctional/ScheduleMethodDlgTest.cs
chambm: Yeah, just do two separate assertions. Even with && that would make more sense.

pwiz_tools/Skyline/Util/DsvStreamReader.cs
```diff
@@ -0,0 +1,113 @@
...
+ * limitations under the License.
+ */
+
+using System.Collections.Generic;
+using System.IO;
+using pwiz.Skyline.Util.Extensions;
+
+namespace pwiz.Skyline.Util
+{
+    public class DsvStreamReader
```
nickshulman: Could you have used DsvFileReader instead of this new class?

rita-gwen: Initially MethodModel class where DsvStreamReader is used was part of the CommonMsData library and I was not able to use DsvFileReader because it is in the main Skyline executable and it would cause a cyclic reference. Since then MethodModel moved to Skyline, so I guess I can replace DsvStreamReader with DsvFileReader now.

pwiz_tools/Skyline/Util/FormEx.cs
```diff
@@ -251,7 +251,7 @@ private static Screen GetScreen(int x, int y)
 
         public virtual void CancelDialog()
         {
-            CancelButton.PerformClick();
+            CancelButton?.PerformClick();
```
nickshulman: This doesn't look right.
"CancelDialog" is almost exclusively used in tests. We would want the test to throw an exception if this method got called and there was no cancel button.

pwiz_tools/Skyline/Model/WatersConnect/MethodModel.cs
```diff
@@ -0,0 +1,278 @@
...
+        public string Polarity { get; set; }
+
+        [JsonProperty("transitions")]
+        public IList<Transition> Transitions { get; set; }
+
+        public override void ParseObject(DsvFileReader reader)
+        {
+            base.ParseObject(reader);
+            var polarityHeader = GetColumnName(@"Polarity");
+            if (reader.TryGetFieldByName(polarityHeader, out var precursorCharge) &&
```
nickshulman: TryParse can handle null, so you could just do:
`if (int.TryParse(reader.GetFieldByName(polarityHeader), out var charge))`

pwiz_tools/Skyline/Util/Extensions/Text.cs
```diff
@@ -981,6 +980,18 @@ public string GetFieldByName(string fieldName)
             return GetFieldByIndex(fieldIndex);
         }
 
+        public bool TryGetFieldByName(string fieldName, out string fieldValue)
```
nickshulman: I don't think this method should exist because you can "GetFieldName" returns null when the field does not exist.

pwiz_tools/Skyline/TestUtil/MockHttpMessageHandler.cs
```diff
@@ -0,0 +1,187 @@
...
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace pwiz.Common.Mock
+{
+    public class MockHttpMessageHandler : HttpMessageHandler
```
nickshulman, line 13: Is this class still necessary now that "HttpClientWithProgress" supports recording and playing back interactions during tests?
Can you use "HttpClientTestHelper" instead?

pwiz_tools/Skyline/Util/Extensions/Text.cs
rita-gwen: This method exists to enable result assignment in an if statement, like in the previous comment and avoid calling GetFieldByName twice.

pwiz_tools/Skyline/TestUtil/MockHttpMessageHandler.cs
rita-gwen: I do not see how I can perform uploaded data validation with the HttpClientWithProgress. I think MockHttpMessageHandler and HttpClientWithProgress should be merged at some point, but in a separate PR.

pwiz_tools/Skyline/Util/Extensions/Text.cs
chambm: Isn't Nick correct though that you could replace that if logic with just TryParse, as he suggested:
`if (int.TryParse(reader.GetFieldByName(polarityHeader), out var charge))` ?

pwiz_tools/Skyline/TestUtil/MockHttpMessageHandler.cs
chambm: Agreed with Rita

pwiz_tools/Shared/CommonMsData/RemoteApi/WatersConnect/WatersConnectAccount.cs
```diff
@@ -15,24 +15,71 @@
...
 {
     [XmlRoot("waters_connect_account")]
     public class WatersConnectAccount : RemoteAccount
     {
+        public static readonly string HANDLER_NAME = @"WatersConnect.Handler.Main";
+        public static readonly string AUTH_HANDLER_NAME = @"WatersConnect.Handler.Authentication";
+        public static readonly string TOKEN_DATA = @"token";
+        public static readonly string GET_FOLDERS = @"/waters_connect/v2.0/folders";
+
+        public class TokenCacheEntry
```
nickshulman: I wonder if the logic for setting "ExpirationDateTime" should be moved into the constructor since all users of this class calculate it the same way:
```
        public class TokenCacheEntry
        {
            public TokenCacheEntry(TokenResponse tokenResponse)
            {
                TokenResponse = tokenResponse;
                ExpirationDateTime = DateTime.UtcNow.AddSeconds(tokenResponse.ExpiresIn);
            }

            public TokenResponse TokenResponse { get; }
            public DateTime ExpirationDateTime { get; }
        }
```
This would enable you to make the "TokenResponse" and "ExpirationDateTime" properties read-only.

PR #3389, opened 2025-03-07
Fix out of memory error with very large RetentionTimes table

nickshulman:
Mike has a .blib file with 60 million rows in the RetentionTimes table.
That .blib file was hitting an OutOfMemoryException in "BiblioSpecLiteLibrary.CreateCache" because that method uses a MemoryStream, and the MemoryStream was more than 2 billion bytes.

This pull request fixes the OutOfMemoryException by writing the MemoryStream out to disk as soon as it grows to larger than 10MB.

If it is not possible to write to disk (because the folder is read-only) then it will keep the entire cache file in memory so the user will still be able to use the .blib file (so long as the cache file is not larger than 2GB).

Next version we should get rid of the .slc file.
I have some ideas about how to speed up reading from the SQLite database using multiple threads so that there won't be any speed improvement having a cache file.


brendanx67:
Score being a double also seems like overkill to me.

PR #3391, opened 2025-03-10
Fix class "IntegerList.Bits".

nickshulman:
The class that I added last week was not storing the correct values at all.
Added unit test "IntegerListTest"


brendanx67:
Do other new additions have unit tests, too? Like factors?

PR #3393, opened 2025-03-11
Skyline: Fix minor issues found testing APD changes

brendanx67:
- Fixed mirror plots resetting mirror combos with every selection change (also complained about at courses)
- Fixed ImmediateWindow to restore disabled cut-copy-paste-delete menu items when losing focus
- Fixed ImmediateWindow to set focus to Targets view when Esc is pressed

PR #3395, opened 2025-03-12
Fix TestMethodEditTutorial

nickshulman:
PR #3393 broke TestMethodEditTutorial because the wrong spectrum ends up being selected when the selected Precursor in the Targets tree changes to something that was found in a different library.

### DIFF
--- a/pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
+++ b/pwiz_tools/Skyline/Controls/Graphs/GraphSpectrum.cs
@@ -454,7 +454,9 @@ namespace pwiz.Skyline.Controls.Graphs
                     {
                         comboSpectrum.Items.Clear();
                         comboMirrorSpectrum.Items.Clear();
+                        selectedSpectrum = null;
                         selectedSpectrumIndex = -1;
+                        selectedMirror = null;
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


brendanx67 (dismissed):
This is too simplistic and breaks other functionality. It seems like we are going to need to store whether the selections in these combo boxes were set by the user and persist them when they are and clear them when they are not.

nickshulman:
It sounds like the way to fix the unit test while keeping your original fix in place is to give this thing a more robust way of deciding whether the currently selected spectrum is still in the list of choices.
TestMethodEditTutorial fails because it goes from selecting a precursor in the Targets tree with one spectrum option "Yeast (GPM)" and then selects a different precursor in the Targets tree where the two choices are "Yeast (Atlas)" and "Yeast (GPM)" and the spectrum that should be selected is "Yeast (Atlas)". If Skyline knew that that "Yeast (GPM)" was a completely different spectrum from the previous "Yeast (GPM)" then the "comboSpectrum.Items.IndexOf(selectedSpectrum) == -1" would cause it to do the right thing.

brendanx67:
I am going to try my idea of whether the user explicitly selected the selected value in the combo or not.

nickshulman:
Way back in 2010, the way that this worked was that Skyline knew that if the selectedIndex was non-zero it meant the user had explicitly chosen a spectrum:
```
// Unless the current selected index is the one matching the one currently
// in use by the precursor (zero), then try to stay viewing the in-use spectrum (zero)
int selectedIndex = comboSpectrum.SelectedIndex;
```
I bet this got broken in 2022 when code was added to set "selectedSpectrumIndex" to -1:
aa8be898e82295abb88b2c4d0cfc33ee162a8603

PR #3401, opened 2025-03-19
Move Associate Proteins work to a background thread

nickshulman:
This pull request makes it so that all of the processing that the Associate Proteins dialog does happens on a background thread so that it does not block the user if they want to cancel it or change settings before it is finished.

When it is doing work, a progress bar appears at the bottom of the form:
image

If an error happens, a big red "X" appears in the lower left corner of the form (it uses the icon called "SystemIcons.Error")
image

If you click on the button it shows you the message box that you would have seen (complete with "More Info" button):
image

I wonder whether the progress bar should be displayed somewhere else. Maybe the "Protein association results" grid should be hidden whenever the dialog is doing work and the progress bar or error button could appear there.


pwiz_tools/Skyline/TestFunctional/ImportPeptideSearchTest.cs
```diff
@@ -551,18 +551,20 @@ private void TestWizardDecoysAndMinPeptides()
...
+                    out unmappedOrRemoved);
                 Assert.AreEqual(2, proteins);
                 Assert.AreEqual(2, peptides);
                 Assert.AreEqual(2, precursors);
                 Assert.AreEqual(6, transitions);
                 Assert.AreEqual(1, unmappedOrRemoved);
                 peptidesPerProteinDlg.MinPeptides = 2;
             });
-            WaitForConditionUI(() => peptidesPerProteinDlg2.DocumentFinalCalculated);
+            WaitForConditionUI(() => peptidesPerProteinDlg.DocumentFinalCalculated && peptidesPerProteinDlg.DocumentFinalCalculated);
```
chambm, line 567: Is this a typo?

chambm:
> I wonder whether the progress bar should be displayed somewhere else. Maybe the "Protein association results" grid should be hidden whenever the dialog is doing work and the progress bar or error button could appear there.

I'm not sure how to do that in a way that would look good. But the error icon's background is a different color than the control's background. Can that be fixed? Shouldn't the results have been cleared if there was an error?

nickshulman:
The background only looks like that when the mouse is hovering over the thing.
When the mouse is not pointing at it, it looks like this:
image
(oops, I see that it needs to be a pixel larger so that it does not get clipped on the right and bottom edges)

Good point about the clearing out the results that are displayed in the DataGridView when there is an error-- I'll fix that.
By the way, the way that the data gets displayed there by handling the "CellValueNeeded" event is an unusual way to populate a grid. I think I will change that to the more conventional way of adding a few DataGridViewRow's and set the Value and Format properties of DataGridViewCell objects.

chambm:
> By the way, the way that the data gets displayed there by handling the "CellValueNeeded" event is an unusual way to populate a grid. I think I will change that to the more conventional way of adding a few DataGridViewRow's and set the Value and Format properties of DataGridViewCell objects.

I'm sure there was a reason I went with virtual mode. It may have just been that I wanted numbers less than 10000 to not have the thousands separator. Is there a numeric format that can do that?

nickshulman:
> Is there a numeric format that can do that?

There is no format string that can do that, but when you set the cell's value you can also give it a different format string depending on the value:
```
cell.Value = value;
cell.Style.Format = value >= 10_000 ? @"N0" : string.Empty;
```

PR #3406, opened 2025-03-23
Show percent complete when copying .skyd file to new name

nickshulman:
Fixed to show percent complete when copying .skyd file when saving document to new name.

Add "CopyFileWithProgress" method to Kernel32.

When you do "File > Save As" in Skyline, the .skyd file needs to be copied to a new name. Currently, while that is happening the user will see the LongWaitDlg saying "Optimizing data file..." without any indication of how close it is to being finished.
This PR adds a wrapper around "CopyFileEx" in Kernel32 which allows passing in a callback which receives updates indicating how many bytes have been copied.

(The reason that it says "Optimizing data file" while copying is because if the user had removed any Replicates before saving then Skyline would create a new .skyd file with only the data for the subset of Replicates that remain, and in that case the user does see the percent complete in the LongWaitDlg)


nickshulman:
I think this will just end up getting committed as part of the Peak Imputation work in PR #3428

PR #3418, opened 2025-04-03
Skyline: Clear up confusion around image resources

brendanx67:
- Added to CodeInspectionTest to error when image files are included in both RESX and CSPROJ files as either EmbeddedResources or Content.
- This is usually a mistake that puts the image in the release twice (embedded or not)


nickshulman (approved):
Looks good to me.

PR #3419, opened 2025-04-03
Skyline: Add "Compound" field to all Thermo isolation lists

brendanx67:
- The BuildThermoMethod.exe supports a Compound field, as do the MassTables it injects our isolation lists into
- Our current isolation lists and methods look empty without these
- Currently the ThermoQExactiveIsolationListExporter puts the text in the comment which does not carry through to a method
- Philip Remes has also agreed that he would like to have the Compound field populated for the Stellar


pwiz_tools/Skyline/Model/Export.cs
```diff
@@ -4257,6 +4256,8 @@ protected override void WriteTransition(TextWriter writer, int fileNumber, Pepti
             TransitionGroupDocNode nodeTranGroup, TransitionGroupDocNode nodeTranGroupPrimary, TransitionDocNode nodeTran,
             int step)
         {
+            writer.Write(GetCompound(nodePep, nodeTranGroup, true));
```
nickshulman, line 4259: Do we need to worry about what if the compound name contains commas?
Would it work to put quotes around the value, or should the comma get replaced with a safe character?

brendanx67: I don't know that we have nailed this everywhere in this file, but it seems like only the small molecule value:

nodeGroup.CustomMolecule.InvariantName

Could have a comma. Or maybe a custom isotope label type? It seems like the solution would be to use:

text.ToDsvField(FieldSeparator, FieldSeparatorReplacement)

As gets used in ThermoMassListExporter and WatersMassListExporter:

            writer.WriteDsvField(GetCompound(nodePep, nodeTranGroup), FieldSeparator, FieldSeparatorReplacement);

Though, If I put this in the base implementation of GetCompound() then, the above could become just writer.Write().

brendanx67: This doesn't rely on the vendor software being able to deal with the full flexibility of CSV format.

brendanx67: Okay, I added code for separator replacement to GetCompound(). I am going to wait for Brian P's return to add a test for this, because I don't see an easy way to change the name of a molecule in the SrmDocument tree. The CustomMolecule is still part of the Peptide which is the ID class for PeptideDocNode. So, changing it seems to require a lot of work. Though, I may be missing something.

pwiz_tools/Skyline/Model/AbstractMassListExporter.cs
```diff
@@ -224,9 +224,14 @@ private void NextFile(FileIterator fileIterator)
...
-        public string GetCompound(PeptideDocNode nodePep, TransitionGroupDocNode nodeGroup)
+        public string GetCompound(PeptideDocNode nodePep, TransitionGroupDocNode nodeGroup, bool showLabelType = false)
         {
-            return nodePep.Peptide.IsCustomMolecule ? nodeGroup.CustomMolecule.InvariantName : Document.Settings.GetModifiedSequence(nodePep).Sequence;
+            string compoundName = nodePep.Peptide.IsCustomMolecule
+                ? nodeGroup.CustomMolecule.InvariantName
+                : Document.Settings.GetModifiedSequence(nodePep).Sequence;
+            if (showLabelType)
+                compoundName = string.Format(@"{0} ({1})", compoundName, nodeGroup.TransitionGroup.LabelType);
+            return compoundName.ToDsvField(FieldSeparator, FieldSeparatorReplacement);
```
nickshulman, line 234: It looks like all the other code in the project that was using "GetCompound" was already escaping it, doing things like this in "ThermoMassListExporter.WriteTransition":
`writer.WriteDsvField(GetCompound(nodePep, nodeTranGroup), FieldSeparator, FieldSeparatorReplacement);`

or even this code in ThermoQuantivaMassListExporter.WriteTransition which does not specify a replacement character and would theoretically be relying on quotes being correctly interpreted:
`writer.WriteDsvField(compound, FieldSeparator);`

brendanx67:
I recognized that, but felt my change could not possibly make things worse.
The comment I found about Xcalibur not supporting quoting makes it seem
like my change would find the case you found for the Quantiva.

If you like, I can remove the redundant calls. I hope you agree they won’t
actually impact the output. Though my change will fix what seems like a bug
in the Quantiva code.

nickshulman (approved):

PR #3421, opened 2025-04-06
Fix "TestThermoNegativeFAIMS"

nickshulman:
TestThermoNegativeFAIMS is expecting that the compound name will be a quoted value in the CSV file with a comma in it.


brendanx67 (approved):
Okay. A quoted Compound field does seem to work on the Altis in the MacCoss lab. So, probably also a Quantiva. It may be that this is true of the modern Thermo method editor on all instruments. We should test and if so adapt for the 25.1 release. It may be that the comment on this line is no longer true about XCalibur.

FieldSeparatorReplacement = @"_";  // For use in formats where quoting the value does not suffice, as reportedly in Excalibur

Though, this will not work for File > Export > Method, since BuildThermoMethod.exe uses this:

 string[] fields = (lines[i].Contains('\t')) ? lines[i].Split('\t') : lines[i].Split(',');

So, no quote handling.

PR #3423, opened 2025-04-07
Crux 4.3 (includes Comet 2025.1.1) (#3416)

chambm:
- updated Crux dependency to 4.3 (includes Comet 2025.1.1)
- removed Comet additional settings that are inapplicable or esoteric
* updated audit logs


brendanx67 (approved):
Looks like the master PR to me, as it should since the branch is virtually identical to master

PR #3428, opened 2025-04-09
Peak imputation settings

nickshulman:
This pull request adds some peak imputation settings to the "Prediction" tab on the Peptide Settings dialog:
image

There is more information about how this feature works here:
https://skyline.ms/wiki/home/software/Skyline/page.view?name=PeakImputation


brendanx67:
Just looking at the image, maybe:
-- Impute peak boundaries for ----
[x] Missing peaks
RT delta greater than:
[  1.0 ] min
Width delta greater than:
[       ] min
--------------------------------------
The control labels need mnemonics and the edit boxes need unit labels.

nickshulman:
> Just looking at the image, maybe: -- Impute peak boundaries for ---- 

That does sound better.
I wonder if it should be "Impute peak boundaries if" instead of "Impute peak boundaries for" because "if" (or maybe "when") works better with "RT delta greater than".
The units of "peak width variation" is actually percent, so I'll add a "%" next to that textbox.

PR #3429, opened 2025-04-11
Fix ObjectDisposedException in ChromatogramCache

nickshulman:
Occasionally, the stream that the chromatogram cache is reading from gets closed while it is using it.
This has resulted in a few ObjectDisposedException's from ChromatogramCache being reported on the exception web.

I ran into this while doing a "Rescore" because every time it's finished with one file it closes the stream, which is still being used by other threads.
This is not supposed to be a problem because a PooledFileStream will open itself back up the next time anyone wants to use it, but "ConnectionId.Disconnect" was missing a lock statement.
(That lock statement used to be there, but I removed it when I fixed a different threading problem PR #2979 last year)

Also, CachedChromatogramDataProvider was reading the entire .skyd file again for every replicate, which did not seem to make sense and is really slow if there are hundreds of replicates.

Added RescoreSimultaneouseTest which tests this scenario.


nickshulman:
Actually, the premise of this fix was flawed.

PR #3442, opened 2025-04-18
Hide the selection in the grid in AssociateProteinsDlg

nickshulman:
A "ClearSelection" call was incorrectly removed in PR #3401


chambm:
Just a reminder that the merge from master without a subsequent squash is what messes up the backport action. (And I don't really know how to fix that voodoo.). But I think you can `<slash>squash` and that should do a squash action after which you can safely merge with backport.

PR #3456, opened 2025-04-29
20250428 add tide

kfattila:
Adds Tide-search engine to Skyline. This PR contains wrapper functions in Skyline to call user-parameterized tide-search from crux 4.2.

### DIFF
--- a/pwiz_tools/Skyline/Model/DdaSearch/TideSearchEngine.cs
+++ b/pwiz_tools/Skyline/Model/DdaSearch/TideSearchEngine.cs
@@ -1,0 +1,856 @@
+﻿/*
+ * Original author: Matt Chambers <matt.chambers42 .at. gmail.com >
+ *
+ * Copyright 2024 University of Washington - Seattle, WA
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using pwiz.BiblioSpec;
+using pwiz.Common.Chemistry;
+using pwiz.Common.Collections;
+using pwiz.Common.SystemUtil;
+using pwiz.Skyline.Model.DocSettings;
+using pwiz.Skyline.Model.Results;
+using pwiz.Skyline.Model.Tools;
+using pwiz.Skyline.Properties;
+using pwiz.Skyline.Util;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Drawing;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading;
+using pwiz.Skyline.Model.AuditLog;
+using pwiz.Skyline.Util.Extensions;
+using MSAmanda.Utils;
+using NHibernate.SqlCommand;
+using Enzyme = pwiz.Skyline.Model.DocSettings.Enzyme;
+using System.Xml.Linq;
+
+namespace pwiz.Skyline.Model.DdaSearch
+{
+    public class TideSearchEngine : AbstractDdaSearchEngine, IProgressMonitor
+    {
+        private List<string> TIDE_SETTINGS = new List<string>();
+
+        // Percolator settings
+        private const string PERCOLATOR_TEST_QVALUE_CUTOFF = "test-fdr";
+        private const string PERCOLATOR_TRAIN_QVALUE_CUTOFF = "train-fdr";
+        private readonly string[] PERCOLATOR_SETTINGS = { PERCOLATOR_TEST_QVALUE_CUTOFF, PERCOLATOR_TRAIN_QVALUE_CUTOFF };
+
+        private const string KEEP_INTERMEDIATE_FILES = "keep-intermediate-files";
+
+        public TideSearchEngine(double percolatorQvalueCutoff)
+        {
+            AdditionalSettings = new Dictionary<string, Setting>
+            {
+                {PERCOLATOR_TEST_QVALUE_CUTOFF, new Setting(PERCOLATOR_TEST_QVALUE_CUTOFF, percolatorQvalueCutoff, 0, 1)},
+                {PERCOLATOR_TRAIN_QVALUE_CUTOFF, new Setting(PERCOLATOR_TRAIN_QVALUE_CUTOFF, 0.01, 0, 1)},
+                {KEEP_INTERMEDIATE_FILES, new Setting(KEEP_INTERMEDIATE_FILES, true)},
+            };
+
+            // Tide-search settings
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("max-precursor-charge", 5, 1, 9));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("min-precursor-charge", 1, 1, 9));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("deisotope", 0.0, 0.0));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("isotope-error", ""));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("min-peaks", 20, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("mz-bin-offset", 0.4, 0, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("fragment-tolerance", 0.02, 0.00001, 2));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("override-charges", false));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("remove-precursor-peak", false));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("remove-precursor-tolerance", 1.5, 0));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("scan-number", ""));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("score-function", "xcorr", "xcorr|combined-p-values".Split('|')));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("skip-preprocessing", false));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("spectrum-max-mz", 1e+09, 0));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("spectrum-min-mz", 0.0, 0));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("use-flanking-peaks", false));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("use-neutral-loss-peaks", true));
+            
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("precursor-window", 50));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("precursor-window-type", "ppm", "mass|mz|ppm".Split('|')));   // No support for 'mass' unit
+
+            // Add param-medic options.
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("pm-charges", "0,2,3,4"));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("pm-max-frag-mz", 1800, 0));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("pm-max-precursor-delta-ppm", 50, 0));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("pm-max-precursor-mz", 1800, 0));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("pm-max-scan-separation", 1000, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("pm-min-common-frag-peaks", 20, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("pm-min-frag-mz", 150, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("pm-min-peak-pairs", 200, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("pm-min-precursor-mz", 400, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("pm-min-scan-frag-peaks", 40, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("pm-pair-top-n-frag-peaks", 5, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("pm-top-n-frag-peaks", 30, 1));
+
+            // Input and output 
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("fileroot", ""));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("mass-precision", 4, 0));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("mzid-output", false));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("mztab-output", false));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("precision", 8, 0));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("print-search-progress", 10000));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("spectrum-parser", "pwiz", "pwiz|mstoolkit".Split('|')));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("sqt-output", false));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("store-index", ""));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("store-spectra", ""));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("top-match", 5, 1));
+//            AddAdditionalSetting(TIDE_SETTINGS, new Setting("txt-output", true));   // Must be true
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("use-z-line", true));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("verbosity", 30));
+
+            // Tide-index settings
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("memory-limit", 4, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("auto-modifications-spectra", ""));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("clip-nterm-methionine", false));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("isotopic-mass", "mono", "mono|average".Split('|')));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("max-length", 50, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("max-mass", 7200, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("min-length", 6, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("min-mass", 200, 1));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("cterm-peptide-mods-spec", ""));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("cterm-protein-mods-spec", ""));
+            AddAdditionalSetting(TIDE_SETTINGS, new Setting("min-mods", 0, 0));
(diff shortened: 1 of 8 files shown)


chambm:
I don't see a new test e.g. TestDdaSearchTide. Did that file get left out of the commit?

chambm:
Also is they're a reason you're using Crux 4.2 instead of 4.3?

chambm:
The params that get set by SetTideParam should not be in AdditionalSettings. The former are the params that are exclusively set by the GUI elements, the latter exclusively edited in the table shown by clicking the "Additional Settings" button. If there's overlap then it would require synchronizing the overlapping settings and that isn't currently handled.

kfattila:
> I don't see a new test e.g. TestDdaSearchTide. Did that file get left out of the commit?

I was following the Comet PR. and I did not see TestDdaSearchComet? Can you point me to an example?

kfattila:
> Also is they're a reason you're using Crux 4.2 instead of 4.3?

There is an issue with 4.3. It hangs up after processing the fasta file. I'll discuss it with Bill. Crux 4.2 works for now.

kfattila:
> The params that get set by SetTideParam should not be in AdditionalSettings. The former are the params that are exclusively set by the GUI elements, the latter exclusively edited in the table shown by clicking the "Additional Settings" button. If there's overlap then it would require synchronizing the overlapping settings and that isn't currently handled.

I removed the duplicated options. Note that, now when the user selects Trypsin (semi), then tide-search will not generate semi-tryptic peptides. It will cleave according to the "digestion" parameter specified by the user in the Additional Settings dialog box.

chambm:
> TestDdaSearchComet

It's a TestMethod in TestDdaSearch.cs. Easiest way to search is `Ctrl+,` and then just type in what you're looking for (I guess the way I phrased it made it sound like a filename).

kfattila:
> > TestDdaSearchComet
> 
> It's a TestMethod in TestDdaSearch.cs. Easiest way to search is `Ctrl+,` and then just type in what you're looking for (I guess the way I phrased it made it sound like a filename).

Hi, Indeed, I forgot to commit it. Now it is there. Sorry.

chambm:
When you run the test do you get ` FATAL: Multiple static mods on same residue detected.` like I do?

kfattila:
> When you run the test do you get ` FATAL: Multiple static mods on same residue detected.` like I do?

Hi,

I get the following output on the consol:
```
TestRunner test=TestDdaSearchTide loop=1

Process: 50720

Running 1 tests...

#!!!!! DEBUGGING STARTED !!!!!

# Pass 2+: Run tests in each selected language.
[13:26]   2.0   TestDdaSearchTide                              (fr)
```
Then it hangs up, does not use CPU process, does not terminate... Just hangs...

kfattila:
Hi, How can I specify a non-specific digestion in Skyline? (so, it is not fully-tryptic, not semi-tryptic, I mean the non-tryptic digestion)
I tried to create a cleavage rule: `[X]|[]` but Skyline made an error saying: `X` is not an amino acid. Thanks.

chambm:
> Hi, How can I specify a non-specific digestion in Skyline? (so, it is not fully-tryptic, not semi-tryptic, I mean the non-tryptic digestion) I tried to create a cleavage rule: `[X]|[]` but Skyline made an error saying: `X` is not an amino acid. Thanks.

I'm not sure Skyline is really designed to support non-specific digestion, but you could always just put all the amino acids in there instead of X. Tedious, but non-specific digestion was always going to be tedious. :)

brendanx67:
Probably best to step back and think about the ideal workflow you are trying to achieve and then discuss that with us. I doubt what Matt just suggested would be satisfying for a user.

chambm:
I expect he's trying to translate enzyme parameters from Skyline to Tide params. But I think it's better to say we just won't support non-specific digestion from Skyline, just as we don't support importing those searches very well either.

Although if a user really wants a non-specific digestion of a single protein, why would my suggestion by unsatisfying?

nickshulman:
> Although if a user really wants a non-specific digestion of a single protein, why would my suggestion by unsatisfying?

In Skyline, specifying that the enzyme cleaves on every amino acid does not work very well because every amino acid in the peptide is then a missed cleavage, and you are not allowed to choose a number greater than "9" for "max missed cleavages".

kfattila:
Thanks for the `offscreen=0` tip, it works, and I found the issue. 
 
 Many thanks for your comments about the non-enzymatic cleavage too. We have recently made several updates to tide-search, and one of them is that tide-search can efficiently handle terabytes of peptide index databases, resulting from non-enzymatic cleavages by outsourcing the data from RAM to disk. Tide has an additional parameter, called `digestion` to control the cleavage:
```
AddAdditionalSetting(TIDE_SETTINGS, new Setting("digestion", "full-digest", "full-digest|partial-digest|non-specific-digest".Split('|')));
```
We could keep this parameter, but then it would overwrite the "semi vs full cleavage" option in the enzyme definition.

How did you handle this in Comet? Can Comet generate non-enzymatic cleavages in Skyline?

wsnoble:
I agree that supporting non-enzymatic digestion is desirable. In comet, this is enzyme number 0 ("cut everywhere"), as described [here](https://uwpr.github.io/Comet/parameters/parameters_202501/search_enzyme_number.html). Is this functionality not supported in Skyline?

brendanx67:
> I agree that supporting non-enzymatic digestion is desirable. In comet, this is enzyme number 0 ("cut everywhere"), as described [here](https://uwpr.github.io/Comet/parameters/parameters_202501/search_enzyme_number.html). Is this functionality not supported in Skyline?

You can certainly work with a spectral library built from a non-emymatic search. Our File > Search wizard does not currently support this workflow, because our code to generate targets from a FASTA file requires at least a semi-cleavage enzyme. That would require new code to support matching arbitrary peptides to protein sequences in our FASTA processing. So, it's not currently a supported flow for File > Search. A user could still build a library from these results and add arbitrary peptides from them to Skyline as targets. I can't recall if Refine > Associate Proteins can then be used to associate these arbitrary peptides with proteins.

So, it is not as straightforward as I am sure a Tide or Comet user would hope approaching Skyline for a search of this type.

brendanx67:
I checked and associate proteins works in the same way. It will not arbitrarily search all FASTA sequences for matching peptide sequences. It also only searches based on the specified protease enzyme. So, if that is just Trypsin, it will fail to match a semi-tryptic peptide. But if you specify Trypsin (semi) then it will match semi-tryptic peptides.

It seems like someone would need to do a fair bit of work to create a nice experience for Skyline users trying to use non-enzymatic cleavage targets.

nickshulman:
> I checked and associate proteins works in the same way. It will not arbitrarily search all FASTA sequences for matching peptide sequences. It also only searches based on the specified protease enzyme.

That used to be the behavior, but, as of October, 2024, Skyline-daily will try to find proteins from which the peptide could have been cleaved by the enzyme, but if no such protein exists, then it will fall back to do a simple substring match.

The code that handles this is in ProteinAssociation.cs:
```
if (!proteinPeptideMatches.EnzymaticPeptides.Contains(peptideSequence))
{
    // The peptide could not have been digested by the enzyme from this protein sequence.
    // Only skip it if there is at least one protein that could produce the digested peptide
    if (allEnzymaticPeptides.Contains(peptideSequence))
    {
        continue;
    }
}
```

chambm:
The enzymatic restraint in Associate Proteins is more of an addon than an inherent part. The underlying matching is Aho-Corasick, which doesn't care about specificity. It looks like Nick changed Associate Proteins so that it ignores max missed cleavages. So the all-amino-acids enzyme pattern I suggested would work, but only for Associate Proteins? I'm going to check this myself because the import wizard uses the same logic as Associate Proteins (unlike Import FASTA).

brendanx67:
Okay. Oops. I checked 24.1. Sounds like it got improved for 25.1.

Anyone care to venture a guess at how big a job it would be to adapt the peptide search wizard to do its best to include all peptides from the spectral library and use the same matching strategy as Associate Proteins, where when a peptide is not found with cleavage, it searches the FASTA for it?

nickshulman:
> Anyone care to venture a guess at how big a job it would be to adapt the peptide search wizard to do its best to include all peptides from the spectral library and use the same matching strategy as Associate Proteins, where when a peptide is not found with cleavage, it searches the FASTA for it?

Part of that is straightforward to implement. It's basically making it so that if "Pick peptides matching" is either "Library" or "Library of Filter", PeptideGroupDocNode.GetPeptideNodes includes all substrings which can be found in the library.

The other part is trickier, where non-enzymatic peptides are only matched if there is no other protein which could produce that peptide. To get that to work with "Import FASTA", that would require two passes through the FASTA file. That might mean that we would need to create a background proteome so that Skyline would continue to choose the same set of peptides when they do "Refine > Advance > Auto select all matching peptides" etc.

chambm:
I just tested it with a MSFragger search within Skyline (had to make some modifications to allow it to do a ntt=0 search) with an "all amino acids" enzyme and as Nick said, the ImportFasta function doesn't allow the non-specific peptides to be created. But the library has them and after "Add All" from Library explorer and Associate Proteins, I've got non-specific peptides associated with proteins as expected. The number of missed cleavages is equal to the peptide length so it should probably just be omitted: I don't think number of missed cleavages is meaningful for a non-specific search.

Why are we digesting the FASTA first for the import search wizard instead of just importing whatever the created library contains (minus document filters/parsimony)?

brendanx67:
Probably, Kaipo just followed File > Import > FASTA. In other words, code
was written a long time ago.

brendanx67:
Others have suggested they would like to see Skyline do a better job of
getting everything detected in the search into the Skyline document…
Though, for a library import, the user may provide a subset FASTA they want
to target.

kfattila:
Thanks a lot for your comments. I'll get tide working without non-enzymatic cleavage and then do this upgrade in a separate step later.

kfattila:
Hi, 

Sorry for the late reply. I managed to fix the error: `FATAL: Multiple static mods on same residue detected.`. Now tide and percolator runs to completion, but I have another issue. 

Tide generates a single output file even when multiple MS input files are passed, so there is just one single percolator.pepXML file after tide-search. But Skyline expects a percolator.pepXML file for every MS input file.  How can we resolve this issue?

chambm:
You change the code so it doesn't loop over the sources, it just adds percolator.pepXML. BTW I don't actually know that Tide pepXML will work with BiblioSpec, that may be another aspect that needs to be fixed.

kfattila:
Ok! Now, I get the following error with BlibBuild:
```
Skyline (64-bit : developer build) 24.1.1.462 (a2438f8de)

System.IO.IOException: ERROR: While searching for spectrum file for the search results file 'percolator.pep.xml', could not find matches for the following basename with any of the supported file extensions (.mz5, .mzML, .mzXML, .raw, .wiff, .wiff2, .d, .lcd, .ms2, .cms2, .bms2, .pms2):
ERROR: NA
ERROR: 
ERROR: In any of the following directories:
ERROR: D:\ProteoWizard\pwiz_tools\Skyline\TestResults\TestRunner results\DdaSearchTest
ERROR: D:\ProteoWizard\pwiz_tools\Skyline\TestResults\TestRunner results\
ERROR: D:\ProteoWizard\pwiz_tools\Skyline\TestResults\

Command-line: D:\ProteoWizard\pwiz_tools\Skyline\bin\x64\Debug\BlibBuild -s -A -H -o -c 0.1 -i TestDdaSearch -K -S "D:\ProteoWizard\pwiz_tools\Skyline\TestResults\TestRunner results\DdaSearchTest\TestDdaSearch.redundant202505221233.stdin.txt" "D:\ProteoWizard\pwiz_tools\Skyline\TestResults\TestRunner results\DdaSearchTest\TestDdaSearch.redundant.blib"
Working directory: D:\ProteoWizard\pwiz_tools\Skyline\TestResults\TestRunner results\DdaSearchTest
Exit code: 1 
```

I guess the issue is that BlibBuilder expects the search results file names derived from the input spectrum file names. Shoudn't the input spectra names be given in the pep.xml file?

nickshulman:
In that error message, the line that says:
`ERROR: NA`
should have a filename instead of `NA`.

The error message is saying that it's trying to find a raw file whose name starts with `NA` and ends with any of several possible extensions including ".mz5", ".mzML", etc. in a couple of different possible folders.

I think that in pepxml files, the filename usually comes from the `base_name=` attribute on a `<msms_run_summary>` tag:
```
   } else if(isElement("msms_run_summary",name)) {
      fileroot_ = getRequiredAttrValue("base_name",attr);
```

kfattila:
> In that error message, the line that says: `ERROR: NA` should have a filename instead of `NA`.
> 
> The error message is saying that it's trying to find a raw file whose name starts with `NA` and ends with any of several possible extensions including ".mz5", ".mzML", etc. in a couple of different possible folders.
> 
> I think that in pepxml files, the filename usually comes from the `base_name=` attribute on a `<msms_run_summary>` tag:
> 
> ```
>    } else if(isElement("msms_run_summary",name)) {
>       fileroot_ = getRequiredAttrValue("base_name",attr);
> ```

Indeed it is NA:
```
<msms_run_summary base_name="NA" msManufacturer="NA" msModel="NA" msIonization="NA" raw_data_type="NA" raw_data="NA" >
```.

What should be in these fields when the search results are obtained from searching couple of spectrum files at once?

I'll try to fix this in Tide.

**Comment by kfattila** [comment]
> At least `base_name` should be the file the spectrum came from, right? Tide used to report this in its TSV output, so that information should be available.

This 'base_name' is in the header of the pep.xml, This is not a filed in the PSM section.

**Comment by kfattila** [comment]
Hi guys, 

Thanks again for your help; I managed to fix the psm converter in crux, so now Skyline and BlibBuild works. Now, I get this exception error, but I cannot figure out what could be the problem, the GUI disappears. Any suggestions?


```
Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException
  HResult=0x80131500
  Message=Assert.Fail failed. Microsoft.VisualStudio.TestTools.UnitTesting.AssertFailedException: Assert.AreEqual failed. Expected:<123>. Actual:<149>. 
   at Microsoft.VisualStudio.TestTools.UnitTesting.Assert.AreEqual[T](T expected, T actual, String message, Object[] parameters)
   at pwiz.SkylineTestFunctional.DdaSearchTest.<>c__DisplayClass20_0.<TestSearch>b__16() in D:\ProteoWizard\pwiz_tools\Skyline\TestFunctional\DdaSearchTest.cs:line 647
   at pwiz.SkylineTestUtil.AbstractFunctionalTest.<>c__DisplayClass37_0.<RunUI>b__0() in D:\ProteoWizard\pwiz_tools\Skyline\TestUtil\TestFunctional.cs:line 358
  Source=Microsoft.VisualStudio.QualityTools.UnitTestFramework
  StackTrace:
   at Microsoft.VisualStudio.TestTools.UnitTesting.Assert.Fail(String message)
   at pwiz.SkylineTestUtil.AbstractFunctionalTest.<>c__DisplayClass37_0.<RunUI>b__0() in D:\ProteoWizard\pwiz_tools\Skyline\TestUtil\TestFunctional.cs:line 362

```

File: pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs [hunk 1]
```diff
@@ -216,6 +216,32 @@ public void TestDdaSearchCometAutoTolerance()
...
+
+            TestSettings = new DdaTestSettings
+            {
+                SearchEngine = SearchSettingsControl.SearchEngine.Tide,
+                FragmentIons = "b,y",
+                Ms2Analyzer = "Default",
+                PrecursorTolerance = new MzTolerance(15, MzTolerance.Units.ppm),
+                FragmentTolerance = new MzTolerance(1.0005),
+                AdditionalSettings = new Dictionary<string, string>(),
+                ExpectedResultsFinal = new ExpectedResults(123, 297, 358, 1074, 144)
```
nickshulman: You will probably need to change these numbers in order for this test to pass.
If you modify the code so that the property "IsRecordMode" (on line 317) is true, then the test will write out the expected numbers instead of failing, and you can copy those numbers from the test output and put them here and then change "IsRecordMode" back to false.

pwiz_tools/Skyline/Model/DdaSearch/TideSearchEngine.cs
```diff
@@ -807,24 +814,10 @@ public override void SetCutoffScore(double cutoffScore)
...
-            else
-            {
-                return Path.Combine(Path.GetDirectoryName(searchFilepath.GetFilePath()) ?? "",
-                    Path.ChangeExtension(@"tide", extension));
-            }
-        }
-
-        private string[] SupportedExtensions = { @".mzml", @".mzxml", @".raw" };
+       
+        private string[] SupportedExtensions = { @".mzml", @".mzxml", @".ms2", @".mgf", @".raw" };
```
chambm: Would the workflow actually work with ms2 and mgf files? We use mzML for the unit tests; I suppose we should really test with all these formats. Possibly by converting them from mzML in the test.

kfattila: I removed the ms2 and mgf file formats

chambm:
Looks like the test is still triggering the `FATAL: Multiple static mods on same residue detected.` error.

kfattila:
Hi, Sorry for the long silence, we were making changes in crux to be compatible with Skyline. Now I am trying to run the tester and I get this error below. We are trying to download crux 4.3.2. https://noble.gs.washington.edu/crux-downloads/crux-4.3.2/ Any help would be appreciated. :
```

TestRunner test=TestDdaSearchTide loop=1 offscreen=0

Process: 704

Running 1 tests...

#!!!!! DEBUGGING STARTED !!!!!

# Pass 2+: Run tests in each selected language.
[12:05]   2.0   TestDdaSearchTide                              (fr) # Downloading test data file https://ci.skyline.ms/skyline_tool_testing_mirror/crux-4.3.2.Windows.AMD64.zip to D:\ProteoWizard\pwiz_tools\Skyline\bin\x64\Debug\CachedDownloadsForTests\crux-4.3.2.Windows.AMD64.zip...


Closing open form of type pwiz.Skyline.Controls.SequenceTreeForm



Closing open MessageDlg: Download failed. Check your network connection or contact Skyline developers.
https://ci.skyline.ms/skyline_tool_testing_mirror/crux-4.3.2.Windows.AMD64.zip
Skyline (64-bit : developer build) 24.1.1.462 (a2438f8de)

System.Reflection.TargetInvocationException: Download failed. Check your network connection or contact Skyline developers.
https://ci.skyline.ms/skyline_tool_testing_mirror/crux-4.3.2.Windows.AMD64.zip ---> System.Exception: Download failed. Check your network connection or contact Skyline developers.
https://ci.skyline.ms/skyline_tool_testing_mirror/crux-4.3.2.Windows.AMD64.zip ---> System.Net.WebException: The remote server returned an error: (403) Forbidden.
   at System.Net.HttpWebRequest.EndGetResponse(IAsyncResult asyncResult)
   at System.Net.WebClient.GetWebResponse(WebRequest request, IAsyncResult result)
   at System.Net.WebClient.DownloadBitsResponseCallback(IAsyncResult result)
   --- End of inner exception stack trace ---
   at pwiz.Skyline.Util.SimpleFileDownloader.DownloadRequiredFiles(IEnumerable`1 filesToDownload, IProgressMonitor waitBroker) in D:\ProteoWizard\pwiz_tools\Skyline\Util\UtilInstall.cs:line 280
   at pwiz.Skyline.Alerts.SimpleFileDownloaderDlg.<>c__DisplayClass0_0.<Show>b__0(IProgressMonitor pm) in D:\ProteoWizard\pwiz_tools\Skyline\Alerts\SimpleFileDownloaderDlg.cs:line 128
   at pwiz.Skyline.Util.ProgressWaitBroker.PerformWork(ILongWaitBroker broker) in D:\ProteoWizard\pwiz_tools\Skyline\Util\UtilUI.cs:line 124
   at pwiz.Skyline.Controls.LongWaitDlg.RunWork(Action`1 performWork) in D:\ProteoWizard\pwiz_tools\Skyline\Controls\LongWaitDlg.cs:line 254
   --- End of inner exception stack trace ---
   at pwiz.Skyline.Util.Helpers.WrapAndThrowException(Exception x) in D:\ProteoWizard\pwiz_tools\Skyline\Util\Util.cs:line 1926
   at pwiz.Skyline.Controls.LongWaitDlg.PerformWork(Control parent, Int32 delayMillis, Action`1 performWork) in D:\ProteoWizard\pwiz_tools\Skyline\Controls\LongWaitDlg.cs:line 202
   at pwiz.Skyline.Controls.LongWaitDlg.PerformWork(Control parent, Int32 delayMillis, Action`1 performWork) in D:\ProteoWizard\pwiz_tools\Skyline\Controls\LongWaitDlg.cs:line 140
   at pwiz.Skyline.Alerts.SimpleFileDownloaderDlg.Show(Control parent, String title, IEnumerable`1 requiredFiles) in D:\ProteoWizard\pwiz_tools\Skyline\Alerts\SimpleFileDownloaderDlg.cs:line 128
   at pwiz.Skyline.FileUI.PeptideSearch.SearchSettingsControl.EnsureRequiredFilesDownloaded(IEnumerable`1 requiredFiles, Func`1 extraDownloadAction) in D:\ProteoWizard\pwiz_tools\Skyline\FileUI\PeptideSearch\SearchSettingsControl.cs:line 194



Closing open form of type pwiz.Skyline.FileUI.PeptideSearch.ImportPeptideSearchDlg


```

kfattila:
Hi, I wanted to see the Skyline code inspection issues (that test failed on TeamCity) but I do not have access to see the output. Could you please send me the output, or grant me access? Thank you. A.

chambm:
> Hi, I wanted to see the Skyline code inspection issues (that test failed on TeamCity) but I do not have access to see the output. Could you please send me the output, or grant me access? Thank you. A.

I can't figure out how to share the code inspection report with you from TeamCity: there's no easy way to export all the issues because you have to click on each category of issue to see those issues, you can't see all categories at once. So I will take care of the code inspection issues.

PR #3459, opened 2025-05-01
Skyline: New test that automates the Webinar 25 Acquisition Comparison tutorial

brendanx67:
Most of this PR is creating the tutorial in our HTML format with automated screenshots. For convenience, I am also committing the Word document I have been editing, and then converted to HTML, because I intend to send that out to the authors for review.

There are also a few important fixes to how Skyline works to simplify this tutorial:
1. Making it so the Import Peptide Search wizard uses natural sort for the imported results files to maintain consistency with the OpenDataSourceDialog. There was a section of the tutorial that had users re-order the replicates to deal with this inconsistency.
2. When importing a DDA search, if the MS/MS filtering settings are not already set to "DDA", they get set to "None". The MS/MS filtering settings are not visible to the user in this wizard and it is confusing to have them end up PRM, DIA, or SureQuant. They are not visible because we were assuming they would always be "None" until we added "DDA". We might want to consider showing them with just those 2 options now.
3. When importing a DDA search, the logic used to be that if precursor was not present, the ion types were set to only precursor. I made this more subtle so that we only set to precursor-only when the MS/MS filtering is set to "None" (i.e. not DDA) and I also remove the "min ion count" filter in this case to avoid removing everything for lack of product ions. Otherwise, the code simply adds precursor if it is missing.


bspratt (approved):
Changes look good - there is one test failure on TC, TestSkylineUninstall in ProteoWizard and Skyline Docker container (Wine x86_64) I've restarted that to see if it clears up

PR #3463, opened 2025-05-06
Ardia File Upload

ekoneil:
Support uploading Skyline documents to Ardia as a .sky.zip archive containing the open .sky document and associated files to the chosen directory on the server.

Highlights:
* Adds a new item to the File menu "Upload to Ardia..." appearing only when an Ardia account is registered
* Uses existing SkylineWindow.ShareDocument(...) to create a .zip file archive with modifications supporting multipart .zips.
* Extends PublishDocumentDlg with customizations for Ardia


pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/ArdiaClient.cs
```diff
@@ -66,6 +66,8 @@ public sealed class ArdiaClient
         private const string PATH_DELETE_FOLDER   = @"/session-management/bff/navigation/api/v1/folders/folderPath";
         private const string PATH_SESSION_COOKIE  = @"/session-management/bff/session-management/api/v1/SessionManagement/sessioncookie";
         private const string PATH_COMPLETE_MULTIPART_REQUEST = @"/session-management/bff/document/api/v1/documents/completeMultiPartRequest";
+        private const string PATH_GET_PARENT_BY_PATH = @"/session-management/bff/navigation/api/v1/navigation/path";
```
ekoneil: Nit: put this with the other ".../navigation/..." paths in the list.

pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/ArdiaClient.cs
```diff
@@ -84,15 +86,17 @@ public static ArdiaClient Create(ArdiaAccount account)
...
-            return new ArdiaClient(account, serverApiUrl, applicationCode, token);
+            return new ArdiaClient(account, serverUrl, serverApiUrl, applicationCode, token);
         }
 
-        private ArdiaClient(ArdiaAccount account, string serverUrl, string applicationCode, string token)
+        private ArdiaClient(ArdiaAccount account, string serverUrl, string serverApiUrl, string applicationCode, string token)
         {
             Account = account;
-            ServerUri = new Uri(serverUrl);
+            ServerUrl = serverUrl;
```
ekoneil: Use clearer property names. ServerUrl and ServerUri are so similar it's hard to tell reading the code that they refer to different hostnames. :) More explicit names will help - e.g. ApiUri and DataExplorerUri.

pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/ArdiaClient.cs
```diff
@@ -102,6 +106,7 @@ private ArdiaClient(ArdiaAccount account, string serverUrl, string applicationCo
         private string ApplicationCode { get; }
         private string Token { get; }
         private Uri ServerUri { get; }
+        private string ServerUrl { get; }
```
ekoneil: Use Uri rather than string. A more specific type makes the purpose of the property clearer and in this case Uri is safer, more explicit, and has the benefit of providing more URL-oriented features.

pwiz_tools/Skyline/Alerts/ArdiaLoginDlg.cs
chambm, line 1: This seems like a lot of churn on something that, at some point, will go away entirely: the ArdiaLoginDlg is doomed because Thermo asked us to use the system browser. We absolutely must, at some point, integrate system browser automation so we can properly test what users are actually doing.

ekoneil: Understand your point, especially wanting tests for what Skyline users actually experience. ArdiaLoginDlg has System Browser support so big parts of it (though certainly not all) will exist somewhere - and imo, cleanup will make those changes easier.

PR #3468, opened 2025-05-08
Fix Japanese ConsoleImportPeptideSearchTest and TestCommandLineExportThermoMethod

nickshulman:
Move "LocalizedResourcesTest" from "TestFunctional" into "TestData" so that it can use "ErrorChecker.cs". Add assert to LocalizedResourcesTest that if English text starts with "Error:", localized text starts with localized version of that.

### DIFF
--- a/pwiz_tools/Skyline/TestData/LocalizedResourcesTest.cs
+++ b/pwiz_tools/Skyline/TestData/LocalizedResourcesTest.cs
@@ -30,7 +30,7 @@ using System.Windows.Forms;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using pwiz.SkylineTestUtil;
 
-namespace pwiz.SkylineTestFunctional
+namespace pwiz.SkylineTestData
 {
     [TestClass]
     public class LocalizedResourcesTest : AbstractUnitTest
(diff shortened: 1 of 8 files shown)


pwiz_tools/Skyline/TestData/LocalizedResourcesTest.cs
```diff
@@ -91,24 +109,34 @@ private void ValidateLocalizedResource(object invariantValue, object localizedVa
...
             var invariantText = invariantValue as string;
             if (invariantText != null)
             {
                 var localizedText = (string)localizedValue;
                 Assert.AreEqual(MaxSubstitutionIndex(invariantText), MaxSubstitutionIndex(localizedText), message);
                 StringAssert.DoesNotMatch(localizedText, RegexLowercaseMnemonic, "Mnemonic should be uppercase: {0}", message);
+                Assert.IsFalse(localizedText.StartsWith(ENGLISH_ERROR_PREFIX), "Localized text should not start with the English '{0}' {1}", ENGLISH_ERROR_PREFIX, message);
+                if (ErrorChecker.IsErrorLine(invariantText))
+                {
+                    Assert.IsTrue(ErrorChecker.IsErrorLine(localizedText), "Localized text should start with localized form of 'Error:' {0}", message);
```
brendanx67, line 123: This seems like a subtle error message for the current case where a colon and space were left in place instead of using the Unicode double-wide colon and no space.

brendanx67:
By my simple find-in-files, it still looks like you are missing this one:

Code	File	Line	Column
    <value>エラー: {0} 再試行回数{1}/{2}。</value>	C:\proj\review\pwiz_tools\Skyline\Model\Koina\KoinaResources.ja.resx	164	12

pwiz_tools/Skyline/TestData/LocalizedResourcesTest.cs
brendanx67: Also, I think you probably want to explicitly check the colon, because to an untrained eye, it is difficult to notice the difference between a Unicode double-wide colon and an ascii colon with a space.

nickshulman:
> By my simple find-in-files, it still looks like you are missing this one:
> 

That one was fixed in the first commit in this pull request.

PR #3478, opened 2025-05-15
Stop creating .slc file for BiblioSpec libraries

nickshulman:
Skyline currently creates a .slc file in the same folder as the .blib file which Skyline uses after opening a Skyline document to read the list of spectra and retention times faster than it could be read from the .blib file.

This change makes it so that Skyline reads from the .blib file fast enough that the .slc file does not seem to be necessary.
With the TEI-REX .blib files which have about 1000 rows in the SpectrumSourceFiles table, the .slc file can be read in 28 seconds, and this PR allows the .blib file to be loaded directly in 29 seconds.

This pull request also uses a library called "Dapper" to do the mapping between column names and object properties.

Another thing that I did in the pull request was make it so that BiblioSpecLite.GetDataFileDetails does most of its work just by adding up numbers that are already in the _libraryEntries and _librarySourceFiles collection. This required adding a fileId field to "BiblioLiteSpectrumInfo".

I have not tested this performance with other sorts of .blib files, which maybe have fewer rows in the RetentionTimes table. With these TEI-REX .blib files, nearly all of the library loading time is spent reading the RetentionTimes table, but other sorts of libraries might have different performance characteristics for which the optimizations in this pull request might not do as well.


chambm:
Is this replacing the NHibernate ORM with Dapper, or adding another ORM in addition?

nickshulman:
> Is this replacing the NHibernate ORM with Dapper, or adding another ORM in addition?

Adding another ORM.
Come to think of it, it might be a good idea to replace our use of NHibernate with something simpler like Dapper. We have never been using NHibernate the way it was intended with lazy loaded collections and modified Entities existing in the Session.

Issue #3479, opened 2025-05-15
In Waters UV spectra, the x-axis is wrongly labelled as m/z


pete-reay-waters:
The x-axis should be a `MS_wavelength_array` with units of `UO_nanometer` whereas it's currently reporting `MS_m_z_array` with units of `MS_m_z`

(I will raise a Pull Request for the fix for this which I have already implemented)

PR #3480, opened 2025-05-15
Write correct units into mz array when writing Waters spectra

pete-reay-waters:
The spectrum list was already determining that the unit should be 'nanometre' in some cases, but this wasn't reflected in the array in the mzML

#3479


chambm:
Looks good, the failure doesn't have anything to do with this PR.

PR #3488, opened 2025-05-23
* fixed verification of result from MSFragger verification code button

chambm:


pwiz_tools/Skyline/Alerts/MsFraggerDownloadDlg.cs
```diff
@@ -241,9 +239,15 @@ private void btnAccept_Click(object sender, EventArgs e)
 
         private void btnRequestVerificationCode_Click(object sender, EventArgs e)
         {
-            RequestVerificationCode();
+            var verificationResultHtml = RequestVerificationCode();
             lblVerificationCode.Enabled = tbVerificationCode.Enabled = true;
-            MessageDlg.Show(this, AlertsResources.MsFraggerDownloadDlg_btnRequestVerificationCode_Click_Check_your_email);
+            var checkEmailDlg =
+                new AlertDlg(AlertsResources.MsFraggerDownloadDlg_btnRequestVerificationCode_Click_Check_your_email, MessageBoxButtons.OK)
```
nickshulman, line 245: You can use the "ShowAndDispose" method here:
`new AlertDlg(AlertsResources.MsFraggerDownloadDlg_btnRequestVerificationCode_Click_Check_your_email, MessageBoxButtons.OK){DetailMessage = verificationResultHtml}.ShowAndDispose(this);`

It is not necessary to set "StartPosition" on AlertDlg. It gets the "CenterParent" value from "CommonAlertDlg.resx".

pwiz_tools/Skyline/Alerts/AlertsResources.resx
```diff
@@ -325,6 +325,8 @@
     <value>Load Login page failed with HTTP status code {0}. It is likely that the Client Registration Code is invalid. A new Client Registration Code was just received from the server so this is probably a bug. Login Page URL: {1}</value>
   </data>
   <data name="MsFraggerDownloadDlg_btnRequestVerificationCode_Click_Check_your_email" xml:space="preserve">
-      <value>Check your email for the verification code and copy it into the box.</value>
+    <value>Check your email for the verification code and copy it into the box.
+
+If you don't get an email, double-check your email address or click "More Info" for troubleshooting.</value>
```
nickshulman, line 330: Personally, I would not bother to mention the "More Info" button, or even the part about checking their email address.

nickshulman (approved):

pwiz_tools/Skyline/Alerts/MsFraggerDownloadDlg.cs
chambm: That's what I thought too but when I was testing I noticed the dialog was showing up outside the parent dialog's bounds. What could be going on there?

pwiz_tools/Skyline/Alerts/AlertsResources.resx
chambm: I didn't want us to get reports with HTML result strings if the user simply got the email address wrong. :)

pwiz_tools/Skyline/Alerts/MsFraggerDownloadDlg.cs
chambm: Maybe it's due to using Show() instead of ShowAndDispose()? It seems to work without the StartPosition when I use ShowAndDispose.

nickshulman: > Maybe it's due to using Show()

The instance method "Show" on a Form should never be used for dialogs that are supposed to be modal. That brings up a non-modal window, and has a void return type.
The instance method to show a modal dialog and get back a DialogResult is "ShowDialog" (which is what "ShowAndDispose" eventually ends up calling).
"CenterParent" only matters for modal dialogs.

chambm: Oof, thanks for the reminder. I knew that at one point. My memory must have been overwritten with the Show overloads in MessageDlg and other Skyline forms. :) Definitely explains the CenterParent thing (which I did NOT know about non-modal dialogs).

PR #3494, opened 2025-06-02
Update AbstractDeepLibraryBuilder.cs

dshteyn:
Fix two inefficiencies


pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -129,12 +129,12 @@ public IEnumerable<string> GetPrecursorTable(bool training)
             }
 
             // Build precursor table row by row
-            foreach (var peptide in Document.Peptides)
+            foreach (var peptide in Document.Peptides.Distinct())
```
nickshulman: The "Equals" method on PeptideDocNode is not really the semantics you would want for "Distinct".
You more just want the distinct set of modified peptide sequences, which is a little different.

nickshulman: Actually, you should probably ignore this comment for now.

AlphapeptDeep really does only care about the modified peptide sequence and charge, so I might do something like:
```
foreach (var peptidesWithSameSequence in Document.Peptides.GroupBy(peptide=>peptide.SequenceKey))
{
    foreach (var charge in peptidesWithSameSequence
                 .SelectMany(peptide => peptide.TransitionGroups.Select(tg => tg.PrecursorCharge))
                 .Distinct())
    {
        // Do something with the modified sequence and charge
    }
```
But, then, Carafe cares about a lot more things on the PeptideDocNode such as the peak boundaries, so it's less clear whether the output should really have one row per peptide sequence.

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -162,19 +162,19 @@ protected abstract string GetTableRow(PeptideDocNode peptide,
...
 
-            var warningMods = GetWarningMods();
+            _warningMods ??= GetWarningMods();
 
             bool unsupportedModification = false;
             for (var i = 0; i < modifiedSequence.ExplicitMods.Count; i++)
             {
                 var mod = modifiedSequence.ExplicitMods[i];
-                var modWarns = warningMods.Where(m => m == mod.Name).ToArray();
+                var modWarns = _warningMods.Where(m => m == mod.Name).ToArray();
```
nickshulman: Does this line of code work?

When I look at the implementation of "GetWarningMods", it looks like every value in the list is going to start with four spaces because of the following line of code in "GetWarningMods":
`resultList[i] = resultList[i].Indent(1);`
The indenting logic should probably be moved into "GetWarning" so that _warningMods only contains modification names without any preceding whitespace.

It looks like you only care whether there are any warnings, and you don't actually need the modification names themselves, so this line of code should probably be:
`var isInWarningMods = _warningMods.Contains(mod.Name);`

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
dshteyn: I switched this back to previous code, where Distinct is used once on the list at the end of the loop.

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
dshteyn: Switched code to use Contains as suggested.

dshteyn:
Added unit test and expanded how it looks at the warning modifications generated and compares to expectation.

PR #3495, opened 2025-06-09
Skyline: Fixed NRE in SequenceTree drag-and-drop audit logging

brendanx67:
- Improve specificity of drop target in audit log when dropping on a peptide and not a protein
- NRE fixed when dropping on the "new" item in the tree


ekoneil (commented):
lgtm

PR #3500, opened 2025-06-11
Fix hint path for Newtonsoft dll

nickshulman:
The relative hint path went outside of the project directories and would not work if there was a sibling folder named "pwiz". Also, remove "AnyCPU" configuration from Skyline.sln.


chambm:
Whoops. I guess this didn't trigger on TC because the checkout directory is `pwiz`! :(

PR #3506, opened 2025-06-12
Agilent libraries update

rita-gwen:


rita-gwen:
This is PR with updated Agilent libraries. If the tests will pass I'll merge it into master right away.

brendanx67 (approved):

PR #3514, opened 2025-06-18
Skyline: Fixed regression graph outlier calculation

brendanx67:
- performance improvement to replace n*log(n) sort with linear QNthItem function used for percentiles in the Statistics class


brendanx67:
Note: the perf only changes to use QNthItem instead of sorting passed all tests (including perf) without needing any changes.

PR #3519, opened 2025-06-21
AbstractLibraryBuilder and AlphaPeptDeep Better PTM Support

dshteyn:
Modularizes modifications (PTMs) supported by models predicting MS2 fragment ion intensities, retention times and collisional cross sections, separately by model type within a given library builder.


pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -30,11 +30,104 @@
 
 namespace pwiz.Skyline.Model.Lib.AlphaPeptDeep
 {
+    public class LibraryBuilderModificationSupport
+    {
+        private IList<ModificationType> _ms2SupportedMods;
```
nickshulman: Instead of having a separate list for modifications which are supported for MS2 fragmentation prediction, retention time prediction and CCS prediction, it might be better to have a class with properties "IsFragmentationPredictionSupported", "IsRetentionTimePredictionSupported" and "IsCcsPredictionSupported", and then one map of modification to that structure.

brendanx67: Maybe just the 2 states we know of: 1) supports spectra, rt, and ccs, 2) supports spectra only.

We have no idea whether we will ever see a model that supports different combinations. So, I don't think it makes the most sense to design for complete flexibility. I would much rather see the code not generate predictions for RT and CCS when they are not available and just warn that only a spectrum will be generated for the modifications with spectra only supported.

dshteyn: Addressed both concerns in the latest commit.

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -30,11 +30,161 @@
 
 namespace pwiz.Skyline.Model.Lib.AlphaPeptDeep
 {
+
+    public class LibraryBuilderModificationSupport
+    {
+        public class PredictionSupport
+        {
+            public bool IsFragmentationPredictionSupported { get; set; }
```
nickshulman: These property names could be much shorter.
Since the class name is "PredictionSupport", the property names could simply be "Fragmentation", "RetentionTime" and "Ccs".

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -30,11 +30,161 @@
...
+            public bool IsRetentionTimePredictionSupported { get; set; }
+            public bool IsCcsPredictionSupported { get; set; }
+        };
+       
+        internal Dictionary<string, PredictionSupport> _predictionSupport; //key is ModificationType.Accession
+
+        public enum SupportType
```
nickshulman: This enum is not actually necessary. You could simply pass around a PredictionSupport object with one or more of its properties set to true.

dshteyn: Refactored.  Thanks Nick!

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
dshteyn: Done.

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -30,11 +30,158 @@
...
+        {
+            if (supportedModifications == null)
+                return;
+
+            foreach (KeyValuePair<ModificationIndex, PredictionSupport> mod in supportedModifications)
+            {
+                var key = mod.Key.Modification.Accession.Split(':')[0];
+
+                if (!_predictionSupport.ContainsKey(key))
+                    _predictionSupport.Add(key, new PredictionSupport());
```
nickshulman: This line of code doesn't really do anything, since the next line of code ends up replacing the same entry in the dictionary.
However, I would still recommend getting rid of the class "LibraryBuilderModificationSupport" since it does not add anything beyond being a dictionary from string to PredictionSupport

nickshulman (commented):
I am not sure I understand why the classes "ModificationIndex" and "ModificationType" both exist. Could those be combined into one class?

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
dshteyn: Thanks for the great suggestions, now implemented.

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -30,12 +30,156 @@
...
-        public static IList<ModificationType> PopulateUniModList(IList<ModificationIndex> supportedList)
+        /// <summary>
+        /// Populate a list of ModificationTypes from those available in UniModData.
+        /// </summary>
+        /// <param name="supportedList">List of modifications.
+        /// - Empty list means *none* will be returned.
+        /// - Null list means *all* in UniModData will be returned.
+        /// </param>
+        /// <returns></returns>
+        public static IList<ModificationType> PopulateUniModList(IList<ModificationType> supportedList)
```
nickshulman: I'm not sure I understand what this function does.
It takes in a list of ModificationType objects, and returns a different list of ModificationType objects.
What is the difference between the list that was passed in and the list that was returned?
Why does this transformation need to happen? What was wrong with the original list that was passed in that needed to be changed?

pwiz_tools/Skyline/Test/AbstractLibraryBuilderModificationSupportTest.cs
```diff
@@ -0,0 +1,194 @@
...
+        };
+        IList<ModificationType> CcsSupportedList => new List<ModificationType>
+        {
+            new ModificationType(1, "001:MOD1", "TestMod1", "TestOnly"),
+            new ModificationType(4, "004:MOD4", "TestMod4", "TestOnly")
+        };
+       
+        [TestMethod]
+        public void AbstractLibraryBuilderModificationSupportTests_Constructor_PopulatesPredictionSupport_Correctly()
+        {
```
chambm: We generally don't use underscores in test names by convention. Certainly the `AbstractLibraryBuilderModificationSupportTests` prefix is unnecessary and don't include the expected result in the test name as a suffix (like ReturnsCorrectSupport or ReturnsFalse). I would personally prefer to see these tests consolidated like:
SupportedModTest() // tests *SupportedMod methods
PredictionSupportTest() // tests constructor and empty list, null list validation

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
dshteyn: The modifications should now be formatted correctly when first initialized from UniModData

pwiz_tools/Skyline/Test/AbstractLibraryBuilderModificationSupportTest.cs
dshteyn: Thanks Matt, I will follow the conventions.

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AlphapeptdeepLibraryBuilder.cs
```diff
@@ -153,14 +143,43 @@ public static PythonInstaller CreatePythonInstaller(TextWriter writer)
...
-        /// List of UniMod Modifications available
-        /// </summary>
-        public static readonly IList<ModificationType> MODIFICATION_NAMES = PopulateUniModList(null);
-
         protected override string ToolName => ALPHAPEPTDEEP;
-        
-        protected override IList<ModificationType> ModificationTypes => MODIFICATION_NAMES;
+        protected override LibraryBuilderModificationSupport libraryBuilderModificationSupport { get; }
+
+        internal static Dictionary<ModificationType, PredictionSupport> MODEL_SUPPORTED_UNIMODS =
```
nickshulman: ModificationType does not override "Equals" and "GetHashCode", so it's unclear whether it makes sense to use it as the key in a Dictionary.
Maybe "PredictionSupport" should be part of the ModificationType class, so "ModificationType" stores information about the modification including what sort of prediction models it supports.
Maybe what you want here is List`<Tuple<ModificationType, PredictionSupport>>`

What do you see as the job of this "MODEL_SUPPORTED_UNIMODS" field?

dshteyn: Switched to using List<ModificationType> for MODEL_SUPPORTED_UNIMODS and moved PredictionSupport inside ModificationType.  Also implemented Equals and GetHashCode for ModificationType and PredictionSupport.  The purpose of MODEL_SUPPORTED_UNIMODS is to track supported modifications by the library builder and also to track which models of the library builder support the modifications in the list.

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -30,42 +30,241 @@
...
+                return false;
+            }
+
+            return type.SupportedModels.Ccs;
+        }
 
+
+        public bool PeptideHasOnlyMs2SupportedMod(string modifiedPeptide)
+        {
+            string[] peptideParts = modifiedPeptide.Split(new[]{'[',']',':'});
```
nickshulman: You should write a function:
`IEnumerable<int> GetModificationIds(string modifiedPeptide)`
Also, this class should have a method:
`ModificationType GetModificationType(int id)`

Then, this function would be implemented as:
`GetModificationIds(modifiedPeptide).All(id=>true == GetModificationType(id)?.SupportedModels.Fragmentation)`

The job of the class "LibraryBuilderModificationSupport" is really just to quickly lookup a ModificationType object by id.

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -30,42 +30,241 @@
 
 namespace pwiz.Skyline.Model.Lib.AlphaPeptDeep
 {
-    /// <summary>
-    /// Abstract base class for AlphaPeptDeep and Carafe library builders so they can share code
-    /// </summary>
-    public abstract class AbstractDeepLibraryBuilder : ILibraryBuildWarning
+    public class PredictionSupport : Immutable, IEquatable<PredictionSupport>
```
nickshulman: We only usually implement "IEquatable" for struct's.
The IEquatable interface enables two struct's to be compared without wrapper objects needing to be constructed, so it really is much faster.
But, for classes, since there is no boxing involved, we don't normally bother to implement IEquatable.
I Equatable is problematic for classes that have sub-classes, because it then becomes possible to accidentally call "Equals(PredictionSupport)" when you meant to call "Equals(object)" on an object which is the subclass, and you end up getting a comparison result which has ignored fields in the subclass.

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -229,21 +491,39 @@ protected virtual string GetModName(ModificationType mod, string unmodifiedSeque
...
+            }
+            warningModString = string.Join(Environment.NewLine, noCcsSupportWarningMods.Select(w => w.Indent(1)));
+            return string.Format(ModelResources.Alphapeptdeep_Warn_limited_modification,
                 warningModString);
+
+
         }
         
-        public IList<string> GetWarningMods()
+        public (IList<string>,IList<string>,IList<string>) GetWarningMods()
```
nickshulman: I think it would be better if this method returned `Dictionary<string, PredictionSupport>`.

When the caller wants to know which modifications were not supported for ms2, they can find:
`dictionary.Where(kvp=>!kvp.Value.Fragmentation).Select(kvp=>kvp.Key)`

The problem with returning three separate lists like this is that it's too easy to make a mistake about which list is which.

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -30,42 +32,221 @@
...
+    };
+    public class ModificationType
+    {
+        public ModificationType(int id, string name, string comment, PredictionSupport supportedModels = null)
+        {
+            Id = id;
+            Accession = id + @":" + name;
+            Name = name;
+            Comment = comment;
+            SupportedModels = supportedModels;
```
nickshulman: Usually you would do something like:
`SupportedModels = supportedModels ?? PredictionSupport.NONE;`

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -255,28 +526,22 @@ public IList<string> GetWarningMods()
...
-                    var modNames = ModificationTypes.Where(m => m.Accession == unimodIdWithName).ToArray();
-                    if (modNames.Length == 0)
+                    else
                     {
-                        var haveMod = resultList.FirstOrDefault(m => m == mod.Name);
-                        if (haveMod == null)
-                        {
-                            resultList.Add(mod.Name);
-                        }
+                        if (!(libraryBuilderModificationSupport.IsMs2SupportedMod(mod.UnimodId) &&
```
nickshulman: I think you should make public the `GetModificationType` method on `LibraryBuilderModificationSupport`.

Then you would be able to do:
```
var support = libraryBuilderModificationSupport.GetModificationType(mod.UnimodId.Value)?.SupportedModels ?? PredictionSupport.NONE;
if (!Equals(support, PredictionSupport.ALL))
{
    _warningModSupports[mod.Name] = support;
}
```

I don't think the LibraryBuildModificationSupport class needs all the variations of "IsMs2SupportedMod", "IsRtSupportedMod" etc.
If you make it really easy to get a PredictionSupport from a unimod ID, the caller will be able to do things more efficiently.

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -30,42 +32,221 @@
...
+                hashCode = (hashCode * 397) ^ (Name != null ? Name.GetHashCode() : 0);
+                hashCode = (hashCode * 397) ^ (Comment != null ? Comment.GetHashCode() : 0);
+                hashCode = (hashCode * 397) ^ (SupportedModels != null ? SupportedModels.GetHashCode() : 0);
+                return hashCode;
+            }
+        }
+    }
+    public class LibraryBuilderModificationSupport
+    {
+        internal List<ModificationType> _predictionSupport; //key is ModificationType.Index
```
nickshulman: This should probably be `Dictionary<int, ModificationType>` so that it's faster to find things by ID.

dshteyn: Check!

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -30,42 +32,239 @@
...
+                var hashCode = Fragmentation.GetHashCode();
+                hashCode = (hashCode * 397) ^ RetentionTime.GetHashCode();
+                hashCode = (hashCode * 397) ^ Ccs.GetHashCode();
+                return hashCode;
+            }
+        }
+    };
+    public class ModificationType
+    {
+        public ModificationType(int id, string name, string comment, PredictionSupport supportedModels = null)
```
nickshulman: I would recommend getting rid of the `= null` here and make `supportedModels` a required argument.

I would also recommend adding an overloaded constructor:
`public ModificationType(UniModModificationData unimodModificationData, PredictionSupport supportedModels)`
That overloaded constructor would make initializing the list `MODEL_SUPPORTED_UNIMODS` easier in `AlphapeptdeepLibraryBuilder`

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
```diff
@@ -161,52 +373,103 @@ protected abstract string GetTableRow(PeptideDocNode peptide,
...
+        protected internal PredictionSupport ValidateSequenceModifications(ModifiedSequence modifiedSequence, out string mods, out string modSites)
         {
             var modsBuilder = new StringBuilder();
             var modSitesBuilder = new StringBuilder();
 
             // The list returned below is probably always short enough that determining
             // if it contains a modification would not be greatly improved by caching a set
             // for use here instead of the list.
-            var warningMods = GetWarningMods();
+            var warningModSupports = GetWarningMods();
```
nickshulman: I don't think the `Dictionary<string, PredictionSupport>` that you get back from "GetWarningMods" really helps you in this function much.
You have the UniModId on each of the Modification objects in `modifiedSequence` so you can call `libraryBuilderModificationSupport.GetModificationType` yourself.

I would have written this something like the following, which takes care of writing out the necessary warning message:
```
var modificationTypes = new List<Tuple<ModifiedSequence.Modification, ModificationType>>();
foreach (var explicitMod in modifiedSequence.ExplicitMods)
{
    ModificationType modificationType = null;
    if (explicitMod.UnimodId.HasValue)
    {
        modificationType =
            libraryBuilderModificationSupport.GetModificationType(explicitMod.UnimodId.Value);
        modificationTypes.Add(Tuple.Create(explicitMod, modificationType));
    }
}
bool ms2SupportedMod = true;
bool rtSupportedMod = true;
bool ccsSupportedMod = true;

foreach (var group in modificationTypes.GroupBy(tuple => tuple.Item1.Name))
{
    if (group.Any(tuple => true != tuple.Item2?.SupportedModels?.Fragmentation))
    {
        var msg = string.Format(ModelsResources.BuildPrecursorTable_UnsupportedModification, modifiedSequence, group.Key, ToolName);
        Messages.WriteAsyncUserMessage(msg);
        ms2SupportedMod = false;
        rtSupportedMod = false;
        ccsSupportedMod = false;
        continue;
    }

    bool msgGenerated = false;
    if (group.Any(tuple => true != tuple.Item2?.SupportedModels?.RetentionTime))
    {
        var msg = string.Format(ModelsResources.BuildPrecursorTable_Unimod_limited_Modification, modifiedSequence, group.Key, group.First().Item1.UnimodIdWithName, ToolName);
        Messages.WriteAsyncUserMessage(msg);
        rtSupportedMod = false;
        msgGenerated = true;
    }

    if (group.Any(tuple => true != tuple.Item2?.SupportedModels?.Ccs))
    {
        if (!msgGenerated)
        {
            var msg = string.Format(ModelsResources.BuildPrecursorTable_Unimod_limited_Modification, modifiedSequence, group.Key, group.First().Item1.UnimodIdWithName, ToolName);
            Messages.WriteAsyncUserMessage(msg);
        }
        ccsSupportedMod = false;
    }
}
```

I didn't quite get to the code to set "mods" and "modSites", but you get the idea.

You don't have to write the code this way, just a suggestion.

nickshulman (commented):
Here are a couple of ideas.

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
dshteyn: Changes applied.

pwiz_tools/Skyline/Model/Lib/AlphaPeptDeep/AbstractDeepLibraryBuilder.cs
dshteyn: Changes applied.

brendanx67 (approved):
After some clean-up, this is now looking okay to me.

PR #3527, opened 2025-06-27
Skyline/work/20250627 python installer virtual environments

dshteyn:
Fix for hanging execution when installing python virtual environments.

### DIFF
--- a/pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
+++ b/pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
@@ -163,7 +163,7 @@ namespace pwiz.Common.SystemUtil
             }
         }
 
-        private const string PREFIX_LONG_PATH = @"\\?\";
+        public const string PREFIX_LONG_PATH = @"\\?\";
 
         /// <summary>
         /// Returns a path with a long-path prefix. If the passed in path already has
@@ -172,7 +172,7 @@ namespace pwiz.Common.SystemUtil
         /// <param name="path">Path to convert to long-path syntax. This must be a fully qualified path.</param>
         public static string ToLongPath(this string path)
         {
-            if (path.StartsWith(PREFIX_LONG_PATH))
+            if (path != null && path.StartsWith(PREFIX_LONG_PATH))
                 return path;
 
             // Note: Using this function requires an already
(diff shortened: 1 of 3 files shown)


pwiz_tools/Skyline/Model/Tools/PythonInstaller.cs
```diff
@@ -1076,13 +1077,13 @@ public RunGetPipScriptTask(PythonInstaller installer)
         public override void DoAction(ILongWaitBroker broker)
         {
             // CONSIDER(brendanx): Path quoting?
-            var cmdLine = TextUtil.SpaceSeparate(PythonInstaller.BasePythonExecutablePath,
-                PythonInstaller.GetPipScriptDownloadPath);
+            var cmdLine = TextUtil.SpaceSeparate(TextUtil.Quote(PythonInstaller.BasePythonExecutablePath),
```
brendanx67: Note that the function prototype:

public static string Quote(this string text)

The "this" means you can just use:

PythonInstaller.BasePythonExecutablePath.Quote()

pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
```diff
@@ -172,11 +172,9 @@ public static String GetDirectoryName(String path)
         /// <param name="path">Path to convert to long-path syntax. This must be a fully qualified path.</param>
         public static string ToLongPath(this string path)
         {
-            // Note that Path.GetFullPath() will be equivalent to
-            // Path.Combine(Directory.GetCurrentDirectory(), path) for a relative path,
-            // which is rarely appropriate. So, using this function requires an already
+            // Note: Using this function requires an already
             // fully qualified path.
-            if (!Equals(path, Path.GetFullPath(path)))
+            if (!Path.IsPathRooted(path))
```
brendanx67: My read of the documentation is that this does not guarantee a fully qualified path. A rooted path can be relative, e.g. C:Document - rooted by drive but relative on directory, or \Documents - rooted on path, but not drive. I think you likely need to write a new PathEx.IsFullyQualified() function for this which starts simple and eventually calls Path.GetFullPath() in an exception handler, because it can throw if path contains invalid characters. All pointed out by ChatGPT. You can start by asking it to write this function for you.

pwiz_tools/Skyline/Model/Tools/PythonInstaller.cs
```diff
@@ -144,7 +144,7 @@ internal List<PythonTaskBase> GetPythonTasks(bool? haveNvidiaGpu = false)
 
             return tasks;
         }
-
```
brendanx67: Please revert unnecesary whitespace and formatting-only changes. These make it more challenging to find the important part of your commit without turning off whitespace diffs.

brendanx67 (changes requested):
The added quoting seems good. I assume there was an issue with Path.GetFullPath() throwing? I think the replacement for this code needs to be more thorough than just checking IsPathRooted().

pwiz_tools/Skyline/Model/Tools/PythonInstaller.cs
dshteyn: Apologies for the whitespace and formatting changes now reverted.

pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
dshteyn: Added code and associated tests for PathEx.IsFullyQualifiedPath().

pwiz_tools/Skyline/Model/Tools/PythonInstaller.cs
dshteyn: Thanks for the suggestion.  Code updated accordingly.

dshteyn:
> The added quoting seems good. I assume there was an issue with Path.GetFullPath() throwing? I think the replacement for this code needs to be more thorough than just checking IsPathRooted().

Switched to using PathEx.IsPathFullyQualified and added unit tests.

pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
chambm: Wow, today I learned... I've thought Windows rooted paths were always absolute for decades. This differs from POSIX where a path is rooted if it starts with a slash, i.e. at the root, and must therefore be absolute.

pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
```diff
@@ -172,14 +172,50 @@ public static String GetDirectoryName(String path)
...
+
             // Avoid adding the long-path prefix to a path that already has it.
-            return path.StartsWith(PREFIX_LONG_PATH) ? path : PREFIX_LONG_PATH + path;
+            return PREFIX_LONG_PATH + path;
+        }
+
+        /// <summary>
+        /// Returns a true if the path is fully qualified
+        /// </summary>
+        /// <param name="path">Path to convert to long-path syntax. This must be a fully qualified path.</param>
```
nickshulman: This part of the comment isn't correct.
The `<param>` tags are not required in these comments so you can just remove it.

pwiz_tools/Skyline/Test/LongPathDirectoryTest.cs
```diff
@@ -30,6 +30,142 @@ namespace pwiz.SkylineTest
     [TestClass]
     public class LongPathDirectoryTest : AbstractUnitTest
     {
+        private const string PREFIX_LONG_PATH = @"\\?\"; // Assuming this is the long-path prefix
+
+        [TestMethod]
+        public void IsPathFullyQualified_ValidDriveLetterPath_ReturnsTrue()
```
nickshulman: I would recommend combining all of the tests in this class into a single function with multiple Assert statements in it.

pwiz_tools/Skyline/Test/LongPathDirectoryTest.cs
```diff
@@ -30,6 +30,142 @@ namespace pwiz.SkylineTest
...
+        {
+            string path = @"C:\Folder\Subfolder";
+            bool result = PathEx.IsPathFullyQualified(path);
+
+            Assert.IsTrue(result, "Expected C:\\Folder\\Subfolder to be fully qualified.");
```
nickshulman: I would recommend:
`Assert.IsTrue(PathEx.IsPathFullyQualified(path), "{0} should be fully qualified", path);`

nickshulman (commented):
Could we write a class called something like "ArgumentList" which is just a list of raw strings.
When ArgumentList is asked for the full command line, it looks at each argument and decides whether the argument needs quoting based on its value (does the value contain spaces quotes or other special characters).
That seems a lot simpler than having to decide whether to call "quote" before adding each argument to a list.

PR #3530, opened 2025-06-30
Simplify HardklorReader explicit RT handling by moving logic into the  parent sslReader class

bspratt:
Solves an issue with reading SSL files with explicit retention times and scan IDs duplicated across multiple files, as well as simplifying unnecessarily complicated (and faulty) logic


bspratt:
It seems so. Though this code appears to prioritize RT values pulled from the scans when available, where I have the new code prioritizing the columnar value.

nickshulman (approved):

PR #3531, opened 2025-07-01
Skyline: Add a handler recommended by Claude.ai

brendanx67:
- For unhandled exceptions in .NET Task objects, which use the .NET thread pool
- Any code using async and await use these, and they get used inside .NET and like other DLLs we use written to .NET CLI


pwiz_tools/Skyline/Program.cs
```diff
@@ -585,16 +589,26 @@ public static void Init()
         private static readonly object _unhandledExceptionLock = new object();
         public static ToolService MainToolService;
 
+        private static void TaskScheduler_UnobservedTaskException(object sender, UnobservedTaskExceptionEventArgs e)
+        {
+            ReportUnhandledException(e.Exception);
```
nickshulman, line 594: If you were to also call "e.SetObserved()" would that rescue the Skyline process from abruptly exiting?

It seems that we would want to do that if it works, and also call "ReportException" instead of "ReportUnhandledException".

nickshulman:
What sort of code could get this event to fire?
If it's possible to prevent Skyline from abruptly exiting, then we should add some code to test to "ReportErrorDlgTest".

If Skyline has to exit if this event ever fires, then we might want to add a hidden menu item to the "Help" menu so that we can verify it works.
There already is the hidden menu item "Crash Skyline" which only shows up if you are holding down Ctrl+Shift when you open the Help menu.

ekoneil:
I may have seen an example of an abrupt exit last week while messing around with using WebClient.UploadFileAsync to upload large documents to Ardia.

My understanding is UploadFileAsync puts work on the .NET thread pool. That work failed, threw an exception, and Skyline insta-crashed. At least, I think that's what happened.

brendanx67:
> What sort of code could get this event to fire? If it's possible to prevent Skyline from abruptly exiting, then we should add some code to test to "ReportErrorDlgTest".
> 
> If Skyline has to exit if this event ever fires, then we might want to add a hidden menu item to the "Help" menu so that we can verify it works. There already is the hidden menu item "Crash Skyline" which only shows up if you are holding down Ctrl+Shift when you open the Help menu.

I have been assuming anything that throws an exception on an unitialized .NET thread pool thread. Our own Assync thread creation mechanism adds exception handling and locale support. We really should never use a thread that doesn't use our thread initialization.

ChatGPT suggests this as a way to test it:

Task.Run(() => throw new InvalidOperationException("Test exception from task"));

nickshulman:
> ChatGPT suggests this as a way to test it:
> 
> Task.Run(() => throw new InvalidOperationException("Test exception from task"));

I tried pretty hard to get anything bad to happen with statements like that, and I couldn't get anything remotely resembling a crash.
For instance, you would think that if you added that line:
`Task.Run(() => throw new InvalidOperationException("Test exception from task"));`
to the beginning of Program.Main, Skyline might have a bit of trouble starting.
But, Skyline still starts up just fine.

brendanx67:
Okay.

PR #3533, opened 2025-07-07
Add "Detailed Info" audit log report

nickshulman:
Added "Detailed Info" audit log report which includes user name and Skyline version.

image


brendanx67 (approved):
Looks good. Wish I had thought of it earlier.

PR #3535, opened 2025-07-07
Add "EI" (electron ionization) option to Settings > Full Scan > MS/MS filtering > Acquisition method…

bspratt:
… for improved GC-MS support. With this setting, MS1 data is treated as all-ions fragmented MS2 data.


pwiz_tools/Skyline/Model/DocSettings/DocSettingsResources.resx
```diff
@@ -178,6 +178,9 @@
   <data name="FullScanAcquisitionMethod_DDA_TOOLTIP" xml:space="preserve">
     <value>Data dependent acquisition: MS2 chromatograms are extracted but not used for quantification</value>
   </data>
+  <data name="FullScanAcquisitionMethod_EI_TOOLTIP" xml:space="preserve">
+    <value>Data independent acquisition: MS1 EI fragmented data is treated as MS2</value>
```
nickshulman: This string gets used in a tooltip that appears on the Full Scan tab in Transition Settings:
image

This text here should start with "Electron Ionization" so that the user knows what "EI" stands for.
I'm not sure what the rest of the text should say, but it should help the user figure out which option they want to choose.

pwiz_tools/Skyline/Model/Results/SpectrumFilter.cs
```diff
@@ -173,15 +173,15 @@ public SpectrumFilter(SrmDocument document, MsDataFileUri msDataFileUri, IFilter
...
                     {
                         if (instrumentInfo != null)
                         {
                             _isWatersMse = _isWatersFile;
                             _isAgilentMse = instrumentInfo.IsAgilentFile;
-                            _isElectronIonizationMse = instrumentInfo.ConfigInfoList != null &&
-                                   instrumentInfo.ConfigInfoList.Any(c => @"electron ionization".Equals(c.Ionization));
+                            _isElectronIonizationMse = (instrumentInfo.ConfigInfoList != null &&
+                                                        instrumentInfo.ConfigInfoList.Any(c => @"electron ionization".Equals(c.Ionization))) ||
+                                                       _acquisitionMethod == FullScanAcquisitionMethod.EI;
```
nickshulman: I wonder whether this code should be moved outside of the `if (instrumentInfo != null)` block, since it does not actually require an instrumentInfo to check the acquisition method.
Maybe this statement should be:
```
_isElectronIonizationMse = 
    true == instrumentInfo?.ConfigInfoList?.Any(c => @"electron ionization".Equals(c.Ionization))
    || _acquisitionMethod == FullScanAcquisitionMethod.EI;
```

pwiz_tools/Skyline/TestData/CommandLineEiTest.cs
```diff
@@ -0,0 +1,78 @@
...
+                "--full-scan-acquisition-method=EI",  // Treat MS1 data as all-ions-fragmentation Ms2 results
+                "--full-scan-product-analyzer=tof",
+                "--full-scan-product-res=30000",
+                "--full-scan-rt-filter=scheduling_windows",
+                "--full-scan-rt-filter-tolerance=5",
+                "--tran-product-ion-charges=1",
+                "--instrument-min-mz=42",
+                "--instrument-max-mz=2000",
+                "--instrument-min-time=" + 0.42,
+                "--instrument-max-time=" + 4.2,
```
nickshulman: The spectra in the file "2325013014_original.mzXML" go from 4.6 minutes to 34 minutes.
`TransitionInstrument.MaxTime` is an integer, and it gets set to the ceiling of the value specified here, so the only spectra being used for this test are from 4.6 to 5.0 minutes.
2325013014_original.mzXML is 12MB compressed.
If this test is only using a small fraction of the spectra in that file, should we make the mzXML file smaller?

nickshulman (commented):
You are going to need to add "EI" to "full_scan_acquisition_method_type" in Skyline_Current.xsd.
You will also need to increase the current version number in "DocumentFormat".

An older version of Skyline would choke if it sees "EI" as the acquisition method.
What we usually do for something like this is add some code to "SrmSettings.RemoveUnsupportedFeatures" so that if the acquisition method is "EI" and the document is being saved to an older version, the acquisition method gets changed to whatever we think would be the closest match. We did not do that when we introduced the "PRM" and "SureQuant" acquisition methods, but we probably should have.

I would recommend adding the line:
`AssertEx.Serializable(doc);`
to the end of ConsoleImportEiTest because that would alert you that Skyline_Current.xsd needs to be updated.

pwiz_tools/Skyline/TestData/CommandLineEiTest.cs
bspratt: Just copy-paste cruft, I'll remove it

pwiz_tools/Skyline/Model/DocSettings/DocSettingsResources.resx
bspratt: Very good, thanks. I always appreciate your detailed reviews.

pwiz_tools/Skyline/Model/Results/SpectrumFilter.cs
bspratt: Cleaner, yes - thanks

nickshulman:
What's the difference between "MSe" and "EI"?
Is the only difference that EI means to always extract chromatograms from MS1, whereas MSe tries to do some smart guessing about whether it's going to be MS1 or MS2?

I wonder whether we should call this new Acquisition Method "MSe" (or is that a trademark, and we need to call it "All Ions") instead of "EI". Then, if they choose that, the "Isolation Scheme" dropdown disappears and is replaced with something where they can choose the MS Level, which might be "Auto", "1", or "2".

The way that we have it now where the way to get MSe is to first choose "DIA" and then "All Ions" as the isolation scheme which is kind of hokey, especially because "All Ions" is not actually an isolation scheme that you can edit.

bspratt:
Interesting - what names would you propose there?

brendanx67:
Precursor spectra:
Fragment spectra:

Maybe 🤔

nickshulman:
I think I understand this a little better than I did before. I have a couple of questions/clarifications:

1. Currently, if you choose "All Ions" as the isolation scheme, Skyline will behave like this new "EI" acquisition method if any of the configs in the "ConfigInfoList" say "electron ionization". We are adding this new acquisition method so that the user can explicitly tell Skyline that chromatograms for fragment transitions should always be extracted from MS1 spectra.

2. The "All Ions" isolation scheme tells Skyline to use a complicated algorithm to decide which MS level fragment transition chromatograms should be extracted from. This algorithm involves looking at the vendor type as well as the instrument config, and there might be some cases where it does not work right. If you know that these chromatograms should be extracted from MS1 spectra, then you should choose "EI" as the acquisition method. If you know that these chromatograms should be extracted from MS2 spectra you should define an isolation scheme with one big isolation window that covers the whole scan range.

3. What should we do if they've chosen something other than "None" in the "Isotope Peaks Included" dropdown, and the Acquisition Method is "EI"? Should we allow this scenario as a way of telling Skyline to extract chromatograms from the full precursor isotope envelope? Or should we change the Isotope Peaks Included dropdown to red and give it a tooltip saying not to do this and that chromatograms for intact precursors will be extracted from the same spectra as the fragments. (While we are at it, we should probably make it so the Isotope Peaks Included dropdown is red if it's "None" and the acquisition method is DDA because that's another scenario that does not work).

4. Is "EI" the best name for the acquisition method? The text that appears in the user interface can be different from the value that gets saved in the .sky file.

bspratt:
Thanks again for your thoughtful consideration. To your points:
1) Yes, this is correct. Until now people have performed various workarounds to handle EI data, this makes that unnecessary, and properly supports the case where the instrument config fails to declare electron ionization.
2) The essential difference between EI and All Ions is that there is only one instrument state in play. No MS  with MS/MS, no varying CE. Just a single stage in which everything gets blasted.
3) The entire Full Scan MS1 section is ignored, probably that should just be greyed out when "EI" is selected. We currently just don't worry about unfragmented precursors.
4) "EI" is what users seem to expect. No more obscure than "DIA" or "DDA", I'd think.

pwiz_tools/Skyline/Model/DocSettings/DocSettingsResources.resx
```diff
@@ -178,6 +178,9 @@
   <data name="FullScanAcquisitionMethod_DDA_TOOLTIP" xml:space="preserve">
     <value>Data dependent acquisition: MS2 chromatograms are extracted but not used for quantification</value>
   </data>
+  <data name="FullScanAcquisitionMethod_EI_TOOLTIP" xml:space="preserve">
+    <value>Electron Ionization: MS1 data is assumed to be EI fragmented and is treated as MS2</value>
```
nickshulman, line 182: Maybe this should say:
Electron Ionization: Fragment transition chromatograms are extracted from MS1 spectra

It's might not be clear exactly what "treated as MS2" would mean.

bspratt: Perhaps "Electron Ionization: Fragment transition chromatograms are extracted from MS1 spectra, which are assumed to be EI fragmented"

chambm:
Hey @bspratt , I have some Agilent GCMS files that I think are natively collected in MassHunter and those DO have the proper metadata so their ionization type is read as "electron ionization":
image

So you can auto-detect that if you want and set your EI option automagically as you mentioned in the support thread. It just won't work on old ChemStation data apparently, so it has to be manually selectable as well.

bspratt:
Thanks! That file should already work just fine - the code that looks for that cue has been in place for quite a while.

PR #3542, opened 2025-07-14
Added a test to check if available disk space is less than 5% of total

aaronbanse:
Necessary to prevent unrelated tests failing inadvertently when an issue causes the disk to fill up.

### DIFF
--- a/pwiz_tools/Skyline/Test/DiskSpaceTest.cs
+++ b/pwiz_tools/Skyline/Test/DiskSpaceTest.cs
@@ -1,0 +1,40 @@
+﻿/*
+ * Original author: Aaron Banse <abanse .at. uw dot edu>,
+ *                  MacCoss Lab, Department of Genome Sciences, UW
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+using System.IO;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using pwiz.Common.SystemUtil;
+using pwiz.SkylineTestUtil;
+
+namespace pwiz.SkylineTest
+{
+    [TestClass]
+    // Test to ensure there is enough space on the disk before starting.
+    public class DiskSpaceTest : AbstractUnitTest
+    {
+        private const double MIN_SPACE_PERCENT = .05;
+
+        [TestMethod]
+        public void TestIsDiskFull()
+        {
+            DriveInfo drive = new DriveInfo(Path.GetPathRoot(TestContext.TestDir));
+            Assert.IsTrue((double)drive.TotalFreeSpace / drive.TotalSize > MIN_SPACE_PERCENT, 
+                $"The disk has less than {MIN_SPACE_PERCENT * 100}% of space remaining.");
+        }
+    }
+}
--- a/pwiz_tools/Skyline/Test/DiskSpaceTest.cs
+++ b/pwiz_tools/Skyline/Test/DiskSpaceTest.cs
@@ -1,6 +1,8 @@
 ﻿/*
  * Original author: Aaron Banse <abanse .at. uw dot edu>,
  *                  MacCoss Lab, Department of Genome Sciences, UW
+ *         
+ * Copyright 2025 University of Washington - Seattle, WA
  * 
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


pwiz_tools/Skyline/Test/DiskSpaceTest.cs
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+using System.IO;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using pwiz.SkylineTestUtil;
+
+namespace pwiz.SkylineTest
+{
+    [TestClass]
+    // Test to see if there is less than 5% of space left on disk.
```
ekoneil: Nit: move the comment above the attribute. This keeps all the code together (attribute + method signature) and is more common than separating them.

pwiz_tools/Skyline/Test/DiskSpaceTest.cs
```diff
@@ -0,0 +1,41 @@
...
+    // Necessary to prevent unrelated tests failing when an issue causes disk to fill up.
+    public class DiskSpaceTest : AbstractUnitTest
+    {
+        private const double MIN_SPACE_PERCENT = .05;
+
+        [TestMethod]
+        public void TestIsDiskFull()
+        {
+            DriveInfo drive = new DriveInfo(Path.GetPathRoot(TestContext.TestDir));
```
ekoneil: Use var.

pwiz_tools/Skyline/Test/DiskSpaceTest.cs
```diff
@@ -0,0 +1,41 @@
...
+            Assert.IsTrue((double)drive.TotalFreeSpace / drive.TotalSize > MIN_SPACE_PERCENT, 
+                $"The disk has less than {MIN_SPACE_PERCENT * 100}% of space remaining.");
```
ekoneil: A more dire, prescriptive message would be good here. :) Example:

> Warning: this machine is running out of disk space. The D:\ drive has less than 5% of space remaining. Running out of space will cause other tests to fail.

Or something like that.

chambm:
I'm a little wary of testing free disk space by percentage. Seems like we could go with something absolute like 10GB, which could change as our needs change. Even trickier though is perftests with their persistent zipfiles and persistent extracted files have quite different space requirements and could even be stored on a different drive (see `PathEx.GetDownloadsPath`). I would think we'd be much more likely to run out of disk space while downloading/extracting perftests than non-perftests. I have 50GB of Perftests extracted and 60GB of Tutorials. Some of that might be unofficial stuff that I only have locally though. So without actually going through all the perftests and checking whether they have their files already downloaded, which seems overkill, it might be reasonable to do a full perftest run on a clean machine, check the space used in the persistent dir, and require that much space to be free on the `PathEx.GetDownloadsPath` drive. That would be a separate check from the current one, which is for the TestResults path which is basically always next to the source code.

chambm:
>Hi Matt,
> I saw your comments on the disk space test pull request, and I totally agree, now that I'm aware of that factor. it would probably be good to add the perftest check as an additional test, although I don't entirely understand what is involved in that process- I don't know enough about the codebase just yet. I can discuss it with Eddie and/or Brendan for advice. 
> 
> In the meantime, Brendan and Eddie agreed that the existing test is better than nothing, and would have caught the issue from last week. Eddie also suggested that it would be good to modify the name of the test to something like AaaDiskSpaceTest so that it shows up first.
> 

I'm fine with adding it without considering perftests but I still think percentage testing is bad. It would require very different space on a 500GB local drive vs. a 10TB RAID array. I use a 2TB SSD and I think 100GB is too high for a free space check if you're not talking about perf tests. I probably hover more around 30GB free. :)

Along with renaming to AaaDiskSpaceTest so it runs first (or right after AaaantivirusTest), which makes sense, another thing to consider is whether the test should outright kill the test process (like `System.Diagnostics.Process.GetCurrentProcess().Kill()`) instead of throwing a typical exception. Does it make sense for the tests to keep running if the disk space test fails? Or is the point to just clearly indicate that disk space was low when the tests were run?

aaronbanse:
> I probably hover more around 30GB free.
What if the tests run on a machine with very little storage space, where 30GB would be far too aggressive? What if the check considers whichever is smaller, 5% of the total drive or 30GB, as the threshold? (Exact numbers could be adjusted)

aaronbanse:
> Along with renaming to AaaDiskSpaceTest so it runs first (or right after AaaantivirusTest), which makes sense, another thing to consider is whether the test should outright kill the test process (like `System.Diagnostics.Process.GetCurrentProcess().Kill()`) instead of throwing a typical exception. Does it make sense for the tests to keep running if the disk space test fails? Or is the point to just clearly indicate that disk space was low when the tests were run?

If the goal of this test is to prevent other tests from failing due to storage that wouldn't have otherwise, this makes sense to me. It would prevent any possible confusion if we prevent the other tests from running if they could potentially yield false positives.

pwiz_tools/Skyline/Test/DiskSpaceTest.cs
```diff
@@ -0,0 +1,42 @@
...
+using pwiz.SkylineTestUtil;
+
+namespace pwiz.SkylineTest
+{
+    // Test to see if there is less than 20GB of space left on disk.
+    // Necessary to prevent unrelated tests failing when an issue causes disk to fill up.
+    [TestClass]
+    public class DiskSpaceTest : AbstractUnitTest
+    {
+        private const long MIN_SPACE_BYTES = 20_000_000_000;
```
ekoneil: Nit: use a more typical way to calculate 20GB is 20 * 1024 * 1024 * 1024.

ekoneil (changes requested):
See minor comment RE calculating 20GB.

ekoneil (approved):

aaronbanse:
I just made it a PR, alongside some byte string readability fixes.

PR #3555, opened 2025-07-21
Keep cookie token encrypted while stored, only decrypt when accessed for API call

aaronbanse:
ArdiaAccount now stores encrypted Token. SetToken method encrypts token. ArdiaClient now stores private encrypted token, Token property returns decrypted token when accessed.

### DIFF
--- a/pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
+++ b/pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
@@ -1,30 +1,39 @@
-﻿using pwiz.Common.Collections;
+﻿/*
+ * Copyright 2025 University of Washington - Seattle, WA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using pwiz.Common.Collections;
 using pwiz.Common.SystemUtil;
 
 namespace pwiz.CommonMsData.RemoteApi.Ardia
 {
     public class EncryptedToken
     {
-        public string Encrypted { get; private set; }
-        public string Decrypted => !IsNullOrEmpty(this) ? CommonTextUtil.DecryptString(Encrypted) : "";
+        public string Encrypted { get; }
+        public string Decrypted => !IsNullOrEmpty(this) ? CommonTextUtil.DecryptString(Encrypted) : string.Empty;
         public static bool IsNullOrEmpty(EncryptedToken token) => token == null || token.Encrypted.IsNullOrEmpty();
 
-        private EncryptedToken(string token)
+        private EncryptedToken(string encryptedToken)
         {
-            Encrypted = CommonTextUtil.EncryptString(token);
+            Encrypted = encryptedToken;
         }
 
-        public static EncryptedToken FromString(string token) => new EncryptedToken(token);
+        public static EncryptedToken FromString(string token) => new EncryptedToken(CommonTextUtil.EncryptString(token));
 
-        public static EncryptedToken FromEncryptedString(string encryptedToken)
-        {
-            var token = new EncryptedToken("");
-            {
-                token.Encrypted = encryptedToken;
-            };
-            return token;
-        }
+        public static EncryptedToken FromEncryptedString(string encryptedToken) => new EncryptedToken(encryptedToken);
 
-        public static EncryptedToken Empty => new EncryptedToken("");
+        public static EncryptedToken Empty => new EncryptedToken(string.Empty);
     }
 }
--- a/pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
+++ b/pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
@@ -22,8 +22,7 @@ namespace pwiz.CommonMsData.RemoteApi.Ardia
     public class EncryptedToken
     {
         public string Encrypted { get; }
-        public string Decrypted => !IsNullOrEmpty(this) ? CommonTextUtil.DecryptString(Encrypted) : string.Empty;
-        public static bool IsNullOrEmpty(EncryptedToken token) => token == null || token.Encrypted.IsNullOrEmpty();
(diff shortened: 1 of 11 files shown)


pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
```diff
@@ -0,0 +1,30 @@
+﻿using pwiz.Common.Collections;
```
ekoneil: Nit: add standard header.

pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
```diff
@@ -0,0 +1,30 @@
...
+using pwiz.Common.SystemUtil;
+
+namespace pwiz.CommonMsData.RemoteApi.Ardia
+{
+    public class EncryptedToken
+    {
+        public string Encrypted { get; private set; }
+        public string Decrypted => !IsNullOrEmpty(this) ? CommonTextUtil.DecryptString(Encrypted) : "";
+        public static bool IsNullOrEmpty(EncryptedToken token) => token == null || token.Encrypted.IsNullOrEmpty();
```
ekoneil: Does this need to be static? Or can it be converted to an instance method?

Makes sense this follows string.IsNullOrEmpty but in this case, seems like it can be simpler. But maybe I'm missing why calling it statically is necessary.

pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
```diff
@@ -0,0 +1,30 @@
+﻿using pwiz.Common.Collections;
+using pwiz.Common.SystemUtil;
+
+namespace pwiz.CommonMsData.RemoteApi.Ardia
+{
+    public class EncryptedToken
+    {
+        public string Encrypted { get; private set; }
+        public string Decrypted => !IsNullOrEmpty(this) ? CommonTextUtil.DecryptString(Encrypted) : "";
```
ekoneil: Nit: instead of "", use string.Empty. Another example at line 28.

pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            Encrypted = CommonTextUtil.EncryptString(token);
+        }
+
+        public static EncryptedToken FromString(string token) => new EncryptedToken(token);
+
+        public static EncryptedToken FromEncryptedString(string encryptedToken)
+        {
+            var token = new EncryptedToken("");
+            {
```
ekoneil: These curly braces can be removed - it's a little funky to call the constructor with empty string only to immediately overwrite that value. It implies that the relationship between the constructor (private constructor is nice!) and static methods could be simplified and the Token field made read-only, which would be a safer and less error-prone contract to have with callers.

pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
```diff
@@ -0,0 +1,30 @@
+﻿using pwiz.Common.Collections;
+using pwiz.Common.SystemUtil;
+
+namespace pwiz.CommonMsData.RemoteApi.Ardia
+{
+    public class EncryptedToken
```
ekoneil, line 22: Add a unit test! :)

ekoneil (changes requested):

pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
aaronbanse: The only reason I made it static was to mimic the analog string IsNullOrEmpty function. I'm happy to change it!

pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
ekoneil: Can be simple and use a random string - e.g. no need to use an actual token value.

pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
ekoneil: Makes sense - nice attention to surrounding code! ;)

chambm: You can't have an "IsNull*" instance method, because you can't call a function on a null object. :) You can make it an extension method though, and indeed using an extension method version of IsNullOrEmpty() for a string is quite nice.

pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
chambm: Why not just pass encryptedToken instead of `""`?

pwiz_tools/Shared/CommonMsData/RemoteApi/Ardia/EncryptedToken.cs
aaronbanse: Here's one reason why static is useful - if the Token is null, an instance method can't be called on it and needs an additional conditional to check if it's non-null wherever the method is used.
edit- didn't see Matt's comment before posting this. Thanks!

ekoneil: @aaronbanse, let's look at an example together in-person.

aaronbanse: > You can't have an "IsNull*" instance method, because you can't call a function on a null object. :) You can make it an extension method though, and indeed using an extension method version of IsNullOrEmpty() for a string is quite nice.

@chambm Where would I define the extension method, in a separate file? Since I'm only using the extension for the EncryptedToken class, it would be nice if I could just throw a static class inside EncryptedToken.cs next to the EncryptedToken class.

chambm: I think it has to be in a separate static namespace. See the example at the end of this file:
pwiz\pwiz_tools\Skyline\Controls\Graphs\GraphSummary.cs

aaronbanse: In the example, the static Extensions class is contained in the same namespace as the main class, do you mean the extension must be in a separate class?

chambm: Sorry yes I should have said class, not namespace.

ekoneil (approved):
Nice work.

PR #3564, opened 2025-07-25
Changes for LongPaths for Nightly

dshteyn:
Changes for LongPaths for Nightly


nickshulman:
Claude said that if we add the following to our app.config, we would not need to prepend "\\?\" to our paths. It would be handled automatically whenever the path was too long:
`<AppContextSwitchOverrides value="Switch.System.IO.UseLegacyPathHandling=false;Switch.System.IO.BlockLongPaths=false" />`

chambm:
According to:
https://learn.microsoft.com/en-us/archive/blogs/jeremykuhne/net-4-6-2-and-long-paths-on-windows-10

You need the app.config line and also a line in your manifest.

pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
```diff
@@ -168,8 +168,8 @@ public static String GetDirectoryName(String path)
         /// <summary>
         /// Returns a path with a long-path prefix. If the passed in path already has
         /// a long-path prefix it is returned as is. Otherwise, one is prepended.
-        /// </summary>
         /// <param name="path">Path to convert to long-path syntax. This must be a fully qualified path.</param>
+        /// </summary>
```
brendanx67: This seems like an incorrect change. Let's not change this file at all.

pwiz_tools/Skyline/SkylineNightly/SkylineNightly.csproj
```diff
@@ -127,6 +127,12 @@
     </Compile>
     <EmbeddedResource Include="SkylineNightly.skytr" />
   </ItemGroup>
+  <ItemGroup>
```
brendanx67: I don't think we need this dependency anymore, which is good, because SkylineNightly.exe generally does not have this DLL around in real-world use.

pwiz_tools/Skyline/TestRunner/app.config
```diff
@@ -5,14 +5,16 @@
...
-      <dependentAssembly>
-        <assemblyIdentity name="System.Threading.Tasks.Extensions" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-4.2.0.1" newVersion="4.2.0.1" />
-      </dependentAssembly>
-      <dependentAssembly>
-        <assemblyIdentity name="System.Runtime.CompilerServices.Unsafe" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-6.0.0.0" newVersion="6.0.0.0" />
-      </dependentAssembly>
-    </assemblyBinding>
+          <dependentAssembly>
```
brendanx67: Seems like this should be a 2-line change. Would be great to put back the original indenting on the rest of these lines.

brendanx67 (changes requested):
Mostly looks good. Just a few clean-up issues and it should be ready to go.

pwiz_tools/Skyline/Test/LongPathDirectoryTest.cs
```diff
@@ -129,13 +129,13 @@ public void DirectoryWithLongPathTest()
             if (IsVerboseMode)
                 Console.WriteLine($@"Creating directory with path ""{inputPath}"" that has length {inputPath.Length} characters");
 
-            DirectoryEx.CreateLongPath(inputPath);
+            Directory.CreateDirectory(inputPath);
```
brendanx67, line 132: Thinking about this more. I think you should leave the testing of DirectoryEx and add this as extra testing which is great confirmation that the config solution works. However, since this PR does not remove the code in DirectoryEx. It still needs to get tested until it is removed, if it is removed. That will be a separate PR.

PR #3567, opened 2025-07-29
Update Ardia import tests

ekoneil:
Use new server URL and test data. Import tests require these paths exist on the server:

/Skyline/Small/BSA_min_21
/Skyline/Small/caffeicquinic acid MSMS


pwiz_tools/Skyline/TestConnected/ArdiaFunctionalTest.cs
```diff
@@ -203,9 +203,20 @@ protected override void DoTest()
             }
             
             WaitForDocumentLoaded();
-            RunUI(() => SkylineWindow.SelectElement(ElementRefs.FromObjectReference(ElementLocator.Parse("Molecule:/Molecules/Reserpine"))));
 
-            ClickChromatogram(0.5, 33000);
```
chambm, line 208: If you swap the raw files above so caffeicquinic acid is first then you don't need to change this part, right?

ekoneil: Yeah, I wondered that too - reversing the order didn't reorder the tabs in the graph. 

The Import Results dialog still puts BSA first - example:

image

chambm: Oh, probably it's alphabetical by default then and I just got "lucky" with the old files. :)

ekoneil:
Right. :) Thanks for reviewing.

PR #3573, opened 2025-07-31
Properties View

aaronbanse:
Adding a properties view analog to Visual Studio's properties view. Clicking on a relevant control, like a FileNode in Files View, will display its properties. Properties view is a dockable form.


pwiz_tools/Skyline/Controls/FilesTree/FilesTreeForm.cs
```diff
@@ -55,7 +57,7 @@ public FilesTreeForm(SkylineWindow skylineWindow) : this()
             filesTree.LabelEdit = true;
             filesTree.AllowDrop = true;
             filesTree.NodeMouseDoubleClick += FilesTree_TreeNodeMouseDoubleClick;
-            filesTree.MouseDown += FilesTree_MouseDown;
+            filesTree.AfterSelect += FilesTree_SelectedNodeChanged;
```
aaronbanse: FilesTree_AfterSelect

pwiz_tools/Skyline/Controls/FilesTree/FilesTreeForm.cs
```diff
@@ -345,6 +347,24 @@ public Rectangle RectToScreen(Rectangle r)
...
+
+        public void NotifyPropertySheetOwnerGotFocus(SkylineWindow skyline, EventArgs e)
+        {
+            SkylineWindow.PotentialPropertySheetOwnerGotFocus(this, e);
+        }
+
+        public GlobalizedObject GetSelectedObjectProperties()
+        {
+            // Return the selected node's model as the object to display in the property sheet
+            if (!(FilesTree.SelectedNode is FilesTreeNode selectedNode))
```
aaronbanse: ftn cast function

pwiz_tools/Skyline/Model/PropertySheets/IPropertySheetOwner.cs
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System;
+
+namespace pwiz.Skyline.Model.PropertySheets
+{
+    public interface IPropertySheetOwner
+    {
```
aaronbanse, line 25: and

pwiz_tools/Skyline/Model/PropertySheets/IPropertySheetOwner.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// Should be called by the implementer when the selection changes or the owner gains focus,
+        /// to notify the main window to update the property sheet.
```
aaronbanse: And invoke SkylineWindow.PropertySheet....

pwiz_tools/Skyline/Controls/FilesTree/FilesTreeNode.cs
```diff
@@ -314,5 +316,10 @@ public bool IsFileInitialized()
             // Nodes not backed by files are considered initialized by default
             else return true;
         }
+
+        public GlobalizedObject GetProperties()
+        {
+            return FileNodeProperties.Create(Model);
+        }
```
aaronbanse, line 395: remove unused

pwiz_tools/Skyline/Model/Files/FileNode.cs
```diff
@@ -81,7 +81,7 @@ protected FileNode(IDocumentContainer documentContainer,
 
         internal IDocumentContainer DocumentContainer { get; }
         internal SrmDocument Document => DocumentContainer.Document;
-        internal string DocumentPath => DocumentContainer.DocumentFilePath;
+        public string DocumentPath => DocumentContainer.DocumentFilePath;
```
aaronbanse: revert

pwiz_tools/Skyline/Model/Files/FileNode.cs
```diff
@@ -91,6 +91,7 @@ protected FileNode(IDocumentContainer documentContainer,
         public abstract string Name { get; }
         public abstract string FilePath { get; }
         public virtual string FileName => Path.GetFileName(FilePath);
+        public string FileType => GetType().Name;
```
aaronbanse: don't do .Name since not localized, also FileType potentially conflicts with other uses

pwiz_tools/Skyline/Model/Files/Replicate.cs
```diff
@@ -160,7 +160,7 @@ public override bool ModelEquals(FileNode nodeDoc)
             return ReferenceEquals(ChromatogramSet, replicate.ChromatogramSet);
         }
 
-        private ChromatogramSet ChromatogramSet
+        public ChromatogramSet ChromatogramSet
```
aaronbanse: make internal and move specific propertyobjects to same namespace

pwiz_tools/Skyline/Model/GlobalizedObject.cs
```diff
@@ -51,7 +51,7 @@ static GlobalizedObject()
                     TypeConverterDictionary.Add(methodKey, method);
             }
         }
-
+
```
aaronbanse: remove

pwiz_tools/Skyline/Model/PropertySheets/Templates/FileNodeProperties.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        protected FileNodeProperties(FileNode fileNode)
+        {
+            FilePath = fileNode.FilePath ?? string.Empty;
+            Name = fileNode.Name ?? string.Empty;
+            FileType = fileNode.FileType ?? string.Empty;
+        }
+
+        public static FileNodeProperties Create(FileNode fileNode)
```
aaronbanse: move onus to property provider

pwiz_tools/Skyline/Model/Results/Chromatogram.cs
```diff
@@ -3,7 +3,7 @@
  *                  MacCoss Lab, Department of Genome Sciences, UW
  *
  * Copyright 2009 University of Washington - Seattle, WA
- *
```
aaronbanse, line 6: remove

pwiz_tools/Skyline/Model/Results/Chromatogram.cs
```diff
@@ -26,12 +26,12 @@
 using pwiz.Common.Collections;
 using pwiz.Common.SystemUtil;
 using pwiz.CommonMsData;
+using pwiz.CommonMsData.RemoteApi;
```
aaronbanse: revert

pwiz_tools/Skyline/Skyline.cs
```diff
@@ -1091,6 +1092,9 @@ protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
                     FindNext(true);
                     SequenceTree.UseKeysOverride = false;
                     return true;
+                case Keys.F4:
+                    ShowPropertyForm(show: !(_propertyForm is { Visible: true }));
```
aaronbanse: also used in litedropdownlist

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
```diff
@@ -23,6 +23,20 @@
 // 
 
 
+using AssortResources;
```
aaronbanse: revert reorder

pwiz_tools/Skyline/Test/CodeInspectionTest.cs
```diff
@@ -674,7 +716,7 @@ void InspectMisplacedResources(string root, List<string> errors)
             }
         }
 
-
```
aaronbanse, line 677: revert

pwiz_tools/Skyline/TestFunctional/PropertySheetTest.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
```
aaronbanse: license

pwiz_tools/Skyline/Controls/FilesTree/FilesTree.cs
```diff
@@ -75,6 +75,8 @@ public FilesTree()
             ImageList.Images.Add(Resources.ViewFile);           // 32bpp
         }
 
+        public FilesTreeNode CurrentlySelectedFTN => SelectedNode as FilesTreeNode;
```
ekoneil: The naming patten for methods like this is to append FTN to the base class's property name. That way, the name in .NET's API documentation matches the name of the convenience method that casts to FilesTreeNode. 

So: rename to SelectedNodeFTN.

pwiz_tools/Skyline/Controls/FilesTree/FilesTreeForm.cs
```diff
@@ -57,7 +58,7 @@ public FilesTreeForm(SkylineWindow skylineWindow) : this()
             filesTree.LabelEdit = true;
             filesTree.AllowDrop = true;
             filesTree.NodeMouseDoubleClick += FilesTree_TreeNodeMouseDoubleClick;
-            filesTree.MouseDown += FilesTree_MouseDown;
```
ekoneil, line 63: The MouseDown event handler was deleted - put it back. :)

pwiz_tools/Skyline/Controls/FilesTree/FilesTreeNode.cs
```diff
@@ -22,7 +22,9 @@
 using System.Windows.Forms;
 using pwiz.Common.SystemUtil;
 using pwiz.Skyline.Controls.SeqNode;
+using pwiz.Skyline.Model;
```
ekoneil: Unused import will fail code inspection. Revert.

pwiz_tools/Skyline/Controls/FilesTree/FilesTreeNode.cs
```diff
@@ -22,7 +22,9 @@
...
 using pwiz.Skyline.Model.Files;
+using pwiz.Skyline.Model.PropertySheets.Templates;
```
ekoneil: Unused import will fail code inspection. Revert.

pwiz_tools/Skyline/Model/Files/FileNode.cs
```diff
@@ -16,6 +16,7 @@
 
 using System.Collections.Generic;
 using System.IO;
+using pwiz.Skyline.Controls.FilesTree;
```
ekoneil: I understand the problem this solves; however, this approach means the model depends on the UI layer - specifically, classes in pwiz.Skyline.Controls.*. See output of the CodeInspection test (or the wiki?) for more detail about those restrictions.

Let's talk offline about how we could do this differently.

pwiz_tools/Skyline/Model/Files/ReplicateProperties.cs
```diff
@@ -0,0 +1,112 @@
...
+using pwiz.Skyline.Model.PropertySheets.Templates;
+
+namespace pwiz.Skyline.Model.Files
+{
+    public class ReplicateProperties : FileNodeProperties
+    {
+        public ReplicateProperties(FilesTreeNode filesTreeNode, Replicate model)
+            : base(filesTreeNode)
+        {
+            if (model == null)
```
ekoneil: Is passing a null model considered a coding error?

pwiz_tools/Skyline/Model/PropertySheets/Templates/FileNodeProperties.cs
```diff
@@ -0,0 +1,63 @@
...
+                FileSize = FormatFileSize(fileInfo.Length);
+            }
+        }
+
+        [Category("FileInfo")] public string FilePath { get; set; }
+        [Category("FileInfo")] public string Name { get; set; }
+        [Category("FileInfo")] public string FileSize { get; set; }
+
+        // TODO: Does this need to be localized?
+        private static string FormatFileSize(long bytes)
```
ekoneil: Take a look at FileSizeFormatProvider, which can help and will be consistent with how file sizes look elsewhere in Skyline.

pwiz_tools/Skyline/Model/PropertySheets/Templates/FileNodeProperties.cs
```diff
@@ -0,0 +1,63 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+using System.ComponentModel;
+using System.Resources;
+using pwiz.Common.Collections;
+using pwiz.Skyline.Controls.FilesTree;
+
+namespace pwiz.Skyline.Model.PropertySheets.Templates
```
ekoneil: Can this move into the Model.Files namespace? Or does it need to be closer to be in Model.PropertySheets? Moving it would keep Files-related classes together.

pwiz_tools/Skyline/Model/GlobalizedObject.cs
```diff
@@ -30,7 +47,7 @@ public UseToCompare(bool isUsed)
     /// </summary>
     public abstract class GlobalizedObject : ICustomTypeDescriptor
     {
-        private PropertyDescriptorCollection globalizedProps;
+        protected PropertyDescriptorCollection globalizedProps;
```
ekoneil: Improve encapsulation by keeping this private and adding a new method so subclasses can add property descriptors.

pwiz_tools/Skyline/Model/SpectrumProperties.cs
```diff
@@ -41,5 +41,9 @@ public void SetFileName(string fileName)
                 FileName = Path.GetFileName(fileName);
             }
         }
+
+        // Test Support - enforced by code check
+        // Invoked via reflection in InspectPropertySheetResources in CodeInspectionTest
+        private static ResourceManager ResourceManager() => SpectrumPropertiesResx.ResourceManager;
```
ekoneil: How should we think about classes like this where there are 2 ways to access SpectrumPropertiesResx.ResourceManager - (1) overriding an abstract method in the base class and (2) implementing the static method used by reflection?

pwiz_tools/Skyline/Model/Files/ReplicateProperties.cs
```diff
@@ -0,0 +1,112 @@
...
+                        GetResourceManager(), Instruments[i], instrumentsCategoryKey, instrumentsCategoryKey + i,
+                        typeof(InstrumentProperties), Instruments[i].Model, null,
+                        new Attribute[] { new TypeConverterAttribute(typeof(ExpandableObjectConverter)) }));
+                }
+            }
+        }
+
+        // Test Support - enforced by code check
+        // Invoked via reflection in InspectPropertySheetResources in CodeInspectionTest
+        private static ResourceManager ResourceManager() => PropertySheetFileNodeResources.ResourceManager;
```
ekoneil: VisualStudio shows this as unused so it appears greyed out and makes it possible someone will just delete it. :) Is there a way to mark it with VS or IntelliJ annotations as implicitly used?

pwiz_tools/Skyline/Model/Files/ReplicateSampleFileProperties.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// List rendering is suboptimal, so we either show all properties of a single instrument at the top level,
+        /// or we show each instrument's properties in a separate expandable section.
+        /// </summary>
+        protected override void AddCustomizedProperties()
+        {
+            const string instrumentsCategoryKey = "Instruments";
+
+            if (Instruments?.Count == 1)
+            {
+                // if only one instrument, add instrument properties in one category at top level
```
ekoneil: Good comments - nice to have the explainers about how these two cases are handled.

pwiz_tools/Skyline/Model/GlobalizedObject.cs
```diff
@@ -1,6 +1,23 @@
-﻿using System;
+﻿/*
```
ekoneil: Thanks for adding - good catch.

pwiz_tools/Skyline/Model/GlobalizedObject.cs
```diff
@@ -359,4 +373,123 @@ public override void SetValue(object component, object value)
...
+        }
+
+        public override bool IsReadOnly
+        {
+            get => true;
+        }
+
+        public override string Name
+        {
+            get => _name;
```
ekoneil: Simplify with expression-bodied property throughout.

pwiz_tools/Skyline/TestFunctional/PropertySheetTest.cs
```diff
@@ -0,0 +1,97 @@
...
+    public class PropertySheetTest : AbstractFunctionalTest
+    {
+        internal const string RAT_PLASMA_FILE_NAME = "Rat_plasma.sky";
+        internal const int REP_FILE_PROP_NUM = 9;
+        internal const int REP_SAMPLE_FILE_PROP_NUM = 9;
+
+        [TestMethod]
+        public void TestPropertySheet()
+        {
+            TestFilesZip = @"TestFunctional\FilesTreeFormTest.zip";
```
ekoneil: Good to reuse the (large) test archive from FilesTreeForm - (1) add a 1-liner explaining why (example in FilesModelTest), (2) remove the RAT_PLASMA_FILE_NAME above and reuse the constant from FilesTreeFormTest, and (3 - bonus points, lol) move the .zip file path to a new constant in FilesTreeFormTest since this is the 3rd place the same path is defined.

pwiz_tools/Skyline/Model/PropertySheets/PropertySheetResources.resx
ekoneil, line 1: Error_NoDisplayNameFound / etc appear to be unused - can this file be removed?

pwiz_tools/Skyline/Model/PropertySheets/Templates/FileNodeProperties.cs
ekoneil: Also for PropertySheetFileNodeResources.resx - can this move to Model\Files?

pwiz_tools/Skyline/Model/GlobalizedObject.cs
```diff
@@ -359,4 +378,123 @@ public override void SetValue(object component, object value)
...
+        private readonly string _category;
+        private readonly string _name;
+        private readonly Type _type;
+        private readonly string _nonLocalizedDisplayName;
+        private readonly Func<string, string> _displayNameFormat;
+
+        private const string DESCRIPTION_PREFIX = @"Description_";
+        private const string CATEGORY_PREFIX = @"Category_";
+
+        public CustomHandledGlobalizedPropertyDescriptor(ResourceManager resourceManager, object value, string category,
```
ekoneil: My $0.02 on ordering: I tend to put arguments passed to a constructor in an order that puts the most important args first and the least import last (followed by optional params because language spec). For me looking at a property descriptor, it's probably: type, name, value (e.g. what property are we dealing with) and then putting other properties in order of most general to most specific (ResourceManager, category, default value). 

I think about writing it that way so the next person coming along to read calls to this constructor able to understand basic type / value info about the property first (example: float foo = 3.14) and so on.

This is getting pretty down in the weeds but I decided to write this out because my initial reaction upon reading this signature was "ResourceManager is the most important thing here" which IMO ended up not being the case. 

YMMV. :)

ekoneil (changes requested):

pwiz_tools/Skyline/Model/Files/ReplicateProperties.cs
aaronbanse: [UsedImplicitly] is a nice Attribute that tells how the function is used and makes it not greyed out.

pwiz_tools/Skyline/Model/PropertySheets/Templates/FileNodeProperties.cs
aaronbanse: Yes, that makes sense. Moving them there only removes dependencies, so I'll move forward with it!

pwiz_tools/Skyline/Model/PropertySheets/PropertySheetResources.resx
aaronbanse: Yes, I believe this was an artifact leftover from my DynamicPropertyObject code, and I forgot to delete it. Good catch!

pwiz_tools/Skyline/Controls/FilesTree/FilesTreeNode.cs
aaronbanse: This doesn't show as outdated, but I think it is as it is no longer unused on my end.

pwiz_tools/Skyline/Controls/PropertyGridForm.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        public PropertyGridForm(SkylineWindow skylineWindow)
+        {
+            InitializeComponent();
+            HideOnClose = true; // Hide the form when closed, but do not dispose it
+            SkylineWindow = skylineWindow;
+            ((IDocumentUIContainer)SkylineWindow).ListenUI(SkylineWindow_OnUiDocumentChanged);
+        }
+
+        public SkylineWindow SkylineWindow { get; }
```
ekoneil: Does this need to be public? if not, improve information hiding by changing to private.

pwiz_tools/Skyline/Controls/SequenceTreeForm.cs
```diff
@@ -21,26 +21,45 @@
...
             sequenceTree.InitializeTree(documentContainer);
             sequenceTree.LockDefaultExpansion = false;
             if (documentContainer.DocumentUI != null)
                 UpdateResultsUI(documentContainer.DocumentUI.Settings, null);
+            sequenceTree.AfterSelect += SequenceTree_AfterSelect;
+        }
+
+        [Browsable(false)]
+        [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
+        public SkylineWindow SkylineWindow { get; }
```
ekoneil: Similar comment as PropertyGridForm - SequenceTreeForm doesn't need to put SkylineWindow on its public API. Marking private improves information hiding and eliminates the risk another class happens to depend on this to access SkylineWindow.

pwiz_tools/Skyline/Controls/SequenceTreeForm.cs
```diff
@@ -21,26 +21,45 @@
...
 using pwiz.Skyline.Util;
 
 namespace pwiz.Skyline.Controls
 {
-    public partial class SequenceTreeForm : DockableFormEx
+    public partial class SequenceTreeForm : DockableFormEx, IPropertyProvider
     {
         public SequenceTreeForm(IDocumentUIContainer documentContainer, bool restoringState)
         {
+            SkylineWindow = (SkylineWindow)documentContainer;
```
ekoneil: This is unexpected :) and will throw an exception should a caller pass a different implementation of IDocumentContainer. Admittedly, the likelihood of that is low - but doing a cast here raises questions that can be avoided by just passing SkylineWindow.

pwiz_tools/Skyline/Executables/DevTools/DocumentConverter
ekoneil, line 1: I don't see these in the FilesView PR. Should these be here and could this cause a problem merging back to main?

pwiz_tools/Skyline/Controls/FilesTree/FilesTreeForm.cs
```diff
@@ -404,6 +408,18 @@ public Rectangle RectToScreen(Rectangle r)
...
         #endregion
 
+        #region IPropertyProvider implementation
+
+        public SkylineObject GetPropertyObject()
+        {
+            var selectedNode = FilesTree.SelectedNodeFTN;
+            var dataSchema = new SkylineWindowDataSchema(SkylineWindow);
+            var identityPath = selectedNode?.Model.IdentityPath;
+            return selectedNode?.Model.PropertyObjectInstancer(dataSchema, identityPath);
```
ekoneil: Can the signature of PropertyObjectInstancer change and only take dataSchema? As is, this reads the model's IdentityPath and passes it right back to the same model instance. Similar happens with SequenceTree reading SelectedPath. Maybe there's a scenario I'm missing, though.

pwiz_tools/Skyline/TestFunctional/FilesTreePropertyGridTest.cs
```diff
@@ -0,0 +1,152 @@
...
+
+        private static void TestSelectedNodeProperties()
+        {
+            // test selecting a replicate node
+            var replicateFolder = SkylineWindow.FilesTree.Folder<ReplicatesFolder>();
+            var replicateNode = SkylineWindow.FilesTree.File<Replicate>(replicateFolder);
+            Assert.IsNotNull(replicateNode);
+            RunUI(() =>
+            {
+                SkylineWindow.FilesTreeForm.FilesTree.SelectNodeWithoutResettingSelection(replicateNode);
```
ekoneil: Does this need to use SelectNodeWithoutResettingSelection? That method is an implementation detail needed during drag-and-drop (and I should have marked it internal, oops). Seems better if the test uses TreeView's SelectedNode, which is public.

ekoneil: Does this need to use SelectNodeWithoutResettingSelection? That method is an implementation detail needed during drag-and-drop (and I should have marked it internal, oops). Seems better if the test uses TreeView's SelectedNode, which is public.

pwiz_tools/Skyline/TestFunctional/FilesTreePropertyGridTest.cs
```diff
@@ -0,0 +1,152 @@
...
+            // Add annotation definitions for replicates
+            RunUI(() => { PropertyGridTestUtil.TestAddAnnotations(SkylineWindow, AnnotationDef.AnnotationTarget.replicate); });
+
+            // Select replicate node
+            var replicateFolder = SkylineWindow.FilesTree.Folder<ReplicatesFolder>();
+            var replicateNode = SkylineWindow.FilesTree.File<Replicate>(replicateFolder);
+            Assert.IsNotNull(replicateNode);
+            RunUI(() =>
+            {
+                SkylineWindow.FilesTreeForm.FilesTree.SelectNodeWithoutResettingSelection(replicateNode);
```
ekoneil: Same comment as above.

pwiz_tools/Skyline/TestFunctional/SequenceTreePropertyGridTest.cs
```diff
@@ -0,0 +1,140 @@
...
+    public class SequenceTreePropertyGridTest : AbstractFunctionalTest
+    {
+        private const int PEPTIDE_EXPECTED_PROP_NUM = 14;
+        private const string SEQUENCE_PROP_NAME = "Sequence";
+        private const string NOTE_PROP_NAME = "Note";
+
+        [TestMethod]
+        public void TestSequenceTreePropertyGrid()
+        {
+            TestFilesZip = SequenceTreeRatioTest.TEST_FILES_ZIP;
```
ekoneil: Is Nick ok with sharing SequenceTreeRatioTest's zip file?

ekoneil: Is Nick ok with sharing SequenceTreeRatioTest's zip file?

pwiz_tools/Skyline/Controls/PropertyGridForm.cs
aaronbanse: Good catch, changed to private.

pwiz_tools/Skyline/Controls/SequenceTreeForm.cs
aaronbanse: Yup, I agree

pwiz_tools/Skyline/Executables/DevTools/DocumentConverter
aaronbanse: This was something I wanted to ask about, I might need help resolving this. They appeared in the git changes after I force reset from another test branch, and for some reason there was no option to undo the changes. It seems like some kind of buggy artifact, but I also tried cleaning and rebuilding to no luck. Maybe Nick has some ideas on how to resolve it, I can ask him.

pwiz_tools/Skyline/Controls/FilesTree/FilesTreeForm.cs
aaronbanse: I've updated SequenceTree to build databinding model objects off of SrmTreeNodes and not DocNodes, as SrmTreeNodes store their IdentityPath. So now both Instancers take just a DataSchema, and derive the IdentityPath from themselves.

pwiz_tools/Skyline/TestFunctional/FilesTreePropertyGridTest.cs
aaronbanse: Fixed, I must have missed the base SelectNode method earlier.

pwiz_tools/Skyline/Controls/SequenceTreeForm.cs
aaronbanse: I think Nick wrote that line in his review, but I have changed the constructor to just take a SkylineWindow. The constructor is only called in one place, and an explicit SkylineWindow is passed in, so I think it is a safe change.

pwiz_tools/Skyline/TestFunctional/SequenceTreePropertyGridTest.cs
aaronbanse: Not sure, I just went with the same strategy I did for FilesTreePropertyGridTest of using a zip that was already in use. I can ask him.

pwiz_tools/Skyline/Executables/DevTools/DocumentConverter
ekoneil: I ran into this too and iirc, resolved it using 'git submodule update'. Try updating submodules locally, though you may need to remove these files first? I think this was related to Matt's checkin, so he may have advice on how to fix.

pwiz_tools/Skyline/Controls/FilesTree/FilesTreeForm.cs
ekoneil: Nice!

pwiz_tools/Skyline/TestFunctional/SequenceTreePropertyGridTest.cs
ekoneil: Thought that might be the case. :) I remember us doing that for Files. That said, it's a pattern I wouldn't repeat since it creates a dependency between two unrelated things - making each harder to evolve. I considered changing FilesTreeFormTest.zip recently and held off to avoid breaking this. Might still change its this week, though. ;)

pwiz_tools/Skyline/TestFunctional/FilesTreePropertyGridTest.cs
ekoneil: Thanks for making that change. I'll change the visibility of FilesTree's method shortly.

pwiz_tools/Skyline/TestFunctional/SequenceTreePropertyGridTest.cs
aaronbanse: So would the correct strategy be to copy the zip and rename it to something specific to my test?

pwiz_tools/Skyline/Executables/DevTools/DocumentConverter
aaronbanse: Yeah I did that before, git submodule update resolved the local changes but didn't change anything on github when I pushed. I will email Matt

chambm: I don't think this has anything to do with any of my commits. Have you tried reverting the submodule locally and then committing it to effectively revert this change on the PR?

pwiz_tools/Skyline/TestFunctional/SequenceTreePropertyGridTest.cs
ekoneil: Yes, that's a good approach and avoids tight coupling between unrelated tests. Avoid reusing the FilesTreeFormTest.zip (which is 90MB and much larger than is typical). Instead (1) use something smaller and (2) use a .data directory rather than a .zip file. The former is easier to maintain. FilesTreeFileSystemTest is an example of both #1 and #2.

pwiz_tools/Skyline/Executables/DevTools/DocumentConverter
aaronbanse: Thanks, I was able to resolve it

pwiz_tools/Skyline/TestFunctional/SequenceTreePropertyGridTest.cs
aaronbanse: What's up with this code? There's no such file, but there is one with a .data suffix. Is TestFunctional doing something weird behind the scenes?

```TestFilesZipPaths = new[] { @"TestFunctional\FilesTreeFileSystemTest.zip" };```

PR #3575, opened 2025-08-06
Enable filtering spectra on Dissociation Method

nickshulman:
This change allows creating spectrum filters that use "Dissociation Method".
image

You can also see the Dissociation Method in the Full Scan property sheet if you click on a point along an MS2 chromatogram
image


pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -1784,6 +1791,12 @@ private static IEnumerable<MsPrecursor> GetMs1Precursors(Spectrum spectrum)
             return null;
         }
 
+        private static CVID? GetPrecursorDissociationMethod(Precursor precursor)
+        {
+            return precursor.activation?.cvParams
+                .FirstOrDefault(param => CV.cvIsA(param.cvid, CVID.MS_dissociation_method))?.cvid;
```
chambm: I'm not sure whether this suffices for cleaning up the CVParam wrappers that get generated. Can you run this through dotMemory in a unit test a few hundred thousand times and see if anything builds up?

chambm: Oh the leak detected on TeamCity seems to confirm my comment. :) Although I'm surprised it was enough to trigger leak detection. Perhaps make the lambda a statement lambda and after checking `cvIsA`, call `param.Dispose()`? Otherwise there must be some way to put it in a `using` scope. We do have another cvParams usage in MsDataFileImpl but it's not used anywhere (ConfigInfo).

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -1793,8 +1794,17 @@ private static IEnumerable<MsPrecursor> GetMs1Precursors(Spectrum spectrum)
 
         private static CVID? GetPrecursorDissociationMethod(Precursor precursor)
         {
-            return precursor.activation?.cvParams
-                .FirstOrDefault(param => CV.cvIsA(param.cvid, CVID.MS_dissociation_method))?.cvid;
+            foreach (var cvParam in (IEnumerable<CVParam>) precursor.activation.cvParams ?? Array.Empty<CVParam>())
+            {
+                using (cvParam)
```
chambm, line 1810: Hah I didn't even know you could do that. Not sure why this is still leaking. But I did realize this could be more efficient. Use: `precursor.activation.cvParamChildren(CVID.MS_dissociation_method)` to get the list of CVParams falling under that term, you don't have to do the matching yourself. Note I switched to the plural version because there can be more than one method for a precursor. For example, EThcD will have HCD and ETD (HCD is a "supplemental" activation). So perhaps the method will need to return List<CVID>.

nickshulman: @chambm, do you have an example of a file with this sort of multiple dissociation methods? I'll add it to my unit test.

The dissociation method ends up being a string by the time Skyline sees it.
If there is more than one dissociation method in a single spectrum (which could theoretically happen if there was more than one precursor) then the value ends up being the methods separated by spaces. That might be how we would want this to work. Since the value is just a string, users could use "contains" in the Spectrum Filter that they define.

chambm: [Rpal-ETciD.zip](https://github.com/user-attachments/files/21687660/Rpal-ETciD.zip)

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
nickshulman: I can't figure out how to use a CVParamList without leaking memory.
I can fix the leak by using `precursor.activation.cvParamChild(CVID.MS_dissociation_method)` but if I try to use `precursor.activation.cvParamChildren(CVID.MS_dissociation_method)` then I do seem to leak.

I have added a test "TestDataFileSpectrumMemoryLeak" in DataFileMemoryLeakTest.cs.
The following code leaks:
```
var list = new List<CVID>();
// TODO (nicksh): This code leaks
foreach (var cvParam in precursor.activation.cvParamChildren(CVID.MS_dissociation_method))
{
    using (cvParam)
    {
        list.Add(cvParam.cvid);
    }
}

return list;
```

Also, I noticed that there is other code in MsDataFileImpl.GetInstrumentConfig which also seems to leak when tested by "TestDataFileConfigInfoMemoryLeak":
```
foreach (CVParam term in fileDescriptionFileContent.cvParams)
    contentTypeSet.Add(term.name);
```
Do you have any ideas how to rewrite either of these blocks of code so they would not leak?

In theory, it should not be necessary to have a `using` block with a CVParam, because the C# finalizer `~CVParam` and the C++ finalizer `CVParam::!CVParam` should both get called so that the thing gets deleted.

chambm: I'm looking into it.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
nickshulman: For that file, for the spectra with more than on dissociation method, the Dissociation Method in the property sheet will show up as "CID ETD supplemental collision-induced dissociation".
image

chambm: We may want to special case this such that if supplemental CID is there, then regular CID is omitted. Or perhaps vice versa if we think users would rather stick with the short names.

PR #3581, opened 2025-08-12
Fix "Use only scans within X minutes of predicted time" includes two extra points after where it should have stopped

nickshulman:
Mike noticed that with his scheduled PRM data, whenever there was a gap between the intended range and some other range that happened to have a matching precursor, the XIC would always have two extra points on the right side of the gap.
image

This change makes it so those two extra points do not get included.

This change affected the expected values in a lot of tests.

Also, increased the timeout in a few tests that seemed to be flaky.
Also, fixed running Koina tests in Test Explorer: was getting error about wrong version of System.Runtime.CompilerServices.Unsafe which required binding redirect in TestConnected.csproj


pwiz_tools/Skyline/TestPerf/DiaSwathTutorialTest.data/TestTtofData_full.json
```diff
@@ -45,7 +45,7 @@
     2142
   ],
   "UnpolishedProteins": 2207,
-  "PolishedProteins": 2394,
+  "PolishedProteins": 2207,
```
brendanx67: Do you understand why this change would cause the loss of 187 proteins? It also seems a bit suspicious that the number is now the same as UnpolishedProteins.

brendanx67: I am even a bit suspious that a fix meant to solve a problem in PRM is causing these changes to DIA data processing.

nickshulman: Good point. It looks like I might have generated that particular number with one flaky run of the test. I will try to reproduce that incorrect number. There probably needs to be a more robust "WaitForCondition" before it verifies/records that particular expected value.

PR #3583, opened 2025-08-18
Changes for dealing with newer relatively complex Bruker DiaPASEF modes, e.g. "Diagonal" PASEF or "MiDia" PASEF…

bspratt:
… (which, significantly, has overlapping isolation ranges for its MS2 scans)

The fundamental change is Skyline's use of the DiaFrameMsMsWindows table found in analysis.tdf. This information is now passed into the mzML model as a set of UserParams.

Equipped with this information Skyline can now choose the ideal WindowGroup for each transition before starting chromaogram extraction. Formerly this was based on observing the data as it passed by, which pretty quickly settled on the proper window group but often meant a few early chhromatogram points pulled from the wrong window group. The observable effect is that the more correct approach tends to produce chromatograms that start slightly later than with the old code.

One significant effect of this is that Skyline can determine which transitions simply do not fit any WindowGroup's combination of precursor isolation and ion mobility filtering. In these cases a message is written to the ImmediateWindow (or console, for commandline use) e.g. "No isolation window found in B_20231027_IO25x75_HeLa_800ng_35min_midiaPASEF_24x12_75ms_frag_Slot2-6_1_6699.d for precursor LLEDGEDFNLGDALDSSNSMQTIQK (m/z=1370,634536025 im=ccs530,301031483079/1,3173898204277Vs/cm^2/he/1/K0 (Vs/cm^2)/w0,087826)"

Skyline requests Bruker DIA MS2 data be passed as entire frames in the 3-array [mz, intensity, ionMobility], omitting any scans in the frame that are not mentioned in the frame's DiaFrameMsMsWindows entry.

Skyline can use the ion mobility values in MS2 frames as markers for precursor isolation range, so Skyline does not request the optional isolation range arrays [mz, intensity, ionMobility, isoLow, isoHigh] - it just asks for [mz, intensity, ionMobility].

N.B. I tried not sorting MS2 full frames on mz, instead make the inherently sorted IM dimension our primary when we have that IM<->Isolation mapping available, but there was no performance benefit.

### DIFF
--- a/pwiz_tools/Skyline/Controls/Graphs/GraphValues.cs
+++ b/pwiz_tools/Skyline/Controls/Graphs/GraphValues.cs
@@ -162,7 +162,7 @@ namespace pwiz.Skyline.Controls.Graphs
 
             public bool TryGetRegressionFunction(MsDataFileUri filePath, out AlignmentFunction regressionFunction)
             {
-                regressionFunction = SrmSettings.DocumentRetentionTimes.GetAlignmentFunction(SrmSettings.PeptideSettings.Libraries, filePath,
+                regressionFunction = SrmSettings.DocumentRetentionTimes.GetRunToRunAlignmentFunction(SrmSettings.PeptideSettings.Libraries, filePath,
                     false);
                 return regressionFunction != null;
             }
--- a/pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
+++ b/pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
@@ -1227,6 +1227,11 @@ namespace pwiz.Skyline.Model.DocSettings
 
         public double[] GetAlignedRetentionTimes(ChromatogramSet chromatogramSet, MsDataFileUri filePath, IList<Target> targets)
         {
+            var alignmentFunction = DocumentRetentionTimes.GetLibraryAlignmentFunction(PeptideSettings.Libraries, filePath, true);
+            if (alignmentFunction == null)
+            {
+                return Array.Empty<double>();
+            }
             var batchNames = new List<string> { null };
             if (!string.IsNullOrEmpty(chromatogramSet?.BatchName))
             {
@@ -1259,22 +1264,17 @@ namespace pwiz.Skyline.Model.DocSettings
                         }
                     }
 
-                    var alignmentFunction = libraryAlignment.Alignments.GetAlignmentFunction(filePath, true);
-                    if (alignmentFunction != null)
+                    var times = new List<double>();
+                    foreach (var normalizedTime in libraryAlignment.GetNormalizedRetentionTimes(spectrumSourceFiles,
+                                 targets))
                     {
-                        var times = new List<double>();
-                        foreach (var normalizedTime in libraryAlignment.GetNormalizedRetentionTimes(spectrumSourceFiles,
-                                     targets))
-                        {
-                            var alignedTime = alignmentFunction.GetY(normalizedTime);
-                            // Console.Out.WriteLine("Mapping normalized time {0} to {1} for file {2}", normalizedTime, alignedTime, filePath.GetFileName());
-                            times.Add(alignedTime);
-                        }
-                        if (times.Count > 0)
-                        {
-                            return times.ToArray();
-                        }
-
+                        var alignedTime = alignmentFunction.GetY(normalizedTime);
+                        // Console.Out.WriteLine("Mapping normalized time {0} to {1} for file {2}", normalizedTime, alignedTime, filePath.GetFileName());
+                        times.Add(alignedTime);
+                    }
+                    if (times.Count > 0)
+                    {
+                        return times.ToArray();
                     }
                 }
             }
@@ -1301,6 +1301,11 @@ namespace pwiz.Skyline.Model.DocSettings
 
         public double[] GetRetentionTimesNotAlignedTo(ChromatogramSet chromatogramSet, MsDataFileUri fileNotAlignedTo, ICollection<Target> targets, ICollection<SignedMz> precursorMzs)
         {
+            if (null != DocumentRetentionTimes.GetLibraryAlignmentFunction(PeptideSettings.Libraries, fileNotAlignedTo,
+                    true))
+            {
+                return Array.Empty<double>();
+            }
             var result = new List<double>();
             foreach (var library in PeptideSettings.Libraries.Libraries)
             {
@@ -1309,11 +1314,6 @@ namespace pwiz.Skyline.Model.DocSettings
                     continue;
                 }
 
-                if (null != DocumentRetentionTimes.GetLibraryAlignment(library.Name))
-                {
-                    // TODO(nicksh): if there is a batch name, then potentially a subset of the times in this library should be returned.
-                    continue;
-                }
                 result.AddRange(library.GetRetentionTimesWithSequences(null, targets).SelectMany(list=>list));
                 if (library is MidasLibrary)
                 {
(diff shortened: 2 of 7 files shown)


pwiz_tools/Skyline/Model/DocSettings/TransitionIonMobilityFiltering.cs
```diff
@@ -1004,11 +1004,18 @@ public IonMobilityFilter Merge(IonMobilityFilter other, bool isMs1)
...
                 return this;
-            return GetIonMobilityFilter(IonMobility.Mobility + offset, IonMobilityUnits, IonMobilityExtractionWindowWidth, CollisionalCrossSectionSqA);
+            // Original bounds
+            var boundsLow = IonMobility.Mobility.Value - 0.5 * IonMobilityExtractionWindowWidth??0;
+            var boundsHigh = boundsLow + IonMobilityExtractionWindowWidth??0;
+            // Apply offsets
+            boundsLow += offsetLow;
+            boundsHigh += offsetHigh;
+            var width = Math.Abs(boundsHigh - boundsLow);
+            return GetIonMobilityFilter(boundsLow + 0.5*width, IonMobilityUnits, width, CollisionalCrossSectionSqA);
```
nickshulman: Since you did "width = Math.Abs(boundsHigh - boundsLow)" on the previous line, I would expect the first argument here to be:
`GetIonMobilityFilter(Math.Min(boundsHigh, boundsLow) + 0.5 * width...`

pwiz_tools/Skyline/TestUtil/ResultsUtil.cs
```diff
@@ -483,5 +483,85 @@ private static double MaxPeakTime(ChromatogramGroupInfo chromGroupInfo)
...
+                var fileNames = doc.MeasuredResults.MSDataFilePaths.ToArray();
+                for (var index = 0; index < fileNames.Length; index++)
+                {
+                    var fileName = fileNames[index].GetFileName();
+                    if (skyFileContents == null ||
+                        skyFileContents.Contains(fileName)) // Did we remove the replicate for debug?
+                    {
+                        foreach (var pair in doc.PeptidePrecursorPairs)
+                        {
+                            if (!results.TryLoadChromatogram(index, pair.NodePep, pair.NodeGroup,
```
nickshulman: Passing "index" in here means that this code is assuming that there is only one result file per Replicate.
You might want to rewrite this so that it has an outer loop that iterates of MeasuredResults.Chromatograms, and an inner loop which iterates over the MSDataFileInfos in that ChromatogramSet.

bspratt: I'll add a comment to this effect, thanks. As it's only debug code it's probably not worth the effort to make it 100% bulletproof against a relatively unusual condition.

bspratt: Looking more closely I do take your point, we could miss important differences and never know about it. I'll implement your suggestion, thanks.

pwiz_tools/Skyline/Model/DocSettings/TransitionIonMobilityFiltering.cs
bspratt: Good point, thanks.

pwiz_tools/Skyline/Model/Results/ChromCacheBuilder.cs
```diff
@@ -203,6 +203,7 @@ public void BuildCache()
                                     .IsEnabledMsMs, // If we don't want MS2, ask reader to totally skip it (not guaranteed)
                             CentroidMs1 = fullScan.IsCentroidedMs,
                             CentroidMs2 = fullScan.IsCentroidedMsMs,
+                            PassEntireDiaPasefFrame = true, // Ask for diaPASEF frames as single chunks instead of split by isolation ranges
```
chambm, line 206: So regular diaPASEF data is now being processed the same way as diagonalPASEF? Was there a notable performance difference?

pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.cpp
```diff
@@ -501,6 +559,45 @@ TimsDataImpl::TimsDataImpl(const string& rawpath, bool combineIonMobilitySpectra
                 ms2Time = frame->rt_;
                 ms2Intensities.push_back(frame->tic_);
             }
+
+            // Note contents of DiaFrameMsMsWindows, with scan range translated to 1/K0 range
+            if (oneOverK0ByScanNumberByCalibration_.size() == 1) // Only works if calibration doesn't change in run
```
chambm, line 564: So I guess we'll fix multi-calibration files when we get one? :)

pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.hpp
```diff
@@ -245,6 +247,14 @@ struct PWIZ_API_DECL TimsDataImpl : public CompassData
     /// returns true if the source is TIMS PASEF data
     virtual bool hasPASEFData() const;
 
+    /// returns true if the source is TIMS diagonal PASEF data (combined spectra will have isolation m/z arrays)
```
chambm: comment should be updated, right?

pwiz/data/vendor_readers/Bruker/SpectrumList_Bruker.cpp
```diff
@@ -429,7 +441,41 @@ PWIZ_API_DECL SpectrumPtr SpectrumList_Bruker::spectrum(size_t index, DetailLeve
...
+                    result->binaryDataArrayPtrs.push_back(isolationMzEnd);
+                    arrayType = MS_scanning_quadrupole_position_upper_bound_m_z_array;
+                    arrayType.units = MS_m_z;
+                    isolationMzEnd->cvParams.emplace_back(arrayType);
+
+                    spectrum->getCombinedSpectrumData(mz, intensity, mobility->data, true, isolationMzStart->data, isolationMzEnd->data, config_.sortAndJitter);
+
+                    // Note the isolation range
+                    double isoLower = isolationMzStart->data[isolationMzStart->data.size() - 1];
+                    double isoUpper = isolationMzEnd->data[0];
```
nickshulman, line 463: Why does "isoLower" get set to a value from the last window, but "isolUpper" gets set to a value from the first window?

pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.cpp
bspratt: Yeah, if and when.

pwiz_aux/msrc/utility/vendor_api/Bruker/TimsData.hpp
bspratt: yes, I'll change that to
// returns true if the source is TIMS diagonal PASEF data (combined spectra will have isolation m/z arrays by default)

pwiz/data/vendor_readers/Bruker/SpectrumList_Bruker.cpp
bspratt: The values are presented in reverse sort order, just a quirk of the format

bspratt:
> Looks good. I don't really understand the activeScans thing though.
I'll improve the comments there. It's just about WindowGroups where some scans aren't included in any part of the isolation scheme, the caller can choose to have those omitted from the data transfer.

pwiz_tools/Skyline/Model/Results/ChromCacheBuilder.cs
bspratt: Correct, it's just easier to maintain a single general solution. I haven't done enough iterations to have solid mean times for master or branch, but in my sampling cherry-picking the best observed times for each the performance is as good as before, and in some cases rather better.

PR #3585, opened 2025-08-20
Files view persistent location bug fix

aaronbanse:
Fixed a bug where if Files View is closed or disabled while docked at a non-default location, it will always reappear at the default location when enabled again.

Note: Extra commits and reverts mixed in here as I had committed work that should have been on a different branch.


ekoneil (approved):
Nice find on this bug. Thanks for the fix.

PR #3587, opened 2025-08-22
Add SearchTool, SearchToolList, Edit Search Tools, customizable Java heap

chambm:
- added SearchTool and SearchToolList to store both user-managed and auto-downloaded search tools (could be refactored to other tools in the future)
- added Tools / Edit Search Tools... menu option
- added automatic copying of auto-downloaded search tools from old locations to current location (e.g. when updating Skyline version, MSFragger from old version will be copied to new version's Tools directory)
- moved Java heap parameters to SearchTool property so it is customizable by user (and defaults to 2/3 of their total memory)
* added GetEnumAttribute overload that throws for null/empty instead of returning defaultValue
* added EXECUTABLE_FILES_FILTER for use in OpenFile dialogs that search for executable files (first option is all common executable types, not just .exe)
* added overloads for OkDialog and CancelDialog that use AcceptButton and CancelButton instead of passing an action

### DIFF
--- a/pwiz_tools/Skyline/Model/SearchTool.cs
+++ b/pwiz_tools/Skyline/Model/SearchTool.cs
@@ -116,7 +116,7 @@ namespace pwiz.Skyline.Model
         public void ReadXml(XmlReader reader)
         {
             // Read tag attributes
-            Name = reader.GetEnumAttribute<SearchToolType>(ATTR.name, XmlUtil.EnumCase.unkown);
+            Name = reader.GetRequiredEnumAttribute<SearchToolType>(ATTR.name, XmlUtil.EnumCase.unkown);
             Path = reader.GetAttribute(ATTR.path);
             ExtraCommandlineArgs = reader.GetAttribute(ATTR.args);
             InstallPath = reader.GetAttribute(ATTR.install_path);
--- a/pwiz_tools/Skyline/Model/DdaSearch/HardklorSearchEngine.cs
+++ b/pwiz_tools/Skyline/Model/DdaSearch/HardklorSearchEngine.cs
@@ -95,7 +95,12 @@ namespace pwiz.Skyline.Model.DdaSearch
 
         public override string SearchEngineBlurb => DdaSearchResources.HardklorSearchEngine_SearchEngineBlurb;
 
-        public override event NotificationEventHandler SearchProgressChanged;
+        //public override event NotificationEventHandler SearchProgressChanged;
+        public override event NotificationEventHandler SearchProgressChanged
+        {
+            add { }
+            remove { }
+        }
 
         public override bool Run(CancellationTokenSource cancelToken, IProgressStatus status) // Single threaded version
         {
--- a/pwiz_tools/Skyline/Model/DdaSearch/TideSearchEngine.cs
+++ b/pwiz_tools/Skyline/Model/DdaSearch/TideSearchEngine.cs
@@ -619,15 +619,6 @@ namespace pwiz.Skyline.Model.DdaSearch
                     if (mod.IsVariable  )
                     {
                         ++modCounter;
-                        const int force = 0;
-                        int distance = mod.Terminus == null ? -1 : 0;
-                        int terminus = mod.Terminus switch
-                        {
-                            null => 0,
-                            ModTerminus.N => 2,
-                            ModTerminus.C => 3,
-                            _ => throw new ArgumentException(nameof(mod.Terminus))
-                        };
                         string massSign = mass > 0 ? @"+" : @"";
                         string res = residues;
                         if (residues == @"n^" || residues == @"c^")


pwiz_tools/Skyline/Util/Xml.cs
```diff
@@ -716,6 +716,20 @@ public static TAttr GetEnumAttribute<TAttr>(this XmlReader reader, Enum name, TA
...
         public static TAttr GetEnumAttribute<TAttr>(this XmlReader reader, Enum name, TAttr defaultValue, EnumCase enumCase)
+        {
+            try
+            {
+                return reader.GetEnumAttribute<TAttr>(name, enumCase);
+            }
+            catch(InvalidDataException x)
+            {
+                if (x.InnerException == null)
+                    return defaultValue;
```
nickshulman: This method is supposed to throw an exception here instead of returning defaultValue.
It's only supposed to return defaultValue if the attribute value is missing or empty.
But if the attribute value is there and does not match any of the options, it's supposed to throw an exception.
We probably don't want to change this behavior because it could hide bugs.

pwiz_tools/Skyline/Model/SearchTool.cs
```diff
@@ -0,0 +1,172 @@
...
+
+using System.Xml;
+using System.Xml.Schema;
+using System.Xml.Serialization;
+using pwiz.Common.Collections;
+using pwiz.Skyline.Util;
+
+namespace pwiz.Skyline.Model
+{
+    public enum SearchToolType
```
nickshulman, line 29: I would recommend getting rid of this enum and defining separate classes for each of these search tool types.
I expect that you will eventually get tired of writing "switch(SearchToolType)" for everything that is slightly different about these tools.
I might be wrong, though.

pwiz_tools/Skyline/Model/SearchTool.cs
```diff
@@ -0,0 +1,172 @@
...
+        DIANN,
+        EncyclopeDIA,
+        Java,
+        Java8,
+        MSFragger,
+        MSGFPlus
+    }
+    
+    [XmlRoot("search_tool")]
+    public sealed class SearchTool : IKeyContainer<string>, IXmlSerializable
```
nickshulman: I would recommend making this class extend Immutable.
We are so used to working with Immutable objects everywhere else in the code we're likely to make a mistake with these objects and accidentally modify it when we should have made a copy.

pwiz_tools/Skyline/FileUI/PeptideSearch/MatchModificationsControl.resx
```diff
@@ -336,9 +336,9 @@
   <data name="&gt;&gt;btnAddModification.ZOrder" xml:space="preserve">
     <value>2</value>
   </data>
-  <data name="menuAddModification.TrayLocation" type="System.Drawing.Point, System.Drawing">
```
chambm, line 339: @nickshulman Not sure what these changes are. Should I revert them?

pwiz_tools/Skyline/FileUI/PeptideSearch/SearchControl.resx
```diff
@@ -357,13 +357,16 @@
...
+  </data>
   <data name="$this.Size" type="System.Drawing.Size, System.Drawing">
     <value>358, 424</value>
   </data>
   <data name="&gt;&gt;$this.Name" xml:space="preserve">
     <value>SearchControl</value>
   </data>
   <data name="&gt;&gt;$this.Type" xml:space="preserve">
-    <value>System.Windows.Forms.UserControl, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    <value>pwiz.Skyline.Controls.WizardPageControl, Skyline-daily, Version=25.1.1.209, Culture=neutral, PublicKeyToken=null</value>
```
chambm, line 370: @nickshulman Should Version be manually set to 0.0.0.0? Or should the whole line be reverted even though the type would be wrong?

nickshulman: > @nickshulman Should Version be manually set to 0.0.0.0? Or should the whole line be reverted even though the type would be wrong?

The values in the .resx file whose names start with "&gt;&gt;" have no impact on Skyline at runtime.
The only reason that values like that are in there is to enable implementing a graphical designer for these .resx files which can do its job without having to look at the .designer.cs file.
We never use design tools like that, so we don't care what these sorts of values look like in the .resx files.
I would recommend checking in this line as you have it in this pull request, because that is the correct current value, but it really does make no difference.

pwiz_tools/Skyline/FileUI/PeptideSearch/MatchModificationsControl.resx
nickshulman: This `<data name="menuAddModificdation.TrayLocation"` is a bogus thing to have in a .resx file. It is good that you are deleting it.

If the tag was "metadata" instead of "data", then this would be the thing that controls where the menu item is displayed in the Visual Studio Form Designer. It's not something we care much about, but it does affect the way things look in the designer.
But I don't think this data tag with a tray location does anything, and was probably the result of a human hand-editing the .resx file.

pwiz_tools/Skyline/Model/SearchTool.cs
chambm: In this case it's really used as a name. I don't think I've needed to use `switch(SearchToolType)` at all. That kind of logic is pretty extensive in SearchSettingsControl though, but that logic isn't new to this PR.

pwiz_tools/Skyline/Util/Xml.cs
chambm: What I wanted was a GetEnumAttribute without a defaultValue option at all. I just re-implemented the existing function in terms of the new one. It should have the same behavior: InnerException is null if string.IsNullOrEmpty. It's not null if there was an incorrect value that caused a parsing error. That is admittedly pretty odd, but can you think of a clearly better solution for differentiating between "exception because value was empty/null" and "exception because the value was not valid for the enumeration"?

nickshulman: > What I wanted was a GetEnumAttribute without a defaultValue option at all.

To be consistent with the other methods in XmlUtil, we would probably want to call a method like that "GetNullableEnumAttribute" and it would be up to the caller to throw an exception if the attribute is missing.

Usually we would not worry about whether the Name attribute is missing here since the only way the user could get in this state would be by hand-editing the user.config file.
If you are actually worried about the user hand-editing the XML and putting an invalid name attribute in there, you should use the "TypeSafeEnum" class to make sure the enum value being returned is actually one of the listed values in the "SearchToolEnum" type.
That is, if you are worried the XML might contain:
`name="-1"`
you should use something like "TypeSafeEnum.Parse"

chambm: I wanted to be able to do this:
`Name = reader.GetEnumAttribute<SearchToolType>(ATTR.name, XmlUtil.EnumCase.unkown);`
If parsing the attribute fails for any reason, it should throw an exception. Is that incorrect behavior for `ReadXml()`?

You think I should use TypeSafeEnum<SearchToolType> instead? Like you said it would probably take hand editing (or enum member renaming in a later version) to cause an error here, so I just tried to keep it simple.

nickshulman: Probably the simplest would be, in "SearchTool.ReadXml" to do:
`Name = reader.GetEnumAttribute(ATTR.name, default(SearchToolType));`

(this is the same as `Name = reader.GetEnumAttribute(ATTR.name, SearchToolType.CruxComet);`, but "default(SearchToolType)" probably makes it more clear that it's not important what happens when the `name` attribute is missing)

Yes, throwing exceptions in ReadXml is the behavior we want.
The user will not see those exceptions, though, because System.Configuration.SettingsBase catches and ignores all during property deserialization.

chambm: But I want it to throw for any reason, not just if it's an invalid value. I.e. if it's missing it's also an error. It seemed needlessly complicated to handle that in the calling code (the easiest way to do that would indeed be to use the nullable version and then throw the exception after a coalescing operator `??`). Perhaps rename the inner method to `GetRequiredEnumAttribute`?

PR #3615, opened 2025-09-26
Skyline: Add Help > Documentation > Keyboard Shortcuts menu item

brendanx67:
- With help from Cursor on Claude and ChatGPT models
- Started creating agent notes and STYLEGUIDE.md for LLM agents
- Also added a test that checks for duplicate menu mnemonics and suggests possible solutions when run in English
- Fixed the cases we had of duplicate mnemonics


pwiz_tools/Skyline/Menus/KeyboardShortcutDocumentation.cs
```diff
@@ -0,0 +1,251 @@
...
+                () => GetMnemonicRows(menuMain), 
+                MenusResources.Keyboard_Mnemonics_Table_Title, 
+                MenusResources.Keyboard_Mnemonics_Header_Mnemonic);
+            
+            sb.Append(@"</body></html>");
+            return sb.ToString();
+        }
+
+        private static void AddShortcutsTable(StringBuilder sb, 
+            Func<List<(string MenuPath, string Shortcut)>> getRows,
```
nickshulman: I would probably make this just `List<(string MenuPath, string Shortcut)>` without the `Func`.

pwiz_tools/Skyline/Menus/KeyboardShortcutDocumentation.cs
```diff
@@ -0,0 +1,251 @@
...
+            return string.Empty;
+        }
+
+        public static string CleanMenuText(string text)
+        {
+            if (string.IsNullOrEmpty(text))
+                return string.Empty;
+
+            // Remove accelerator ampersands and trailing ellipses
+            var noAmp = text.Replace(@"&", string.Empty);
```
nickshulman: If you wanted you could do:
`return text.Replace(@"&", string.Empty).TrimEnd('…', '.');`
"TrimEnd" removes any number of occurrences of those characters, so it's not exactly the same as looking for three "." characters, but we probably don't care.

nickshulman (approved):

PR #3619, opened 2025-10-01
Prevent issues when passing a zero-length list to MeasuredResults.ChangeChromatograms

ekoneil:
Prevent a strange issue in MeasuredResults when calling ChangeChromatograms. Passing a zero-length list causes issues with SequenceTree, undo/redo, Audit Log, and FilesTree.

Instead:

(1) Require callers to pass null to clear a non-empty list of chromatograms. Throw an exception except for one special case described below.

(2) Allow passing a zero-length list when MeasuredResults already has no chromatograms. This case happens during XML deserialization. MoleculeDocumentSerializeTest fails unless this is allowed.

### DIFF
--- a/pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
+++ b/pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
@@ -531,7 +531,15 @@ namespace pwiz.Skyline.Model.DocSettings
 
         public SrmSettings ChangeMeasuredResults(MeasuredResults prop)
         {
+            ValidateMeasuredResults(prop);
+
             return ChangeProp(ImClone(this), im => im.MeasuredResults = prop);
+        }
+
+        private void ValidateMeasuredResults(MeasuredResults measuredResults)
+        {
+            if (measuredResults != null && measuredResults.Chromatograms.Count == 0)
+                throw new ArgumentException(@"MeasuredResults must have at least one ChromatogramSet. Use null to remove all results.");
         }
 
         public SrmSettings ChangeIsResultsJoiningDisabled(bool prop)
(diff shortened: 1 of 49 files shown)


ekoneil:
Below is the stack trace for the failure during XML deserialization.

Stacktrace

There is an error in XML document (40, 4).
   at System.Xml.Serialization.XmlSerializer.Deserialize(XmlReader xmlReader, String encodingStyle, XmlDeserializationEvents events)
   at System.Xml.Serialization.XmlSerializer.Deserialize(TextReader textReader)
   at pwiz.SkylineTestUtil.AssertEx.Deserialize[TObj](String s) in C:\pwiz\pwiz_tools\Skyline\TestUtil\AssertEx.cs:line 300
   at pwiz.SkylineTest.SrmDocumentTest.MoleculeDocumentSerializeTest() in C:\pwiz\pwiz_tools\Skyline\Test\SrmDocumentTest.cs:line 136
------- Stdout: -------
 74. MoleculeDocumentSerializeTest

nickshulman:
What was the code in your branch that was causing the problem?
I don't think that giving the zero-length list to MeasuredResults is the problem.
I think the problem is that SrmSettings never wants to have a non-null MeasuredResults with a zero-length Chromatograms list.
So, calling "ChangeMeasuredResults(MeasuredResults.Empty)" is the thing that we want to prevent.

pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
```diff
@@ -2280,15 +2288,20 @@ public override void ReadXml(XmlReader reader)
             // 10.23.12 -- The order of <measured_results> and <data_settings> has been switched to enable parsing (in Panorama)
             // of all annotation definitions before reading any replicate annotations in <measured_results>.
             // We want Skyline to be able to read older documents where <measured_results> come before <data_settings>
+            MeasuredResults NullIfEmpty(MeasuredResults val)
```
nickshulman: It would probably be cleaner if "NullIfEmpty" were a method on MeasuredResults:
`public MeasuredResults NullIfEmpty() { return Chromatograms.Count == 0 ? null : this);}`

Then you would do:
`MeasuredResults = reader.DeserializeElement<MeasuredResults>()?.NullIfEmpty();`

brendanx67: My original implementation was in this direction, but I forgot about using ?.NullIfEmpty(). Yeah. That seems cleaner.

PR #3626, opened 2025-10-04
Skyline: Extended keyboard shortcut help

brendanx67:
- Added keyboard help for grids
- Replaced deprecated WebBrowser (IE-based) with WebView2 (Chromium based)
- Enabled DocumentationViewer test in StandardTypeTest

### DIFF
--- a/pwiz_tools/Shared/Common/DataBinding/Controls/Editor/DocumentationViewer.cs
+++ b/pwiz_tools/Shared/Common/DataBinding/Controls/Editor/DocumentationViewer.cs
@@ -1,6 +1,7 @@
 ﻿/*
  * Original author: Nicholas Shulman <nicksh .at. u.washington.edu>,
  *                  MacCoss Lab, Department of Genome Sciences, UW
+ * AI assistance: Cursor (Claude Sonnet 4) <cursor .at. anysphere.co>
  *
  * Copyright 2016 University of Washington - Seattle, WA
  *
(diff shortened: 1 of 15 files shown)


pwiz_tools/Shared/Common/DataBinding/Controls/Editor/DocumentationViewer.cs
```diff
@@ -17,31 +18,216 @@
...
+                @"--disable-backgrounding-occluded-windows",
+                @"--disable-renderer-backgrounding");
+            
+            // Create environment with custom user data folder on UI thread
+            return CoreWebView2Environment.CreateAsync(null, _testDataFolder, options).Result;
+        }
+
+        private void RunUI(Action act)
+        {
+            Invoke(act);
```
nickshulman: I would recommend using "CommonActionUtil.SafeBeginInvoke" here because it handles the case where the form has already been closed.
If it were important that you use "Invoke" instead of "BeginInvoke" then you would not want to use SafeBeginInvoke, but I don't think we care in this case.

pwiz_tools/Shared/Common/DataBinding/Controls/Editor/DocumentationViewer.cs
```diff
@@ -17,31 +18,216 @@
...
+        }
+
+        private void CleanupTestDataFolder()
```
nickshulman: Can you move this code into "DocumentationViewerHelper"?

pwiz_tools/Skyline/TestFunctional/KeyboardShortcutHelpTest.cs
```diff
@@ -22,15 +22,14 @@
...
 using pwiz.SkylineTestUtil;
 
 namespace pwiz.SkylineTestFunctional
 {
     [TestClass]
-    public class KeyboardShortcutHelpTest : AbstractFunctionalTest
+    public class KeyboardShortcutHelpTest : AbstractFunctionalTestEx
     {
         [TestMethod, NoParallelTesting(TestExclusionReason.WEB_BROWSER_USE)]
         public void TestKeyboardShortcutHelp()
```
nickshulman, line 35: I wonder if we should have two different [TestMethod]'s in this class: "TestKeyboardShortcutHelp" and "TestDebugKeyboardShortcutHelp" where only the "TestDebug" method sets "TestWebView2EnvironmentDirectory" and is able to assert more things about the HTML content.
The regular "TestKeyboardShortcutHelp" might not be able to assert as many things but would at least be able to verify that nothing bad happens when WebView2 is invoked without all those debug arguments.

brendanx67: Not in this PR. Willing to consider a PR from you later that improves on what I have done.

pwiz_tools/Shared/Common/DataBinding/Controls/Editor/DocumentationViewer.cs
brendanx67: Cursor/Claude left things a bit messy. I moved the DocumentationViewHelper the IDisposable I asked the LLM agent to create out to AbstractFunctionalTestEx fairly late in the process, and clearly didn't move everything I could have to it. After moving a bunch more, I was also able to get rid of the extra debug flags on the WebView2 environment. Much cleaner and simpler now without the complicating need of trying to clean up the test environment in Form_Closed(), and it ran thousands of times without a failure. The test code is now responsible for creating and deleting the environment temp folder, and tries the folder deletion in a WaitForCondition() loop.

pwiz_tools/Shared/Common/DataBinding/Controls/Editor/DocumentationViewer.cs
brendanx67: Sure. I also changed the function name to RunUIAsync to be clearer about what it is doing.

pwiz_tools/Skyline/TestUtil/AbstractFunctionalTestEx.cs
```diff
@@ -922,5 +922,66 @@ public static SrmDocument AddToDocumentFromSpectralLibrary(string libName, strin
...
+        {
+            private readonly string _originalDirectory;
+
+            public DocumentationViewerHelper(TestContext testContext, Action showViewer)
+            {
+                _originalDirectory = DocumentationViewer.TestWebView2EnvironmentDirectory;
+                DocumentationViewer.TestWebView2EnvironmentDirectory = testContext.GetTestResultsPath(@"WebView2");
+                Directory.CreateDirectory(DocumentationViewer.TestWebView2EnvironmentDirectory);
+
+                DocViewer = ShowDialog<DocumentationViewer>(SkylineWindow.ShowKeyboardShortcutsDocumentation);
```
nickshulman: The passed-in Action "showViewer" should be used instead of SkylineWindow.ShowKeyboardShortcutsDocumentation

brendanx67: Oops

PR #3631, opened 2025-10-08
* moved HasMissingDependencies to UI thread

chambm:
* moved HasMissingDependencies to UI thread to test Nick's theory

### DIFF
--- a/pwiz_tools/Skyline/TestFunctional/DiaSearchTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/DiaSearchTest.cs
@@ -101,7 +101,7 @@ namespace pwiz.SkylineTestFunctional
             };
         }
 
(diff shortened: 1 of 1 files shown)


pwiz_tools/Skyline/TestFunctional/DiaSearchTest.cs
```diff
@@ -103,6 +101,8 @@ private void SetupDiaSearchVariableWindows()
             };
         }
 
+        public bool HasMissingDependencies => CallUI(() => !SearchSettingsControl.HasRequiredFilesDownloaded(_testDetails.SearchEngine));
```
nickshulman: This should probably not be a property and instead be the method "bool HasMissingDependencies()" since it has the side effect of updating Settings.Default.SearchToolList.

The Visual Studio debugger tends to automatically evaluate properties in the Locals window and that has the potential to cause confusing behavior.
If you ever cannot change something from a property to a method, but you still want to prevent the debugger from automatically evaluating a property you can mark it with `[DebuggerBrowsable(DebuggerBrowsableState.Never)]`

PR #3636, opened 2025-10-10
Skyline: Convert WebClient use to HttpClient

brendanx67:
- WebClient is based on old technology and is deprecated
- A new wrapper class HttpClientWithProgress is introduced to support the IProgressMonitor interface

### DIFF
--- a/pwiz_tools/Skyline/Test/Test.csproj
+++ b/pwiz_tools/Skyline/Test/Test.csproj
@@ -178,7 +178,6 @@
     <Compile Include="ChromatogramGroupIdsTest.cs" />
     <Compile Include="CmdLineAssociateProteinsTest.cs" />
     <Compile Include="CommandLineThermoMethodTest.cs" />
-    <Compile Include="HelpDocumentationContentTest.cs" />
     <Compile Include="ElibQuantifiableTransitionTest.cs" />
     <Compile Include="EncryptedTokenTest.cs" />
     <Compile Include="ErrorIfPersistentFilesDirModifiedTest.cs" />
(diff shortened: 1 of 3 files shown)


pwiz_tools/Shared/CommonUtil/SystemUtil/HttpClientWithProgress.cs
```diff
@@ -0,0 +1,346 @@
...
+        /// <summary>
+        /// Gets the cancellation token from the progress monitor, or None if no monitor is available.
+        /// </summary>
+        private CancellationToken CancellationToken
+        {
+            get
+            {
+                if (_progressMonitor is SilentProgressMonitor spm)
+                    return spm.CancellationToken;
+                // IProgressMonitor doesn't expose CancellationToken directly; use None and rely on IsCanceled checks
```
nickshulman: There are some places where we have to use an IProgressMonitor but we also want a CancellationToken, we explicitly pass in both the IProgressMonitor and a CancellationToken like in `KoinaModel.PredictBatches`

pwiz_tools/Skyline/Controls/LongWaitDlg.cs
```diff
@@ -189,12 +189,12 @@ public void PerformWork(Control parent, int delayMillis, Action<ILongWaitBroker>
                 }
                 completionEvent?.Dispose();
 
-                if (IsCanceled && null != x)
+                // Treat any OperationCanceledException as cancellation, regardless of whether
+                // the user clicked Cancel or the operation itself threw the exception
+                if (null != x && (x is OperationCanceledException || x.InnerException is OperationCanceledException))
```
nickshulman: I think it would be better to keep this the way it was where the user will see "The operation was cancelled" if they did not actually press the cancel button.
Otherwise, I worry that there may be some weird bugs where the user sees no error message and the thing just fails.

If there's a case where we want it to behave like this I think we should put a catch block in the specify performWork method which cancels the progressMonitor if it caught an OperationCanceledException.

Do you know of a scenario where an OperationCanceledException is getting thrown by someone else that we want to treat as the user canceled?

PR #3651, opened 2025-10-20
Skyline/work/20251019 utf8 no bom

brendanx67:
This work removes a byte order mark (BOM) from many files in our project, fixes our own code that adds these marks and adds a test to the CodeInspectionTest that should fail any file with a BOM within the set of files this test inspects, mostly under the Skyline folder.

Differences between files when only the BOM-v-noBOM changes has long been a pain, but becomes more painful with the introduction of LLM coding tools that seem to prefer noBOM and which is apparently standard in 'nix coding environments. The BOM may be largely going out of style and it is certainly unnecessary, appearing in less than 25% of our files before this change.

This branch also adds a new script scripts/misc/analyze-bom-git.ps1 to analyze the files that have a BOM.


chambm:
Are you sure about making files that VS saves no-BOM? When you explicitly save as UTF-8 in Visaul Studio, it does so with BOM. I don't see an option for no-BOM:
image

chambm:
Oh there it is is, way at the bottom...
image

todos/backlog/TODO-compress_vendor_test_data.md
```diff
@@ -0,0 +1,92 @@
...
+
+3. **Update build system**
+   - Ensure extraction tools available in CI environment
+   - Add extraction step if needed for build-time test data generation
+
+## Benefits
+
+- **Binary integrity**: Compressed archives are explicitly binary, preventing accidental text modifications
+- **Repository size**: Significant space savings (XML compresses extremely well)
+- **Consistency**: Matches how vendor reader DLLs are stored in `pwiz_aux/msrc/utility`
```
chambm, line 75: Vendor DLLs are stored encrypted. I would love to not have to do that because we could update individual DLLs instead of the entire archive. I suppose we could have encrypted each DLL separately, but that would have been pretty awkward.

todos/backlog/TODO-compress_vendor_test_data.md
```diff
@@ -0,0 +1,92 @@
...
+     - Clean up temp directory after test
+   - Handle extraction failures gracefully with clear error messages
+
+3. **Update build system**
+   - Ensure extraction tools available in CI environment
+   - Add extraction step if needed for build-time test data generation
+
+## Benefits
+
+- **Binary integrity**: Compressed archives are explicitly binary, preventing accidental text modifications
```
chambm, line 73: We used to zip (tarball) all vendor test data and their reference mzMLs, and I stopped doing that because it was annoying to update, maintain, and diff. There is already architecture in place to use .tar.bz2 files for this. A lot of the old test data is still maintained in tar.bz2 files, and only the updated reference mzMLs are submitted plaintext. I would definitely want to keep doing that for reference mzMLs. The vendor data itself probably would be ok to put back in tarballs, but I'm not sure it's really necessary or useful except for "modify everything in the repository" scripts like you've done here, which are exceedingly unusual. For git itself, there's a gitattributes file that tells it to treat everything in vendor_readers as binary:
pwiz\data\vendor_readers\.gitattributes

todos/backlog/TODO-compress_vendor_test_data.md
```diff
@@ -0,0 +1,92 @@
...
+- **Repository size**: Significant space savings (XML compresses extremely well)
```
chambm, line 74: There are no large vendor test data XMLs. Only the reference mzMLs are large and I am against going back to zipping those unless there's a way to easily diff stuff within zips.

todos/backlog/TODO-compress_vendor_test_data.md
```diff
@@ -0,0 +1,92 @@
...
+- **Consistency**: Matches how vendor reader DLLs are stored in `pwiz_aux/msrc/utility`
+- **Protection**: Prevents inadvertent BOM removal, line ending changes, or other text transformations
+- **Git performance**: Fewer objects to track, faster clones/pulls
```
chambm, line 77: This is probably a fair point for the directory-based vendor data (not the mzMLs), because those _can_ have tons of small files. But the biggest ones are already stored in tar.bz2 files. I see one Agilent .d, TOFsulfasMS4GHzDualMode+DADSpectra+UVSignal272-NoProfile.d, got committed outside the tarball probably because updating the tarball would mean a 9mb change when the new .d was only 1mb. So if we're going to go back to tarballing all vendor data (and I'm not convinced we should for formats like Bruker TDF with only 2-4 files per .d), it should be per raw file rather than per vendor. That's really the only infrastructure change that would be needed in this TODO, because the existing infrastructure is per vendor.

PR #3659, opened 2025-10-24
- fixed crashes in Wine/Docker container when importing Bruker TIMS

chambm:
- fixed crashes in Wine/Docker container when importing Bruker TIMS data into Skyline (sorting by m/z)

### DIFF
--- a/pwiz_tools/Skyline/Util/Util.cs
+++ b/pwiz_tools/Skyline/Util/Util.cs
@@ -29,7 +29,6 @@ using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using System.Windows.Forms;
-using pwiz.Common.Collections;
 using pwiz.Common.SystemUtil;
 using pwiz.Skyline.Alerts;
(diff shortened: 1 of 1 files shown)


nickshulman:
I think we should try to get rid of the "if (Install.IsRunningOnWine)" and use the same algorithm on regular Windows and Wine just because it would be difficult to debug a wine-only bug.
The amount of time it takes to run BrukerDiaPasefImportTest with these different algorithms are:
Array.Sort: 30s
Enumerable.OrderBy: 32s
SystemLinqExtensionMethods.Sort: 75s

I'm still surprised that sorting has any measurable impact on this scenario. I feel like there must be something we could do with either threading, or taking advantage of the fact that these ion mobility spectra are created by concatenating lists that were already sorted by m/z.

chambm:
OK, I'm not really sure we should anticipate a wine-only bug with the new code, but since it's so close in performance (at least in our limited testing), it may be worth unifying the approach until it's shown to be robust on both Wine and Windows.

PR #3662, opened 2025-10-28
Improved handling of Unicode in file paths

bspratt:
Many external tools, and some vendor reader APIs, can't deal with Unicode characters in file paths. This is often troublesome for users on non-English-language systems. This PR solves that problem by selective use of Windows 8.3 short filename conversion, at least on systems where NTFS 8.3 short filename support is enabled.

e.g. for the mz5 reader, which can't deal with Unicode in paths, we try to use the Windows API to change
   "C:\Program Files\Common Files\my files with ünicode\foo.mz5"
(note the umlaut U) to
   "C:\Program Files\Common Files\MYFILE~1\foo.mz5"
before passing the path into the reader API.

This only works on NTFS volumes, and only those with 8.3 support enabled. So, not on Docker instances, or under Wine, or on NTFS volumes with 8.3 short filename support disabled.

Conversions are only attempted where required - any external tool or vendor reader that is known to handle Unicode paths won't get this treatment applied to any file paths we hand to it. That would be unnecessarily confusing.

We already have code for testing Skyline with Unicode paths, and a test attribute "NoUnicodeTesting" to skip that for tests of Skyline features known to have trouble with Unicode - usually because they spin up processes for tools that can't deal with Unicode paths. With this PR that test attribute is not currently in use - there are no tests that fail when Unicode paths are involved. The Skyline code now converts file paths to 8.3 as required by each external tool before launch. The pwiz wrappers for vendor readers do 8.3 conversion where needed, too.

Obviously we don't test Unicode path handling in environments where we know 8.3 conversion isn't available.

And obviously a user running Skyline on a system without 8.3 short filename support is still going to experience difficulty with Unicode paths as before. But this should help in the majority of cases.

### DIFF
--- a/pwiz.vcxproj
+++ b/pwiz.vcxproj
@@ -580,6 +580,7 @@
       <PreprocessorDefinitions Condition="'$(Configuration)|$(Platform)'=='BrowseOnly|Win32'">PWIZ_READER_THERMO;PWIZ_READER_AGILENT;PWIZ_READER_BRUKER;PWIZ_READER_WATERS;PWIZ_READER_ABI;PWIZ_READER_ABI_T2D;WIN32;USE_RAW_PTR;PWIZ_BINDINGS_CLI_COMBINED;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <PreprocessorDefinitions Condition="'$(Configuration)|$(Platform)'=='BrowseOnly|x64'">PWIZ_READER_THERMO;PWIZ_READER_AGILENT;PWIZ_READER_BRUKER;PWIZ_READER_WATERS;PWIZ_READER_ABI;PWIZ_READER_ABI_T2D;WIN32;USE_RAW_PTR;PWIZ_BINDINGS_CLI_COMBINED;%(PreprocessorDefinitions)</PreprocessorDefinitions>
     </ClCompile>
+    <ClCompile Include="pwiz\utility\bindings\CLI\common\util.cpp" />
     <ClCompile Include="pwiz\utility\bindings\CLI\Diff.cpp" />
     <ClCompile Include="pwiz\utility\bindings\CLI\DiffTest.cpp" />
     <ClCompile Include="pwiz\utility\bindings\CLI\MSData.cpp" />
@@ -1035,6 +1036,7 @@
     <ClInclude Include="pwiz\utility\bindings\CLI\common\SharedCLI.hpp" />
     <ClInclude Include="pwiz\utility\bindings\CLI\common\Unimod.hpp" />
     <ClInclude Include="pwiz\utility\bindings\CLI\common\unit.hpp" />
+    <ClInclude Include="pwiz\utility\bindings\CLI\common\util.hpp" />
     <ClInclude Include="pwiz\utility\bindings\CLI\common\vector.hpp" />
     <ClInclude Include="pwiz\utility\bindings\CLI\common\virtual_map.hpp" />
     <ClInclude Include="pwiz\utility\bindings\CLI\msdata\Diff.hpp" />
--- a/pwiz.vcxproj.filters
+++ b/pwiz.vcxproj.filters
@@ -322,6 +322,7 @@
     <ClCompile Include="pwiz\utility\bindings\CLI\common\SharedCLITest.cpp" />
     <ClCompile Include="pwiz\utility\bindings\CLI\common\Unimod.cpp" />
     <ClCompile Include="pwiz\utility\bindings\CLI\common\UnimodTest.cpp" />
+    <ClCompile Include="pwiz\utility\bindings\CLI\common\util.cpp" />
     <ClCompile Include="pwiz\utility\bindings\CLI\Diff.cpp" />
     <ClCompile Include="pwiz\utility\bindings\CLI\DiffTest.cpp" />
     <ClCompile Include="pwiz\utility\bindings\CLI\MSData.cpp" />
@@ -1041,6 +1042,7 @@
     <ClInclude Include="pwiz\analysis\dia_umpire\PeakCurve.hpp" />
     <ClInclude Include="pwiz\analysis\dia_umpire\ScanData.hpp" />
     <ClInclude Include="pwiz\analysis\spectrum_processing\SpectrumList_DiaUmpire.hpp" />
+    <ClInclude Include="pwiz\utility\bindings\CLI\common\util.hpp" />
     <ClInclude Include="pwiz\utility\bindings\CLI\timstof_prm_scheduler\prmscheduler.h" />
     <ClInclude Include="pwiz\utility\bindings\CLI\timstof_prm_scheduler\PrmScheduler.hpp" />
     <ClInclude Include="pwiz\utility\misc\sort_together.hpp" />
--- a/pwiz/utility/bindings/CLI/Jamfile.jam
+++ b/pwiz/utility/bindings/CLI/Jamfile.jam
@@ -117,6 +117,7 @@ local clr-requirement = <conditional>@msvc-requirement <define>UNICODE <define>_
 
 obj ParamTypes        : common/ParamTypes.cpp        : $(data_libs) $(clr-requirement) <cxxflags>"/doc$(DOC_DATA_PATH)/ParamTypes.xdc" ;
 obj IterationListener : common/IterationListener.cpp : $(data_libs) $(clr-requirement) <cxxflags>"/doc$(DOC_DATA_PATH)/IterationListener.xdc" ;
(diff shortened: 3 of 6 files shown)


pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
```diff
@@ -376,6 +377,38 @@ public static string GetRandomFileName()
             return result;
         }
 
+        /// <summary>
+        /// Like PathEx.GetRandomFileName(), but tries to ensure that no unicode characters are in the name
+        /// </summary>
+        /// <returns>a random filename containing only ASCII characters</returns>
+        public static string GetNonUnicodeRandomFileName()
```
chambm: This function seems odd to me. Couldn't we just give GetRandomFileName() an optional argument like `bool addUnicodeTestString` or something and then this function could call that with `false`, or just whatever is going to use this function would directly call GetRandomFileName like that?

pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
```diff
@@ -394,6 +427,58 @@ public static string GetTempFileNameWithExtension(string ext)
...
+        /// We often encounter tools that can't deal with Unicode characters in file paths, this method
+        /// will try to convert such paths to a non-Unicode version using the 8.3 format short path name.
+        /// Converts only the segments that need it, to avoid trashing filename extensions.
+        /// e.g. "C:\Program Files\Common Files\my files with ünicode\foo.mzml" (note the umlaut U) => ""C:\Program Files\Common Files\MYFILE~1\foo.mzml"
+        ///
+        /// Only works on NTFS volumes, with 8.3 support enabled. So, for example, not on Docker instances.
+        /// </summary>
+        /// <param name="path">Path to an existing file</param>
+        /// <returns>Path with unicode segments changed to 8.3 representation, if possible</returns>    
+        public static string GetNonUnicodePath(string path)
```
chambm, line 414: Could we add a CLI binding for the C++ function and call that here? That way the code isn't duplicated.

pwiz_tools/Skyline/Model/DiaUmpireDdaConverter.cs
```diff
@@ -143,16 +143,18 @@ public override bool Run(IProgressMonitor progressMonitor, IProgressStatus statu
                     if (File.Exists(outputFilepath))
                         FileEx.SafeDelete(outputFilepath);
 
-                    string tmpFilepath = Path.Combine(Path.GetTempPath(), PathEx.GetRandomFileName() + MsConvertOutputExtension); // N.B. FileEx.GetRandomFileName adds unusual characters in test mode
-                    string tmpParams = Path.Combine(Path.GetDirectoryName(outputFilepath) ?? string.Empty, @$"diaumpire-{DateTime.Now.ToString(@"yyyyMMddhhmm")}.params");
+                    // DiaUmpire doesn't handle unicode paths
```
chambm: It should, actually. It's all pwiz C++ code. Did you actually run into problems with it or it just got caught up in a "uses ProcessRunner" search-and-replace?

pwiz_tools/Skyline/TestFunctional/DdaSearchTest.cs
```diff
@@ -790,7 +789,7 @@ private void TestDependencyErrors(ImportPeptideSearchDlg importPeptideSearchDlg)
             searchSucceeded = null;
             editToolDlg = ShowDialog<EditSearchToolDlg>(() => importPeptideSearchDlg.ClickNextButton());
             RunUI(() => editToolDlg.ToolPath = editToolDlg.ToolPath.Replace(".42.exe", ""));
-            editToolDlg.OkDialog();
+            RunUI(editToolDlg.OkDialog); // Purposely using RunUI instead of OkDialog here
```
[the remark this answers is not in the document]
chambm: Hah, I was thinking the same thing as Copilot. Does that mean I'm artificially intelligent? :(

pwiz_tools/Skyline/Model/DiaUmpireDdaConverter.cs
bspratt: I only made changes where tests were failing with Unicode decoration. Possibly something has changed with DiaUmpire since I started messing with this idea last spring? Or I just got confused - I'll check again.

pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
bspratt: Somewhat more complicated implementation, but a cleaner final result. Good idea.

pwiz_tools/Skyline/Model/DiaUmpireDdaConverter.cs
chambm: Maybe it's using mz5 in the backend and that failed. But it should work now with your mz5 changes, right?

pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
bspratt: I'll have a look, but to my mind the use of an argument instead of a separate well named function makes for less intuitive code.

chambm: It's fine to keep a separate function, but on second thought, just call `Path.GetRandomFileName()` since that would be exactly the same as giving PathEx one an optional argument. It's pretty well understood I would think that `Path.GetRandomFileName()` won't include any non-ASCII characters.

chambm: And "tries to ensure" in the comment is probably wrong. It should certainly not have any non-ASCII characters in it (according to your implementation). Using `Path.GetRandomFileName()` I suppose it's not guaranteed (looking at the docs) but it's implemented with `Path.ToBase32StringSuitableForDirName` so it's definitely just ASCII.

pwiz_tools/Skyline/Model/DiaUmpireDdaConverter.cs
bspratt: > Maybe it's using mz5 in the backend and that failed. But it should work now with your mz5 changes, right?

The thing with the NTFS 8.3 conversion is that it only works on paths that exist. That 8.3 name is determined at creation time, and it's not guaranteed to give the same result twice even if the long path+file name is identical. 

Do you know if the mz5 code will simply overwrite the output file without first deleting it? If so this could be made to work if the mz5 connection code first creates an empty file then passes its 8.3 name to the mz5 library.

chambm: It looks like it's configurable in the code, but currently it always uses Connection_mz5::RemoveAndCreate => H5F_ACC_TRUNC, which truncates the file if it exists. So it should be possible to create the empty file with (potential) Unicode characters, then convert it to 8.3 before passing it to H5File.

bspratt: Excellent, I'll try it.

bspratt: With your help I was able to back out this change - Connection_mz5 now deals with the 8.3 conversion directly, including creating an empty mz5 file if needed to enable the 8.3 conversion before writing.

pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
bspratt: hah, the whole thing turns out to be cruft, deleting it

pwiz_tools/Shared/CommonUtil/SystemUtil/PathEx.cs
bspratt: What's your sense of where this should be implemented? Would you expect to find it in  pwiz\utility\bindings\CLI\common\SharedCLI.hpp ?

chambm: Yeah, that's fine for now since we don't have a Filesystem.hpp bindings file (because this is the first time we've needed the same function in C++ and C#). If we ever add one it's easy to move it over.

bspratt: After spending some time on this I have realized that it's the wrong move - there are a number of places in our C# code where this capability is useful that do not currently depend on the pwiz CLI. This is an instance where code duplication is actually justified.

chambm: Oh yes, pulling in the full pwiz_data_CLI.dll into CommonUtil definitely seem overkill for this. It would be possible to build a separate (and much smaller) bindings DLL just for simple functions like this, but I'll take care of that if another "DRY" case comes up.

PR #3669, opened 2025-11-08
Skyline: Phase 3 clean-up of WebClient/HttpWebRequest use in favor of HttpClientWithProgress

brendanx67:
- initial work to remove unnecessary dependencies on WebException where none can be thrown directly
- HttpClient does not throw WebException, though for DNS errors it throws an HttpRequestException with a WebException as the InnerException
- PanoramaClient now wraps HttpClientWithProgress which only throws NetworkRequestException

All tests pass for Skyline - Panorama tests, SkylineBatch, and AutoQC

### DIFF
--- a/ai/todos/backlog/TODO-httpclient_to_progress_continued.md
+++ b/ai/todos/backlog/TODO-httpclient_to_progress_continued.md
@@ -1,0 +1,200 @@
+# TODO-httpclient_to_progress_continued.md
+
+## Branch Information
+- **Branch**: `Skyline/work/YYYYMMDD_httpclient_to_progress_continued` (to be created when work starts)
+- **Status**: Backlog
+- **Objective**: Continue consolidating Skyline HTTP usage onto `HttpClientWithProgress`, migrating remaining `HttpClient` and `HttpWebRequest` usage in Ardia integration and other areas
+
+## Background
+
+This TODO continues the work from `TODO-20251107_httpclient_to_progress.md`, which successfully migrated `WebEnabledFastaImporter` to use `HttpClientWithProgress` with comprehensive HTTP recording/playback infrastructure. This follow-up work focuses on:
+
+- **Ardia Integration**: Migrating `EditRemoteAccountDlg` and `ArdiaLoginDlg` from bare `HttpClient` to `HttpClientWithProgress`
+- **HttpWebRequest Retirement**: Replacing remaining `HttpWebRequest` usage throughout the codebase
+- **Code Inspection**: Adding automated detection of bare `HttpClient` and `HttpWebRequest` usage
+- **Connected Tests Organization**: Comprehensive review and proper organization of tests requiring network access
+
+## Prerequisites
+- ✅ `HttpClientWithProgress` complete with upload/download/cookie/header support
+- ✅ Comprehensive testing infrastructure (`HttpClientTestHelper`) with recording/playback
+- ✅ PanoramaClient migration complete (establishes cookie/session patterns)
+- ✅ WebEnabledFastaImporter migration complete (establishes HTTP recording/playback pattern)
+
+## Scope
+
+### In-Scope: Core Skyline.exe
+1. **`ToolsUI/EditRemoteAccountDlg.cs`** - Ardia API session management
+   - Uses `HttpClient` with `CookieContainer` for session cookies
+   - Similar to PanoramaClient pattern
+   - Should use `HttpClientWithProgress` for consistency
+
+2. **`Alerts/ArdiaLoginDlg.cs`** - Ardia OAuth authentication
+   - Uses bare `HttpClient` for OAuth flows
+   - **Contains async/await keywords** (see `TODO-remove_async_and_await.md`)
+   - OAuth library integration may require special handling
+   - Consider: Can we refactor to remove async/await while migrating to HttpClientWithProgress?
+
+3. **Remaining `HttpWebRequest` usage** - Throughout codebase
+   - Replace with `HttpClientWithProgress` where applicable
+   - Document acceptable usage (e.g., status codes, header enums only)
+
+4. **Connected Tests Organization** - Comprehensive review
+   - Identify all tests requiring network access
+   - Move appropriate tests to `TestConnected` project
+   - Fix TeamCity configuration for connected test runs
+
+### Out-of-Scope (Separate TODOs)
+- **SkylineTester** - Build/test infrastructure (separate from product)
+- **ScreenshotPreviewForm** - Test utilities only
+- **Executables/DevTools** - Developer tools (see `TODO-tools_webclient_replacement.md`)
+
+## Related Work
+
+**This TODO continues:**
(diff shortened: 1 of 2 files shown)


nickshulman (approved):
Looks good to me.

pwiz_tools/Skyline/CommonTest/FastaImporterTest.cs
```diff
@@ -893,78 +1199,408 @@ public void DoTestFastaImport(bool useActualWebAccess, bool doNegTests) // call
...
+            WebEnabledFastaImporter.WebSearchProvider provider = null;
+            if (mode != LookupTestMode.normal)
+            {
+                provider = new QuickFailWebSearchProvider();
+            }
+            else if (useFastProvider)
+            {
+                provider = new FastWebSearchProvider(); // Skip politeness delays during playback
+            }
+            // TODO(Claude): Test progress monitor cancellation by linking a CancellationToken to HttpClientTestBehavior with a ResponseFactory function that triggers cancellation.
```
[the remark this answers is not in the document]
brendanx67: Changed to CONSIDER(brendanx). Nice to have but not high enough priority right now

pwiz_tools/Shared/ProteomeDb/Fasta/WebEnabledFastaImporter.cs
```diff
@@ -493,6 +624,21 @@ public FastaRegExSearchtermPair(string comment, string regex, string webSearchte
         /// </summary>
         public class WebSearchProvider 
         {
+            private const string SKYLINE_USER_AGENT = @"Skyline";
+
+            protected virtual HttpClientWithProgress CreateHttpClient(IProgressMonitor progressMonitor, int timeout)
+            {
+                // Progress status for this request is ignored. The progress monitor is passed only
+                // to support cancellation. Progress for the langer operation that is making the
```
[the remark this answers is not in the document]
brendanx67: typo fixed

pwiz_tools/Shared/ProteomeDb/Fasta/WebEnabledFastaImporter.cs
```diff
@@ -128,6 +217,47 @@ public static ProteinSearchInfo Intersection(IEnumerable<ProteinSearchInfo> list
             }
             return result;
         }
+
+        public override string ToString()
+        {
+            var name = PreferredName ?? ProteinDbInfo?.PreferredName ?? string.Empty;
+            name = Accession ?? ProteinDbInfo?.Accession ?? string.Empty;
```
bspratt: Seems like this is always going to stomp the initial assignment of "name"?

pwiz_tools/Shared/ProteomeDb/Fasta/WebEnabledFastaImporter.cs
```diff
@@ -929,13 +1101,367 @@ private static string NullForEmpty(string strIn)
...
+                    return null;
+                }
+                return bracketed;
+            }
+            return null;
+        }
+
+        private static string BuildUniprotSearchTerm(string accession, string taxId)
+        {
+            if (string.IsNullOrEmpty(accession))
```
bspratt, line 1287: This feels off - is that not meant to be if not null or empty?

pwiz_tools/Shared/ProteomeDb/Fasta/WebEnabledFastaImporter.cs
```diff
@@ -929,13 +1101,367 @@ private static string NullForEmpty(string strIn)
...
+            return $@"{accession} AND (organism_id:{taxId})";
+        }
+
+        private void ProcessEntrezSummaryResult(List<string> searchterms, List<ProteinSearchInfo> responses,
+            ref ProteinSearchInfo response, string id, string caption, string replacedBy, string title,
+            string taxId, string length, string urlString)
+        {
+            // can we transfer this search to UniprotKB? Gets us the proper accession ID,
+            // and avoids downloading sequence data we already have or just don't want
+            string newSearchTerm = null;
```
bspratt, line 1303: Hard question to answer - the UniProt API moves around with some frequency, I hate to disturb an implementation that already works. Also note that UniProt seems to have won the standards battle so in pratice we don't see a lot of data that use these other APIs. I feel like it ain't broke, so don't fix it.

pwiz_tools/Shared/ProteomeDb/Fasta/WebEnabledFastaImporter.cs
bspratt, line 1: Pretty massive refactor. Seems to throw away a fair number of comments that provide valuable insight on why the code was the way it was. Looking at the comments that are in there, many don't follow our convention of capitalizing the first word in the comment so I'm guessing this was at least in part AI assisted? This one makes me nervous.

bspratt (changes requested):
A few places worth looking at, as already noted

pwiz_tools/Shared/ProteomeDb/Fasta/WebEnabledFastaImporter.cs
brendanx67: Good catch. This is for debugging only. I think I ended up preferring to see the Accession in the debugger, since it ended up being more critical to the Uniprot requests, but I am fine with the way Cursor chained these values to use PefferredName first. The next person debugging through the code can tweak this output to what works best for them.

pwiz_tools/Shared/ProteomeDb/Fasta/WebEnabledFastaImporter.cs
brendanx67: Added a comment and made this condition return null to clarify that it is not possible to build a UniprotSearchTerm without an accession. Good catch.

pwiz_tools/Shared/ProteomeDb/Fasta/WebEnabledFastaImporter.cs
brendanx67: Okay. Question came from LLM that was helping working on the code.

pwiz_tools/Shared/ProteomeDb/Fasta/WebEnabledFastaImporter.cs
brendanx67: Good call. Biggest missing comment was the Etrez XML explanation. I did a pass to clean up comment capitalization. I have added new comment guidelines to our .md files under ai.

The existing code made me nervous, and it didn't seem like it was getting well tested with the Web version disabled and the recorded version out of date. It was a bit challenging from a maintenance perspective to make it use HttpClientWithProgress and improve its overall testing. I discovered issues as I tried to increase the amount of validation it was doing and reduce the amount of code that it skipped when the internet was not available. Long-term, I think this makes the code more maintainable and more likely to trigger a test failure when the websites it queries break it.

This was also a nice proving ground for developing recordind and playback for HttpClientWithProgress that should benefit all HTTP network use in the project, which is a growing area with Ardia, waters_connect, Koina, Panorama, etc.

Now we have central place that all of that communication can flow through with record and playback support for disconnected testing. This code felt old and brittle and not really current best practices.

PR #3672, opened 2025-11-18
Skyline/work/20251116 vs2026 support

bspratt:
Adding support for building with VS2026


bspratt:
Doesn't break VS2022 but there's some kind of problem with Skyline and pwiz_data_cli. Hoping @chambm will have a look.

PR #3673, opened 2025-11-18
NIST (.msp) format reader now handles n and c terminal notation e.g. "Name: n[43]ALAVLALLSLSGLEAIQR/5" or "Name: AACDEFGHIKc[17]/3"

bspratt:
In practice we just remove "n" and/or "c" before parsing and Skyline applies the modification to  the AA at the appropriate end.

### DIFF
--- a/ai/docs/build-and-test-guide.md
+++ b/ai/docs/build-and-test-guide.md
@@ -8,46 +8,90 @@ Detailed reference for building, testing, and analyzing Skyline from LLM-assiste
 
 - Visual Studio 2022 Community/Professional installed
 - Initial full build completed with `bs.bat` (Boost.Build + native dependencies)
-- Working directory: `C:\proj\pwiz\pwiz_tools\Skyline`
+
+## ⚠️ CRITICAL: Never Call MSBuild Directly
+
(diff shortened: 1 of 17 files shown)


brendanx67:
Those mods should get translated to A[+43.0] and K[+17.0] in our nomenclature. We don't distinguish between mods on the terminus and mods on the terminal AAs. This we inherited from Ron Beavis and X! Tandem.

brendanx67:
No. I am pretty sure there is code in BlibBuild that changes that format to the one I mentioned. Any code in our repository that produces the format you mention is in error and creating bugs.

bspratt:
OK, I'll have to make it a bit more sophisticated that just removing the "n", no problem.

brendanx67:
I have realized that what we already accept from a NIST .msp or Spectrast .sptxt file is entirely source-specific. They are not required to match our expectations internally. So, if their format is K[17], then you probably just need that. And, in fact, if before we had implemented support for [43]A, you would indeed only need to remove the n for the previous fix to convert that to A[43]. Internally to Skyline, they will be converted to K[+17] and A[+43].

PR #3674, opened 2025-11-19
Skip peptides which are missing particular feature scores when training mProphet model

nickshulman:
This change allows you to train a model where some of the peptides might be missing particular feature scores.
image
The peptides with missing scores will show up as the little "unknown" bit in the bar graph.
If there are any missing combined scores then the binoculars button is shown.
Clicking that button will make the peptides with missing scores show up in the Find window where it will display the comma-separated list of scores that each peptide is missing.
image

Another change is that the binoculars button on the Feature Scores tab show up whenever they are missing scores instead of only showing up when you mouse over where the binoculars would be.


brendanx67:
Maybe the choices should be:
Exclude
Impute
Ignore

I feel like "Fail" implies that the entire model will fail, when it really just means the score gets excluded for lack of complete data. Doing anything to complete the matrix (even assigning zero) would be imputing a score. For the "ignore" case, that can still give you a usable model, but I am not clear how that peak gets scored, then. Does it end up with #N/A for its score and q value? q value zero? If you don't have an imputed value, you can't actually give it a score.

PR #3676, opened 2025-11-21
Mark document dirty when new files are imported

nickshulman:
Changed to mark document dirty when Skyline finishes importing results (not reported by anyone)

When Skyline finishes importing chromatograms from raw files, Skyline marks the document as dirty. This does not add anything new to undo stack-- it just increments the document version so Skyline will later prompt you to save it.


brendanx67:
This seems like a good idea, of course. I guess the case in question would be a file saved after the new results are added but before they have completely imported. Otherwise, the addition of the results already made the document dirty.

PR #3689, opened 2025-12-01
Avoid potential test deadlocks in SkylineInvoke() by not calling Invoke() if already on the UI thread. 

bspratt:
This can occur in nested calls to SkylineInvoke(). Nested calls to SkylineBeginInvoke() seem like a more serious misconstruction of a test, so these throw an exception.

### DIFF
--- a/pwiz_tools/Skyline/TestUtil/TestFunctional.cs
+++ b/pwiz_tools/Skyline/TestUtil/TestFunctional.cs
@@ -382,32 +382,14 @@ namespace pwiz.SkylineTestUtil
 
         private static void SkylineInvoke(Action act)
         {
-            if (null != SkylineWindow)
+            var form = (SkylineWindow as FormEx) ?? FindOpenForm<StartPage>();
+            if (form.InvokeRequired)
             {
-
-                if (SkylineWindow.InvokeRequired)
-                {
-                    SkylineWindow.Invoke(act);
-                }
-                else
-                {
-                    Console.WriteLine(@"# SkylineInvoke() called on UI thread (SkylineWindow)"); // CONSIDER:(bspratt) remove before merge?
-                    act();
-                }
+                form.Invoke(act);
             }
             else
             {
-                var startPage = FindOpenForm<StartPage>();
-
-                if (startPage.InvokeRequired)
-                {
-                    startPage.Invoke(act);
-                }
-                else
-                {
-                    Console.WriteLine(@"# SkylineInvoke() called on UI thread (StartPage)"); // CONSIDER:(bspratt) remove before merge?
-                    act();
-                }
+                act(); // Already on the UI thread
             }
         }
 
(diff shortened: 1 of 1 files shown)


nickshulman:
Which test is calling SkylineInvoke from the event thread? Should we just change it to fail the test immediately if that happens?

bspratt:
Well, I was trying to track down some hangs in tests, and noticed in one instance that after attaching a debugger it was in a nested call to RunUI(). I thought was was weird, so I asked ChatGPT "are nested calls to invoke OK?". The response:

-----------------

Short answer: Yes — nested Task.Invoke() (or Invoke-style) calls work, but they are dangerous and almost always a bad idea unless you know exactly what you're doing.
They can easily lead to deadlocks, re-entrancy bugs, and stack blowups, depending on the framework and context.

Because you didn’t specify the platform, here are the common cases:
🧵 If you mean WinForms: Control.Invoke() / BeginInvoke()
✔ Allowed?
Yes, WinForms allows nested Invoke calls.
⚠ But: re-entrant calls & deadlocks are very common
Example of a deadlock:
lock (someLock) {
    control.Invoke(() => {
        lock (someLock) { /* deadlock */ }
    });
}
Example of accidental re-entrancy:
control.Invoke(() =>
{
    // This runs on the UI thread
    // If you block inside this, calling Invoke again causes freeze
    control.Invoke(...); // UI thread waiting on itself → deadlock
});

Rule of thumb:
Use BeginInvoke if calling from UI thread.
Never call Invoke from the UI thread (it blocks).
Avoid Invoke chaining; prefer posting messages via SynchronizationContext.Post.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -384,23 +384,51 @@ private static void SkylineInvoke(Action act)
         {
             if (null != SkylineWindow)
             {
-                SkylineWindow.Invoke(act);
+
```
brendanx67: I guess I would prefer a function to code duplication. That would keep this block a one-liner. SkyineInvoke(SkylineWindow, act);

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -384,23 +384,51 @@ private static void SkylineInvoke(Action act)
...
+                    act();
+                }
             }
         }
 
         private static void SkylineBeginInvoke(Action act)
         {
             if (null != SkylineWindow)
             {
+                if (!SkylineWindow.InvokeRequired)
```
brendanx67: Function prefered to keep the code in each block one line.

bspratt:
> P.S. - ChatGPT doesn't seem to think this could be responsible for a deadlock. 

Interesting, since ChatGPT thought it was dodgy when I asked. At any rate, it can't hurt, and when I had the Console.Writeline in there to flag nested Invoke() it showed that a number of our flakier tests were doing that.

pwiz_tools/Skyline/TestUtil/TestFunctional.cs
bspratt: OK, I can easily simplify.

PR #3695, opened 2025-12-04
Add "Protein Abundances" and "Peptide Normalized Areas" default reports.

nickshulman:
This change adds two new reports: "Protein Abundances" and "Peptide Normalized Areas".
Both of these reports are "Pivot by replicate name".

Here's what the Protein Abundances report looks like:
image

Here's what the Peptide Normalized Areas report looks like:
image

The Protein Abundances report shows the "Protein Abundance" column, and the Peptide Normalized Areas report shows the "Normalized Area" column. Both of these columns sometimes have the red exclamation point and an asterisk when the value has missing or truncated transitions. When you export the report to CSV, the asterisk is still there but the red exclamation mark is not written to the CSV.


brendanx67 (approved):
Looks good. Nice to add these simple quantitative reports. I think people have found it somewhat strange that Skyline didn't offer anything like this by default. Hope they are the right reports and become as widely used outside the MacCoss lab as it sounds like they are inside.

PR #3700, opened 2025-12-05
Allow exporting reports beyond what can be seen in the Document Grid

nickshulman:
Currently, the "File > Export > Report" dialog (as well as command-line export etc) only allows you to export custom reports that you can see in the Document Grid. This does not include the audit log, group comparisons, the Results Grid or the Candidate Peaks window.

Will Thompson's team wanted to be able to export the Candidate Peaks for all of the molecules in the document.

This change also eliminates the last of the usages of "UI" classes from the model namespaces.


nickshulman:
Subsumed by PR #3968

PR #3706, opened 2025-12-10
Add some extra status output to help track down a crash in NextFlow

nickshulman:
A new docker image will be made from the TeamCity artifacts for this pull request and that will help track down a crash that seems to be happening. (Most likely cause of the crash is that it's running out of memory, but even if that's the cause we would like to eventually make it so Skyline outputs a better reason. Current behavior is that the last thing Skyline writes to the console before it dies is "Updating peak statistics")


nickshulman:
We figured out that the problems have something to do with symbolic links and this is being addressed in PR #4175

PR #3709, opened 2025-12-11
Tweak jamfiles to reduce compiler warnings in C and C++ code that we don't maintain

bspratt:
Reduce compiler warnings in C and C++ code that we don't maintain.

Most of this is just disabling warnings we were already disabling in C++ code.

### DIFF
--- a/pwiz_tools/Skyline/TestUtil/TestFunctional.cs
+++ b/pwiz_tools/Skyline/TestUtil/TestFunctional.cs
@@ -1183,27 +1183,13 @@ namespace pwiz.SkylineTestUtil
         public static bool WaitForConditionUI(int millis, Func<bool> func, Func<string> timeoutMessage = null, bool failOnTimeout = true, bool throwOnProgramException = true)
         {
             int waitCycles = GetWaitCycles(millis);
-            var extraNote = string.Empty;
             for (int i = 0; i < waitCycles; i++)
             {
                 if (throwOnProgramException)
                     Assert.IsFalse(Program.TestExceptions.Any(), "Exception while running test");
 
                 bool isCondition = false;
-
-                // Avoid "System.InvalidOperationException: Invoke or BeginInvoke cannot be called on a control until the window handle has been created."
-                if (Program.MainWindow.IsHandleCreated && !Program.MainWindow.IsDisposed)
-                {
-                    Program.MainWindow.Invoke(new Action(() => isCondition = func()));
-                    extraNote = string.Empty;
-                }
-                else
-                {
-                    extraNote = Program.MainWindow.IsDisposed ?
-                        @" Program.MainWindow is disposed,":
-                        @" Program.MainWindow handle not created,";
-                }
-
+                Program.MainWindow.Invoke(new Action(() => isCondition = func()));
                 if (isCondition)
                     return true;
                 Thread.Sleep(SLEEP_INTERVAL);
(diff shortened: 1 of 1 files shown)


pwiz_tools/Skyline/TestUtil/TestFunctional.cs
```diff
@@ -1183,13 +1183,27 @@ public static bool TryWaitForConditionUI(int millis, Func<bool> func)
...
             for (int i = 0; i < waitCycles; i++)
             {
                 if (throwOnProgramException)
                     Assert.IsFalse(Program.TestExceptions.Any(), "Exception while running test");
 
                 bool isCondition = false;
-                Program.MainWindow.Invoke(new Action(() => isCondition = func()));
+
+                // Avoid "System.InvalidOperationException: Invoke or BeginInvoke cannot be called on a control until the window handle has been created."
+                if (Program.MainWindow.IsHandleCreated && !Program.MainWindow.IsDisposed)
```
nickshulman: What are you trying to protect against here? Wouldn't the "InvalidOperationException" be a clear enough exception if someone called this before the SkylineWindow was created or after it was closed? Does any test actually do that? Is the idea that this method should wait if the SkylineWindow has not yet been created?

I was seeing something on my computer where it would get stuck waiting for the "Program.MainWindow.Invoke" to finish. The action that it was supposed to run never got executed as if the message sent to the event thread never happened. You would not happen to be seeing a failure like that?

bspratt: I encountered a situation where the test was getting out in front of the Skyline window being ready, with an early call to WaitForConditionUI().

nickshulman: Which test was that? Usually that would be solved by adding `WaitForOpenForm<SkylineWindow>()`.

bspratt: I don't actually remember which test it was. But this has the same effect as WaitForOpenForm<SkylineWIndow>() without having to comb through all out tests looking for places where this might be an issue.

nickshulman: It has to be one of the tests that overrides "ShowStartPage", right? Nearly all of those already have a `WaitForOpenForm<SkylineWindow>()` so I'm pretty stumped as to which one it could be that failed for you.

bspratt: What does it matter? The problem was noticed on a machine that has a lot of flaky behavior which apparently gives rise to race conditions we hadn't noticed before. This change fixes at least one test and potentially many more (our functional tests pretty much all assume that the Skyline window is immediately good to go), without changing the semantics of WaitForConditonUI().  I don't understand the objection.

nickshulman: If Program.MainWindow.IsHandleCreated is false, then I am pretty sure it's never going to be created and we should just throw immediately.

The SkylineWindow handle gets created in the SkylineWindow constructor (it happens as a side effect of creating the SequenceTree). So, by the time that Program.MainWindow is not null, Program.MainWindow.IsHandleCreated is already true.

So, if Program.MainWindow.IsHandleCreated is false, it must mean that the SkylineWindow has been destroyed. We would not expect Program.MainWindow.IsHandleCreated to become true after that.

There might be some way for SkylineWindow to behave differently from the way I think it does.
If you know of a test where it's possible to wait for Program.MainWindow.IsHandleCreated to go from false to true I would really like to understand what is going on.

bspratt: Very well, I'll back out that change for now and try to get a solid repro. The reduced compiler noise is nice all on its own.

PR #3712, opened 2025-12-12
Fix synchronized zoom with multiple panes and/or inactive tabs

nickshulman:
Fixed y-axis of synchronized zooming with split graph (reported by Alex)
Fixed synchronized zooming of replicates which are not currently displayed (reported by Simon)


brendanx67:
You should test tutorial screenshots with synchronized zooming on this branch to make sure they don't change.

PR #3717, opened 2025-12-17
Fix all replicates have same q-value in .blib from DIA-NN results

nickshulman:
All of the rows in the Retention Times table for a given peptide have the same "Score" value in a library built from DIA-NN results.
This change makes it so that value from the "Q.Value" column instead of "Global.Q.Value" gets stored in the "RetentionTimes" table.
I don't think the fix in this PR is quite right, since I think we want to still store the Global.Q.Value value in the "RefSpectra" table, and I was not sure what to do about the `// update bestQValue and bestPSM if this row is better than any previous one` code.


brendanx67:
I hate trying to make this change without Matt around, but it does sound broken. My expectation would be that the global q value gets stored in the "run" that represents the canonical predicted library spectra, and the replicates with true file-ids get their run-specific q values. I have not read this code. So, if you don't see this separation of canonical and measured spectra, I may have to look more closely, but this is my understanding of what we settled on based on feedback from Mike MacCoss. It is important that when Skyline constructs its target list it uses both canonical spectra and overall q values, so that a user can filter for overall detected peptide precursors and their canonical spectra for choosing the peptides, precursors, and transitions that go into the targets view.

pwiz_tools/BiblioSpec/src/DiaNNSpecLibReader.cpp
```diff
@@ -1222,7 +1223,7 @@ bool DiaNNSpecLibReader::parseFile()
             psm->fileId = retentionTimes.back().fileId;
```
chambm, line 1222: This would need to change to globalQValue as well to keep the logic the same (or the if statement above needs to keep using redundantPSM.score).

brendanx67:
Note that we worked out for Proteome Discoverer that global q value works
best for the overall q value cut-off used to decide what goes into the
spectral library. Spectronaut was getting much better results that Skyline
for a particular dataset and it came down to the starting spectral library.
When we used theirs we got the same results, and that turned out to be due
to our keeping peptides based on local q values and their using an
experiment-wide q value.

We need to be careful not to end up preserving low quality data by
switching to local q values. Certainly we need the local q values for our
workflow, but we need to be sure globalQValue isn’t just a value that
Skyline never consults, or we will end up handicapping ourselves again.

brendanx67:
> Well, with this change, the RefSpectra table isn't changing, only the RetentionTimes table. That's what we want, right?

Good point. Yes, that means the canonical q value stays the same (if there is one), and per run q values change. That should be fine. What is the canonical q value here? I think the canonical spectrum is the predicted spectrum now, right?

PR #3719, opened 2025-12-17
Read extended Bruker chromatogram values not directly provided by the API

bspratt:
Including work for showing generalized QC trace types in SeeMS and Skyline.

### DIFF
--- a/Jamroot.jam
+++ b/Jamroot.jam
@@ -537,10 +537,17 @@ rule day_of_year ( Y M D )
 
 # YMMDD where Y is number of years since 2016
 # Uses git commit date for reproducible versioning (not build machine date)
+# Falls back to JAMDATE if git command fails (e.g., old git version, shallow clone)
 rule MAKE_BUILD_TIMESTAMP ( )
 {
     local git_date = [ SHELL "git log -1 --format=%cs HEAD" ] ;
     local ymd = [ MATCH "([0-9][0-9][0-9][0-9])-([0-9][0-9])-([0-9][0-9])" : $(git_date) ] ;
+    if ! $(ymd)
+    {
+        # Fallback to build machine date if git date extraction fails
+        local utc = [ modules.peek : JAMDATE ] ;
+        ymd = [ MATCH "([0-9][0-9][0-9][0-9])-([0-9][0-9])-([0-9][0-9])" : $(utc) ] ;
+    }
     local d = [ MATCH "[0-9][0-9]([0-9][0-9])" : $(ymd[0]) ] [ day_of_year $(ymd) ] ;
     return $(d:J=) ;
 }


bspratt:
And in SeeMS

image

nickshulman (commented):
Displaying just the units on the Y-axis looks unscientific. 

I think the label should be either "Pressure" or "Pressure (pascal)" instead of "pascal".

We could store both "QcTrace.MeasuredQuality" and the units on the ChromatogramGroupId.

pwiz/data/vendor_readers/Bruker/ChromatographyDataSqliteReader.cpp
```diff
@@ -0,0 +1,366 @@
...
+                }
+                double dummyValue = 0.0;
+                chrom->units = traceUnitToCVID(unitValue, dummyValue);
+            }
+
+            sqlite3_stmt* chunksStmt = nullptr;
+            std::ostringstream chunksQuery;
+            chunksQuery << "SELECT Times, Intensities FROM TraceChunks WHERE Trace=? ORDER BY ROWID";
+
+            if (sqlite3_prepare_v2(db.connected(), chunksQuery.str().c_str(), -1, &chunksStmt, nullptr) == SQLITE_OK)
```
chambm: Why use raw sqlite API here?

pwiz/data/vendor_readers/Bruker/ChromatographyDataSqliteReader.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+using pwiz::vendor_api::Bruker::Chromatogram;
+using pwiz::vendor_api::Bruker::ChromatogramPtr;
+
+// Forward declarations for local helpers used below
+static CVID traceTypeToCVID(int type, int unit, const std::string description);
+static CVID traceUnitToCVID(int unit, double& value);
+static TraceUnit getDefaultUnitForTraceType(CVID chromatogramType);
+
+std::vector<vendor_api::Bruker::ChromatogramPtr> readChromatographyDataSqlite(const std::string rootpath)
```
chambm: This class/feature really belongs in vendor_api/Bruker where the implementation can be hidden behind our CompassData API. The problem is then you don't get the CV so you have to either duplicate all the CV concepts you need or keep traceType as integer and convert it in the vendor_reader. The utility functions here would move to Reader_Bruker_Detail. Does this seem like a prohibitive amount of work?

pwiz_aux/msrc/utility/vendor_api/Bruker/CompassData.hpp
```diff
@@ -272,10 +273,23 @@ struct PWIZ_API_DECL Chromatogram
...
+    pwiz::cv::CVID units;
+    std::string description; // Expressed as a UserParam
+    std::string instrument;  // Expressed as a UserParam
+    
+    Chromatogram() : type(pwiz::cv::CVID_Unknown), units(pwiz::cv::CVID_Unknown) {}
 };
 
 typedef boost::shared_ptr<Chromatogram> ChromatogramPtr;
 
+struct PWIZ_API_DECL LCTraceSource
```
chambm: Is this dead code? Seems like ChromatogramPtr could contain a boost::optional<LCTraceSource> to make the metadata on ChromatogramPtr a little clearer. Probably not worth making Chromatogram polymorphic at this point.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -1057,13 +1087,11 @@ public List<QcTrace> GetQcTraces()
...
             {
-                CVID chromatogramType;
                 using (var chromMetaData = ChromatogramList.chromatogram(i, minDetailLevel))
                 {
-                    using var cvParamChild = chromMetaData.cvParamChild(CVID.MS_chromatogram_type);
-                    chromatogramType = cvParamChild.cvid;
-                    if (chromatogramType != CVID.MS_pressure_chromatogram &&
-                        chromatogramType != CVID.MS_flow_rate_chromatogram)
+                    // Skip over TIC, BPC, SIC, SIM, SRM, etc as they are not QC traces
+                    using var cvParamChild = chromMetaData.cvParamChild(CVID.MS_ion_current_chromatogram);
```
chambm, line 1177: The chromatogram CV reorganization that makes this logic possible was done as a result of the original work I did here, but it wasn't completed until a few weeks after I committed the PR. I never got around to doing what you've done here. This is good.

pwiz/data/vendor_readers/Bruker/ChromatographyDataSqliteReader.cpp
bspratt: Open to other approaches - just modeling on the python that Sven provided

pwiz/data/vendor_readers/Bruker/ChromatographyDataSqliteReader.cpp
bspratt: Open to other implementations, but the use of CV representation is important to low-maintenance handling of generalized chromatogram types in SeeMS and Skyline, and duplicating stuff doesn't sound conducive to low maintenance. Or perhaps I have misunderstood your note?

pwiz_aux/msrc/utility/vendor_api/Bruker/CompassData.hpp
bspratt: yeah, probably cruft from my AI "assistant" that I missed :/

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
bspratt: yeah, super convenient, thanks!

chambm:
FYI I have refactored this locally as I described above. Will commit next week.

bspratt:
UI updates per @nickshulman suggestion:

image



Note that "Fraction A - [psi]" is of type "Chromatogram" in the mzML model, where the others here are subtypes like "Temperature Chromatogram" etc, so we just use its name instead of its type on the Y axis:
image



image



image

PR #3720, opened 2025-12-18
Fixed AlertDlg.ShowDialog() to enforce test timeout

brendanx67:
## Summary
- Shadowed `Form.ShowDialog()` methods on `AlertDlg` with timeout-aware versions
- During functional tests, dialogs now throw `TimeoutException` after 10 seconds if not dismissed
- Prevents silent test hangs from unexpected `AlertDlg` dialogs (19 existing usages now protected)

## Background
Investigation of a test hang on KAIPOT-PC1 running `TestFilesTreeForm` revealed the issue. The `SkylineWindow.AskForLogEntry()` method shows an `AlertDlg` asking whether to add an audit log entry when the document hash doesn't match. On KAIPOT-PC1, a CRLF line-ending issue caused this dialog to appear unexpectedly during document loading.

The problem: `AlertDlg.ShowDialog()` bypassed the timeout mechanism used by `MessageDlg.Show()` and `MultiButtonMsgDlg.Show()`. The familiar pattern `using (var dlg = new AlertDlg()) { dlg.ShowDialog(...) }` had spread to 19 locations, all of which could cause silent test hangs.

By shadowing `ShowDialog()` with `new`, all existing and future usages automatically get timeout behavior in tests.

## Test plan
- [x] Build succeeds
- [ ] TeamCity build passes
- [ ] KAIPOT-PC1 `TestFilesTreeForm` now fails with TimeoutException instead of hanging silently

### DIFF
--- a/pwiz_tools/Skyline/Alerts/AlertDlg.cs
+++ b/pwiz_tools/Skyline/Alerts/AlertDlg.cs
@@ -77,6 +77,25 @@ namespace pwiz.Skyline.Alerts
             }
         }
 
+        /// <summary>
+        /// Shadows Form.ShowDialog() to prevent parentless dialogs.
+        /// Use ShowDialog(IWin32Window) or ShowAndDispose(IWin32Window) instead.
+        /// </summary>
+        public new DialogResult ShowDialog()
+        {
+            // Parentless dialogs can leak handles. Use ShowDialog(parent) instead.
+            throw new InvalidOperationException(@"Not supported.");
+        }
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


brendanx67:
Thanks for the suggestion @Copilot. We intentionally chose to throw `InvalidOperationException` instead of falling back to `base.ShowDialog()`. This matches the pattern established in `FormEx.ShowDialog()` which also throws to prevent parentless dialogs that can leak handles.

The parameterless `ShowDialog()` currently has zero callers, and we want to keep it that way by making it fail loudly rather than silently allowing the anti-pattern to spread.

PR #3730, opened 2025-12-23
Improved Relative Abundance graph performance with background computation and incremental updates

brendanx67:
## Summary
- Moved expensive O(n log n) sort and Y-value calculation to background thread
- Added ReplicateCachingReceiver for seamless, flash-free replicate switching
- Implemented O(n + k log k) incremental merge algorithm using ReferenceEquals on immutable nodes
- Fixed intermittent "duplicate key" exception caused by RuntimeHelpers.GetHashCode() collisions
- Added progress bar throttling (300ms delay, 100ms updates) to reduce UI flicker

## Test plan
- [x] Tested with ExtracellularVesicalMagNet.sky (~5,000 proteins, 48,000 peptides)
- [x] Verified replicate switching is instant after first calculation
- [x] Verified protein selection is O(1) via dictionary lookup
- [x] Ran PeakImputationDiaTutorial test 43 times with 0 failures (2x longest pre-fix run)
- [x] CodeInspection passes

See ai/todos/active/TODO-20251221_relative_abundance_perf.md

### DIFF
--- a/ai/todos/completed/TODO-20251221_relative_abundance_perf.md
+++ b/ai/todos/completed/TODO-20251221_relative_abundance_perf.md
@@ -4,8 +4,8 @@
 - **Branch**: `Skyline/work/20251221_relative_abundance_perf`
 - **Base**: `master`
 - **Created**: 2025-12-21
-- **Completed**: (pending)
-- **Status**: 🔍 In Review
+- **Completed**: 2025-12-30
+- **Status**: ✅ Complete
 - **PR**: [#3730](https://github.com/ProteoWizard/pwiz/pull/3730)
 - **Objective**: Improve Relative Abundance graph performance by moving computation to background threads
 


pwiz_tools/Skyline/Controls/Graphs/ReplicateCachingReceiver.cs
```diff
@@ -0,0 +1,310 @@
...
+    /// invalidated when the document or settings change.
+    ///
+    /// Additionally, when switching away from a calculation in progress, this wrapper
+    /// keeps a "completion listener" attached so the calculation continues in the
+    /// background and its result is cached. This means switching away and back
+    /// doesn't lose progress - like browser tabs that keep loading when you switch away.
+    /// </summary>
+    /// <typeparam name="TParam">Parameter type for the producer (must provide document, settings, and replicate info)</typeparam>
+    /// <typeparam name="TResult">Result type from the producer</typeparam>
+    public class ReplicateCachingReceiver<TParam, TResult> : IDisposable where TResult : class
```
nickshulman: The way that the ProductionFacility stuff would want this to be accomplished is that there are two different producers, and one of the producers overrides GetInputs to return the other as an "Input".
For instance, the GetInputs method in  `AssociateProteinResults.Producer` requests something that depends only on `ParsimonyIndependentParameters`, which enables switching between parsimony settings without having to recalculate everything.

pwiz_tools/Skyline/Controls/Graphs/ReplicateCachingReceiver.cs
```diff
@@ -0,0 +1,310 @@
...
+        /// <summary>
+        /// Creates a caching wrapper around a Receiver.
+        /// </summary>
+        /// <param name="receiver">The underlying Receiver for background computation</param>
+        /// <param name="getDocument">Function to extract document from parameters</param>
+        /// <param name="getSettings">Function to extract settings from parameters (used for cache invalidation)</param>
+        /// <param name="getCacheKey">Function to extract cache key (replicate index) from parameters</param>
+        public ReplicateCachingReceiver(
+            Receiver<TParam, TResult> receiver,
+            Func<TParam, SrmDocument> getDocument,
```
nickshulman: The methods "getDocument", "getSettings" and "getCacheKey" only ever get used on objects of type `TParam`.

A different way to accomplish this would be to have a base class with properties "Document", "Settings" and "CacheKey" and give TParam the constraint that it inherit from that base class.

I would probably make the types of "Settings" and "CacheKey" generic in that parameter base class, instead of always using "object" and "int".
`public class BaseParam<TKey, TSettings>`

pwiz_tools/Skyline/Controls/Graphs/SummaryRelativeAbundanceGraphPane.cs
```diff
@@ -647,76 +712,352 @@ protected GraphData(SrmDocument document, GraphSettings graphSettings, Productio
...
+            private List<(PointPair point, PeptideGroupDocNode docNode)> CalculateChangedPoints(
+                List<PeptideGroupDocNode> moleculeGroups,
+                IReadOnlyDictionary<int, NodePosition> priorNodePositions,
+                SkylineDataSchema schema,
+                int? resultIndex,
+                ProductionMonitor productionMonitor)
+            {
+                var changedPoints = new List<(PointPair point, PeptideGroupDocNode docNode)>();
+                double maxY = 0;
+                var minY = double.MaxValue;
```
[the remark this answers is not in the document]
brendanx67: This is unrelated to the changes in this PR. It is not worth the testing effort to change this right now. The values are provided as ref parameters and may have side effects beyond being intended for use after the loop completes.

pwiz_tools/Skyline/Controls/Graphs/ReplicateCachingReceiver.cs
brendanx67: Thanks for the pointer to \ - that's an interesting pattern for splitting computation into reusable stages. Our current \ wrapper provides immediate value for the replicate-switching use case with minimal integration changes. If we find the current approach limiting (e.g., if we need to share intermediate results across different graph types), the nested producer pattern would be the natural evolution. For now, the wrapper approach keeps the changes localized.

pwiz_tools/Skyline/Controls/Graphs/ReplicateCachingReceiver.cs
brendanx67: Good suggestion - we've implemented this\! See the latest commits which introduce ICachingParameters and ICachingResult interfaces in CachingContracts.cs. The parameter types now implement ICachingParameters (with Document, CacheKey, and CacheSettings properties) and the result types implement ICachingResult (with Document property). The constructor no longer needs lambdas - it uses type constraints instead.

pwiz_tools/Shared/zedgraph/ZedGraph/ZedGraphControl.Events.cs
```diff
@@ -1,6 +1,6 @@
 //============================================================================
 //ZedGraph Class Library - A Flexible Line Graph/Bar Graph Library in C#
-//Copyright � 2007  John Champion
+//Copyright � 2007  John Champion
```
nickshulman: Something has gone wrong with this copyright symbol.

pwiz_tools/Skyline/Controls/Graphs/SummaryRelativeAbundanceGraphPane.cs
```diff
@@ -647,76 +814,439 @@ protected GraphData(SrmDocument document, GraphSettings graphSettings, Productio
...
                     pointPairList.Add(pointPair);
                     maxY = Math.Max(maxY, groupMaxY);
                     minY = Math.Min(minY, groupMinY);
                 }
 
+                productionMonitor.SetProgress(80);
                 pointPairList.Sort(CompareYValues);
+
+                productionMonitor.SetProgress(90);
+                var identityToIndex = new Dictionary<Identity, int>(pointPairList.Count);
```
nickshulman: I think you need to use a different `IEqualityComparer` for this dictionary (i.e. `ReferenceValue.EQUALITY_COMPARER`).
Otherwise, you're going to be comparing things using PeptideGroup.Equals which is not what you want.

When I want to use Identity objects as keys in a dictionary, I usually declare the dictionary as `Dictionary<ReferenceValue<Identity>, Value>` because the struct `ReferenceValue<>` overrides Equals and GetHashCode to use reference equality.

If you have a `Dictionary<Identity, Value>` you have to always remember to construct it with the correct IEqualityComparer.

pwiz_tools/Shared/zedgraph/ZedGraph/ZedGraphControl.Events.cs
brendanx67: Reverted this file and applied the simple change of moving SetCursor() below the MouseMoveEvent handler so that it appears as a minimal change. Nice catch.

pwiz_tools/Skyline/Controls/Graphs/SummaryRelativeAbundanceGraphPane.cs
brendanx67: Nice catch! Definitely not what I was trying to get Claude Code to implement. I decided to stay with Identity.GlobalIndex as a guaranteed unique key for and Identity object, but I finally understand your alternative approach, and I added it to the ai\docs\architecture-data-model.md documentation I have been creating with and for Claude Code.

PR #3739, opened 2025-12-27
Fix FilesTree deadlock during disposal

### LINKED ISSUE #3738: FilesTree deadlock during disposal causes TestTreeRestoration hang
## Summary

Deadlock in FilesTree disposal causing TestTreeRestoration to hang on nightly tests.

## Root Cause Analysis (from Nick Shulman)

Two threads are deadlocked:

**Thread 1 (UI Thread)** - Disposing FilesTree:
```
DestroyFilesTreeForm → FilesTree.Dispose → LocalFileSystemService.StopWatching 
→ ManagedFileSystemWatcher.DisposeWatcher → FileSystemWatcher.Dispose 
→ Monitor.Enter (BLOCKED)
```

**Thread 2 (Background Thread)** - FileSystemWatcher callback:
```
FileSystemWatcher.CompletionStatusChanged → Control.BeginInvoke 
→ Control.get_Handle → Control.CreateHandle 
→ WaitMessage (BLOCKED - waiting for UI thread)
```

The UI thread is disposing the FileSystemWatcher while the FSW's background callback thread is trying to `BeginInvoke` back to the FilesTree control (which requires the UI thread to create/get the window handle).

## Test Failure

- **Test**: TestTreeRestoration
- **Machine**: EKONEIL01
- **Run ID**: 79671
- **Link**: https://skyline.ms/home/development/Nightly%20x64/testresults-showRun.view?runId=79671

## Related Work

This is a regression in the FilesTree feature (PR #3867). Previous fixes addressed:
- CancellationTokenSource disposal timing (2025-12-17)
- FileSystemWatcher race conditions (2025-12-12)
- Thread lifecycle management (2025-12-03)

See completed TODO: `ai/todos/completed/TODO-20251126_files_view.md`

## Getting Started

When starting this work:
1. Create `ai/todos/active/TODO-YYYYMMDD_filestree_deadlock.md`
2. Link this issue in the TODO file

brendanx67:
## Summary
- Fixes deadlock in FilesTree disposal that caused TestTreeRestoration to hang on nightly tests
- Removed `FileSystemWatcher.SynchronizingObject` which caused FSW to internally call `BeginInvoke` during disposal
- Thread marshaling is already handled safely via `BackgroundActionService.RunUI()` → `SafeBeginInvoke()`

## Root Cause
Setting `SynchronizingObject` on a `FileSystemWatcher` causes .NET to automatically call `BeginInvoke` on that control to marshal events. During disposal, the UI thread waits for FSW.Dispose() while the FSW callback thread tries to BeginInvoke back to the UI thread → deadlock.

## Test plan
- [x] TestFilesModel (16s)
- [x] TestFilesTreeFileSystem (4s)
- [x] TestFilesTreeForm (40s)
- [x] TestSkylineWindowEvents (0s)

Fixes #3738

See ai/todos/active/TODO-20251227_filestree_deadlock.md

### DIFF
--- a/pwiz_tools/Skyline/Model/ReintegrateResultsHandler.cs
+++ b/pwiz_tools/Skyline/Model/ReintegrateResultsHandler.cs
@@ -66,14 +66,12 @@ namespace pwiz.Skyline.Model
             return MProphetResultsHandler.GetPeakFeatureStatistics(peptide, fileId);
         }
 
-        public ChromPeak GetBestPeak(PeptideDocNode peptideDocNode, TransitionGroupDocNode transitionGroupDocNode, 
-            ChromatogramSet chromatogramSet, ChromatogramInfo chromatogramInfo,
-            ref PeakGroupIntegrator peakGroupIntegrator,
-            out UserSet userSet)
+        public ChromPeak GetBestPeak(PeptideDocNode peptideDocNode,
+            TransitionGroupIntegrator transitionGroupIntegrator, Transition transition, out UserSet userSet)
         {
-            var chromFileInfoId = chromatogramSet.FindFile(chromatogramInfo.FilePath);
+            var chromFileInfoId = transitionGroupIntegrator.ChromFileInfoId;
             var peakFeatureStatistics = MProphetResultsHandler.GetPeakFeatureStatistics(peptideDocNode.Peptide, chromFileInfoId);
-            int bestIndex = chromatogramInfo.BestPeakIndex;
+            int bestIndex = transitionGroupIntegrator.ChromatogramGroupInfo.BestPeakIndex;
             userSet = UserSet.FALSE;
             if (peakFeatureStatistics != null)
             {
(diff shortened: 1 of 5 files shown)


brendanx67:
If a flag in OnClosing() is our best answer for this, we should implement the flag in CommonFormEx and make the SafeInvoke pattern use it.

brendanx67:
How does that help? Taking a lock after the handle is destroyed? Still
seems prone to the same race condition. A flag at the end of OnClosing
seems better. That is the earliest I can think of that you truly know the
window is going away and invoking to it will have no benefit. Give up and
exit before the handle is destroyed.

nickshulman:
> How does that help? Taking a lock after the handle is destroyed? 

Yes, it does sound like "OnHandleDestroyed" would be fired after the handle is destroyed.

However, Control.IsHandleCreated still seems to be true in the event handler. It's not until the very end of "Control.DestroyHandle" that "window.DestroyHandle" gets called. So, shutting off events during OnHandleDestroyed does seem to be soon enough to avoid resurrecting the window handle.

PR #3748, opened 2026-01-03
Add "Replicate" column to new Peptide Normalized Areas and Protein Abundances reports.

nickshulman:
Since there hasn't been a Skyline-daily release since when these reports were added, it is not necessary to increment the version number in PersistedViews.cs

Adding the "Replicate" column gives the user something to click on to navigate to the replicate.

image
image


brendanx67 (approved):
Sounds good to me.

PR #3763, opened 2026-01-07
Complicated fix for intermittent failure in TestFilesTreeForm and new unit test

nickshulman:
In TestFilesTreeForm, if the Library Match window updates while OKing a library change in the Peptide Settings dialog, then Skyline will fail to close a PooledSqliteConnection for the library.

I added a test "ChangePickedLibrariesTest" which does only this part of TestFilesTreeForm.

I have changes in GraphSpectrum which does fix the failure, but is ugly.
It might make sense to add a method that everyone can use to SrmDocument to close all streams that are not being used by another document.
Currently, the official code for closing streams not owned by another document is the private method "BackgroundLoader.CloseRemovedStreams". However, any code that interacts with a document which might not be the current document in the SkylineWindow runs the risk of opening a stream that will never be closed.


nickshulman:
This isn't important enough to worry about for now.

PR #3766, opened 2026-01-07
Populated FileCount and File columns in localization CSV export

brendanx67:
## Summary
* Fixed ExportLocalizationCsv to track file paths and populate FileCount/File columns
* For single-file entries: FileCount=1, File=full path to .resx file
* For consolidated entries (same English text in multiple files): FileCount=N, File=semicolon-separated list (up to 3 files, with "..." if more)
* Updated import logic to handle semicolon-separated file lists
* Added GenerateLocalizationCsvFiles Jamfile target for convenience

This provides translators with context about where each string is used, while maintaining backward compatibility with CSV import.

## Test plan
- [x] Verified TestImportLocalizationCsv round-trip test passes
- [x] Verified CSV line count unchanged (487 entries)
- [x] Verified FileCount and File columns populated correctly

### DIFF
--- a/pwiz_tools/Skyline/TestFunctional/FilesTreeFormTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/FilesTreeFormTest.cs
@@ -688,6 +688,10 @@ namespace pwiz.SkylineTestFunctional
                 });
                 OkDialog(libListDlg, libListDlg.OkDialog);
                 RunUI(() => peptideSettingsUI.PickedLibraries = peptideSettingsUI.PickedLibraries.Append(name).ToArray());
+
+                // TODO(nicksh): If GraphSpectrum updates while the OK button is being pressed it might leak a PooledSqliteConnection
+                WaitForConditionUI(() => !SkylineWindow.GraphSpectrum.IsGraphUpdatePending);
+
                 OkDialog(peptideSettingsUI, peptideSettingsUI.OkDialog);
             }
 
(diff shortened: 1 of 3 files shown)


pwiz_tools/Skyline/Executables/DevTools/ResourcesOrganizer/ResourcesOrganizer/ResourcesModel/ResourcesDatabase.cs
```diff
@@ -489,18 +501,27 @@ public void ExportLocalizationCsv(string path, string language, out int entryCou
...
                     if (unqualifiedRecords.Count == 1)
                     {
                         records.Add(unqualifiedRecords[0]);
                         continue;
                     }
                 }
-                records.AddRange(individualRecords);
+
+                // For non-consolidated entries, add file info
+                records.AddRange(individualRecords.Select(r => r with { FileCount = fileCount, File = uniqueFiles[0] }));
```
nickshulman: The `File = uniqueFiles[0]` here is wrong.

This is the code path for either `individualRecords.Count == 1` or the case where `individualRecords.Count > 1` but some of the files have different translations (i.e. the code path of not hitting the "continue" inside of `if (unqualifiedRecords.Count == 1)`).

The original File value must be retained because different files have different translations, and we need to remember which row came from which file.

In this code path, each row represents only a single file, so FileCount should probably still be set to 1.

nickshulman (changes requested):

PR #3774, opened 2026-01-08
Fix VS 2026 (msvc-14.5) build toolset detection

brendanx67:
## Summary
- Fix msvc.jam toolset 14.5 path patterns to use explicit `Hostx64/x64` and `Hostx86/x86` instead of `Host*/*` glob
- VS 2026 added ARM64 host compilers (HostArm64) which sorted alphabetically before Hostx64, causing cl.exe not to be found on x64 machines
- Surface elapsed time to console output in build-apps.bat
- Add b.bat and bs.bat to .gitignore (personal build scripts)

## Test plan
- [x] Verified build completes successfully with VS 2026 (msvc-14.5 toolset)
- [x] Confirmed cl.exe is found in correct Hostx64/x64 directory

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -428,8 +428,10 @@ Version.cpp
 /pwiz_tools/Skyline/SkylineTester test list.txt
 /pwiz_tools/Skyline/SkylineTester.log
 SkylineTester Results
+/b.bat
 /b64.bat
 /b32.bat
+/bs.bat
 /bs64.bat
 /bs32.bat
 /bso64.bat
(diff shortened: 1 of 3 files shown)


chambm:
I'm going to look into this. I want to keep msvc.jam reasonably in sync with the official Boost source except for the C++/CLI customizations I have to make. I'm guessing Brian and I didn't run into this because we chose not to install ARM[64] compiler support (IIRC VS2022 also has this support, optionally, not sure if it defaults to on or not).

PR #3798, opened 2026-01-12
UniCode file handling tweak for Waters and Mobilion…

bspratt:
… (whose reader DLLs do not handle UniCode filenames), msconvert was using the 8.3 version of UniCode-containing filenames for output filename and for run id in the mzML file.

e.g. 091204_NFDM_008-试验.raw should produce 091204_NFDM_008-试验.mzML but was creating 091204\~1.mzML instead, and within the file where it should read id="091204_NFDM_008-试验.raw" it used id="091204\~1.raw"


bspratt:
It’s actually good that the test deals with both possible conditions, and that we have systems that exercise one or the other.

PR #3800, opened 2026-01-13
Improved waters_connect server connection exception handling.

rita-gwen:


pwiz_tools/Skyline/FileUI/BaseFileDialogNE.cs
```diff
@@ -216,16 +219,73 @@ public MsDataFileUri CurrentDirectory
...
+                            var remoteAccount = GetRemoteAccount(remoteUrl);
+                            if (remoteAccount == null) // If no accounts found
+                            {
+                                // Can happen if user selects a remote URL but then deletes the associated account.
+                                // Show error message, set the currentDirectory to RemoteUrl.EMPTY
+                                var message = TextUtil.LineSeparate(
+                                    FileUIResources
+                                        .ExportMethodDlg_OkDialog_Cannot_find_waters_connect_account_for_the_selected_URL_,
+                                    remoteUrl.Username + @"@" + remoteUrl.ServerUrl);
+                                MessageDlg.ShowError(this, message);
```
nickshulman: We don't normally show messages inside of a property setter. The more usual way to do this would be to throw an exception and the caller (i.e. the code that tried to set the property) would catch that exception and show the message. Alternatively, the caller might be expected to validate the value before setting the property.

rita-gwen: I had to wrap the property setter into an explicit setter method and hide the former. The error handling logic there is too complicated to implement everywhere the CurrentDirectory property is set. The idea was to suppress unnecessary exceptions and show less scary messages instead. Let me know if this is good enough. I can convert it into an explicit validation method and replace the setter method calls with if (IsValidDirectory(uri)) dlg.CurrentDirectory = uri it this is preferred. But I think it's not very user-friendly.

PR #3816, opened 2026-01-16
Fix FilesDirTest, TestFastaImport, TestKeyboardShortcutHelp, TestManagingSearchTools, TestWatersConnectExportMethodDlg

nickshulman:


pwiz_tools/Skyline/TestFunctional/ManagingSearchToolsTest.cs
```diff
@@ -134,7 +134,7 @@ private void TestInputErrors()
                 editToolDlg.ToolPath = TestFilesDir.GetTestPath("msfragger.bat");
             });
             var errorDlg = ShowDialog<MessageDlg>(editToolDlg.OkDialog);
-            Assert.AreEqual("The tool MSFragger is already configured.", errorDlg.Message);
```
chambm, line 137: Oof. I wonder if we can give directions to an AI code review so it checks for a list of things like: "No string literals inside Assert or AssertEx statements". So this started failing when the translations came back for 26.1?

chambm: Oh I see Brendan already made copilot-instructions.md. And it has detailed instructions for when string literals are allowed in Asserts. But it was added 12/30, after these two commits. I was going to add instructions in a PR that also had string literal asserts to test that that the review actually would catch what it should catch.

PR #3819, opened 2026-01-16
Add search settings preset list in SearchSettingsControl

chambm:
## Search Settings Preset

Adds a search settings preset system to the Import Peptide Search wizard that saves and restores all search configuration across wizard pages.

### Features
- **Preset dropdown** at bottom-left of wizard (visible on all DDA search pages)
- **Stores**: search engine, tolerances, enzyme, missed cleavages, FASTA path, modifications (full StaticMod definitions), workflow type, IRT standard, and engine-specific additional settings
- **Built-in presets**: Default (MSAmanda), Comet high-high (QE/Q-Tof), Comet high-low (Velos-Orbitrap), Comet low-low (ion trap)
- **Import from .params files**: `<Import...>` option in dropdown parses Comet and MSFragger .params files and creates presets from them
  - Auto-detects engine type from file contents
  - Maps tolerances, enzyme, modifications (via UniMod matching), and engine-specific settings
- **Save**: captures current settings from all wizard pages into a named preset

### Screenshot

preset-dropdown-screenshot

### Changes
- `SearchSettingsPreset.cs` - Preset model with FASTA, mods, workflow, IRT properties
- `SearchSettingsParamsFileParser.cs` - **NEW** - Parser for Comet/MSFragger .params files
- `ImportPeptideSearchDlg.cs` - Preset UI (combo, save, import), ApplyPreset, BuildPresetFromCurrentSettings
- `SearchSettingsControl.cs` - ApplySearchSettingsPreset public method

- `MatchModificationsControl.cs` - Initialize accepts optional preset parameter
- `ImportFastaControl.cs` - ClearFastaContent, Enzyme setter fixes
- `CometSearchEngine.cs` - GetDefaultPresets with instrument-specific configs
- `DdaSearchTest.cs` - Comprehensive preset tests including import from .params files


maccoss:
I did a search on this file and fasta and got what appear like reasonable results. https://panoramaweb.org/_webdav/MacCoss/maccoss/Shared_w_lab/%40files/RawFiles/CometTestFiles

I found the UI awkward.  I thought the preset was just going to be on the Comet search page.  Unless the settings is going to include settings for the entire workflow ... like with Fragpipe (https://fragpipe.nesvilab.org/docs/tutorial_fragpipe_workflows.html) it probably shouldn't go first. 

We need a way for someone who has their comet parameters to be able to use them or for Alexey or Fengchao to provide parameters.  I was hoping that the preset would support importing a comet.params or a fragger.params file (https://github.com/Nesvilab/MSFragger/tree/master/parameter_files).  Users of these tools are familiar with the parameter files.  We can provide the defaults and ones from Jimmy, Alexey, etc... We also need tool tips or something otherwise I have to have the Comet or MSFragger github page open to be able to change the parameters. 

Also, the MS2 tolerance for Comet needs to be changed to a bin size and an offset.  There is no tolerance like with most tools and someone who doesn't know that there is a difference will struggle with Comet. The Comet, SEQUEST, Tide, Proteome Discoverer documentation all describe it as a bin and offset. Also a user shouldn't change the bin size without changing the offset -- this requires the user knowing that it is a bin at a specific location and not a window around a predicted fragment ion m/z like other tools.

Plus this UI step has to state somewhere clearly that this has nothing to do with the search but it is for the chromatogram extraction of the peptides detected by the search. When I watch people use Skyline they assume that this is the MS1 search parameters. Only really experienced Skyline users will recognize that this looks similar to the chromatogram extraction UI.
image

 It would be ideal if the wizard went in the order that the analysis happened. Files, Search, Percolator, Protein Parsimony, Chromatogram extraction, etc... 

One last request is the "Finish" button at the end of the search.  For a long analysis most of us kick of the search at night when we aren't using our computer but it stops 1/2 way through the analysis so we then need to complete the analysis in the morning as opposed to having everything completed in the morning. 

We also need to filter on the PSM, Peptide or Protein level.  Percolator provides q-values for all 3 of these. At a minimum it needs to be clear how the filtering was provided.  Otherwise we need to filter outside of Skyline, create a blib that has the FDR controlled correctly and then reimport.

pwiz_tools/Skyline/FileUI/PeptideSearch/MatchModificationsControl.cs
```diff
@@ -152,8 +156,35 @@ public bool Initialize(SrmDocument document)
                 menuItemAddHeavyModification.Text = PeptideSearchResources.MatchModificationsControl_Initialize_Edit__heavy_modifications___;
             }
 
+            // Add any missing mods from the preset to the global settings lists
+            if (preset != null)
+            {
+                foreach (var mod in preset.StructuralModifications)
+                {
+                    if (!Settings.Default.StaticModList.Any(m => m.Name == mod.Name))
```
nickshulman: You should use `Settings.Default.StaticModList.ContainsKey(mod.Name)` because it's faster than doing a linear search.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.resx
```diff
@@ -1337,6 +1337,87 @@ injected into the mass spectrometer.</value>
...
+    <value>200, 605</value>
+  </data>
+  <data name="btnSavePreset.Size" type="System.Drawing.Size, System.Drawing">
+    <value>45, 23</value>
+  </data>
+  <data name="btnSavePreset.TabIndex" type="System.Int32, mscorlib">
+    <value>12</value>
+  </data>
+  <data name="btnSavePreset.Text" xml:space="preserve">
+    <value>Save</value>
```
nickshulman: I think this should have "..." so:
`Save...`
since pushes the button brings up a dialog asking them for a new name.

Without the ellipsis, users might be afraid to push the button because they might worry it will overwrite their existing preset.

pwiz_tools/Shared/CommonUtil/Collections/MappedList.cs
```diff
@@ -33,6 +33,14 @@ public interface IKeyContainer<out TKey>
         TKey GetKey();
     }
 
+    /// <summary>
+    /// Implement on an element to support renaming in Edit List dialogs.
+    /// </summary>
+    public interface IRenameable<out T>
```
nickshulman: I wonder if we should just change SearchSettingsPreset to be a subclass of XmlNamedElement instead of introducing this new interface.

pwiz_tools/Skyline/FileUI/PeptideSearch/ImportPeptideSearchDlg.cs
```diff
@@ -229,6 +256,334 @@ private static void AddPageControl<TControl>(TControl pageControl, TabPage tabPa
...
+                heavyModifications: heavyMods,
+                workflowType: (SearchWorkflowType)BuildPepSearchLibControl.WorkflowType,
+                irtStandardName: BuildPepSearchLibControl.IrtStandards?.Name,
+                hasExplicitModifications: true);
+        }
+
+        private bool ShowPresetNameInputDialog(string title, string defaultValue, out string result)
+        {
+            result = null;
+            using (var form = new Form())
```
nickshulman: I would recommend adding a new Form class with .designer.cs and .resx files to the project.

nickshulman (commented):
Looks nice!

chambm:
> I did a search on this file and fasta and got what appear like reasonable results. https://panoramaweb.org/_webdav/MacCoss/maccoss/Shared_w_lab/%40files/RawFiles/CometTestFiles
> 
> I found the UI awkward. I thought the preset was just going to be on the Comet search page. Unless the settings is going to include settings for the entire workflow ... like with Fragpipe (https://fragpipe.nesvilab.org/docs/tutorial_fragpipe_workflows.html) it probably shouldn't go first.

Because we are including workflow type, FASTA, and mods, the preset needs to go before those pages. It just makes sense to show it on every page because it affects settings on several pages (and there is now a warning when you change preset on a later page if it will change settings on previous pages, e.g. if you're on the search settings page and you change preset, it will warn you if it will change mods or FASTA).


> We need a way for someone who has their comet parameters to be able to use them or for Alexey or Fengchao to provide parameters. I was hoping that the preset would support importing a comet.params or a fragger.params file (https://github.com/Nesvilab/MSFragger/tree/master/parameter_files). Users of these tools are familiar with the parameter files. We can provide the defaults and ones from Jimmy, Alexey, etc... We also need tool tips or something otherwise I have to have the Comet or MSFragger github page open to be able to change the parameters.

That ability was already in the build but it was kind of obscure in the "Import" option in the preset dropdown. Now though, it's a separate button on the search settings page, because that button really does just import the search engine settings. And it only shows up for the engines we support importing settings from (currently Comet and MSFragger).

We are also shipping default configs for Comet and MSFragger for several workflow/instrument combos. Agreed about tooltips: I assume you mean in the additional settings dialog? For another PR I'll see about adding tooltips for each row.


> Also, the MS2 tolerance for Comet needs to be changed to a bin size and an offset. There is no tolerance like with most tools and someone who doesn't know that there is a difference will struggle with Comet. The Comet, SEQUEST, Tide, Proteome Discoverer documentation all describe it as a bin and offset. Also a user shouldn't change the bin size without changing the offset -- this requires the user knowing that it is a bin at a specific location and not a window around a predicted fragment ion m/z like other tools.

We disagree about this: I think the average user that wants to try Comet is not going to know about its bin size and offset. They will understand m/z tolerance though. However, as a compromise I have adjusted Comet's search settings UI so the MS2 tolerance is greyed out and instead the user configures the MS2 analyzer as either "High resolution" or "Low resolution" which will set the bin and offset accordingly.

> Plus this UI step has to state somewhere clearly that this has nothing to do with the search but it is for the chromatogram extraction of the peptides detected by the search. When I watch people use Skyline they assume that this is the MS1 search parameters. Only really experienced Skyline users will recognize that this looks similar to the chromatogram extraction UI.

Agreed. I updated the text to make it clearer it's for extraction.

> It would be ideal if the wizard went in the order that the analysis happened. Files, Search, Percolator, Protein Parsimony, Chromatogram extraction, etc...

Agreed, but that's out of scope for this PR and will be done separately.

> One last request is the "Finish" button at the end of the search. For a long analysis most of us kick of the search at night when we aren't using our computer but it stops 1/2 way through the analysis so we then need to complete the analysis in the morning as opposed to having everything completed in the morning.

Agreed but also out of scope for this PR. Will do separately.

> We also need to filter on the PSM, Peptide or Protein level. Percolator provides q-values for all 3 of these. At a minimum it needs to be clear how the filtering was provided. Otherwise we need to filter outside of Skyline, create a blib that has the FDR controlled correctly and then reimport.

Still not clear how this would work. Is it not possible to just pick one q-value and say it's the right one to filter on for any situation?

maccoss:
On Tue, Apr 28, 2026, 10:45 AM Matt Chambers ***@***.***>
wrote:

> We disagree about this: I think the average user that wants to try Comet
> is not going to know about its bin size and offset. They will understand
> m/z tolerance though. However, as a compromise I have adjusted Comet's
> search settings UI so the MS2 tolerance is greyed out and instead the user
> configures the MS2 analyzer as either "High resolution" or "Low resolution"
> which will set the bin and offset accordingly
>
> Chat with Jimmy about this. It is not possible to do a tolerance with the
MS2 data in Comet/Tide/SEQUEST and those tools all use a bin and bin offset
in all of their parameters. Not even sure how you could hack a tolerance to
work with these tools. Anyone who has used these tools will expect a bin
size. The offset was something we found years ago to be essential based on
the known forbidden zones.

Mike

PR #3821, opened 2026-01-16
Add support for creating spectrum filters on "Source Offset Voltage"

nickshulman:
Added support for defining spectrum filters on Source Offset Voltage (requested by Philip)

### DIFF
--- a/pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
+++ b/pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
@@ -605,6 +605,17 @@ namespace pwiz.Skyline.FileUI
             SetComboBoxText(columns.PrecursorNoteColumn, Resources.PasteDlg_UpdateMoleculeType_PrecursorNote);
             SetComboBoxText(columns.MoleculeNoteColumn, Resources.PasteDlg_UpdateMoleculeType_MoleculeNote);
             SetComboBoxText(columns.MoleculeListNoteColumn, Resources.PasteDlg_UpdateMoleculeType_MoleculeListNote);
+
+            // Set combo boxes for additional fragment-oriented column assignments beyond the first.
+            // When multiple columns are mapped to the same product type (e.g. three columns all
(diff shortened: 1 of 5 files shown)


pwiz_tools/Skyline/Model/Databinding/Entities/ColumnToolTips.resx
```diff
@@ -1588,6 +1588,9 @@ file, or the file name again for other file types.</value>
   <data name="SMILES" xml:space="preserve">
     <value>Standardized molecular identifier</value>
   </data>
+  <data name="SourceOffsetVoltage">
+    <value>DV voltage applied at the ion source</value>
```
chambm: Typo? Should we use the definition from the CV?

pwiz_tools/Skyline/TestFunctional/SourceOffsetVoltageTest.cs
```diff
@@ -0,0 +1,123 @@
...
+                {
+                    SkylineWindow.SelectedPath = identityPath;
+                });
+                WaitForGraphs();
+            }
+            Assert.IsNotNull(fullPointCount);
+            Assert.AreEqual(fullPointCount.Value, totalFilteredPointCount);
+        }
+    }
+}
```
chambm, line 123: Dangit. I made a small subset RAW for this but forgot to add it to Reader_Thermo_Test.data. The subset RAW is just 300KB. Perhaps we can re-use it in this test from its location (or a symbolic link?) at Reader_Thermo_Test.data? Just in the interest of avoiding test file repository bloat.

pwiz_tools/Shared/CommonUtil/Spectra/SpectrumMetadata.cs
```diff
@@ -33,6 +33,7 @@ private enum Flags
             HasTotalIonCurrent = 4,
             HasInjectionTime = 8,
             HasConstantNeutralLoss = 16,
+            HasSourceOffsetVoltage = 32,
```
chambm, line 36: I notice there's a lot of Skyline classes that need to be touched to add a new metadata type. Could we consolidate these into some key-value dictionary?

PR #3839, opened 2026-01-20
Fixed ProductionFacility.Unlisten race condition

### LINKED ISSUE #3832: ProductionFacility.Unlisten throws InvalidOperationException when entry already removed
## Summary

`ProductionFacility.Unlisten` throws `InvalidOperationException: Entry WorkOrder not found` when attempting to unlisten from a WorkOrder that was already removed by another thread.

## Stack Trace

From TestPeakPickingTutorial failure on KAIPO-DEV (2026-01-18):

[View full stack trace on skyline.ms](https://skyline.ms/home/development/Nightly%20x64/testresults-showFailures.view?end=01%2F18%2F2026&failedTest=TestPeakPickingTutorial)

```
System.InvalidOperationException: Entry pwiz.Common.SystemUtil.Caching.WorkOrder not found
   at pwiz.Common.SystemUtil.Caching.ProductionFacility.GetEntry(WorkOrder key) in ProductionFacility.cs:line 61
   at pwiz.Common.SystemUtil.Caching.ProductionFacility.Unlisten(WorkOrder key, IProductionListener listener) in ProductionFacility.cs:line 42
   at pwiz.Skyline.Controls.Graphs.ReplicateCachingReceiver`2.ClearCache() in ReplicateCachingReceiver.cs:line 372
   at pwiz.Skyline.Controls.Graphs.RTLinearRegressionGraphPane.<RTScoreCalculatorList_ListChanged>b__21_0() in RTLinearRegressionGraphPane.cs:line 107
```

## Root Cause

Race condition between entry cleanup and unlisten operations:

1. Thread A calls `AfterLastListenerRemoved()` which executes `Cache.RemoveEntry(this)` (line 183)
2. Thread B calls `Unlisten()` for the same or a dependent WorkOrder
3. Thread B's `GetEntry()` runs AFTER Thread A removed the entry, throwing the exception

The current code at line 42:
```csharp
public void Unlisten(WorkOrder key, IProductionListener listener)
{
    if (key == null)
        return;
    lock (this)
    {
        GetEntry(key).RemoveListener(listener);  // Throws if key not in _entries
    }
}
```

## Suggested Fix

Make `Unlisten` tolerant of missing entries (already cleaned up by another thread):

```csharp
public void Unlisten(WorkOrder key, IProductionListener listener)
{
    if (key == null)
        return;
    lock (this)
    {
        if (_entries.TryGetValue(key, out var entry))
            entry.RemoveListener(listener);
        // Silently ignore if entry doesn't exist - already cleaned up
    }
}
```

## Frequency

This specific failure pattern has been observed **once** (2026-01-18). It's a rare race condition that requires specific thread interleaving timing to manifest.

## Context

This appears to be related to increased concurrency in the caching system after the Files View feature (PR #3687). A similar race condition in `RTLinearRegressionGraphPane` was fixed in PR #3822.

brendanx67:
## Summary

* Added `_isListening` flag to `CompletionListener` in `ReplicateCachingReceiver`
* Uses `Interlocked.Exchange` to prevent double-Unlisten when `OnProductAvailable` and `ClearCache` race

Per Nick's review: the fix belongs in `ReplicateCachingReceiver.CompletionListener` which should track its listening state, not in `ProductionFacility` which should be able to assume balanced Listen/Unlisten calls.

Fixes #3832

## Test plan

- [x] Build succeeds
- [x] TestPeakPickingTutorial passes (50+ runs, original failure scenario)

See ai/todos/active/TODO-20260120_ProductionFacility_Unlisten.md

### DIFF
--- a/pwiz_tools/Skyline/Controls/Graphs/ReplicateCachingReceiver.cs
+++ b/pwiz_tools/Skyline/Controls/Graphs/ReplicateCachingReceiver.cs
@@ -393,6 +393,7 @@ namespace pwiz.Skyline.Controls.Graphs
             private readonly ReplicateCachingReceiver<TParam, TResult> _owner;
             private readonly int _cacheKey;
             private readonly WorkOrder _workOrder;
+            private int _isListening = 1;  // 1 = listening, 0 = unlistened
 
             public CompletionListener(ReplicateCachingReceiver<TParam, TResult> owner, int cacheKey, WorkOrder workOrder)
             {
@@ -429,6 +430,9 @@ namespace pwiz.Skyline.Controls.Graphs
 
             public void Unlisten()
             {
+                // Atomic check-and-set to prevent double Unlisten from concurrent threads
+                if (Interlocked.Exchange(ref _isListening, 0) == 0)
+                    return;
                 _owner._receiver.Cache.Unlisten(_workOrder, this);
             }
         }


nickshulman (commented):
Is the "Unlisten" call in "CompletionListener.OnProductAvailable" necessary?

Listeners don't normally remove themselves when they receive the value that they asked for, because that causes the ProductionFacility to discard the value that was calculated as well as the inputs that were used in the calculation.

This means that if you later ask for the same data, or for data that used the same inputs, the calculation does not have to start back over from scratch.

This does not make much difference in the scenarios where ReplicateCachingReceiver gets used, because those producers do not do much with "IProducer.GetInputs".

I think the right fix for this is that CompletionListener needs to remember whether it is still listening. Trying to remove the same listener twice was never supposed to be able to happen.

PR #3847, opened 2026-01-21
Improving label layout algorithm

rita-gwen:
The main goal is to replace the existing algorithm with simulated annealing and optimize it for user-friendly UI experience.

### DIFF
--- a/pwiz_tools/Skyline/Controls/Graphs/SummaryRelativeAbundanceGraphPane.cs
+++ b/pwiz_tools/Skyline/Controls/Graphs/SummaryRelativeAbundanceGraphPane.cs
@@ -646,11 +646,11 @@ namespace pwiz.Skyline.Controls.Graphs
             {
                 if (!Settings.Default.GroupComparisonSuspendLabelLayout)
                 {
-                    AdjustLabelSpacings(_labeledPoints);
+                    StartLabelLayoutAsync(_labeledPoints);
                 }
                 else
                 {
-                    AdjustLabelSpacings(_labeledPoints, _labelsLayout);
+                    StartLabelLayoutAsync(_labeledPoints, _labelsLayout);
                 }
                 if (_labeledPoints.Any())
                     _labelsLayout = GraphSummary.GraphControl.GraphPane.Layout?.PointsLayout;


rita-gwen:
The changes so far:
* Introduced simulated annealing label placement with helper utilities (allowed/clamping, screen↔label conversions, base/pair costs) and integrated via GraphPane.AdjustLabelSpacings -> PlaceLabelsSimulatedAnnealing.
* Added per-point base cost and pairwise cost caching so move steps reuse prior contributions instead of recomputing them twice each iteration.
* Inlined hotspot helpers in PairCost (rect construction, area calc, bounding-box overlap check) and only build vectors/do intersection when boxes overlap.
* Pruned the old sequential PlaceLabel search path and its constants; EvaluateLabelBaseCost is now purely single-label (no crossover), with all crossover handled in PairCost.
* Added layout signature caching to skip redundant annealing runs when the chart/axes/points state hasn’t changed (guards duplicate zoom-triggered calls).

pwiz_tools/Skyline/Controls/Graphs/LabelLayoutRunner.cs
```diff
@@ -0,0 +1,545 @@
...
+        private LabelLayoutRequest _pendingRequest;
+
+        private sealed class LabelLayoutWorkItem
+        {
+            public int RequestId;
+            public GraphPane Pane;
+            public LabelLayout LabelLayout;
+            public List<LabeledPoint> VisiblePoints;
+            public List<LabeledPoint.PointLayout> SavedLayout;
+            public CancellationTokenSource Cancellation;
```
nickshulman: This should be a CancellationToken instead of a CancellationTokenSource because work items only need to check whether they are cancelled; they don't actually have to tell anything else to be cancelled.

pwiz_tools/Skyline/Controls/Graphs/LabelLayoutRunner.cs
```diff
@@ -0,0 +1,545 @@
...
+using System.Linq;
+using System.Threading;
+using pwiz.Common.Collections;
+using pwiz.Common.SystemUtil;
+using pwiz.Skyline.Util.Extensions;
+using ZedGraph;
+
+namespace pwiz.Skyline.Controls.Graphs
+{
+    internal sealed class LabelLayoutRunner
```
nickshulman, line 33: I would recommend that you give this class a constructor where you pass in the things that will never change during the lifetime of the LabelLayoutRunner (i.e. the ZedGraphControl, and maybe the GraphPane).
If you do this, then methods like "Start" do not need to take as many arguments.

pwiz_tools/Skyline/Controls/Graphs/LabelLayoutRunner.cs
```diff
@@ -0,0 +1,545 @@
...
+    {
+        private const int LABEL_LAYOUT_PROGRESS_INTERVAL_MS = 200;
+        private const int LABEL_LAYOUT_DEBOUNCE_MS = 100;
+        private const int MAX_LABELS_PER_CELL = 2;
+        private const double MAX_LABEL_AREA_RATIO = 0.3;
+
+        private Thread _labelLayoutWorkerThread;
+        private CancellationTokenSource _labelLayoutCts;
+        private IProgressStatus _labelLayoutStatus;
+        private int _labelLayoutRequestId;
```
nickshulman: I recommend changing this to a LabelLayoutRequest instead of an integer.
The problem with using integers to uniquely identify your objects is that it's very easy to accidentally use the wrong integer since all integers look the same.
If possible, it's generally better to use the object itself as its own unique identifier.

rita-gwen:
Addressed the review comments in the latest commit:

**Redundant `Start()` overload (Nick #1)**: Merged the verbose overload into the `ZedGraphControl` one — the long-parameter version was only ever called from the shorter one, so the indirection was unnecessary. All callers already pass a `ZedGraphControl`, so the derived values (`GraphPane`, `Invalidate`, `IsDisposed`, etc.) are computed directly in the single `Start()` method.

**Numeric request ID (Nick #2)**: Replaced `_labelLayoutRequestId` (int) and `RequestId` fields on both `LabelLayoutRequest` and `LabelLayoutWorkItem` with a `_currentRequest` reference field on the runner. Staleness is now detected by `ReferenceEquals(request, _currentRequest)` — each `StartDebounced` call sets `_currentRequest` to the request it is processing, so any in-flight worker whose captured `request` no longer matches knows it has been superseded. No `Equals`/`GetHashCode` overrides needed.

**`CancellationTokenSource` on work item (Nick #3)**: `LabelLayoutWorkItem.Cancellation` is now a `CancellationToken`. The runner (owner of the `CancellationTokenSource`) still calls `.Cancel()`; the work item and `WorkerProgress` only ever read the token, so giving them the source was unnecessary exposure. `WorkerProgress` action signature simplified from `Action<int, int>` to `Action<int>` accordingly.

pwiz_tools/Shared/zedgraph/ZedGraph/LabelLayout.cs
```diff
@@ -99,63 +110,133 @@ private void FillDensityGrid()
...
                         {
-                            cell._density += intersect.Height * intersect.Width;
+                            var intersect = RectangleF.Intersect(markerRect, cell.Bounds);
+                            if (intersect != Rectangle.Empty)
+                            {
+                                cell.Density += intersect.Height * intersect.Width;
+                                // TODO(ritach): Uncomment to enable per-cell sampling in SamplePointsByDensityGrid.
+                                // Currently cuts TestLabelLayoutDeterminism labels from 32 to 7 -- tune sampling
+                                // parameters (MAX_LABELS_PER_CELL, MAX_LABEL_AREA_RATIO) before enabling.
+                                // cell.PointCount++;
```
brendanx67: `GridCell._pointCount` was never incremented in `FillDensityGrid`, so the per-cell sampling cutoff in `SamplePointsByDensityGrid` was inert (`cutoff` was always `1.0`). Enabling it (uncommenting `cell.PointCount++`) cuts `TestLabelLayoutDeterminism` labels from 32 to 7, which suggests `MAX_LABELS_PER_CELL` and/or `MAX_LABEL_AREA_RATIO` need retuning now that this filter is active alongside `areaSamplingRate`. The two filters multiply together (`cutoff * areaSamplingRate`), so the combined effect is much more aggressive than either alone.

Please tune the sampling parameters and update the test expectations before uncommenting this line.

PR #3858, opened 2026-01-22
Updating the CV terms, and the scripts that actually perform that task.…

bspratt:
… In particular, adding error checking for .obo downloads - we'd missed some updates to the units obo due to a name change in its git repo.

This adds several new machine types for Shimadzu and Thermo, and some useful new unit types like "Bar".

### DIFF
--- a/pwiz_tools/Shared/CommonUtil/SystemUtil/ProcessEx.cs
+++ b/pwiz_tools/Shared/CommonUtil/SystemUtil/ProcessEx.cs
@@ -35,35 +35,29 @@ namespace pwiz.Common.SystemUtil
         /// <summary>
         /// Returns true iff the process is running under an OS and on volume(s) that support windows 8.3 format conversion
         /// </summary>
-        private static bool? _canConvertUnicodePaths;
         private static Dictionary<string, bool> _volumesTested = new Dictionary<string, bool>();
 
-        public static bool? CanConvertUnicodePaths
+        public static bool CanConvertUnicodePaths
         {
-            set => _canConvertUnicodePaths = value;
             get
             {
-                if (!_canConvertUnicodePaths.HasValue)
+                if (IsRunningOnWine)
                 {
-                    if (IsRunningOnWine)
-                    {
-                        _canConvertUnicodePaths = false;
-                    }
-                    else
-                    {
-                        // Systems may have 8.3 conversion enabled on some volumes but not others
-                        // So don't assume we can convert unicode paths unless we can actually do it everywhere we might need to
-                        _canConvertUnicodePaths =
-                            CanConvertUnicodePathsInDirectory(Path.GetTempPath()) &&
-                            CanConvertUnicodePathsInDirectory(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)) &&
-                            CanConvertUnicodePathsInDirectory(Directory.GetCurrentDirectory()) &&
-                            CanConvertUnicodePathsInDirectory(PathEx.GetDownloadsPath());
-                    }
+                    return false; // Never supported on Wine
                 }
-                return _canConvertUnicodePaths;
+                else
+                {
+                    // Systems may have 8.3 conversion enabled on some volumes but not others
+                    // So don't assume we can convert Unicode paths unless we can actually do it everywhere we might need to
+                    return
+                        CanConvertUnicodePathsInDirectory(Path.GetTempPath()) &&
+                        CanConvertUnicodePathsInDirectory(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location)) &&
+                        CanConvertUnicodePathsInDirectory(Directory.GetCurrentDirectory()) &&
+                        CanConvertUnicodePathsInDirectory(PathEx.GetDownloadsPath());
+                }
             }
         }
-        static public bool CanConvertUnicodePathsInDirectory(string baseDir)
+        private static bool CanConvertUnicodePathsInDirectory(string baseDir)
         {
             var volume = Path.GetPathRoot(baseDir);
             if (_volumesTested.TryGetValue(volume, out var result))
--- a/pwiz_tools/Skyline/Model/Tools/ToolDescription.cs
+++ b/pwiz_tools/Skyline/Model/Tools/ToolDescription.cs
@@ -486,7 +486,7 @@ namespace pwiz.Skyline.Model.Tools
 
         public static string GetToolsDirectoryBasis()
         {
-            return (Program.UnitTest && ProcessEx.CanConvertUnicodePaths.Value) ? @"Tööls" : @"Tools"; // Helps catch Unicode path issues on Windows
+            return (Program.UnitTest && !Program.DoNotTestUnicodeHandling) ? @"Tööls" : @"Tools"; // Helps catch Unicode path issues on Windows
         }
 
         /// <summary>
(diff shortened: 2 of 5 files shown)


pwiz/data/common/cvgen.cpp
```diff
@@ -185,9 +184,12 @@ void writeHpp(const vector<OBO>& obos, const string& basename, const bfs::path&
         {
             for(const string& synonym : term.exactSynonyms)
             {
+                auto synonym_enum_name = enumName(term.prefix, synonym, term.isObsolete);
+                if (eName == synonym_enum_name) // Should not happen, but just in case
+                    cerr << "Warning: synonym enum name collision for term id " << term.id << ": " << synonym << ": " << eName << endl;
                 os << ",\n\n"
```
chambm, line 187: Should this be an else block? A name collision wouldn't compile.

pwiz/data/common/cvgen.cpp
```diff
@@ -154,10 +150,13 @@ void writeHpp(const vector<OBO>& obos, const string& basename, const bfs::path&
...
-        os << "// [" << filename << "]\n";
-        string fileDefine = bal::to_upper_copy(filename);
-        transform(fileDefine.begin(), fileDefine.end(), fileDefine.begin(), toAllowableChar);
+        string oboFilename = bfs::path(obo->filename).filename().string();
+        os << "// [" << oboFilename << "]\n";
+        string upper = bal::to_upper_copy(oboFilename);
+        string fileDefine;
+        fileDefine.reserve(upper.size());
+        for (char c : upper)
+            fileDefine.push_back(toAllowableChar(c));
```
chambm: What's the effect of this change?

pwiz/data/common/obo.cpp
```diff
@@ -304,7 +376,10 @@ Term parseTerm(istream& is)
     Term result;
 
     for (string line; getcleanline(is,line) && !line.empty();)
-        parseTagValuePair(line, result);
+    {
+       // if (!isIgnoredRelationship(result.prefix)) // PEFF etc terms are not interesting for describing mass spectrometry data
```
chambm: Why is this commented out? If we keep the commented out condition please add a comment why it's commented out.

pwiz/data/common/obo.cpp
```diff
@@ -320,8 +395,8 @@ void parseStanza(istream& is, OBO& obo)
     {
         Term term = parseTerm(is);
 
-        // ignore UO terms in MS obo
-        if (bal::iends_with(obo.filename, "psi-ms.obo") && term.prefix == "UO")
+        // ignore UO and non-followed terms in MS obo
+        if (/*isIgnoredRelationship(term.prefix) || term.prefix.empty() ||*/ (bal::iends_with(obo.filename, "psi-ms.obo") && term.prefix == "UO"))
```
chambm: Ditto

pwiz/data/vendor_readers/Shimadzu/Reader_Shimadzu_Detail.cpp
```diff
@@ -29,49 +29,88 @@ namespace msdata {
...
-    {"AXIMA PERFORMANCE", MS_AXIMA_Performance_MALDI_TOF_TOF, Contains}, // predicted
-    {"AXIMA CONFIDENCE", MS_AXIMA_Confidence_MALDI_TOF, Contains}, // predicted
-    {"AXIMA ASSURANCE", MS_AXIMA_Assurance_Linear_MALDI_TOF, Contains}, // predicted
-    {"QP2010SE", MS_GCMS_QP2010SE, Contains}, // predicted
-    {"IT-TOF", MS_LCMS_IT_TOF, Contains}, // predicted
-    {"LCMS", MS_Shimadzu_Scientific_Instruments_instrument_model, Exact}, // instrument model not specified, use different fallback type to avoid error
-    // need CVID {"9050", , Contains},
+    {"2010E", MS_LCMS_2010EV}, // predicted
+    {"2010A", MS_LCMS_2010A}, // predicted
+    {"2010", MS_LCMS_2010},
```
chambm: Anything new here that hasn't been seen in an actual file should be `// predicted`. What's the reason for switching to a "longest match" approach?

pwiz/data/common/obo.hpp
```diff
@@ -30,7 +30,6 @@
 #include <set>
 #include <map>
 #include <string>
-#include <limits>
```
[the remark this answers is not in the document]
bspratt: (in case CoPilot learns from responses) 
This is not useful feedback. If that include was necessary then removing it would break the build.

pwiz/data/common/cvgen.cpp
bspratt: It's just to help the development process. Catching it at compile time still works but this makes the problem easier to understand.

pwiz/data/common/obo.cpp
bspratt: Oops, that's just development/debugging cruft. Good catch, thanks, will remove

pwiz/data/vendor_readers/Shimadzu/Reader_Shimadzu_Detail.cpp
bspratt: Eliminating list order requirement just makes it harder to screw up, as I initially did while adding new items.

I'll add // predicted as needed, thanks.

pwiz/data/common/cvgen.cpp
bspratt: No functional change -` transform(fileDefine.begin(), fileDefine.end(), fileDefine.begin(), toAllowableChar)` is tidier, of course, but unfortunately throws an exception when passed non-ASCII strings in a debug build.

pwiz/data/common/obo.cpp
bspratt: yup, thanks

pwiz/data/vendor_readers/Shimadzu/Reader_Shimadzu_Detail.cpp
chambm: I just realized the matching change breaks the fallback behavior. Previously an exact match to "LCMS" (seen in the wild) would not give a warning/error because there's an "instrument" specified, but it's not useful to match to a specific instrument model, so there's no sense giving a warning asking the user to report the name so we can add a mapping for it. With your change, "LCMS" and "LCMS 2099" would both match to the fallback case, but in the latter case we would actually want a warning to be issued. I suppose you could hardcode the exact matches you want to skip warnings for with your approach (currently just "LCMS").

bspratt: You're 100% correct - I had convinced myself those were all "Contains", but that one is the sole "Exact". It's probably still good that "Contains" not be declaration-order-dependent, but I have certainly broken the "exact" case. Will fix.

pwiz/data/common/cvgen.cpp
chambm: Hmm. Is this actually going to handle multibyte characters properly? Remember `char` in C/C++ is a byte, not a 2-byte wchar_t/C# char. So won't it try to treat each byte of a multibyte character as a separate output char? And wait, this is just the OBO filename. What circumstance did you have where the OBO filename had non-ASCII characters? 🤨

chambm: And my comment goes for transform as well. I don't think it was intended to handle non-ASCII OBO filenames!

bspratt: yeah, it's probably pointless, so long as nobody introduces fancier obo filenames. I'll back it out. I think it was just CoPilot being "helpful".

Issue #3814, opened 2026-01-16
IndexOutOfRangeException in ImportTransitionListColumnSelectDlg.AssociateProteins with malformed transition list


brendanx67:
## Problem

When importing a transition list with "Associate Proteins" checkbox enabled, an `IndexOutOfRangeException` is thrown if a data row has fewer columns than expected.

**Exception report**: https://skyline.ms/home/issues/exceptions/announcements-thread.view?rowId=73747

**Frequency**: 8 reports from 6 users since 2025-05-28

## Stack Trace

```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at pwiz.Skyline.FileUI.ImportTransitionListColumnSelectDlg.<>c__DisplayClass43_0.<AssociateProteins>b__2(String[] fields)
      in ImportTransitionListColumnSelectDlg.cs:line 299
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)
   at pwiz.Skyline.FileUI.ImportTransitionListColumnSelectDlg.<>c__DisplayClass43_0.<AssociateProteins>b__0(ILongWaitBroker progressMonitor)
```

## Root Cause

At line 317 in `AssociateProteins()`:

```csharp
peptides.AddRange(_originalLines.Skip(hasHeaders ? 1 : 0).Take(count)
    .Select(line => line.ParseDsvFields(Importer.Separator))
    .Select(fields => fields[_originalPeptideIndex]));  // No bounds check!
```

The code assumes all rows have at least `_originalPeptideIndex + 1` columns, but:
- Trailing empty columns may be trimmed by the CSV parser
- Some rows may be malformed or incomplete
- User may have pasted partial data

## Suggested Fix

Add bounds checking before accessing the field:

```csharp
.Select(fields => fields.Length > _originalPeptideIndex 
    ? fields[_originalPeptideIndex] 
    : string.Empty)
```

Or filter out short rows:
```csharp
.Where(fields => fields.Length > _originalPeptideIndex)
.Select(fields => fields[_originalPeptideIndex])
```

## Environment

- **Version**: 25.1.0.237-519d29babc (also seen in earlier versions)
- **File**: `pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs`
- **Owner**: Brian Pratt (per git blame)

PR #3869, opened 2026-01-24
Fix for github issue #3814: …

bspratt:
…When importing a transition list with "Associate Proteins" checkbox enabled, an IndexOutOfRangeException is thrown if a data row has fewer columns than expected.


bspratt:
@chambm  can you give this the blessing? It involves some of your work, I think.

Issue #3868, opened 2026-01-24
SkylineTester: InvalidOperationException in TestInfo constructor when loading tests


brendanx67:
## Description

SkylineTester shows a MessageBox with an exception frequently during test loading. The exception occurs in `TestInfo..ctor` when `ProcessEx.CanConvertUnicodePaths` returns null and `.Value` is accessed.

## Stack Trace

```
at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
at TestRunnerLib.TestInfo..ctor(Type testClass, MethodInfo testMethod, MethodInfo testInitializeMethod, MethodInfo testCleanupMethod) in C:\pwiz\pwiz_tools\Skyline\TestRunnerLib\RunTests.cs:line 87
at TestRunnerLib.RunTests.<GetTestInfos>d__126.MoveNext() in C:\pwiz\pwiz_tools\Skyline\TestRunnerLib\RunTests.cs:line 1228
at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
at System.Linq.OrderedEnumerable`1.<GetEnumerator>d__1.MoveNext()
at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
at SkylineTester.SkylineTesterWindow.BackgroundLoad(Object sender, DoWorkEventArgs e) in C:\pwiz\pwiz_tools\Skyline\SkylineTester\SkylineTesterWindow.cs:line 385
```

## Root Cause

In `TestInfo` constructor (lines 68-69):

```csharp
ProcessEx.CanConvertUnicodePaths = (RunTests.GetAttribute(testMethod, "NoUnicodeTestingAttribute") != null) ? false : (bool?)null;
DoNotUseUnicode = !ProcessEx.CanConvertUnicodePaths.Value; // <-- Throws if getter returns null
```

Line 68 sets `_canConvertUnicodePaths` to null, expecting the getter to calculate a value. But if the calculation fails (e.g., `PathEx.GetDownloadsPath()` returns null, or path operations throw), `_canConvertUnicodePaths` remains null and `.Value` throws `InvalidOperationException`.

## Immediate Fix

Change line 69 to use null-coalescing:
```csharp
DoNotUseUnicode = !(ProcessEx.CanConvertUnicodePaths ?? false);
```

[...]

PR #3875, opened 2026-01-26
Fix for https://github.com/ProteoWizard/pwiz/issues/3868 - Unicode path test support should not be a process-level static.…

bspratt:
… Cached information about which system volumes can support 8.3 conversion is still suitable as a process-level static but how that information is used depends on test attributes.


brendanx67:
It was in a wonky place before inside of an IEnumerable call. So that a simple ToArray() would have broken the entire implementation. I posted this and reviewed it with Claude Code. This implementation puts the setting of the global test seam Program.DoNotTestUnicodeHandling in the right place, inside of AbstractUnitTest.MyTestInitialize().

PR #3885, opened 2026-01-28
Fix for the "layout too big" exception

rita-gwen:
This is the original exception reported by Mike Riffle: https://skyline.ms/home/issues/exceptions/announcements-thread.view?rowId=73063
Bug fix implemented. Replaced Uri.Encode with a dedicated encoding helper function. Fixed the Settings.Default.GroupComparisonAvoidLabelOverlap behavior on application exit.


pwiz_tools/Skyline/Controls/Graphs/GraphSummary.cs
```diff
@@ -294,7 +295,7 @@ protected override string GetPersistentString()
             var panelLayouts = GraphPanes.OfType<ILayoutPersistable>().FirstOrDefault()?.GetPersistentString();
             if (panelLayouts != null)
             {
-                res = res + '|' + Uri.EscapeDataString(panelLayouts);
+                res = res + '|' + TextUtil.EscapeForXmlAttribute(panelLayouts);
```
nickshulman: Why are you switching from URL encoding to XML escaping?
I think the reason we were URL encoding before was so that "|" would get replaced. It might have been overkill since that might have been the only character that needed replacing.
I'm not sure that XML escaping makes any sense here, since the value is going to end up in an XML attribute anyway which will be escaped in the .sky.view file.

I wonder if it would be better to just put a try/catch around the Uri.EscapeDataString call and swallow the exception if it's too long. Do we really want to be writing 32K to the .sky.view file when there are that many labeled points being displayed? Should we maybe only be serializing these label positions if the user has done something to customize where they are?

pwiz_tools/Skyline/Controls/Graphs/SummaryRelativeAbundanceGraphPane.cs
```diff
@@ -613,14 +614,13 @@ private void this_AxisChangeEvent(GraphPane pane)
...
                     AdjustLabelSpacings(_labeledPoints);
-                    _labelsLayout = GraphSummary.GraphControl.GraphPane.Layout?.PointsLayout;
                 }
                 else
                 {
                     AdjustLabelSpacings(_labeledPoints, _labelsLayout);
-                    _labelsLayout = GraphSummary.GraphControl.GraphPane.Layout?.PointsLayout;
-                    //UpdateConnectors();
                 }
+                if (_labeledPoints.Any() && GraphSummary.GraphControl.GraphPane.Layout != null)
```
nickshulman: Is the purpose of adding "_labeledPoints.Any()" to this if statement to make it so "_labelsLayout" keeps its old value?
I'm a little confused about why "_labelsLayout" is a static variable, since that seems like it would make it easy for it to be inconsistent with the data that is being displayed in the graph.

PR #3902, opened 2026-01-29
Improved error message when user tries to open a mass spec data file instead of a Skyline file

### LINKED ISSUE #3881: Improved error message when user tries to File>Open a mass spec data file in Skyline
## Summary

A new user hasn't grasped the fundamental purpose of Skyline and is trying to open a mass spec file rather than importing it atop a targets list.

Improve the error message to explain that Skyline needs targets before it can extract chromatograms. Could also suggest that if the user simply wants to inspect the data, they can download SeeMS.

**Context:** https://skyline.ms/announcements/home/support/thread.view?rowId=62846

## Getting Started

Use /pw-startissue <number> to begin work.

---
Transferred from: skyline.ms Issue #982

bspratt:
## Summary
- When a user tries to open a mass spec data file (e.g. .mzML, .raw, .wiff) via File > Open in the UI,  or with `--in=` on the command line, Skyline now shows a helpful message explaining that the file is a data file rather than a Skyline document
- The message guides users toward File > Import > Results, File > Search, or SeeMS as appropriate alternatives
- Falls back to the existing generic "not a Skyline document" message for unrecognized file types

Fixes #3881

## Test plan
- [x] `TestSkylineCmd` - verifies command-line path shows friendly message for .mzML files
- [x] `TestDocumentFileLocking` - verifies UI path shows friendly message for .mzML files
- [x] `CodeInspection` - no new violations
- [x] Manual test of command-line with .mzML file
- [x] Manual test of File > Open with .mzML file in UI

### DIFF
--- a/pwiz_tools/Skyline/Model/Files/SkylineFile.cs
+++ b/pwiz_tools/Skyline/Model/Files/SkylineFile.cs
@@ -111,14 +111,14 @@ namespace pwiz.Skyline.Model.Files
                 var peptideSettings = document.Settings.PeptideSettings;
                 if (peptideSettings is { HasLibraries: true })
                 {
-                    var librarySpecs = peptideSettings.Libraries.LibrarySpecs.ToList();
+                    var librarySpecs = peptideSettings.Libraries.LibrarySpecs.Where(s => s != null).ToList();
                     if (librarySpecs.Count == 1)
                     {
                         // Single library - show without folder, using same format as other library types
                         var file = SpectralLibrary.Create(documentFilePath, librarySpecs[0], includeTypePrefix: true);
                         list.Add(file);
                     }
-                    else
+                    else if (librarySpecs.Count > 1)
                     {
                         // Multiple libraries - show in folder
                         var files = librarySpecs.Select(library => SpectralLibrary.Create(documentFilePath, library)).ToList();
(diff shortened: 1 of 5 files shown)


bspratt:
UI message:
image
cmd message:
image

brendanx67 (commented):
I pushed a commit with three changes:

**1. Logic fix in `IsSkylineFile`** — The original PR added `.sky`/`.sky.zip` extension checks to the outer `if` condition, which meant files with those extensions that lacked `<srm_settings>` would pass validation silently (returning `true` with no explanation). This is a regression from the original behavior — `IsSkylineFile` has always rejected any file missing `<srm_settings>`, regardless of extension. Users do rename files to `.sky` hoping Skyline will open them.

I restructured the logic so the outer condition stays as the original `!probeString.Contains("<srm_settings>")`. The extension check now only gates whether to show the mass-spec-specific message vs. the generic one. The generic "not a Skyline document" message always fires as a fallback, restoring the original invariant. Also switched from `path.EndsWith()` to `PathEx.HasExtension()` for consistency with the codebase.

**2. Reworded the user-facing message** — Led with `File > Search` (the most likely action for someone trying to open a data file), moved `File > Import > Results` to "otherwise", and fixed the URL from `proteowizard.sourceforge.io` to `proteowizard.org`.

**3. Added regression tests** — Both `TestDocumentFileLocking` (UI) and `TestSkylineCmd` (command line) now test opening a `.sky` file containing mzML content, verifying the generic "not a Skyline document" error appears. These tests fail with the original PR logic and pass with the fix.

PR #3915, opened 2026-01-31
Fixed ArgumentOutOfRangeException in transition list protein tooltip

bspratt:
## Summary

* Fixed `ArgumentOutOfRangeException` in `dataGrid_MouseMove` when hovering over row 0 (combo-box placeholder) or the trailing "..." row while Associate Proteins is active in the Import Transition List dialog
* Root cause: guard `e.RowIndex >= 0` allowed row 0 through, causing `_proteinList[-1]` access
* Extracted `proteinIndex` variable for clarity and added bounds check against `_proteinList.Count`
* Added `TestMouseMoveToGridCell` test support method and test coverage in `InsertTest`

Fixes exception [#73851](https://skyline.ms/home/issues/exceptions/announcements-thread.view?entityId=e3b9ced7-df59-103e-9ec9-22f535564afc&_anchor=73851)

## Test plan

- [x] InsertTest - verifies mouse move on row 0 and last row no longer crashes

### DIFF
--- a/pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
+++ b/pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
@@ -490,6 +490,7 @@ namespace pwiz.Skyline.FileUI
 
             for (var i = 0; i < numColumns; i++)
             {
+                dataGrid.Columns[i].SortMode = DataGridViewColumnSortMode.NotSortable;
                 if (headers != null && i < headers.Length)
                 {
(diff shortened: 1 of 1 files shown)


brendanx67 (approved):
Looks good. I added disabling sorting on the column headers because that really messes with the tips and even sorts the column of "..." cells under the type combo boxes into the list and a real protein under the combo boxes. I couldn't get this one to happen with manual mouse overs, because I couldn't figure out how to get that to happen in row 0 with the combo boxes over it. But, I trust that it happened in realworld use. And this looks like the right fix for it.

PR #3935, opened 2026-02-02
Fixed protoc path resolution in Claude-initiated Skyline build

bspratt:
## Summary

* Used `.\protoc` in both `generatecode.bat` files to ensure the local `protoc.exe` is found during MSBuild pre-build events, where the current directory is not searched for executables by default

## Test plan

- [x] CodeInspection - passes with all three changes applied
- [x] Full solution build - clean output with no protoc warnings


nickshulman (commented):
You wouldn't happen to have set the "NoDefaultCurrentDirectoryInExePath" environment variable on your computer, would you?
That's the only thing Claude could think of that would cause a .bat file on Windows to not look in the current directory when looking for an executable.

PR #1048, opened 2020-05-06
Crosslinking

nickshulman:
Support for crosslinked peptides in Skyline

You can designate that a modification is a crosslinker using a checkbox on EditStaticModDlg.
You can create a crosslinked peptide using the Edit Modifications dialog
The "Modified Sequence" column uses the new crosslinked peptide sequence (e.g. "DDSPDLPKLKPDPNTLC[+57]DEFK-[+138.1@10,4]-SLGKVGTR")
For linked peptides, the amino acid of the fragment location is displayed to either the right or the left of the ion type, like this:
image

Known issues:
1. Skyline looks up peptide sequences in the .blib file using the syntax where the PeptideModSeq has a modification masses that include the mass of the linked peptide, instead of using the new sequence.

PR #3938, opened 2026-02-03
Translate graph context menus for UI mode (fixes #3937)

### LINKED ISSUE #3937: Graph context menus not UI mode aware - show Peptide in small molecule mode
## Summary

The "View > Retention Times > Molecule Comparison" menu navigation correctly displays "Molecule Comparison", but the graph's right-click context menu and its submenus inconsistently use "Peptide" terminology instead of adapting to the UI mode setting.

**Transferred from**: [skyline.ms Issue #1048](https://skyline.ms/home/issues/issues-details.view?issueId=1048)

## Problems

1. **Right-click context menu items** on RT, peak area, and mass error comparison graphs are not translated for UI mode
2. **"Graph" submenu** shows "Peptide Comparison" instead of "Molecule Comparison" in small molecule/mixed mode

## Root Cause

- The `BuildRTGraphMenu`, peak area, and mass error context menu builders in `SkylineGraphs.cs` did not call `PeptideToMoleculeTextMapper.TranslateMenuItems` after populating context menus
- `TranslateMenuItems` only translates top-level menu items — it does not recurse into `DropDownItems` of submenus, so nested items like "Peptide Comparison" under the "Graph" submenu were never translated

## Fix

Added a `CleanupZedGraphMenuItems` helper method in `SkylineGraphs.cs` that:
1. Removes unwanted ZedGraph menu items (`set_default`, `show_val`)
2. Calls `PeptideToMoleculeTextMapper.TranslateMenuItems` on top-level context menu items
3. Recurses into dropdown submenus to translate nested items (e.g., "Peptide Comparison" → "Molecule Comparison")

Applied to all 3 affected context menu builder methods. This also eliminates code duplication — the remove + translate pattern was repeated identically in each method.

## Files Modified

- `pwiz_tools/Skyline/SkylineGraphs.cs`

## Verification

Tested with `TestMultiSelectPeakAreaGraphAsSmallMolecules` using `PauseTest()` to visually confirm:
- Graph title says "Molecule Comparison"
- Right-click context menu items use "Molecule" terminology
- "Graph" submenu shows "Molecule Comparison" instead of "Peptide Comparison"

## Getting Started
Use /pw-startissue <number> to begin work.

---
Transferred from: skyline.ms Issue #1048

bspratt:
## Summary
- Graph right-click context menus showed "Peptide" terminology in small molecule/mixed UI mode
- Graph right-click context menu submenu items (e.g. "Peptide Comparison" under "Graph") were not translated because `TranslateMenuItems` doesn't recurse into dropdown children
- Extracted shared `RemoveZedGraphMenuItemsAndCheckModeUI` helper to eliminate duplicated cleanup+translate code across 3 context menu builders

## Test plan
- [x] Build succeeds
- [x] `TestMultiSelectPeakAreaGraphAsSmallMolecules` passes
- [x] Visual verification with `PauseTest()`: graph title, context menu items, and "Graph" submenu all show "Molecule" in small molecule mode

Fixes #3937
Transferred from: [skyline.ms Issue #1048](https://skyline.ms/home/issues/issues-details.view?issueId=1048)

### DIFF
--- a/pwiz_tools/Skyline/Util/PeptideToMoleculeTextMapper.cs
+++ b/pwiz_tools/Skyline/Util/PeptideToMoleculeTextMapper.cs
@@ -21,7 +21,6 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
-using System.Runtime.CompilerServices;
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
@@ -42,12 +41,6 @@ namespace pwiz.Skyline.Util
         public class PeptideToMoleculeTextMapper
         {
             public static readonly PeptideToMoleculeTextMapper SMALL_MOLECULE_MAPPER = new PeptideToMoleculeTextMapper(SrmDocument.DOCUMENT_TYPE.small_molecules, null);
-
-            // Stores original (pre-translation) texts for persistent menu items translated in place.
-            // Allows restoring to proteomic text before each translation pass so that switching
-            // back from Mixed/SmallMolecule to Proteomic mode correctly reverts item texts.
-            private static readonly ConditionalWeakTable<ToolStripItem, OriginalMenuItemText> _originalMenuTexts =
-                new ConditionalWeakTable<ToolStripItem, OriginalMenuItemText>();
 
             private readonly List<KeyValuePair<string, string>> TRANSLATION_TABLE;
             private List<ToolTip> ToolTips; // Used when working on an entire form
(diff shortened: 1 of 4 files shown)


pwiz_tools/Skyline/Util/PeptideToMoleculeTextMapper.cs
```diff
@@ -256,6 +256,12 @@ public static void TranslateMenuItems(ToolStripItemCollection items, SrmDocument
                         item.Visible = isActive;
                     }
                     mapper.Translate(activeItems); // Update the menu items that aren't inherently wrong for current UI mode
+
+                    // Recursively translate submenu items (e.g. "Peptide Comparison" under the Graph submenu)
+                    foreach (var menuItem in activeItems.OfType<ToolStripMenuItem>().Where(m => m.HasDropDownItems))
+                    {
+                        TranslateMenuItems(menuItem.DropDownItems, modeUI, extender);
+                    }
```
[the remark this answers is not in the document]
bspratt: Made recursion optional, defaulting to false

bspratt:
@nickshulman I think claude+git lost its mind - no point looking at my latest changes until I figure out why it's claiming more than 2000 files changed

pwiz_tools/Skyline/Util/PeptideToMoleculeTextMapper.cs
```diff
@@ -42,6 +43,12 @@ public class PeptideToMoleculeTextMapper
         {
             public static readonly PeptideToMoleculeTextMapper SMALL_MOLECULE_MAPPER = new PeptideToMoleculeTextMapper(SrmDocument.DOCUMENT_TYPE.small_molecules, null);
 
+            // Stores original (pre-translation) texts for persistent menu items translated in place.
+            // Allows restoring to proteomic text before each translation pass so that switching
+            // back from Mixed/SmallMolecule to Proteomic mode correctly reverts item texts.
+            private static readonly ConditionalWeakTable<ToolStripItem, OriginalMenuItemText> _originalMenuTexts =
```
nickshulman: Neat! I did not know about ConditionalWeakTable. It seems to be very similar to Java's WeakHashMap.

In the case of transposing the text on context menus, we should actually be constructing a brand new context menu each time.
The reason we should create a new context menu each time is that with the current way we are doing things, the menu items get stripped of their children every time a ZedGraphControl gets disposed. That's the reason that we have things like:
```
// Sometimes child menuitems are stripped from the parent
if (transitionsContextMenuItem.DropDownItems.Count == 0)
```
I'll start an email thread about refactoring the context menus in SkylineWindow so that we don't have to worry about menu items being removed when ZedGraphControl's are destroyed which will also mean that we don't need to worry about their original text.

If we did want to make it so that PeptideToMoleculeTextMapper remembered the original text of menu items, then I would recommend change the class PeptideToMoleculeTextMapper, make the dictionary an instance variable, and make it so that each Form constructs a new PeptideToMoleculeTextMapper and hangs onto it for the lifetime of the Form.

bspratt: Sounds like there might be upcoming changes that will conflict with this? Should we just hold off here?

PR #3954, opened 2026-02-05
Fix Copy Data for IMS heatmap to output clean 3-column format

### LINKED ISSUE #3953: "Copy Data" for IMS heatmap display in Full Scan view outputs poorly formatted data
## Summary

User reported issues with the "Copy Data" feature when extracting data from Full Scan viewer for ion mobility spectrometry (IMS) data.

**Priority:** 3  
**Area:** Skyline  
**Reported:** 2025-10-29 by Brian Pratt

## Problem

The output formatting presents multiple problems:

- Columns for m/z and 1/K0 (Vs/cm²) are repeated excessively
- Intensity value assignments are unclear
- Not user-friendly for analysis in R or Python

## Requested Improvement

User requested simplified output format:
- For IMS data: three columns - m/z, 1/K0 (Vs/cm²), and intensity
- For 2D scans: m/z, intensity, and annotation

## Getting Started

---
Transferred from: [skyline.ms Issue #1056](https://skyline.ms/issues/home/software/Skyline/issues-details.view?issueId=1056)

bspratt:
## Summary

- Fixes "Copy Data" for IMS heatmap in Full Scan viewer to output clean, usable data format
- Before: Hundreds of repeated m/z and ion mobility columns (one per intensity-colored curve)
- After: Clean 3-column output (m/z, ion mobility, intensity) suitable for R/Python analysis

## Changes

- `HeatMapGraphPane.GetHeatMapDataForClipboard()` - New method to extract raw data directly
- `HeatMapData.GetAllPoints()` - Expose raw point data for clipboard copy
- `GraphPaneData.TryGetHeatMapData()` - Detect HeatMapGraphPane and use direct extraction
- `FullScanGraphTest` - Extended test to verify 3-column output

## Test plan

- [x] `TestFullScanGraph` passes - verifies heatmap Copy Data has exactly 3 columns
- [x] Manual testing - Copy Data from IMS heatmap produces clean tab-separated data
- [x] CodeInspection passes

Fixes #3953

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/pwiz_tools/Skyline/SkylineNightly/LogFileMonitor.cs
+++ b/pwiz_tools/Skyline/SkylineNightly/LogFileMonitor.cs
@@ -18,12 +18,14 @@
  */
 
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Timers;
+using Microsoft.Diagnostics.Runtime;
 using Timer = System.Timers.Timer;
 
 namespace SkylineNightly
@@ -189,6 +191,7 @@ namespace SkylineNightly
                                 message.Append(_logTail);
                                 message.AppendLine();
                             }
+                            AppendTestRunnerStacks(message);
                             SendEmailNotification(subject, message.ToString());
                         }
                         return;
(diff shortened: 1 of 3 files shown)


pwiz_tools/Skyline/TestFunctional/FullScanGraphTest.cs
```diff
@@ -196,6 +197,15 @@ protected override void DoTest()
...
             WaitForOpenForm<GraphFullScan>();   // For localization testing
 
+            // Test Copy Data output for heatmap - should have 3 columns (m/z, ion mobility, intensity)
+            // not 100+ columns from separate intensity curves (issue #3953)
+            var graphData = GraphData.GetGraphData(SkylineWindow.GraphFullScan.ZedGraphControl.MasterPane);
+            Assert.AreEqual(1, graphData.Panes.Count, "Expected 1 pane in heatmap");
+            Assert.AreEqual(1, graphData.Panes[0].DataFrames.Count, "Expected 1 DataFrame in heatmap");
+            var dataFrame = graphData.Panes[0].DataFrames[0];
+            Assert.AreEqual(3, dataFrame.ColumnCount, "Heatmap Copy Data should have 3 columns (m/z, ion mobility, intensity)");
+            Assert.IsTrue(dataFrame.RowCount > 0, "Heatmap should have data rows");
```
brendanx67: Should we be able to make more ambitious assertions here? It seems like the code could be pretty broken and still pass these assertions.

bspratt: Indeed. I have added checks for actual output values. While doing that I realized that the UI doesn't provide any way to verify that manually, so I added a tooltip that shows the current cursor position (then disappears after 3 seconds of no activity)
image

brendanx67: I have been noticing that recently added custom tooltips like this don't always go away as easily as I would expect. I feel like we need a standardized way of showing a tooltip on a graph pane where we can fix everything in one place. Does such a thing exist, or is everyone rolling their own at this point?

bspratt: I started out with an actual ToolTip here but ended up using Label instead - ToolTip operation does always seem a bit indeterministic. This particular implementation is set to disappear after 3 seconds of no mouse movement - that would be the AutoPopDelay property value if this was an actual ToolTip.

PR #3966, opened 2026-02-09
Fix assumption exception in RetentionTimeRegression.RecalcRegression

nickshulman:
Fixed error doing "Run to Run" regression when replicates may have missing peaks (reported by Alex)

Added "TestRunToRunRecalcRegression" which exercises this scenario


pwiz_tools/Skyline/Model/RetentionTimes/RetentionTimeRegressionGraphData.cs
```diff
@@ -414,7 +414,7 @@ private void Refine(CancellationToken cancellationToken)
                     {
                         validPoints.Add(pt);
                     }
-                    else if (_regressionIncludesMissingValues)
+                    else if (_regressionIncludesMissingValues && !IsRunToRun)
```
brendanx67: I would prefer to see a new getter that uses this logic with a comment that explains it.

PR #3969, opened 2026-02-12
Tell ProteoWizard to centroid all levels above 2 when centroiding MS2

nickshulman:
Fixed MS3 data not centroided for chromatogram extraction (reported by Lilian)


chambm (approved):
An alternate approach would add CLI bindings for IntegerSet, but I don't think it's worth it.

PR #3972, opened 2026-02-13
 Fix ArgumentOutOfRangeException in GraphHelper.GetGraphPane during synchronized zoom

### LINKED ISSUE #3970: ArgumentOutOfRangeException in GraphHelper.GetGraphPane - GraphPaneKeys/PaneList invariant violation
## Exception Report

[**#73957**](https://skyline.ms/home/issues/exceptions/announcements-thread.view?rowId=73957) — 2 reports from 1 user since 2026-01-22

**Fingerprint**: `9a576598ddd670e7`
**Version**: 26.0.9.032-e672b40dd (release branch)

## Stack Trace

```
System.ArgumentOutOfRangeException: Index was out of range.
   at GraphHelper.DisplayState.GetGraphPane(MSGraphControl graphControl, PaneKey graphPaneKey)
        in GraphHelper.cs:line 461
   at GraphHelper.DisplayState.GetOrCreateGraphPane(MSGraphControl graphControl, PaneKey graphPaneKey)
        in GraphHelper.cs:line 474
   at GraphHelper.AddChromatogram(PaneKey paneKey, ChromGraphItem chromGraphItem)
        in GraphHelper.cs:line 352
   at GraphChromatogram.DisplayTransitions(...) in GraphChromatogram.cs:line 1674
   at GraphChromatogram.UpdateUI(Boolean selectionChanged) in GraphChromatogram.cs:line 1070
   at GraphChromatogram.OnDocumentUIChanged(...) in GraphChromatogram.cs:line 613
   ...
   at GraphChromatogram.EndDrag(Boolean commit) in GraphChromatogram.cs:line 3562
   at GraphChromatogram.graphControl_MouseUpEvent(...) in GraphChromatogram.cs:line 3458
```

Triggered by: user drags peak bounds → EndDrag → document change → chromatogram graph UpdateUI → tries to add chromatogram to a pane that doesn't exist.

## Root Cause Analysis

`DisplayState` maintains a `GraphPaneKeys` list that is expected to stay in 1:1 correspondence with `graphControl.MasterPane.PaneList` — `GraphPaneKeys[i]` is the key for `PaneList[i]`. The exception tells us these got out of sync: `BinarySearch` found the key at an index that is valid for `GraphPaneKeys` but beyond the end of `PaneList`.

```csharp
int index = GraphPaneKeys.BinarySearch(graphPaneKey);
if (index >= 0)
{
    return (MSGraphPane) graphControl.MasterPane.PaneList[index];  // BOOM
}
```

### Why a simple bounds check is insufficient

A naive fix (return null when `index >= PaneList.Count`) would cause `GetOrCreateGraphPane` to call `BinarySearch` again, get a **positive** index (the key IS in `GraphPaneKeys`), then compute `~index` on that positive value — producing a negative insert position and another crash. More fundamentally, once the invariant is broken, we don't know which keys map to which panes. Any "repair" at this point is guessing.

### Underlying issue: broken Immutable contract

`DisplayState` derives from `Immutable` but `GraphPaneKeys` is a `List<PaneKey>` — a mutable reference type. When `ImClone` shallow-copies the `DisplayState`, both the original and clone share the **same list object**. Mutations to `GraphPaneKeys` (via `Add`, `Insert` in `GetOrCreateGraphPane`) affect all copies. This violates the `Immutable` contract and makes it very difficult to reason about when and how the two lists diverge.

## Suggested Fix

Two complementary changes:

**1. Fix the Immutable contract.** Either:
- Make `GraphPaneKeys` an `ImmutableList<PaneKey>` and clone the `DisplayState` whenever the list changes, or
- Decide `DisplayState` is not truly immutable and stop deriving from `Immutable`

The current implementation is broken — it claims immutability but mutates shared state.

**2. Add diagnostics to catch the desync at its source.** Tag each pane in `MasterPane.PaneList` with its corresponding `PaneKey` (e.g. via `Tag` property). Add an `Assume.IsTrue(PanesMatchKeys())` assertion that verifies the two lists are in sync. Place these assertions at mutation points (`GetOrCreateGraphPane`, `InsertMsGraphPane`, `SetDisplayState`) so the invariant violation is caught where it happens, not downstream in `GetGraphPane`.

This won't fix the user-facing crash immediately, but it will identify the actual sequence of operations that breaks the invariant, which is needed to find the real fix.

## Files to Change

- `pwiz_tools/Skyline/Controls/Graphs/GraphHelper.cs`

nickshulman:
- Fixed crash where ZoomTo replaced _displayState with a foreign DisplayState from another graph tab via the
  ZoomSynchronizer
  - The foreign DisplayState's GraphPaneKeys did not match this graph's PaneList, causing ArgumentOutOfRangeException
  when indexing into PaneList
  - ZoomTo now just marks ZoomStateValid = true on the graph's own DisplayState instead of replacing it
  - Removed Immutable base class from DisplayState since it is no longer cloned via ImClone
  - Removed unused SynchronizedDisplayState property from ZoomSynchronizer and GetDisplayState from GraphHelper

  Fixes #3970

### DIFF
--- a/pwiz_tools/Skyline/TestFunctional/PeakAreaRelativeAbundanceGraphTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/PeakAreaRelativeAbundanceGraphTest.cs
@@ -360,8 +360,15 @@ namespace pwiz.SkylineTestFunctional
             });
 
             // === Test 4: Undo - verify the restored peptides are recalculated ===
-            RunUI(SkylineWindow.Undo);
-            WaitForConditionUI(() => pane.IsComplete);
+            // Track cached results to investigate whether multiple DocumentChanged events
+            // cause parallel calculations during undo (same pattern as Test 6/PR #3830)
+            using (new ScopedAction(
+                       () => GraphDataCachingReceiver.TrackCaching = true,
+                       () => GraphDataCachingReceiver.TrackCaching = false))
+            {
+                RunUI(SkylineWindow.Undo);
+                WaitForConditionUI(() => pane.IsComplete);
+            }
 
             RunUI(() =>
             {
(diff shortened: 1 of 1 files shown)


brendanx67:
Are you able to add a regression test that would cause this exception with the old code and not with the new code? Were you testing against a document that caused this state? It would be great to add that to the PR.

PR #3982, opened 2026-02-15
Fix memory leak in TestIrtTutorial

### LINKED ISSUE #3980: Memory and handle leaks in iRT tests after TaskScheduler change in Receiver.cs
## Summary

Both `IrtFunctionalTest` and `TestIrtTutorial` began leaking memory and handles starting Feb 11-12, 2026. The regression coincides with PR #3959 (commit `fad394269`), which replaced `CommonActionUtil.SafeBeginInvoke` with `Task.Factory.StartNew` using a captured `TaskScheduler` in `Receiver.cs`.

### IrtFunctionalTest (Performance Tests)
- **13 leak events** over 3 days (Feb 12-14), zero leaks in the preceding 4 weeks
- **First seen**: 2026-02-12 on git `fad394269`
- **Machines**: BRENDANX-UW7 (memory + handles), BRENDANX-UW25, SKYLINE-DEV6
- Memory leaks range from 8,502 to 650,678 bytes per event
- Handle leaks averaging ~2-5 handles per event

### TestIrtTutorial (Nightly x64)
- **10 leak events** over 4 days
- One isolated event on Jan 29 (BRENDANX-UW6, git `e034132cb`) — possibly pre-existing borderline
- Consistent cluster starts **Feb 11** on git `fad394269` — same commit
- **Machines**: BOSS-PC, BRENDANX-UW25, BRENDANX-UW6, BRENDANX-UW7

### Regression Commit

PR #3959 changed `Receiver.cs` (`pwiz_tools/Shared/CommonUtil/SystemUtil/Caching/Receiver.cs`):

```csharp
// Before: lightweight fire-and-forget via Control.BeginInvoke
CommonActionUtil.SafeBeginInvoke(OwnerControl, () => { ... });

// After: Task objects with captured TaskScheduler
Task.Factory.StartNew(action, CancellationToken.None, TaskCreationOptions.None, _taskScheduler);
```

The `Receiver` class bridges background caching/production with UI updates, which iRT regression calculations use heavily.

## Likely Mechanism

Each `Task.Factory.StartNew` call creates a `Task` object that:
- Has finalizer overhead and `CancellationToken` state
- Captures the `Action` closure and holds a reference to the `TaskScheduler`
- Unlike `Control.BeginInvoke`, `Task` objects are not lightweight — they may prevent GC of related objects if not properly completed/disposed

The PR author noted awareness of thread pool leak concerns ("The thread pool which caused us memory leak tracking problems only gets used if you pass null into StartNew"), but the non-thread-pool path via `TaskScheduler` still appears to leak.

## Suggested Investigation

1. Profile `Receiver` usage during iRT calculations to count `Task` allocations
2. Check whether completed `Task` objects are being collected or accumulating
3. Consider whether the `_taskScheduler` reference itself prevents cleanup when the `Receiver` outlives its `OwnerControl`
4. The `HandleDestroyed` event handler disconnects the control, but `_taskScheduler` remains valid (it uses a hidden window) — tasks queued after logical disposal may hold references

nickshulman:
Fixed memory leak in TestIrtTutorial
The switch to using "TaskScheduler" instead of "SafeBeginInvoke" enabled a notification to be sent after the graph window had been disposed. The graph window would end up asking for more data that would never get released.
Also, this change switches to just using "SynchronizationContext" instead of "TaskScheduler.FromCurrentSynchronizationContext" because that makes the code a little simpler, but the actual memory leak fix is not notifying after OwnerControl has been set to null in Receiver.cs.

Fixes #3980

### DIFF
--- a/pwiz_tools/Shared/CommonUtil/SystemUtil/Caching/Receiver.cs
+++ b/pwiz_tools/Shared/CommonUtil/SystemUtil/Caching/Receiver.cs
@@ -74,7 +74,11 @@ namespace pwiz.Common.SystemUtil.Caching
 
         private void BeginInvoke(Action action)
         {
-            _synchronizationContext?.Post(_ => action(), null);
+            _synchronizationContext?.Post(_ =>
+            {
+                if (OwnerControl != null)
+                    action();
+            }, null);
         }
 
         private void OnProductStatusChanged()


brendanx67:
## Review

The fix looks correct and clean. The root cause analysis is sound — `Task.Factory.StartNew` callbacks queued before `Dispose()` could still fire after `OwnerControl` was set to null, causing the graph to request data that would never be released. The `if (OwnerControl != null)` guard inside the posted callback is the right fix.

The switch from `TaskScheduler` to `SynchronizationContext.Post` is a nice simplification — `TaskScheduler.FromCurrentSynchronizationContext()` wraps `SynchronizationContext.Current` internally, so going direct eliminates the `Task` object overhead.

Setting `_synchronizationContext = null` in `Dispose()` is good belt-and-suspenders — even if a callback gets queued after disposal, `?.Post` becomes a no-op.

### On Copilot's suggestions

**Null check on `SynchronizationContext.Current`** — Minor defensive improvement. The constructor already checks `OwnerControl.InvokeRequired` which verifies we're on the UI thread, and WinForms always installs a `WindowsFormsSynchronizationContext` before controls exist. But adding the explicit throw preserves the fail-fast behavior that `TaskScheduler.FromCurrentSynchronizationContext()` used to provide. Harmless to add, not worth blocking on.

**BeginInvoke return value for pending flags** — Over-engineering. After `Dispose()`, `Cache.Unlisten` was already called so no more callbacks arrive, and nobody checks pending flags on a disposed Receiver. Not worth the added complexity.

LGTM.

PR #3984, opened 2026-02-16
Fix Apply Peak to All crash when replicates lack chromatogram data

### LINKED ISSUE #3983: "Apply Peak to All" crashes when some replicates lack chromatogram data at target RT
## Summary

"Apply Peak to All" (chromatogram context menu) crashes with `ArgumentOutOfRangeException` when some replicates don't have chromatogram data at the target retention time. The user sees the crash/exception reporting dialog instead of the operation silently skipping replicates that don't have matching data.

30 reports from 21 users since June 2025. Exception report: [#73976](https://skyline.ms/home/issues/exceptions/announcements-thread.view?rowId=73976)

## Stack Trace

```
ArgumentOutOfRangeException: No peak found at 13.0
  at TransitionGroupDocNode.ChangePeak()        — TransitionGroupDocNode.cs:3013
  at SrmDocument.ChangePeak()                   — SrmDocument.cs:1832
  at PeakMatcher.PeakMatch.ChangePeak()         — PeakMatcher.cs:437
  at PeakMatcher.ApplyPeak()                    — PeakMatcher.cs:164
  at EditMenu.ApplyPeak()                       — EditMenu.cs:886
  at ChromatogramContextMenu.applyPeakAllMenuItem_Click()
```

## Analysis

The `ArgumentOutOfRangeException` throws in `TransitionGroupDocNode.ChangePeak` (line 3013) and `SrmDocument.ChangePeak` (lines 1888-1906) were originally designed for the single-replicate path — when `GraphChromatogram` fires an event to change a peak in the replicate the user is currently viewing. In that context, the data is expected to be present (the user is looking at it), and a mismatch between the UI event and the underlying data is genuinely a programming defect.

However, "Apply Peak to All" (`PeakMatcher.ApplyPeak`) iterates across **all** replicates and applies the peak. There is no expectation that every replicate will have chromatogram data at the target RT — missing data in some replicates is completely normal. The current code in `PeakMatch.ChangePeak` (PeakMatcher.cs:437-439) calls `doc.ChangePeak()` without any protection, so the first replicate with missing data crashes the entire operation.

## Suggested Fix

Wrap the `doc.ChangePeak()` call in `PeakMatch.ChangePeak` (PeakMatcher.cs:437-439) with a try-catch for `ArgumentOutOfRangeException`. On catch, return `doc` unchanged — silently skip replicates that don't have matching data and continue with the rest.

```csharp
// PeakMatcher.cs, PeakMatch.ChangePeak(), around line 437
try
{
    doc = _retentionTime.HasValue
        ? doc.ChangePeak(groupPath, nameSet, filePath, null, _retentionTime.Value, UserSet.TRUE)
        : doc.ChangePeak(groupPath, nameSet, filePath, null, StartTime, EndTime, UserSet.TRUE, null, false);
}
catch (ArgumentOutOfRangeException)
{
    return doc; // Skip replicates without matching chromatogram data
}
```

The throws in `TransitionGroupDocNode.ChangePeak` and `SrmDocument.ChangePeak` should remain as-is — they correctly catch programming errors in the single-replicate UI path.

## Files to Change

- `pwiz_tools/Skyline/Model/PeakMatcher.cs` — `PeakMatch.ChangePeak` method

## Related

- #3949 — Different bug in same code path (AssumptionException when libraries not yet loaded)

Co-Authored-By: Claude Code <noreply@anthropic.com>

brendanx67:
## Summary

* Fixed crash in "Apply Peak to All" when some replicates don't have chromatogram data at the target retention time (30 reports from 21 users)
* Wrapped `doc.ChangePeak()` call in `PeakMatch.ChangePeak` with try-catch for `ArgumentOutOfRangeException`, returning `doc` unchanged to skip replicates without data
* Left the throws in `TransitionGroupDocNode.ChangePeak` and `SrmDocument.ChangePeak` intact — they correctly catch programming errors in the single-replicate UI path

Fixes #3983

## Test plan

- [x] ApplyPeakToAllTest - existing test passes with fix applied

See ai/todos/active/TODO-20260215_apply-peak-crash.md

### DIFF
--- a/pwiz_tools/Skyline/Model/PeakMatcher.cs
+++ b/pwiz_tools/Skyline/Model/PeakMatcher.cs
@@ -434,9 +434,18 @@ namespace pwiz.Skyline.Model
 
                 var groupPath = new IdentityPath(peptideGroup, peptideDocNode.Peptide, nodeTranGroup.Id);
 
-                doc = _retentionTime.HasValue
-                    ? doc.ChangePeak(groupPath, nameSet, filePath, null, _retentionTime.Value, UserSet.TRUE)
-                    : doc.ChangePeak(groupPath, nameSet, filePath, null, StartTime, EndTime, UserSet.TRUE, null, false);
+                try
+                {
+                    doc = _retentionTime.HasValue
+                        ? doc.ChangePeak(groupPath, nameSet, filePath, null, _retentionTime.Value, UserSet.TRUE)
+                        : doc.ChangePeak(groupPath, nameSet, filePath, null, StartTime, EndTime, UserSet.TRUE, null, false);
+                }
+                catch (ArgumentOutOfRangeException)
+                {
+                    // Skip replicates without chromatogram data at the target RT.
+                    // This is normal when Apply Peak to All iterates across all replicates.
+                    return doc;
+                }
 
                 var activeTransitionGroup = (TransitionGroupDocNode) doc.FindNode(groupPath);
                 if (activeTransitionGroup.RelativeRT != RelativeRT.Matching)


brendanx67:
Correction to the closing comment: the fix for issue #3983 was PR #3911 (NullReferenceException in ApplyPeakWithLongWait), not PR #3646 (which fixed a different ArgumentOutOfRangeException involving duplicate precursor m/z). See the final comment on #3983 for details.

Additionally, a new exception report ([#73977](https://skyline.ms/home/issues/exceptions/announcements-thread.view?rowId=73977)) on version 26.0.9.032 — which includes both PR #3646 and PR #3911 — shows that `PeakMatch.ChangePeak` can still hit an unprotected `ArgumentOutOfRangeException` in `SrmDocument.ChangePeak` via a different code path ("Apply Peak Subsequent" rather than "Apply Peak All"). This is now tracked as #3996.

The try-catch fix proposed in this PR remains the right approach for #3996.

PR #4001, opened 2026-02-18
Fix an attempt to use the address of the first element of a zero length array in writeMzMLbExtra

bspratt:
This was causing failures in TC debug builds


chambm (changes requested):
Good catch. There's a case in `void writeMzMLbExtra<BinaryDataArray>` that needs to have the same fix.  Line 1733

bspratt:
I'll have a look, thanks.

PR #4011, opened 2026-02-19
Added ClrMD stack trace capture for hung tests in SkylineNightly

### LINKED ISSUE #3850: SkylineNightly: Capture thread stack traces for hung tests using ClrMD
## Summary

When SkylineNightly detects a hung test (TestRunner not responding for 1 hour), the hang alert email only tells developers which test was running. Developers must RDP to the machine and attach a debugger to diagnose - but by then the process has usually been killed or the machine rebooted.

## Proposed Solution

Use **ClrMD** (Microsoft.Diagnostics.Runtime) to capture managed thread stack traces from the hung TestRunner.exe process before killing it, and include them directly in the hang alert email.

### Why ClrMD?

- **NuGet package** - no tools to install on test machines
- **Attaches externally** - SkylineNightly can attach to TestRunner.exe from outside
- **Returns strings** - stack traces can go directly into the email, no separate files
- **Managed stacks** - gets the .NET call stacks we care about

### Implementation

1. Add `Microsoft.Diagnostics.Runtime` NuGet reference to SkylineNightly
2. When hang is detected (1-hour timeout):
   - Attach to TestRunner.exe using ClrMD
   - Enumerate all managed threads
   - Capture stack traces as text
   - Include in hang alert email
   - Then kill the process

### Additional Improvement: Flush Partial Line on Test Start

Currently, the log only shows the last *completed* test because lines flush on completion. Change to flush the first part of the line when the test starts:

```
[Flush] "[00:24]   1.877 TestSkyp                                       (en)"
... test runs ...
[Append+Flush] "   0 failures, 94.93/61.86/425.5 MB, 175/1632 handles, 4 sec.\n"
```

Same single line in the final log, but checking during a hang shows which test is currently running.

## Context

PR #3731 added inline stack trace capture for RunUI/WaitForConditionUI hangs (from inside the process). This issue extends similar capability to SkylineNightly's external hang detection.

## Alternatives Considered (for reference, not for implementation now)

- **Minidumps**: Infrastructure exists in TestRunnerLib/MiniDump.cs, but creates .dmp files requiring WinDbg analysis - not as actionable as inline stack traces
- **ProcDump**: Requires installing SysInternals tool on all test machines, also creates files

## Value

Transforms hang alerts from "notification to act manually" to "diagnosis included." Developers would see the actual stuck call stacks in the email, enabling faster triage without RDP access.

brendanx67:
## Summary

* Added Microsoft.Diagnostics.Runtime (ClrMD) reference to SkylineNightly project
* When a hang is detected, SkylineNightly now captures managed thread stack traces from the TestRunner.exe process using ClrMD passive attach
* Stack traces are included in the hang alert email, providing diagnostic info when the internal HangDetection watchdog doesn't fire
* Added ClrMD DLL and XML to SkylineNightly.zip artifact packaging (CreateZipInstallerWindow.cs)

Fixes #3850

## Test plan

- [x] Solution build - verified clean build with all projects succeeding
- [ ] Verify TeamCity build produces SkylineNightly.zip containing Microsoft.Diagnostics.Runtime.dll
- [ ] Verify hang alert email includes thread stacks (next real hang occurrence)

### DIFF
--- a/pwiz_tools/Skyline/SkylineNightly/LogFileMonitor.cs
+++ b/pwiz_tools/Skyline/SkylineNightly/LogFileMonitor.cs
@@ -296,7 +296,7 @@ namespace SkylineNightly
                     }
                     catch (Exception ex)
                     {
-                        message.AppendFormat("Failed to capture stacks for PID {0}: {1}", process.Id, ex.Message);
+                        message.AppendFormat("Failed to capture stacks for PID {0}: {1}", process.Id, ex);
                         message.AppendLine();
                     }
                 }
--- a/pwiz_tools/Skyline/SkylineTester/CreateZipInstallerWindow.cs
+++ b/pwiz_tools/Skyline/SkylineTester/CreateZipInstallerWindow.cs
@@ -110,7 +110,6 @@ namespace SkylineTester
                         "SkylineNightly.exe.config",
                         "SkylineNightly.pdb",
                         "Microsoft.Diagnostics.Runtime.dll",
-                        "Microsoft.Diagnostics.Runtime.xml",
                         "Microsoft.Win32.TaskScheduler.dll",
                         "DotNetZip.dll"
                     };


pwiz_tools/Skyline/SkylineNightly/LogFileMonitor.cs
```diff
@@ -248,6 +251,63 @@ private void Log(string message)
...
+                                    message.AppendLine();
+                                }
+
+                                message.AppendLine();
+                            }
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        message.AppendFormat("Failed to capture stacks for PID {0}: {1}", process.Id, ex.Message);
```
nickshulman: I would pass "ex" here instead of "ex.Message" just in case the callstack or inner exception is important for understanding what went wrong.

pwiz_tools/Skyline/SkylineTester/CreateZipInstallerWindow.cs
```diff
@@ -109,6 +109,8 @@ public static void CreateZipFile(string zipPath, bool addTestZipFiles = false)
                         "SkylineNightly.exe",
                         "SkylineNightly.exe.config",
                         "SkylineNightly.pdb",
+                        "Microsoft.Diagnostics.Runtime.dll",
+                        "Microsoft.Diagnostics.Runtime.xml",
```
nickshulman: The .xml file just contains documentation. We don't need to include it in the .zip file.

pwiz_tools/Skyline/SkylineNightly/LogFileMonitor.cs
brendanx67: Makes sense.

pwiz_tools/Skyline/SkylineTester/CreateZipInstallerWindow.cs
brendanx67: Fair enough.

PR #4012, opened 2026-02-19
Fix thermo ms3+ activation

chambm:
Fixed per-precursor activation type reporting for Thermo multi-precursor scans

  * Replaced scan-level activationType() with per-precursor precursorActivationType(index) on ScanInfo
  * Each precursor now stores its own ActivationType via GetActivation(i) in filter parsing
  * Fixed SPS/MSX scans incorrectly inheriting a single activation type for all precursors
  * Updated IT-HCD-SPS test data to reflect correct CID activation on SPS precursors


pwiz_aux/msrc/utility/vendor_api/thermo/RawFile.h
```diff
@@ -396,7 +397,7 @@ struct PWIZ_API_DECL InstrumentData
     // Device suggested label of X axis
     std::string AxisLabelX;
 
-    // Device suggested label of Y axis (name for units of data, such as "�C")
+    // Device suggested label of Y axis (name for units of data, such as "�C")
```
nickshulman: The degree symbol in this comment got corrupted here.

chambm: Geeze, trying to get Copilot to fix that from the web interface was ridiculously ineffective.

PR #4024, opened 2026-02-20
Generalize heatmap Copy Data for all heatmap graph types

### LINKED ISSUE #4021: Generalize heatmap Copy Data for all heatmap graph types
## Problem

Commit 8c6eefb665 fixed "Copy Data" for the ion mobility heatmap (`HeatMapGraphPane`) to output clean 3-column data (x, y, intensity) instead of the 100+ intensity-colored curves. However, the fix only works for `HeatMapGraphPane` instances.

The two other heatmap pane types — `AreaCVHistogram2DGraphPane` and `MassErrorHistogram2DGraphPane` — extend `SummaryGraphPane`, not `HeatMapGraphPane`. They use `HeatMapGraphPane.GraphHeatMap()` as a static helper but are not themselves `HeatMapGraphPane` instances, so `TryGetHeatMapData()` in `GraphPaneData.cs` does not recognize them.

## Solution

Introduce an `IHeatMapDataProvider` interface in `pwiz.MSGraph` and have all three heatmap pane types implement it. Update `GraphPaneData.TryGetHeatMapData()` to check for the interface instead of the concrete `HeatMapGraphPane` type.

## Affected Graph Types

- `HeatMapGraphPane` (ion mobility heatmap) — already working
- `AreaCVHistogram2DGraphPane` (CV vs Log10 Mean Area) — **not working**
- `MassErrorHistogram2DGraphPane` (Mass Error vs m/z or RT) — **not working**

bspratt:
## Summary
- Add `IHeatMapDataProvider` interface so `GraphPaneData.TryGetHeatMapData()` recognizes all heatmap pane types, not just `HeatMapGraphPane`
- Implement the interface on `AreaCVHistogram2DGraphPane` and `MassErrorHistogram2DGraphPane` (in addition to the existing `HeatMapGraphPane`)
- Move shared clipboard extraction logic to `HeatMapData.GetClipboardData()` to avoid code duplication
- Add `ZAxisName` to `HeatMapData` so each heatmap type can label its z-axis column appropriately (Intensity, Frequency, Count)

Fixes #4021

## Test plan
- [x] `TestAreaCVHistograms` — passes, now verifies 3-column Copy Data output with correct column headers and spot-checked values
- [x] `TestMassErrorGraphs` — passes, now verifies 3-column Copy Data output with correct column headers and spot-checked values
- [x] `TestFullScanGraph` — passes (existing ion mobility heatmap Copy Data test, regression check)
- [x] `CodeInspection` — passes

### DIFF
--- a/pwiz_tools/Skyline/TestFunctional/AreaCVHistogramTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/AreaCVHistogramTest.cs
@@ -24,7 +24,6 @@ using System.Linq;
 using System.Windows.Forms;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using pwiz.Skyline.Controls;
-using pwiz.Common.Graph;
 using pwiz.SkylineTestUtil;
 using pwiz.Skyline.Controls.Graphs;
 using pwiz.Skyline.EditUI;
--- a/pwiz_tools/Skyline/TestFunctional/FullScanGraphTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/FullScanGraphTest.cs
@@ -22,7 +22,6 @@ using System.Linq;
 using System.Text;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using pwiz.Common.Chemistry;
-using pwiz.Common.Graph;
 using pwiz.MSGraph;
 using pwiz.Skyline.Alerts;
 using pwiz.Skyline.Controls.Graphs;
--- a/pwiz_tools/Skyline/TestFunctional/MassErrorGraphsTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/MassErrorGraphsTest.cs
@@ -18,7 +18,6 @@
  */
 
 using Microsoft.VisualStudio.TestTools.UnitTesting;
-using pwiz.Common.Graph;
 using pwiz.Skyline.Controls.Graphs;
 using pwiz.Skyline.EditUI;
 using pwiz.Skyline.Model;


pwiz_tools/Shared/Common/Graph/GraphPaneData.cs
```diff
@@ -97,40 +97,42 @@ public GraphPaneData(string title, IEnumerable<DataFrame> dataFrames)
...
-        /// Check if graphPane is a HeatMapGraphPane and extract its data directly.
+        /// Check if graphPane implements IHeatMapDataProvider and extract its data directly.
+        /// This provides clean 3-column output for all heatmap types instead of the
+        /// default curve-based extraction which would output data for each intensity-colored curve.
         /// </summary>
         private static GraphPaneData TryGetHeatMapData(GraphPane graphPane)
         {
-            var heatMapPane = graphPane as HeatMapGraphPane;
-            if (heatMapPane == null)
+            var heatMapProvider = graphPane as IHeatMapDataProvider;
```
nickshulman: I wonder if it would be simpler to move this logic into "CopyGraphDataToolStripMenuItem.CopyGraphData" in the Skyline project.

pwiz_tools/Shared/MSGraph/HeatMapGraphPane.cs
```diff
@@ -19,15 +19,23 @@
...
 
 namespace pwiz.MSGraph
 {
+    /// <summary>
+    /// Interface for graph panes that can provide raw heatmap data for clipboard copy.
+    /// Implementors return 3-column data (x, y, z) instead of the default curve-based extraction.
+    /// </summary>
+    public interface IHeatMapDataProvider
+    {
+        Tuple<string, string, string, IEnumerable<Point3D>> GetHeatMapDataForClipboard();
```
nickshulman: It would probably be simpler to just have this return a HeatMapData object, so you don't have to worry about which Tuple member is which.

PR #4034, opened 2026-02-23
Added GarbageCollectionTracker and fixed DuplicatedPeptideFinder GC leak

brendanx67:
## Summary

* Added WeakReference-based GarbageCollectionTracker that verifies SkylineWindow and SrmDocument are garbage collected after each functional test, preventing silent memory leak regressions
* Fixed static GC leak in `DuplicatedPeptideFinder._lastSearchedDocument` - was holding a strong reference to the last searched SrmDocument indefinitely via `Finders.LST_FINDERS` static list. Changed to `WeakReference<SrmDocument>` to preserve caching while allowing GC
* Every SrmDocument holds SkylineWindow alive via AuditLogEntry undo closures (Skyline.cs ModifyDocument), so the DuplicatedPeptideFinder leak was transitively holding both objects
* Cleared audit log entry caches (`_setSeenEntries`, `_lastLoggedEntries`) in `CleanupAuditLogs()` to break secondary reference chains from test objects through undo closures to SkylineWindow
* Added `PinSurvivors()` method for promoting tracked WeakReferences to strong refs during dotMemory profiling sessions, enabling retention path analysis
* Added single-snapshot dotMemory support (`-MemoryProfileWaitRuns 0`) for faster profiling workflow

## Test plan

- [x] All 395 TestFunctional tests pass (full English run, serial and parallel)
- [x] Previously failing tests verified individually: TestAssociateProteins, TestEncyclopeDiaSearch, TestKoinaSkylineIntegration, TestLibraryBuild, TestLibraryExplorer, TestLibraryExplorerAsSmallMolecules, TestPeakAreaRelativeAbundanceGraph, TestRetentionTimeManager, TestStandardType, UpgradeBasicFunctionalTest, UpgradeCancelFunctionalTest, UpgradeErrorsFunctionalTest
- [x] dotMemory profiling confirmed retention paths and verified fixes

### DIFF
--- a/pwiz_tools/Skyline/Test/ConnectionPoolTest.cs
+++ b/pwiz_tools/Skyline/Test/ConnectionPoolTest.cs
@@ -1,0 +1,194 @@
+/*
+ * Original author: Brendan MacLean <brendanx .at. uw.edu>,
+ *                  MacCoss Lab, Department of Genome Sciences, UW
+ * AI assistance: Claude Code (Claude Opus 4.6) <noreply .at. anthropic.com>
+ *
+ * Copyright 2026 University of Washington - Seattle, WA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+using System;
+using System.IO;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using pwiz.Skyline.Model;
+using pwiz.Skyline.Util;
+using pwiz.SkylineTestUtil;
+
+namespace pwiz.SkylineTest
+{
+    [TestClass]
+    public class ConnectionPoolTest : AbstractUnitTestEx
+    {
+        /// <summary>
+        /// For visual validation of report output. Not really recorded to the test itself.
+        /// </summary>
+        protected override bool IsRecordMode => false;  // Set to true to enable console output of reports for visual inspection
+
+        private void Log(string s)
+        {
+            if (IsRecordMode)
+                Console.WriteLine(s);
+        }
+
+        [TestMethod]
+        public void TestConnectionPoolReportAndTracking()
+        {
+            // Save and restore TrackHistory to avoid order-dependent test behavior
+            using var restoreTracking = new ScopedAction(
+                () => ConnectionPool.TrackHistory = false,
+                () => ConnectionPool.TrackHistory = false);
+
+            var pool = new ConnectionPool();
+
+            // Verify empty pool reports nothing
+            Assert.IsFalse(pool.HasPooledConnections);
+            Assert.AreEqual(string.Empty, pool.ReportPooledConnections());
+
+            // Create fake connections
+            var id1 = new TestConnectionId(@"C:\TestResults\data.skyd");
+            var id2 = new TestConnectionId(@"C:\TestResults\lib.blib");
+            pool.GetConnection(id1, () => new MemoryStream());
+            pool.GetConnection(id2, () => new MemoryStream());
+            Assert.IsTrue(pool.HasPooledConnections);
+
+            // Report should show connection lines with file paths but no event history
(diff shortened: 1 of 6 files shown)


pwiz_tools/Skyline/Model/Find/DuplicatedPeptideFinder.cs
```diff
@@ -26,7 +27,7 @@ namespace pwiz.Skyline.Model.Find
     /// </summary>
     public class DuplicatedPeptideFinder : AbstractDocNodeFinder
     {
-        private SrmDocument _lastSearchedDocument;
+        private WeakReference<SrmDocument> _lastSearchedDocument;
```
nickshulman: SrmDocument has a property on it called "ReferenceId" which changes every time the document changes. It's something you can use instead of a WeakReference to know whether the document you looked at is the same as the current one. Objects in the Document Grid use it to know whether they need to recalculate things.

pwiz_tools/Skyline/Program.cs
```diff
@@ -891,4 +892,14 @@ public void Start(string arg0)
...
     }
+
+    /// <summary>
+    /// Interface for tracking objects via WeakReferences to verify they become
+    /// eligible for garbage collection after test cleanup. Set on
+    /// <see cref="Program.GcTracker"/> during tests; null in production.
+    /// </summary>
+    public interface IGarbageCollectionTracker
+    {
+        void Register(Type type, object target);
```
nickshulman: The signature for this method should probably be:
`void Register<T>(T target);`
because it's less wordy.

pwiz_tools/Skyline/TestRunnerLib/GarbageCollectionTracker.cs
```diff
@@ -0,0 +1,138 @@
...
+        // Strong references to survivors, kept alive for dotMemory inspection
+        // ReSharper disable once CollectionNeverQueried.Local
+        private static readonly List<object> _pinnedSurvivors = new List<object>();
+
+        /// <summary>
+        /// Promotes surviving (not yet GC'd) objects from weak to strong references
+        /// so they appear in dotMemory snapshots for retention path analysis.
+        /// Use instead of Clear() when memory profiling is active.
+        /// </summary>
+        public static void PinSurvivors()
```
nickshulman, line 110: A different way to implement this would be to have this method be called "GetSurvivors" and return the collection of objects that would be put into "_pinnedSurvivors". The caller could hang onto that collection for as long as it needs to (i.e. is that until the "TakeDotMemorySnapshotIfNeeded" call?) and then let go of the collection. The advantage of making this return the collection is that it's then more obvious how long the caller wants the survivors to be held onto.

But, this way works too.

nickshulman (commented):
Looks good to me.

pwiz_tools/Skyline/Model/Find/DuplicatedPeptideFinder.cs
brendanx67: Great suggestion! Switched to `SrmDocument.ReferenceId` instead of `WeakReference` - much cleaner and uses the existing pattern designed for exactly this purpose.

pwiz_tools/Skyline/Program.cs
brendanx67: Done - changed to `void Register<T>(T target)`. Callers now just write `Register(this)` and `Register(docNew)`.

pwiz_tools/Skyline/TestRunnerLib/GarbageCollectionTracker.cs
brendanx67: Good point about making the lifetime explicit. Keeping as-is for now since the survivors need to stay pinned across the snapshot call and the current approach keeps it self-contained, but will revisit if the pattern gets more complex.

pwiz_tools/Skyline/TestRunnerLib/GarbageCollectionTracker.cs
```diff
@@ -0,0 +1,138 @@
...
+    /// Tracks objects via WeakReferences to verify they become eligible for
+    /// garbage collection after test cleanup. Registration happens from
+    /// Skyline code via <c>Program.GcTracker</c>; verification happens in
+    /// RunTests.Run after FlushMemory.
+    ///
+    /// Inspired by a Java pattern of using weak-reference maps to assert
+    /// that primary objects (like the main window and document) are properly
+    /// released after each test, preventing silent memory leak regressions.
+    /// </summary>
+    public static class GarbageCollectionTracker
```
nickshulman: If this were not a static class then it could implement IGarbageCollectionTracker and you wouldn't need GcTrackerAdapter. You could just have a static field "INSTANCE".

PR #4037, opened 2026-02-24
Scoped ConnectionPool.TrackHistory to functional test lifetime

brendanx67:
## Summary

* PR #4033 added `ConnectionPool.TrackHistory = true` in `TestFunctional.RunTest()` but never turned it off
* Since `TrackHistory` is static, it stayed on for all subsequent unit tests in the same process
* Unit tests accumulated `StackTrace` strings in `_history` without clearing, causing 93 tests to report as memory leaks in nightly runs
* Used `ScopedAction` to scope `TrackHistory` to `RunTest()` lifetime, ensuring it is turned off even when exceptions are thrown

## Test plan

- [x] Reproduction test (TestAlertDlg followed by AgilentFormatsTest) passes clean with no leak report
- [x] Previously required 25 iterations to detect leak; now completes Pass 1 in 8 iterations with no leak

### DIFF
--- a/pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
+++ b/pwiz_tools/Skyline/FileUI/ImportTransitionListColumnSelectDlg.cs
@@ -605,6 +605,17 @@ namespace pwiz.Skyline.FileUI
             SetComboBoxText(columns.PrecursorNoteColumn, Resources.PasteDlg_UpdateMoleculeType_PrecursorNote);
             SetComboBoxText(columns.MoleculeNoteColumn, Resources.PasteDlg_UpdateMoleculeType_MoleculeNote);
             SetComboBoxText(columns.MoleculeListNoteColumn, Resources.PasteDlg_UpdateMoleculeType_MoleculeListNote);
+
+            // Set combo boxes for additional fragment-oriented column assignments beyond the first.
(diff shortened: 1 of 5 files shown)


nickshulman (commented):
Do you want setting "TrackHistory" to false to behave like "pause tracking history" so that it maintains the history that it's already saved but just stops collecting information for new streams, or do you want it to be more like "I am done looking at history"?

If you want it to be like "Pause tracking history", then ConnectionPool.ReportPooledConnections should always look in "_history" regardless of the current value of "TrackHistory".

If you want it to be more like "I am done tracking history" then it might be better to give the ConnectionPool two static methods:
```
public static void StartTrackingHistory() { ClearHistory(); TrackHistory = true; }
public static void EndTrackingHistory() { TrackHistory=false; ClearHistory(); }
```
And then make the setter for TrackHistory private.

One advantage of making these things methods is that you can do:
`using var restoreTracking = new ScopedAction(ConnectionPool.StartTrackingHistory, ConnectionPool.EndTrackingHistory);`
Without needing the "()=>" to turn them into Actions that can be passed to ScopedAction.

"ClearHistory" should be called both at the beginning of the test and the end of the test, right? It needs to be called at the beginning of the test in case the previous test misbehaved and it should be called at the end of the test because we want to clean up static resources.

PR #4050, opened 2026-03-03
Fix spiky chromatograms from narrow scan window MS/MS spectra

chambm:
* tracked per-target scan window coverage in FilterSpectrumList and marked uncovered targets with NaN intensity in ProcessExtractedSpectrum
  * filtered NaN-marked time points in ChromCollector.ReleaseChromatogram so each transition only contains points from spectra whose scan window covered its product m/z
  * added mzMLb format support to DataSourceUtil and OpenDataSourceDialog
  * added TestImportFullScanNarrowScanWindows functional test with narrow scan window test data

### DIFF
--- a/pwiz_tools/Shared/CommonMsData/DataSourceUtil.cs
+++ b/pwiz_tools/Shared/CommonMsData/DataSourceUtil.cs
@@ -168,7 +168,7 @@ namespace pwiz.CommonMsData
 
         public static bool IsWiffFile(string filePath)
         {
-            return PathEx.HasExtension(filePath, EXT_WIFF);
+            return IsWiffOrWiff2File(filePath);
         }
 
         public static bool IsWiffOrWiff2File(string filePath)


pwiz_tools/Skyline/Model/Results/ChromCollector.cs
```diff
@@ -114,6 +114,43 @@ public void ReleaseChromatogram(byte[] bytesFromDisk, out TimeIntensities timeIn
             var scanIds = Scans != null
                 ? Scans.ToArray(bytesFromDisk)
                 : null;
+
+            // Filter out NaN intensities from narrow scan window coverage.
+            // NaN marks time points where the target's product m/z was outside the
+            // spectrum's scan window and should not contribute to the chromatogram.
+            int nanCount = 0;
```
nickshulman, line 121: `int nanCount = intensities.Count(float.IsNaN);`

pwiz_tools/Skyline/Model/Results/ExtractedSpectrum.cs
```diff
@@ -55,5 +55,12 @@ public ExtractedSpectrum(ChromatogramGroupId chromatogramGroupId,
         public float[] Intensities { get; private set; }
         public float[] MassErrors { get; private set; }
         public ChromExtractor Extractor { get; private set; }
+
+        /// <summary>
+        /// Per-target flag indicating whether the target's product m/z was within the
+        /// scan window of any contributing spectrum. Null means all targets are covered
+        /// (no narrow scan window filtering was needed).
+        /// </summary>
+        public bool[] HasScanWindowCoverage { get; set; }
```
nickshulman, line 64: Do you really need these booleans? Could you accomplish the same thing by putting NaN into Intensities?

One thing you might want to do is change Intensities to float?[] instead of float[].
Nullable floats use twice as many bytes as regular floats, but Intensities is such a small array here that it does not matter.
Making the Nullable might make it easier to make sure that you have found all the places that need to be checking whether they are missing.

PR #4067, opened 2026-03-10
Add support for Zstd-based compression methods to ProteoWizard

mobiusklein:
This PR implements the compression methods introduced at https://github.com/HUPO-PSI/psi-ms-CV/pull/444.

### Actual content:
- Added zstd-1.5.7 to the 3rd-party libraries, including a vendor'd tarball. 
- Added the single header-only `span` implementation from https://github.com/tcbrindle/span to polyfill the C++20 `std` implementation because my implementation of these compressors required it. I am not sure if the vendor'd version of Arrow also has a polyfill but by the time I noticed that I had already gotten the `tcb` implementation work.
- A much improved, version of the single header-only https://github.com/mobiusklein/mzd.cpp. It was a stark reminder for me of just how easy undefined behavior is to introduce and hard to catch once you have it.
- All the plumbing to connect those new compression methods to `BinaryDataEncoder`, and the requisite bits in `IO` to make them available in mzML.
- New tests to exercise the new compression in `BinaryDataEncoderTest`.
- Integration with command line `msconvert.exe`

I haven't touched MSConvertGUI yet, and before I go there I wanted to get feedback on what is here already.

### Notes

The new compression codecs are a bit more complicated than the existing ones because they handle endianness as they re-arrange the data for better compressibility. `mzd::compress_buffer` technically doesn't but I wanted to have parity with the other methods. The byte shuffling method will directly shuffle bytes into little-endian order, and the dictionary encoding method reuses the byte shuffling encoding as well. 

For this reason, I had to add some extra logic to `BinaryDataEncoder::Impl` to make the endian changes conditional on not using any of these new codecs as well, but still permit the precision to be adjusted up/down. I was tempted to make parts of the `encode` and `decode` methods templates, but I know that's more disruptive to C++ than in the languages I am used to so I refrained. Of the three codecs, only `Compression_Zstd` is compatible with Numpress, and I try to throw errors when those other configurations mix.

If this doesn't go catastrophically, now that I've seen how to get ProteoWizard bootstrapped it's feasible for me to make more changes in the future.

Thank you


pwiz/data/msdata/BinaryDataEncoderTest.cpp
```diff
@@ -34,10 +34,10 @@ using namespace pwiz::msdata;
 namespace bfs = boost::filesystem;
 
 
-ostream* os_ = 0;
+ostream* os_ = &std::cout;
```
chambm: Left over debug code? Should only be set to cout (actually just `cout` which resolves to `boost::nowide::cout`) when the `-v` flag is given and that should already be handled.

pwiz_tools/commandline/msconvert.cpp
```diff
@@ -386,6 +389,19 @@ Config parseCommandLine(int argc, char** argv)
         ("zlib,z",
             po::value<bool>(&zlib)->implicit_value(true),
             ": use zlib compression for binary data (add =off to disable compression)")
+        ("zstd",
```
chambm, line 392: We can't have both -z and --zstd as implicit_value(true). That means if you don't specify, it defaults to true. Let's leave -z as the default for now. Specifying --zstd will override -z (that should be handled by the order of if(zlib) and if(zstd) below).

pwiz/utility/misc/mzd.hpp
```diff
@@ -0,0 +1,991 @@
+#include <iostream>
```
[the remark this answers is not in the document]
chambm: Huh, I didn't know that about compilation times.

PR #4110, opened 2026-03-27
Volcano/RelAbundance: independent per-trait formatting rules

rita-gwen:
This feature was suggested by Michael MacCoss.

## Summary

- **New feature**: Formatting rules on volcano plots and relative abundance graphs now resolve traits (color, symbol, size, labeled) independently. The first rule that matches a point AND has a given trait set wins for that trait — letting users mix rules freely, e.g. *color by protein name, shape by p-value cutoff*.
- Nullable `PointSymbol?` and `PointSize?` on `MatchRgbHexColor` represent "not set" — rules with null traits are skipped for that trait during per-point resolution.
- UI: Symbol and size combos in `VolcanoPlotFormattingDlg` now include a **None** option (first item) so users can create rules that set only specific traits.
- Backward compatible: existing documents with `symbol_type`/`point_size` attributes deserialize to non-null values, preserving old behavior.
- Matched-point curves are ordered by first matching rule index; unmatched points go to the default gray curve added last.

## Test plan

- [x] `TestVolcanoPlotFormatting` — existing test covers multi-rule scenarios; passes with new per-trait resolution
- [x] `TestVolcanoPlotLayout` — label layout unaffected
- [x] `TestVolcanoPlotParsing` — parsing unaffected
- [x] `TestGroupComparisonVolcanoPlot` — volcano plot general behavior unaffected
- [x] `TestPeakAreaRelativeAbundanceGraph` — relative abundance graph passes

### DIFF
--- a/pwiz_tools/Skyline/TestFunctional/VolcanoPlotFormattingTest.cs
+++ b/pwiz_tools/Skyline/TestFunctional/VolcanoPlotFormattingTest.cs
@@ -150,6 +150,50 @@ namespace pwiz.SkylineTestFunctional
                     }, false),
             },
 
+            // Nullable trait test cases: verify that the rendering code applies the correct defaults
+            // (Color.Gray, PointSymbol.Circle, PointSize.small) when traits are omitted from a rule.
+
+            new[]
+            {
+                // Rule specifies color only (null symbol, null size) — matched points get the default symbol+size
+                new MatchExprInfo(new MatchExpression(string.Empty, new[] {MatchOption.BelowLeftCutoff}),
+                    new VolcanoPlotPointsInfo(17, Color.Cyan, false, null, null),
+                    new List<string>
+                    {
+                        "VFWIEVALFWR",
+                        "SDFQVPCQYSQQLK",
+                        "WWGQEITELAQGPGR",
+                        "AGDQILAINEINVK",
+                        "AGSWQITMK",
+                        "FAEDHFAHEATK",
+                        "NLAPLVEDVQSK",
+                        "CSSLLWAGAAWLR",
+                        "ETGLMAFTNLK",
+                        "MLSGFIPLKPTVK",
+                        "LQTEGDGIYTLNSEK",
+                        "SVVDIGLIK",
+                        "IAELFSDLEER",
+                        "FSISTDYSLK",
+                        "EVLPELGIK",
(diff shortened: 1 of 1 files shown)


pwiz_tools/Skyline/ToolsUI/ColorGrid.cs
```diff
@@ -167,6 +168,45 @@ public interface IColorGridOwner
             BindingList<T> GetCurrentBindingList();
         }
 
+        private int _useColorColumnIndex = -1;
+
+        /// <summary>
+        /// Adds a "Use color" checkbox column bound to <c>UseColor</c> immediately after the color swatch column.
+        /// When the user checks it with no color set the color picker opens automatically.
+        /// </summary>
+        public void AddUseColorColumn(string headerText)
```
nickshulman: I don't think you need a separate column for the user to indicate whether or not they want to set a color.
If they blank out the text in the RGB column that means they don't want to set the color.

pwiz_tools/Skyline/Controls/Graphs/DotPlotUtil.cs
```diff
@@ -207,5 +225,51 @@ public static bool IsTargetSelected(SkylineWindow skylineWindow, Peptide peptide
...
+
+        /// <summary>
+        /// Resolves per-trait formatting for a single data point by scanning the ordered rule list.
+        /// Each trait (color, symbol, size, labeled=true) is set by the first matching rule that
+        /// explicitly provides that trait. <c>Labeled=false</c> is treated as "not set" so a later
+        /// rule can still enable labeling.
+        /// Returns <c>null</c> when no rule contributes any trait — the caller should add the point
+        /// to the default "other" collection rather than a matched curve.
+        /// </summary>
+        public static (Color? color, PointSymbol? symbol, PointSize? size, bool labeled, int firstRuleIndex)?
```
nickshulman: Usually you would want to define a class for this, instead of using such a complicated Tuple.
And, you would put a method "Merge" on that class, which would return a new object where the null property values were change to the values from the second object.

"MatchRgbHexColor" would be the perfect candidate to add such a "Merge" method to, except for the fact that MatchRgbHexColor is mutable, and has a lot of other baggage for sending property notifications that are used by the DataGridView.

PR #4111, opened 2026-03-27
Refactor context menus from Skyline.designer.cs into separate files in Menus subfolder

nickshulman:
Also, context menus get constructed every time they get used, so we don't need the code which adds menu items back to their parent if they got removed when ZedGraphControl was disposed.


rita-gwen:
Apparently, this PR changed the structure of the GraphSpectrum context menu. The menu in the release looks like this:
image
 but after this PR it lost the Copy Metafile and Copy Data items and a separator. Also, it acquired a couple of new items.
image

nickshulman:
> Apparently, this PR changed the structure of the GraphSpectrum context menu.

Oops! Thanks for catching this, Rita.
I need to add a call to "PrepareZedGraphContextMenu" which adds the Copy Data menu item and removes "Show Point Values" and "Set Scale to Default".
I'll open a new PR.

PR #4136, opened 2026-04-03
Improve performance of saving .sky file

nickshulman:
Improved speed of saving .sky file by using multiple threads

Multiple threads are used to create XElement's for each of the PeptideDocNode, and then those are written out to disk in a single thread.
Also, hashing the output stream (for the Audit Log) is done on a separate thread so that it does not slow down writing the bytes to disk.


nickshulman:
The performance problem that Mike was experiencing was fixed by increasing the buffer size on the write stream. The improvements here are just marginal and not worth the complexity.

PR #4144, opened 2026-04-08
Fix full scan graph IM heatmap display issues

bspratt:
## Summary
This fixes a couple of longstanding unreported issues with ion mobility heatmap display, as well as one recently introduced.

 Extraction bar scaling — The extraction box minimum-width enforcement  (added in #4087) used ReverseTransform to compute 1 pixel in m/z units,  but AxisChange (called by base.SetScale) does not call SetupScaleData, leaving the pixel-to-data transform stale from the previous zoom level.
  This caused boxes to disappear when zooming out and become wildly oversized when zooming back in. Fix: call SetupScaleData after AxisChange to refresh the transform.

  Blank MS/MS heatmap when Zoom To Selection is on — Switching scan type from MS1 to MS/MS in heatmap mode with magnify on produced a completely blank graph (no axes, no data). GetIonMobilityFilterRange required the transition at TransitionIndex to match the requested source type, which fails when the user clicks a precursor chromatogram and then  switches to MS/MS (TransitionIndex still points to the ms1 transition). This left the IM range inverted (MaxValue/MinValue), collapsing the Y axis. Latent since the 2019 per-transition IM high-energy offset change.
  Fix: fall back to all matching-source transitions when the clicked transition's source doesn't match, preserving the original single-transition narrowing when sources do match.

  StackOverflowException in HeatMapData quad-tree — When all points share the same coordinates, CreateCells produces a child cell identical to its parent, causing infinite recursion. Fix: detect when subdivision makes no spatial progress and treat the cell as a leaf.

## Test plan
- [ ] Run `TestFullScanGraph` 
- [ ] Open full scan graph, toggle Zoom To Selection on/off, verify extraction bars remain visible and correctly scaled at all zoom levels

### DIFF
--- a/pwiz_tools/Shared/MSGraph/HeatMapData.cs
+++ b/pwiz_tools/Shared/MSGraph/HeatMapData.cs
@@ -174,6 +174,15 @@ namespace pwiz.MSGraph
                 if (_cells == null)
                     _cells = CreateCells();
 
+                // If subdivision made no progress (all points share the same
+                // coordinates), treat this cell as a leaf.
+                if (_cells[0] == null && _cells[1] == null &&
+                    _cells[2] == null && _cells[3] == null)
+                {
+                    returnedPoints.Add(_maxPoint);
+                    return;
+                }
+
                 // Add the maximum intensity points from each of the 4 cells inside this cell (and recurse to
                 // the right cell size).
                 for (int i = 0; i < 4; i++)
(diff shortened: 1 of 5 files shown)


bspratt:
this is the problem being fixed:

image

nickshulman (approved):
Looks good to me.

PR #4151, opened 2026-04-14
msconvert: write to .partial file, rename on success; clearer error message

chambm:
## Summary
- msconvert now writes converted output to `<outputFilename>.partial` and renames to the final name only after the write completes successfully. Previously, when a conversion failed mid-run (e.g. vendor library fails to centroid a scan), a partial file with the final extension was left on disk and could be mistaken for a valid conversion.
- The per-run error message now notes that conversion of that run is aborting, and suggests `--continueOnError` as a way to skip problematic spectra and continue writing.
- Refactored shared write logic into a `writeAtomically()` helper used by both `mergeFiles()` and `processFile()`.

## Example before/after

Before (on a Thermo RAW file where vendor centroiding fails on some scans):
```
Error writing run 1:
[SpectrumList_Thermo::spectrum()] Error retrieving spectrum "controllerType=0 controllerNumber=1 scan=137414": [RawFileThreadImpl::getMassList()] failed to centroid scan
```
...and a 2 GB `.mzML` file left behind that looks valid but is truncated.

After:
```
Error writing run 1 (aborting conversion of this run): [SpectrumList_Thermo::spectrum()] Error retrieving spectrum "controllerType=0 controllerNumber=1 scan=137414": [RawFileThreadImpl::getMassList()] failed to centroid scan
  To skip problematic spectra and write the remaining data, re-run with --continueOnError.
```
...and the truncated file is left as `.mzML.partial` so it is obviously not a completed conversion.

## Test plan
- [x] Verified failure case: truncated output is named `.mzML.partial`, error message includes aborting hint.
- [x] Verified success case with `--continueOnError`: final `.mzML` is written, no `.partial` file remains.
- [ ] CI build across platforms.

### DIFF
--- a/pwiz_tools/commandline/msconvert.cpp
+++ b/pwiz_tools/commandline/msconvert.cpp
@@ -917,6 +917,22 @@ void calculateSourceFilePtrSHA1(const SourceFilePtr& sourceFilePtr)
 }
 
 
+/// Writes an MSData object to a .partial file first, then renames to the final filename on success,
+/// so that incomplete output from a failed conversion is not mistaken for a valid file.
+void writeAtomically(const MSData& msd,
+                     const string& outputFilename,
+                     const MSDataFile::WriteConfig& writeConfig,
+                     const IterationListenerRegistry* pILR)
+{
+    string partialFilename = outputFilename + ".partial";
+    MSDataFile::write(msd, partialFilename, writeConfig, pILR);
+    // bfs::rename fails on Windows if the destination already exists, so remove it first.
+    if (bfs::exists(outputFilename))
+        bfs::remove(outputFilename);
+    bfs::rename(partialFilename, outputFilename);
+}
+
+
 /// Combines multiple input files into a single MSData object. Called
 /// when the --merge argument is present on the command line.
 int mergeFiles(const vector<string>& filenames, const Config& config, const ReaderList& readers)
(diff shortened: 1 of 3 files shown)


pwiz_tools/commandline/msconvert.cpp
```diff
@@ -917,6 +917,43 @@ void calculateSourceFilePtrSHA1(const SourceFilePtr& sourceFilePtr)
...
+/// new file has been promoted, and restored if the final rename fails.
+void writeAtomically(const MSData& msd,
+                     const string& outputFilename,
+                     const MSDataFile::WriteConfig& writeConfig,
+                     const IterationListenerRegistry* pILR)
+{
+    string partialFilename = outputFilename + ".partial";
+    MSDataFile::write(msd, partialFilename, writeConfig, pILR);
+
+    // bfs::rename fails on Windows if the destination already exists, so preserve any existing
```
nickshulman, line 932: Is this really true? Claude thinks this problem of not being able to overwrite a file with bfs::rename on Windows got fixed in Boost 1.74 and we are using 1.86.

chambm: You're right — I checked Boost 1.86's source (`libs/filesystem/src/operations.cpp`): `rename` calls `MoveFileExW(old, new, MOVEFILE_REPLACE_EXISTING | MOVEFILE_COPY_ALLOWED)` on Windows (and plain `::rename` on POSIX), both of which atomically replace an existing destination. The flag is baked in — no version guard, nothing the caller passes.

So the backup-and-restore logic was both unnecessary and counterproductive (it reintroduced the non-atomic data-loss window the bot worried about). I've simplified `writeAtomically` down to: write to `.partial`, then `bfs::rename` onto the output. Since `.partial` lives in the same directory, the replace is same-volume and therefore atomic.

Done in e22a4ed.

PR #4162, opened 2026-04-23
Fix method export crash for explicit ion mobility without units

bspratt:
## Summary

Most users of ion mobility data have just one kind of equipment, so when ion mobility units are not specified it is reasonable for Skyline to infer units by looking at the current document's loaded replicate data, or explicit ion mobility units on any precursors, or values in spectral libraries or ion mobility libraries. If more than one kind of IM unit is mentioned, or none, then Skyline reverts to its old behavior of requiring user input. But if the document consistently refers to a single kind of ion mobility, then that IM unit is assumed, resulting in a more intuitive UX. Overrides are still possible, as before, but are rarely needed.

This also fixes the "InvalidOperationException: Nullable object must have a value" thrown during Bruker timsTOF method export when a peptide has `ExplicitIonMobility` set but `ExplicitIonMobilityUnits` is `none` as reported by Nick Shulman via email on 2026-04-20; user Todd Greco hit the crash after setting Explicit Ion Mobility values in the Document Grid without also setting the Units column.

See `ai/todos/active/TODO-20260423_deduce_im_units.md` for design notes.


pwiz_tools/Skyline/Model/DocSettings/TransitionIonMobilityFiltering.cs
```diff
@@ -136,6 +136,79 @@ public eIonMobilityUnits GetFirstSeenIonMobilityUnits()
...
+                    }
+                }
+            }
+
+            // Ion mobility library.
+            var imFiltering = settings.TransitionSettings.IonMobilityFiltering;
+            if (imFiltering != null)
+            {
+                var libUnits = imFiltering.GetFirstSeenIonMobilityUnits();
+                if (libUnits != eIonMobilityUnits.none)
```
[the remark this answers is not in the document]
bspratt: Good catch - fixed in 58f69f11. The IM library block now filters both `none` and `unknown`, matching the guards on the results and spectral-library sources.

pwiz_tools/Skyline/Test/IonMobilityUnitTest.cs
```diff
@@ -18,13 +18,18 @@
  */
 
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using pwiz.Common.Chemistry;
+using pwiz.Common.SystemUtil;
```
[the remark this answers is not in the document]
bspratt: Fixed in 58f69f11 - the `pwiz.Common.SystemUtil` using was indeed unused and has been removed.

pwiz_tools/Skyline/Test/IonMobilityUnitTest.cs
```diff
@@ -18,13 +18,18 @@
...
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using pwiz.Common.Chemistry;
+using pwiz.Common.SystemUtil;
+using pwiz.CommonMsData;
 using pwiz.Skyline.Model;
 using pwiz.Skyline.Model.DocSettings;
 using pwiz.Skyline.Model.IonMobility;
 using pwiz.Skyline.Model.Lib;
+using pwiz.Skyline.Model.Results;
+using pwiz.Skyline.Properties;
```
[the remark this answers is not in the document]
bspratt: Declining this one - `pwiz.Skyline.Properties` is used by the test for `SrmSettingsList.GetDefault()` (defined in `Properties/Settings.cs`). Removing the using would break the build.

pwiz_tools/Skyline/Model/Databinding/Entities/Precursor.cs
```diff
@@ -354,8 +354,21 @@ public double? ExplicitIonMobility
...
+                var unitsForSet = DocNode.ExplicitValues.IonMobilityUnits;
+                if (value.HasValue && unitsForSet == eIonMobilityUnits.none)
+                {
+                    // The user may not have set the IonMobilityUnits column yet. If the document
+                    // unambiguously implies a single unit, silently apply it along with the value -
+                    // saves the user a step and makes the audit log reflect the deduced unit.
+                    // If deduction is empty or ambiguous, accept the value with units=none and
+                    // let the user set the units column next; the safety net in
+                    // SrmSettings.GetIonMobilityFilter guards consumption of an unresolved state.
+                    var candidates = TransitionIonMobilityFiltering.GetDocumentIonMobilityUnits(SrmDocument);
```
nickshulman: This call to "GetDocumentIonMobilityUnits(SrmDocument)" will be slow if the document has thousands of peptides in it, which means that pasting a large number of values into the "Explicit Ion Mobility" column could take a long time.

Part of the problem is that "Library.TryGetIonMobilityInfos" is particularly slow in the way it creates a new LibKeyIndex each time, which we should fix anyway since it can make bringing up the Library Explorer slow for big libraries.

pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
```diff
@@ -1044,8 +1045,43 @@ public IonMobilityFilter GetIonMobilityFilter(PeptideDocNode nodePep,
             }
             if (nodeGroup.ExplicitValues.IonMobility.HasValue)
             {
+                // Explicit IM value without units produces an empty IonMobilityValue and later
+                // crashes. The document grid setter prevents this going forward, but legacy
```
nickshulman: This comment is a little misleading because the document grid will leave the units as "none" if there were more than one unit being used. Also, the user can explicitly set the units back to "none".

pwiz_tools/Skyline/Model/Databinding/Entities/Precursor.cs
bspratt: The number of users with access to more than one flavor of ion mobility is pretty small, I think, so we could probably just as profitably ask "does any library in use contain IM values at all, and if so are there more than one kind" and it would nearly always come back with zero or one kinds found. No real need to find an actual library match.

pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
bspratt: good point

pwiz_tools/Skyline/Model/Databinding/Entities/Precursor.cs
bspratt: Fixed in 6d9eda82. Replaced the per-libkey spectral library scan with a new `Library.GetDistinctIonMobilityUnits()` that uses the null-targets fast path (skipping the `LibKeyIndex` rebuild you flagged) and caches the result on the library instance. Subsequent calls during a bulk paste hit the cache. Also added a `> 1` short-circuit so the deducer bails as soon as it knows the answer is ambiguous, and reordered `GetDocumentIonMobilityUnits` to scan sibling transition groups first (cheap) before consulting settings-level sources. The underlying `TryGetIonMobilityInfos` / `LibKeyIndex` perf issue you mentioned is still worth fixing on its own, but the deducer no longer triggers it.

pwiz_tools/Skyline/Model/DocSettings/SrmSettings.cs
bspratt: Fixed in 6d9eda82. Rewrote the comment to cover all three paths that can land here in the invalid state: legacy documents loaded from disk, documents where the grid setter accepted the value without being able to deduce a unique unit, and documents where the user explicitly set units back to `none` via the Units column.

PR #4163, opened 2026-04-23
OspreySharp: Stage 5 dump text format matches Rust ryu byte-for-byte

brendanx67:
## Summary

* New `pwiz.OspreySharp.Core.Diagnostics.FormatF64Roundtrip(double)` helper. Uses `"R"` + `double.Parse` round-trip check, falls back to `"G17"`, then expands any scientific-form output to fixed decimal form so the text matches Rust's `format!("{}", v)` byte-for-byte on `.NET Core 3+` / `.NET 5+` / `.NET 8`.
* Replaces 10 `.ToString("G17", inv)` call sites in the three Stage 5 dump writers (`standardizer`, `SVM weights`, end-of-Stage-5 `Percolator`).
* Reconciles the class-level `OspreyDiagnostics` XML doc: Stages 1-4 use `F10`; Stages 5+ use `Core.Diagnostics.FormatF64Roundtrip`. Yesterday's (#4160) note that "Stage 5 uses G17 to match Rust's G17 choice" was incorrect — Rust uses `format!("{}", v)` which is ryu-shortest, not G17.
* Adds `DiagnosticsTest.cs`: special-value handling, integer-valued f64, a real Stellar standardizer value (`5.598374948209159`), and a `#if NETCOREAPP || NET5_0_OR_GREATER` byte-match assertion on `0x41583D9959B55C3F`.

### Framework caveat

`.NET Framework 4.7.2`'s `Double.ToString("R")` has a long-standing bug where a minority of values don't round-trip. The helper detects this via parse-check and falls back to G17, so net472 output still round-trips but may be slightly longer than ryu-shortest for the affected values. The parity harness defaults to net8.0 for correct byte-identity.

### Evidence

Before the fix, an independent SHA-256 byte-parity check across all 3 Stellar single-files failed on 3 of 4 Stage 5 dumps (only the integer-only subsample dump passed). After the fix, **3/3 Stellar files byte-identical on all four dumps**:

```
Stellar Ste-2024-12-02_HeLa_4mz_sDIA_400-900_20   PASS
Stellar Ste-2024-12-02_HeLa_4mz_sDIA_400-900_21   PASS
Stellar Ste-2024-12-02_HeLa_4mz_sDIA_400-900_22   PASS
```

## Test plan

- [x] `Build-OspreySharp.ps1 -RunInspection -RunTests` — all unit tests pass on both net472 and net8.0, zero R# warnings.
- [x] `Compare-Stage5-AllFiles.ps1 -Dataset Stellar` — 3/3 byte-identical.
- [ ] `Compare-Stage5-AllFiles.ps1 -Dataset Astral` — pending Astral Parquet regeneration.

### DIFF
--- a/pwiz_tools/OspreySharp/OspreySharp.Core/Diagnostics.cs
+++ b/pwiz_tools/OspreySharp/OspreySharp.Core/Diagnostics.cs
@@ -1,0 +1,130 @@
+/*
+ * Original author: Brendan MacLean <brendanx .at. uw.edu>,
+ *                  MacCoss Lab, Department of Genome Sciences, UW
+ * AI assistance: Claude Code (Claude Opus 4) <noreply .at. anthropic.com>
+ *
+ * Based on osprey (https://github.com/MacCossLab/osprey)
+ *   by Michael J. MacCoss, MacCoss Lab, Department of Genome Sciences, UW
+ *
+ * Copyright 2026 University of Washington - Seattle, WA
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


chambm:
WTF does `Copilot was unable to run its full agentic suite in this review.` mean? Google search only pulls up one hit for that. I guess it has something to do with the smart trigger pwiz_tools check I added, but that's just a guess (because I don't see the expected review comment!).

chambm:
There we go. Must be a fluke, or let's see if it happens on the next review requested by you (I think it's per account even though it's automatically triggered).

PR #4172, opened 2026-04-28
Upgraded Parquet.Net 3.0.0 to 4.25.0 with Zstd as default compression

### LINKED ISSUE #4171: Upgrade Parquet.Net to 5.5.0 to enable Zstd compression
## Summary

Upgrade the bundled Parquet library from Parquet.Net 3.0.0 (currently shipped at `pwiz_tools/Shared/Lib/Parquet/ParquetNet.dll`) to **Parquet.Net 5.5.0** so we can write Parquet files with Zstandard compression. The current 3.x version only supports Snappy/Gzip; Zstd compression was first added in 4.1.0 and reimplemented as a fully-managed (no native dep) `ZstdSharp.Port` codec in 5.5.0.

5.5.0 is the **last 5.x release that still ships a `netstandard2.0` target**, so it is the latest version compatible with Skyline's net472 target framework.

## Scope

- [ ] Refresh `pwiz_tools/Shared/Lib/Parquet/` with the Parquet.Net 5.5.0 closure (10 DLLs)
- [ ] Bump 6 polyfill DLLs in `pwiz_tools/Shared/Lib/` to versions Parquet.Net 5.5.0 was compiled against
- [ ] Update `Skyline.csproj` references; drop now-unused `System.Reflection.Emit.Lightweight` (inbox in net472)
- [ ] Update `app.config` bindingRedirects in `Skyline`, `TestRunner`, `TestFunctional` (Parquet PKT and Version both change)
- [ ] Update installer `Product-template.wxs` and `FileList64-template.txt`
- [ ] Bump `OspreySharp.IO` and `OspreySharp.Test` PackageReference 3.10.0 → 5.5.0
- [ ] Pin transitive NuGet versions in `CommonMsData.csproj` so older copies don't clobber Lib/ during Copy Local
- [ ] Rewrite `ParquetReportExporter.cs` and `ParquetScoreCache.cs` for the new async-only API (without `async`/`await` keywords per Skyline rules)
- [ ] Switch the default compression to `CompressionMethod.Zstd`

## Context

- Parquet.Net 5.x dropped support for `DataTypeOffset` storage; calls writing `DateTime` columns must now store as `DateTime?` rather than `DateTimeOffset?`.
- Parquet.Net 4.0 switched to an async-only public API (`ParquetWriter.CreateAsync`, `WriteColumnAsync`, `ReadColumnAsync`). Per Skyline's no-`async`/`await` rule (`ai/CRITICAL-RULES.md`), call sites use `.Result` / `.Wait()` to synchronously block.
- Parquet.Net 5.5.0 exposes `CustomMetadata` directly on `ParquetWriter`/`ParquetReader`, so the reflection-based metadata workaround in `ParquetScoreCache.cs` can be replaced with a direct property access.
- The strong-name key was rotated in Parquet.Net 4.x: PKT changes from `de28deb604dd91c9` to `d380b3dee6d01926`, and `Version` from 3.0.0.0 to 5.0.0.0. All bindingRedirect/codeBase entries must be updated accordingly.

## Getting Started

nickshulman:
## Summary

* Upgrade bundled Parquet.Net 3.0.0 → 4.25.0 in `pwiz_tools/Shared/Lib/Parquet/` and `OspreySharp.IO`/`OspreySharp.Test` so Parquet exports can use **Zstandard** compression on Skyline's net472 / netstandard2.0 target.
* Switch default compression to `CompressionMethod.Zstd` in `ParquetReportExporter` and `ParquetScoreCache`.
* Originally targeted Parquet.Net 5.5.0; pivoted to 4.25.0 after `TestParquetArrays` failed at runtime — Parquet.Net 5.5.0's netstandard2.0 build has stub `CopyToAsync` extension methods that throw `NotImplementedException`, breaking every write through `DataColumnWriter.CompressAndWriteAsync`. See [parquet-dotnet#710](https://github.com/aloneguid/parquet-dotnet/issues/710), closed `NOT_PLANNED` ("I'm deprecating .net standard support"). 5.6.0+ dropped netstandard2.0 entirely; 4.25.0 is the last release on a long, mature 4.x line that still works on net472.
* 4.25.0 has the same 5.x-style API surface as 5.5.0 (`Parquet.Schema.ParquetSchema`, `DataField(string, Type, bool? isNullable, bool? isArray)`, public `CustomMetadata` get/set on `ParquetWriter`/`ParquetReader`, async-only API), so the prior code rewrites in `ParquetReportExporter.cs` and `ParquetScoreCache.cs` carry over unchanged.
* Refresh `Lib/Parquet/` for the 4.25.0 closure: ParquetNet 4.0.0.0, IronCompress 1.5.2 + native `nironcompress.dll` (win-x64), Snappier 1.1.6, ZstdSharp.Port 0.8.1, Microsoft.IO.RecyclableMemoryStream 3.0.1, System.Text.Json 8.0.5, System.Text.Encodings.Web 8.0.0, System.Reflection.Emit.Lightweight 4.7.0.
* Revert the polyfill bumps in `Lib/` and the transitive-version pins in `CommonMsData.csproj` from the 5.5.0 attempt — 4.25.0's references are satisfied by master polyfill versions, no churn needed.
* Update three `app.config` files for Parquet `version=4.0.0.0` (PKT `d380b3dee6d01926` — same as 5.x, only 3.x→4.x rotated), add IronCompress redirect, target master polyfill versions in bindingRedirects.
* Update installer `Product-template.wxs` + `FileList64-template.txt` for the 4.25.0 closure (drop CommunityToolkit / K4os / Bcl.HashCode / IO.Pipelines, add IronCompress / nironcompress / System.Reflection.Emit.Lightweight back).

Fixes #4171.

See `ai/todos/active/TODO-20260428_parquet_zstd.md` for full engineering context, the 5.5.0 → 4.25.0 pivot rationale, and remaining validation work.

## Updates from code review

* **Copilot feedback addressed (commit dcd8d5973):**
  * Replaced all `.Result` / `.Wait()` sync-over-async patterns with `ConfigureAwait(false).GetAwaiter().GetResult()` so original exception types are preserved (no `AggregateException` wrapping) and `SynchronizationContext` deadlocks are avoided.
  * Added `RunSync<T>(Task<T>)` / `RunSync(Task)` helpers in `ParquetScoreCache` so call sites stay terse.
  * `TestConvertToStorageType` now asserts `typeof(DateTime)` rather than `typeof(DateTime?)` — boxing `Nullable<T>` with `HasValue` strips the wrapper.
* **Pre-existing fixes also folded in:**
  * `LoadCwtCandidatesFromParquet` was added in PR #4173 (Stage 6) after the original 5.5.0 attempt and missed during the merge — still used `new ParquetReader(stream)` / `groupReader.ReadColumn(...)`. Migrated to the async API like the rest of the file.
  * `TestParquetScoreCacheRoundTrip` was failing on the new 4.x strict "field must be attached to schema" check. Two-part fix: (a) build the 21 PIN feature `DataField`s once and pass the same instances to both `BuildWriteSchema` and the `WriteColumnAsync` calls (was building two unrelated sets); (b) on the read side, look up `DataField`s by name from `reader.Schema.GetDataFields()` instead of relying on static `FIELD_*` instances that may not be attached in fresh-process reads (and `FIELD_COELUTION_SUM`, the reader-only alias, was never in any schema).

## Test plan

- [x] Build `Skyline.sln` Release|x64 — clean
- [x] Build `OspreySharp.sln` Release for net472 + net8.0 — clean (all 8 projects)
- [x] `TestFunctional / ParquetReportExporterTest.TestParquetArrays` — 0 failures (~1 sec; round-trips a parquet through 4.25.0 reader/writer)
- [x] `TestFunctional / ParquetReportExporterTest.TestConvertToStorageType` — 0 failures
- [x] `OspreySharp.Test` full suite — 279/282 passing, 3 skipped, 0 failures (includes `TestParquetScoreCacheRoundTrip` and `TestParquetScoreCacheEmptyWrite`)
- [ ] Cross-impl byte-parity tests against Rust osprey (Stage 5/6). 4.25.0 routes Zstd through native IronCompress + ZstdSharp.Port 0.8.1 fallback, slightly different framing than 5.5.0's direct ZstdSharp.Port — parquet-level encoding should still match.
- [ ] Smoke test: export a Skyline report as Parquet, open in DuckDB / Python `pyarrow`, verify Zstd compression metadata.

## Notes

* **Parquet.Net 5.5.0 is broken on netstandard2.0.** Confirmed at the IL level (`StreamExtensions.CopyToAsync` is 47 bytes — a stub that throws `NotImplementedException`) and at the source level (`#if NETSTANDARD2_0 throw new NotImplementedException(); #endif` on master). `DataColumnWriter.CompressAndWriteAsync` always calls `pageData.Memory.CopyToAsync(_stream)` so every write fails. Maintainer is not fixing it.
* **Compression backend trade-off vs. 5.5.0:** 5.5.0 used fully-managed `ZstdSharp.Port`; 4.25.0 routes through `IronCompress 1.5.2` which P/Invokes a bundled native `nironcompress.dll` (~4 MB win-x64) for performance and falls back to managed Snappier / ZstdSharp.Port if native loading fails (per upstream issue #574). Acceptable; the alternative is no Parquet 5.x on net472 at all.
* **Strong-name key:** PKT `de28deb604dd91c9` → `d380b3dee6d01926` was a 3.x → 4.x rotation, not 4.x → 5.x. So the rotated PKT from the original 5.5.0 attempt is still correct for 4.25.0.
* **`DateTimeOffset` storage dropped** in Parquet.Net 5.x (and unchanged from the original 5.5.0 attempt). 4.25.0 still supports `DateTimeOffset`, but `ParquetReportExporter._storageTypes` continues to map `DateTime` → `DateTime?` since the old call sites only ever wrapped `new DateTimeOffset(dt)` with a local-kind assumption.
* **`DataType` enum** is gone in 4.x as well — `new DataField(name, typeof(T), isNullable:..., isArray:...)` is the right form for both 4.25.0 and 5.5.0.
* **Reflection metadata workaround removed** in `ParquetScoreCache.SetWriterMetadata` — 4.25.0 also exposes `CustomMetadata` as a public property (`IReadOnlyDictionary<string,string>` setter on writer; `Dictionary<string,string>` getter on reader), collapsing ~70 lines of `_footer`/`_fileMeta`/`Key_value_metadata` reflection to one-line property assignments.


brendanx67 (approved):
Approved. Solid migration — the bulk of the diff is dependency mechanics, but the actual code change is small, well-factored, and the PR description does an excellent job documenting the 5.5.0 → 4.25.0 pivot rationale.

What I verified locally:
- Clean rebuild: only pre-existing MSB3331 (`microsoft.vc90.crt` publish) remains; the new MSB3836 (`System.Text.Json`) is now cleared.
- `TestConvertToStorageType` and `TestParquetArrays` pass.
- Async-bridge pattern (`ConfigureAwait(false).GetAwaiter().GetResult()`) is the right idiom — preserves exception types, avoids deadlocks.
- The schema-attached `DataField` requirement in 4.x is handled cleanly via name-based lookup (`BuildFieldLookup` / `ReadColumnByName`).
- Removal of the reflection-based `SetWriterMetadata` in favor of the public `CustomMetadata` property is a nice cleanup.
- Binding redirects + installer file list look correct for the 4.25.0 closure.

Pushed three small follow-ups in `bd4dfa123`:
1. `TestParquetArrays` now opens the row group and reads the column back — the previous version only asserted the schema field count, which wouldn't catch a list/array write or decode regression.
2. Removed the explicit `System.Text.Json` redirect from all three app.config files. `AutoGenerateBindingRedirects` supplies it (8.0.0.5) and the explicit 8.0.0.0 was being overridden anyway, triggering MSB3836.
3. Replaced the stale "Parquet.Net 5.x dropped DateTimeOffset support" comment in `ParquetReportExporter._storageTypes` with the real reason — old call sites wrapped `DateTime` as `DateTimeOffset` with an invalid local-Kind assumption.

The two unchecked items in your test plan (DuckDB/pyarrow smoke test on a real exported report, Rust osprey byte-parity) are the riskiest remaining work — your call on whether to gate merge on those.

PR #4174, opened 2026-04-29
Fixed times/intensities mismatch on multi-event SRM Shimadzu import

bspratt:
## Summary

* In IsSingleTime mode each `ChromCollector` owns its own time array, so the
  zero-fill paths in `SpectraChromDataProvider.ProcessExtractedSpectrum`
  that added intensities without matching times desynced the arrays.
* Guarded the new-collector `FillZeroes` (line 1759) and the trailing
  missing-ion zero-fill (line 1796) so they only run in grouped/shared
  time modes; SRM (single-time) collectors now stay length-balanced.
* Reported by Wesley Vermaelen on the support thread:
  https://skyline.ms/home/support/announcements-thread.view?rowId=66356
* The trigger is a Shimadzu LabSolutions native-export mzML with two
  scheduled events at the same Q1 but disjoint Q3 sets (e.g., DOC +
  17α-OH-progesterone share Q1=331.2 with Q3 sets {81,97,109} and
  {97,109,121}). Without the fix, import aborted with
  `InvalidDataException: Times (390) and intensities (779) disagree in
  point count`.

## Test plan

- [x] Manual: Wesley's `Labsolutions_PackA_MA_alt.mzML` imports cleanly
      via `SkylineCmd --import-file=...`; previously aborted.
- [x] Regression: `ShimadzuFormatsTest` (Release, en + fr) — 7400+ passes,
      0 failures, no handle leaks.
- [ ] Add a small synthetic regression test exercising the multi-event
      same-Q1 pattern (follow-up).


bspratt:
@nickshulman your intuition is better than anyone's in this area of the code, what say you?

pwiz_tools/Skyline/Model/Results/SpectraChromDataProvider.cs
```diff
@@ -1754,7 +1754,10 @@ public void ProcessExtractedSpectrum(float time, SpectraChromDataProvider.Collec
                     chromCollector = new ChromCollector(chromIndex, IsSingleTime, spectrum.MassErrors != null);
                     // If more than a single ion scan, add any zeros necessary
                     // to make this new chromatogram have an entry for each time.
-                    if (ionScanCount > 1 && lenTimes > 1)
+                    // Only when times are shared across collectors (grouped/shared mode);
```
brendanx67: Note that IsSingleTime is passed to new ChromCollector(). Seems like this would be better placed inside the ChromCollector, since both operations end up calling chromCollector.Add*(). Maybe the ChromCollector can just do the right thing or it could certainly ignore the operation based on IsSingleTime, if that is truly the right solution. This approach is better encapsulation / separation of concern than requiring the calling code to know that these methods do not apply in IsSingleTime mode.

bspratt: Good call. Done in 5e2a4777a — moved the no-op into `ChromCollector.FillZeroes` (skips when `Times != null`) and `ChromCollector.AddPoint` (skips when called without a preceding `AddTime` in single-time mode). Both callsites in `SpectraChromDataProvider` are back to their pre-fix shape.

pwiz_tools/Skyline/Model/Results/ChromCollector.cs
```diff
@@ -81,9 +88,14 @@ public void AddPoint(int chromatogramIndex, float intensity, float? massError, B
 
         /// <summary>
         /// Fill a number of intensity and mass error values for the given chromatogram with zeroes.
+        /// In single-time mode this collector owns its own time array and back-filling
+        /// only intensities would desync the arrays, so the operation is ignored — the
+        /// caller is responsible for keeping times and intensities aligned via AddTime.
         /// </summary>
         public void FillZeroes(int chromatogramIndex, int count, BlockWriter writer)
         {
+            if (Times != null)
```
nickshulman: Does "Times != null" really tell you whether this ChromCollector owns the times list?
I thought that when the times were being shared between several chromatograms, Times would not be initialized in the ChromCollector constructor, but it would get set to a shared list by the call to "ChromCollector.SetTimes".

(I am not at all familiar with this SRM chromatograms from spectra scenario, and I'm not sure at all how it's intended to work)

bspratt: I am out of my depth - this is Claude's work. At this point I'd like to hand off the investigation to someone familiar with the code. Looking at git history, @brendanx67 is probably best qualified to assess this.

bspratt: > Does "Times != null" really tell you whether this ChromCollector owns the times list? I thought that when the times were being shared between several chromatograms, Times would not be initialized in the ChromCollector constructor, but it would get set to a shared list by the call to "ChromCollector.SetTimes".
> 
> (I am not at all familiar with this SRM chromatograms from spectra scenario, and I'm not sure at all how it's intended to work)

@nickshulman I just committed a change to be explicit about ownership

chambm:
Shall we get Wesley to try this build and see how well it fixes it for him? Also, does this affect anything other than reading SRM as spectra? Because if not, it should be pretty low impact (risk?).

PR #4178, opened 2026-04-30
Port ProteoWizard core to .NET 8

chambm:
## Summary

Adds `pwiz-sharp/`, a .NET 8 C# port of the ProteoWizard core, and wires it into TeamCity via the new `ProteoWizard_CoreWindowsNet` build config.

The port currently covers:
- **Core**: `Pwiz.Util`, `Pwiz.Data.Common`, `Pwiz.Data.MsData` (mzML / mzXML / MGF read+write, indexed envelopes, SHA-1 footers, ParamContainer / CV term machinery, lossy round-trip diff)
- **Analysis**: peak-picking (CWT centroider) + DDA processor + chromatogram extractor + lockmass refiner + threshold / metadata / zero-samples filters + spectrum-list factory
- **Vendor readers** (`src/Vendor/`): Thermo (RawFileReader), Bruker (BAF / TSF / TDF + PASEF, plus YEP / FID through CompassXtract — no Bruker format is now unported), Waters (MassLynx + IMS / MSe / SONAR / lockmass), Agilent (Cecil-patched on .NET 8 to remove `Delegate.BeginInvoke`; MS scans + IM frames + UV/DAD + SRM/SIM transitions + pump/sampler chromatograms), Sciex (`.wiff` via legacy AnalystDataProvider, `.wiff2` via `SCIEX.Apis.Data.v1` in a side-by-side `AssemblyLoadContext`), Shimadzu, UNIFI, UIMF, Mobilion, Bruker PRM scheduling
- **`msconvert-sharp`**: command-line entry point exercising the full stack
- **SeeMS port**: WinForms spectrum/chromatogram viewer + MSGraph + ZedGraph + StlContainers
- **~655 tests** across 21 test projects (Util / Common / MsData / MsData.NativeAot / IdentData / TraData / Analysis / BiblioSpec / Installer / MsConvert / MsConvertGUI + the vendor suites)

## cpp parity

The port is measured against C++ `msconvert` over a 492-file corpus (every vendor file under 100 MB in the shared test tree), converting both with identical flags and comparing the mzML **semantically** rather than byte-wise.

| status | 10 Aug | 11 Aug | 12 Aug | current |
|---|---:|---:|---:|---:|
| identical | 95 | 298 | 334 | **359** |
| differs | 320 | 124 | 87 | 62 |
| C#-only failures | 8 | 1 | 1 | 1 |

"Current" is the 12 Aug sweep plus 25 files individually re-verified against C++ since it ran (Sciex 7, Agilent 8, Waters 6, Thermo 4). Per vendor: Sciex 180/226, Thermo 93/109, Agilent 33/50, Waters 32/56, Bruker 17/20, Mobilion 3/7, PerkinElmer 1/1, Shimadzu 0/23.

`cpp-failed` (35) and `both-failed` (33) are files C++ itself cannot convert; they are excluded from the comparison rather than counted against the port.

### Differences that are not port defects

Roughly half the remaining 62 are blocked on something other than our code, and are documented at the relevant call sites:

| files | cause |
|---:|---|
| 20 Waters | The two trees ship **different MassLynx SDKs** — cpp 4.9.0.0, pwiz-sharp 5.0.0.0. 4.9 renders scan-stat strings at 6 significant digits, 5.0 at float round-trip, so SET_MASS-derived values differ (`1261.77` vs `1261.77332`). Verified by running the C# binary against cpp's 4.9 DLL, which reproduces cpp exactly. A 5.0 upgrade for the cpp tree is staged on a branch; note 5.0 drops the `setRawPath` export and the `MassLynxRawProcessor` header, so it is a small API migration rather than a DLL swap. |
| 10 Shimadzu | The Shimadzu SDK leaves `SampleInfo.AnalysisDate` at `MinValue` under .NET 8, so `startTimeStamp` falls back to `FilePropTag.GeneratedDateTime`, 15 s away. Not BinaryFormatter — `EnableUnsafeBinaryFormatterSerialization` is already set and the field is still empty. |
| 3 Thermo | cpp emits `MS:1000803 analyzer scan offset` for PDA scans from uninitialized memory: two consecutive runs of the same binary emit 689 of them, then 122. Deliberately not reproduced. |
| 2 Thermo | RawFileReader 5.0.0.93 (cpp) vs 8.0.6.0 (ours) round half-way precursor m/z differently in the filter string (`598.125` → `598.13` vs `598.12`). Both sides call the same SDK method. |
| 2 Agilent | The Agilent SDK couples `GetChromatogram` to subsequent spectrum reads (~5e-10 in the m/z axis). No single reader state satisfies both the files that need the chromatogram-sourced TIC and the files that need an unperturbed m/z axis; the safer half of the trade is taken and documented. |

### Recent parity work

Vendor-reader fixes landed since the last sweep, each verified by converting the affected corpus files with both binaries:

- **Agilent** — pump/sampler/instrument-curve chromatograms (cpp reads both signal *tables*, not `GetSignalInfo`); TIC intensities rounded through `float` as cpp caches them; spectrum TIC / base-peak taken from the cached chromatograms rather than the scan record (they disagree by up to 12% on centroided QTOF runs)
- **Thermo** — PDA instrument configuration renamed `IC{n}` with its param-group and software refs; instrument model resolved from the method text (cpp's first source, which we lacked); TIC ms-level array written as the raw `MSOrder` (negative for neutral-loss/parent scans); CAD intensities scaled by 1e-6
- **Sciex** — empty cycles dropped as cpp drops them, in two populations: a sentinel base peak with zero TIC (free), and cycles that read back empty despite plausible chromatogram values (behind `ReaderConfig.VerifyNonEmptySpectraAtIndex`, on for conversion, off for interactive callers)
- **Waters** — SIM chromatograms tagged with polarity; `--srmAsSpectra` made functional (it was hardcoded off, so the flag silently produced 15 spectra where cpp produced 5,112; the other four vendors were checked and were already correct)
- **Cross-vendor** — corrupt timestamps floored at year 1400 as cpp does; Shimadzu acquisition date converted to UTC (cpp's only `ToUniversalTime` call); empty `spectrumList` omitted
- **Build** — `$(OutputPath)wiff2` was string-concatenated, so `dotnet build -o <dir>` staged the wiff2 plugin into a *sibling* directory and every `.wiff2` silently failed to open

## Port status

pwiz module dependencies

Auto-generated from `pwiz/**/Jamfile.jam` and `pwiz_tools/{commandline,MSConvertGUI,SeeMS}/Jamfile.jam` — each box is a pwiz module directory; arrows aggregate at the directory level after transitive reduction. Generator script: [`pwiz-sharp/docs/jamdep.py`](https://github.com/ProteoWizard/pwiz/blob/chambem2/pwiz-sharp/pwiz-sharp/docs/jamdep.py).

**Legend** — green = ported and parity-checked, yellow = ported with known gaps, red = not started.

### Notes on partials

| Module | What's ported | Known gap |
|---|---|---|
| `data/msdata` | mzML / mzXML / MGF / **mz5** / **mzMLb** read+write, BinaryDataEncoder, Diff, MSData/Run/Spectrum model, indexed envelopes + SHA-1 footers | multi-run reads — `ReaderConfig.RunIndex` round-trips but every reader still loads run 0 (SeeMS already parses `path:N`); IndexedMzMLBuffer optimisations |
| `data/identdata` | mzIdentML model + IO, Diff, DBSequence / SequenceCollection / ProteinDetection / AnalysisSoftware | not parity-swept against cpp — no equivalent of the msconvert corpus run |
| `data/tradata` | TraML model, IO, Diff | as above |
| `analysis` | CWT peak picking, MSX + overlap demultiplexing (NNLS), DIA-Umpire, chromatogram + spectrum list filters, threshold / metadata / zero-samples filters, Savitzky-Golay smoother, list factories | `--filter` coverage is not exhaustive against cpp's set |
| `analysis/.../LockmassRefiner` | Generic `SpectrumList_LockmassRefiner` + `ChromatogramListLockmassRefiner` under `Pwiz.Analysis` | — (was Waters-only; now matches cpp's generic shape) |
| `vendor_readers/Bruker` | BAF, TDF (PASEF), TSF, **YEP + FID via the CompassXtract COM server** (registration-free, Windows-only), PRM scheduling | non-PASEF TDF variants; IsolationMzAndMobilityFilter on combineIMS+peakPicking. CompassXtract resolves its plugin DLLs relative to the **host executable's** directory, so the SDK payload is staged app-local on first use and a read-only install directory needs one elevated conversion — the one vendor SDK the download-to-cache model cannot serve unaided |
| `vendor_readers/Sciex` | legacy WIFF + wiff2, SRM/SIM, empty-cycle suppression matching cpp | swath centroid aggressiveness; one `no-spectra` WIFF still fails to open (`STG_E_FILENOTFOUND`) |
| `vendor_readers/Agilent` | MS path, IM frames, source files, TIC / SRM / SIM / absorption / pressure / flow chromatograms, UV/DAD spectra (after Cecil-patching the SDK) | precursor-ion-scan child ordering; IM instrument-model strings; 2 files on the SDK chromatogram/m-z state trade |
| `vendor_readers/Waters` | MS + IMS / MSe / SONAR / lockmass, SIM + SRM chromatograms, `--srmAsSpectra` | SET_MASS precision pending the MassLynx 5.0 upgrade on the cpp side |
| `vendor_readers/Shimadzu` | spectra + SRM chromatograms, `--srmAsSpectra` | `startTimeStamp` blocked on the SDK (see above) |
| `vendor_readers/UNIFI`, `UIMF`, `Mobilion` | reading + the vendor reference suites | UNIFI/waters_connect share `AbstractWatersHttpReader`; not corpus-swept (network / niche formats) |
| `utility/proteome` | AminoAcid, Peptide, Modification, Fragmentation, **Digestion** | — |
| `utility/misc` | IterationListener, Filesystem, FloatingPoint, SHA1Calculator, date parsing, basic time helpers | full DateTime range; threading helpers |
| `tools/msconvert` | arg parser, conversion pipeline (input → readers → filters → writers), bundled with Skyline as the managed `msconvert` | `--filter` completeness gaps mirror the analysis-module gaps |
| `tools` (other) | BiblioSpec (BlibBuild / BlibFilter / BlibSearch / BlibToMs2), BullseyeSharp, MSConvertGUI, SeeMS | — |

## TeamCity wiring

- `pwiz-sharp/build.bat` — TeamCity entry point. Runs `dotnet restore` + `dotnet build Pwiz.sln` + `dotnet test Pwiz.sln --no-build` with `##teamcity[progressMessage]` markers, `--logger:trx`, and `--logger:teamcity` (TeamCity.VSTest.TestAdapter — emits per-test `##teamcity[testStarted/Finished/Failed]` so individual tests show up in the TC UI). Defaults to Release; pass `Debug` as the first arg to switch.
- `pwiz-sharp/global.json` pins the SDK to .NET 8 (latestFeature) so dotCover 2023.3.3 (which doesn't support .NET 9/10) keeps working.
- `pwiz-sharp/build/AgilentPatcher/` — Mono.Cecil tool that rewrites the Agilent SDK to remove `Delegate.BeginInvoke` / `EndInvoke` calls (removed from .NET 5+) and to fix a partial-read bug that made large spectra unreadable on .NET Core. Runs after `ExtractAgilentAssemblies`; patched DLLs are gitignored.
- `pwiz-sharp/i-agree-to-the-vendor-licenses.bat` — writes a per-user `Directory.Build.user.props` setting `IAgreeToVendorLicenses=true` so vendor SDKs work in Visual Studio without per-build property injection.
- `scripts/misc/vcs_trigger_and_paths_config.py` — added a `CoreWindowsNet` target that maps to `ProteoWizard_CoreWindowsNet`, plus a `pwiz-sharp/.*` matchPath placed before the generic `.bat` / `scripts` / `libraries` patterns. Net effect: changes under `pwiz-sharp/` trigger only the .NET config; cpp changes outside that tree don't trigger the .NET config.

## Test plan

- [ ] `cd pwiz-sharp && build.bat Debug` succeeds locally
- [ ] `ProteoWizard_CoreWindowsNet` build config exists in TeamCity and runs `pwiz-sharp/build.bat` (or `tcbuild.bat`)
- [ ] PR-triggered build of `ProteoWizard_CoreWindowsNet` passes
- [ ] Verify a cpp-only change in a sibling PR does NOT trigger `ProteoWizard_CoreWindowsNet`
- [ ] Verify a `pwiz-sharp/`-only change does NOT trigger the cpp Core/Skyline/Bumbershoot chain


maccoss:
Testing the Sciex reader against a **ZenoTOF 8600** dataset: the file reads correctly — 300,570 spectra, 429 distinct isolation windows, isolation window present on every MS2 — but the instrument model is not recognised:

```
[Reader_Sciex.FillInMetadata] unable to determine instrument model
    (unknown instrument type: ZenoTOF 8600 System)
```

`ParseInstrumentName` in `Reader_Sciex_Detail.cs` ends at `ZENOTOF7600`, so an 8600 falls through to `Unknown`.

That is non-fatal by design, but the downstream effect is larger than the message suggests: with the model unrecognised the reader emits an instrument configuration with **no components at all**, so there is no `MS_time_of_flight` analyzer term. Anything that classifies an instrument from `instrumentConfiguration` — rather than from the model string — sees nothing and cannot tell a TOF from an ion trap.

Looks like a one-line addition alongside the existing entry:

```csharp
if (n.Contains("ZENOTOF7600", StringComparison.Ordinal)) return SciexInstrumentModel.ZenoTOF7600;
```

plus a `ZenoTOF8600` member mapping to `MS_time_of_flight` the way `ZenoTOF7600` does.

Happy to open a PR against the branch if that is useful. Thanks for the port — the Thermo, Bruker and Sciex readers have all worked against real data here.

chambm:
Already done in https://github.com/ProteoWizard/pwiz/pull/4598 but not ported to this branch yet.

chambm:
Superseded by #4619, which is the same content at the same head commit (`14e8820ac5`) but opened from `Skyline/work/20260612_net8_port` instead of `chambem2/pwiz-sharp`, so the port and the PRs stacked on it share one integration branch. GitHub cannot repoint an existing PR at a different head branch, so this one is closed rather than edited. Review discussion here is preserved and still readable; please continue on #4619.

PR #4200, opened 2026-05-12
Fix SRM import problems with multiple compounds that share a precursor m/z

bspratt:
## Summary
- `PeptideFinder.FindPeptides` returns *all* peptides whose precursor m/z falls
  within tolerance (existing `FindPeptide` left as-is for non-SRM callers).
- `SpectraChromDataProvider.ExtractChromatogramsLocked` fans each SRM spectrum
  out to every matched peptide, keying its collector dict by
  `(SignedMz, PeptideDocNode)` so each peptide gets its own
  `ChromDataCollector` and `ChromatogramGroupId`. Previously the first peptide
  found "won" the chromatogram and the others received empty placeholder
  `TransitionGroupChromInfo` entries — Wesley's "only one compound shows up"
  symptom.
- Extended `ShimadzuSrmDuplicateQ1Test` with both group-level
  (every same-Q1 peptide has non-empty `Results[0]`) and transition-level
  (shared Q3 across colliding peptides has non-empty `Results[0]`)
  assertions; uses `ExtensionTestContext.ExtShimadzuRaw` so it exercises
  the real `.lcd` vendor reader on dev/UI runs and the `.mzML` fallback
  on offscreen nightlies.

Reported by Wesley — support thread
[66356](https://skyline.ms/home/support/announcements-thread.view?rowId=66356).

## Test plan
- [x] `ShimadzuSrmDuplicateQ1Test` (added regression coverage)
- [x] `ShimadzuSrmTest`, `AgilentCEOptTest`, `AgilentMixTest`, `AgilentMseTest`, `ImportSimTest`


bspratt:
@brendanx67 @nickshulman the reporting user has been provided a special build and reports it working well

brendanx67 (commented):
Thanks for the quick turnaround on this -- it does fix the reported same-Q1 collapse, and the regression coverage is solid.

One design concern with the approach: fanning each SRM spectrum out to every peptide whose Q1 matches is substantially different from how SRM that arrives from ProteoWizard as chromatograms is handled. On the chromatogram path, `ChromCacheBuilder.GetMatchingGroups`/`FilterMatchingGroups` matches by product-ion overlap and discards a precursor that matches only a subset of another's transitions. The blind fan-out has no such guard, so a precursor that merely shares a Q1 -- e.g. a compound targeted only by a different SRM method, sharing 1-2 of another compound's transitions -- can pick up that compound's signal as noise. That cross-method bleed is exactly what the chromatogram-side matching was designed to avoid, and it has been Skyline's behavior on SRM for a long time.

I put up #4305 as an alternative that keeps the spectra path closer in principle to the chromatogram path: it pre-scans the file for the aggregate set of product channels measured per Q1, then assigns a spectrum's data only to peptides that match a majority of their own transitions. It preserves your fix -- and your duplicate-Q1 assertions, adopted unchanged -- and passes the same test plan, while not handing an incidental Q1 neighbor another compound's signal.

These are intended as alternatives, with the expectation that one merges and the other closes. Wanted to give the team a concrete comparison point rather than just flag the concern. Happy to discuss which way to go.

bspratt:
I am going to abandon this PR in favor of Brendan's [#4305](https://github.com/ProteoWizard/pwiz/pull/4305)

But I would also like to draw your attention to [some thoughts on the applicability of Spectrum Filters to this kind of data (and possibly actual chromatograms as well)](https://github.com/ProteoWizard/pwiz/issues/4309)

PR #4207, opened 2026-05-14
Fix BiblioSpec failing on blank 1/K0 cells in MaxQuant evidence.txt

bspratt:
## Summary

* Treats blank (or whitespace-only) `1/K0` cells in MaxQuant `evidence.txt` as "no IM" (0.0) instead of letting `boost::lexical_cast` throw
* Preserves row alignment because `inverseK0_` is indexed by `evidenceID`; the consumer at `MaxQuantReader.cpp:760` already treats `ionMobility == 0` as "no IM" (no `ionMobilityType` set)
* Adds a bounds check for malformed (short) rows
* Reported by Ani via skyline.ms support thread #74557

## Test plan

- [x] Manual verification with the reporter's evidence.txt — bspratt 2026-05-13

### DIFF
--- a/pwiz_tools/BiblioSpec/src/MaxQuantReader.cpp
+++ b/pwiz_tools/BiblioSpec/src/MaxQuantReader.cpp
@@ -402,7 +402,14 @@ void MaxQuantReader::initEvidence()
                 break;
             boost::split(columns, line, boost::is_any_of("\t"));
             if (colInvK0 >= 0)
-                inverseK0_.push_back(boost::lexical_cast<double>(columns[colInvK0]));
+            {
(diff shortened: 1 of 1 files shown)


bspratt:
@brendanx67 Claude chose the "cherry pick to release" label here - I don't think I would have thought to do that myself. It's true the problem exists in release, but it doesn't feel like something we'd normally bother backporting. Who is right here?

nickshulman (approved):
Looks good to me.
In general, we don't bother to back-port fixes to the release branch if the bug has existed for more than one version unless there's some reason to believe things have changed so the bug will start happening more often.
There's some weird stuff going on with this dataset in terms of retention times in the msms.txt file so I have asked the user for more information about their peptide search. There might be additional work we want to do in the future, but this fix looks good.

Issue #4209, opened 2026-05-14
IndexOutOfRangeException in AreaCV 2D histogram heat map


bspratt:
## Problem

Activating the Peak Area CV 2D histogram dock pane throws `IndexOutOfRangeException` from `HeatMapGraphPane.GraphHeatMap`. The crash fires from `GraphSummary_VisibleChanged` -> `GraphSummary.UpdateGraph` -> `AreaCVHistogram2DGraphPane.UpdateGraph` -> `HeatMapGraphPane.GraphHeatMap` when the dock pane becomes visible.

## Exception Report

- **Exception ID**: #74715
- **Source**: https://skyline.ms/home/issues/exceptions/announcements-thread.view?rowId=74715
- **Title / fingerprint**: `IndexOutOfRangeException | AreaCVHistogram2DGraphPane.cs:line 178 | 26.1.1.097-922725ca01 | d176487c229d`
- **Version**: 26.1.1.097-922725ca01 (64-bit)
- **Installation ID**: f857f40e-f993-49d6-b772-d176487c229d
- **Reported**: 2026-05-14

## Stack Trace

```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at pwiz.MSGraph.HeatMapGraphPane.GraphHeatMap(GraphPane graphPane, HeatMapData heatMapData, Int32 maxDotRadius, Int32 minDotRadius, Single yMin, Single yMax, Boolean logScale, Int32 cutoff) in HeatMapGraphPane.cs:line 51
   at pwiz.Skyline.Controls.Graphs.AreaCVHistogram2DGraphPane.UpdateGraph(Boolean selectionChanged) in AreaCVHistogram2DGraphPane.cs:line 178
   at pwiz.Skyline.Controls.Graphs.GraphSummary.UpdateGraph(Boolean selectionChanged) in GraphSummary.cs:line 387
   at pwiz.Skyline.Controls.Graphs.GraphSummary.UpdateUIWithoutToolbar(Boolean selectionChanged) in GraphSummary.cs:line 319
   at pwiz.Skyline.Controls.Graphs.GraphSummary.GraphSummary_VisibleChanged(Object sender, EventArgs e) in GraphSummary.cs:line 252
   ...
   at DigitalRune.Windows.Docking.DockPaneStripBase.OnMouseDown(MouseEventArgs e)
```

[...]

PR #4210, opened 2026-05-14
Fix IndexOutOfRangeException in AreaCV 2D histogram heat map (issue #4209)

bspratt:
## Summary
- Fixed IndexOutOfRangeException when AreaCV histogram max frequency equals 1
- Guarded heat map scale calculations against degenerate Math.Log(1) = 0 condition
- Added array bounds clamping to prevent negative indices from NaN conversions
- Created regression test to verify fix and prevent future occurrences

## Root Cause
When the maximum histogram bin frequency was exactly 1, `Math.Log(1) = 0` caused `fullScale = +Infinity`. In discrete legend mode, the color remapping calculation `0 * +Infinity = NaN`, and `(int)NaN` produces `int.MinValue` on .NET Framework, leading to array bounds violation in `_heatMapColors[colorIndex]`.

## Test Plan
- [x] Added regression test `TestHeatMapCrashRegression` integrated into existing `TestAreaCVHistograms`
- [x] Verified existing `TestAreaCVHistograms` functionality unchanged
- [x] Verified related `ClusteredHeatMapTest` functionality unchanged
- [x] All tests pass with the fix applied

See ai/todos/active/TODO-20260514_areacv_heatmap_index.md


pwiz_tools/Skyline/TestFunctional/AreaCVHistogramTest.cs
```diff
@@ -603,5 +607,57 @@ private void AssertDataCorrect(SummaryGraphPane pane, int statsIndex)
...
+            HeatMapGraphPane.GraphHeatMap(graphPane, heatMapData, 17, 2, 0.0f, 50.0f, true, 0);
+
+            // Verify we created the problematic condition (max frequency = 1)
+            AssertEx.AreEqual(1.0f, heatMapData.MaxPoint.Point.Z,
+                "Test should use max frequency = 1 to trigger the original Math.Log(1) = 0 bug");
+
+            // Verify GraphHeatMap actually processed the data (didn't return early)
+            // This ensures the test exercises the crash-prone discrete legend remap logic
+            AssertEx.IsTrue(graphPane.CurveList.Count > 0,
+                "GraphHeatMap should have populated curves to exercise the crash-prone code path");
```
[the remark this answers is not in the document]
bspratt: Fixed in 1367f54 — the assertion now sums plotted points across CurveList (`Sum(c => c.Points.Count)`) and requires > 0, so it proves a data point actually landed in a curve (the discrete legend-remap path where #4209 crashed), not just that the always-present color curves were inserted.

pwiz_tools/Shared/MSGraph/HeatMapGraphPane.cs
```diff
@@ -70,11 +75,26 @@ public static void GraphHeatMap(GraphPane graphPane, HeatMapData heatMapData, in
...
             double fullScale = (_heatMapColors.Length - 1.0) / maxZValue;
             double scale = isDiscrete
                 ? Math.Min(legendStep, fullScale)
                 : fullScale;
 
+            // Ensure scale is finite and positive to avoid division by zero in legend calculations
+            if (double.IsNaN(scale) || double.IsInfinity(scale) || scale <= 0)
+            {
+                scale = 1.0;
+            }
```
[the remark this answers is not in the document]
bspratt: Fixed in b9c52d5 — folded the negative-log case into the maxZValue guard (now `maxZValue <= 0`), so maxZValue is clamped to a positive value *before* fullScale is derived. fullScale and scale can no longer go negative, and the later `scale <= 0` check is now just belt-and-suspenders. No behavior change for real inputs: count/frequency callers have Math.Log(maxZ) >= 0, and non-log callers have maxZ > 0 (MaxPoint tracks only Z > 0 points).

pwiz_tools/Shared/MSGraph/HeatMapGraphPane.cs
```diff
@@ -106,15 +127,18 @@ public static void GraphHeatMap(GraphPane graphPane, HeatMapData heatMapData, in
             var usedIntensities = new HashSet<int>();
             foreach (var heatPoint in points) {
                 // A log scale produces a better visual display.
-                int intensity = Math.Min((int)((logScale ? Math.Log(heatPoint.Point.Z) : heatPoint.Point.Z) * scale), curves.Length - 1);
+                double intensityValue = (logScale ? Math.Log(heatPoint.Point.Z) : heatPoint.Point.Z) * scale;
```
rita-gwen, line 134: Well, if you are obsessively checking for log's validity, you probably should make sure heatPoint.Point.Z is not 0 as well.

bspratt: Strictly speaking that can't happen - 0 values get filtered out beforehand - but it's a cheap check so why not go for belt and suspenders. Thanks! Otherwise OK?

bspratt: Good point — added a `heatPoint.Point.Z <= 0` skip at the top of the point loop in 2eeb15fb6, so the log scale never evaluates Math.Log(0).

PR #4222, opened 2026-05-17
Added inline image return for MCP screenshot and graph tools

### LINKED ISSUE #4220: Add inline image return for screenshot / graph MCP tools
## Summary

Add a `returnFormat` parameter to the three image-producing MCP tools — `skyline_get_form_image`, `skyline_get_graph_image`, and `skyline_get_tutorial_image` — so the model can receive the PNG directly in the tool response via MCP's `ImageContentBlock` instead of a file path it then has to Read. Mirrors the validation-loop motivation behind `get_report_rows` / `get_report_from_definition_rows`: same MCP call, model sees the image, no file round-trip.

## Motivation

The three tools today serialize the image to a file under `%LOCALAPPDATA%\Skyline\mcp\tmp\` and return the path. The model then has to issue a Read tool call to actually see the image — two MCP turns where one should do. For graph re-renders during iterative analysis (compare-by-eye, "did my settings change move the peak?") that file step is friction.

MCP supports binary content in tool responses via `ImageContentBlock` (base64 `data` + `mimeType`). The SDK we use (ModelContextProtocol 0.8.0-preview.1) already exposes `CallToolResult` with a `Content` array of `ContentBlock`, so the server side is straightforward.

The catch: **not every MCP client renders image content correctly today.** Some clients only display text content; some choke on large base64 payloads. The new path must default to inline (fastest for clients that support it) with a clear escape hatch back to file-on-disk.

## Design

A single parameter — `returnFormat` — controls return shape on each existing tool:

```
skyline_get_form_image(formId, returnFormat="auto" | "inline" | "file", filePath=null)
skyline_get_graph_image(graphId, returnFormat="auto" | "inline" | "file", filePath=null)
skyline_get_tutorial_image(name, imageFilename, returnFormat="auto" | "inline" | "file", language="en", filePath=null)
```

- `returnFormat="auto"` (default): try inline; fall back to file if the rendered payload would exceed the server-side cap, or if the paired Skyline is too old to serve the inline JSON-RPC method (see [Version compatibility](#version-compatibility)).
- `returnFormat="inline"`: always inline; error if over cap. Surfaces a version-mismatch error if Skyline is too old, so the caller can decide explicitly to retry with `auto` or `file`.
- `returnFormat="file"`: existing behavior (back-compat for callers that already pass `filePath`). Always works against any supported Skyline version.
- `filePath` is honored only on `file` or `auto-fell-back-to-file`; otherwise ignored, with a one-line note in the response so the caller isn't surprised.

Why parameter over companion tools: the only difference between modes is "encode the same bitmap as a file or as base64." Same source, same permission flow, same form ID. Adding three companion tools (precedent: `_rows`) would grow the surface to 50 tools for what's really a return-format toggle — the `_rows` precedent applies when the two variants have genuinely different parameter sets and mental models, which is not the case here.

### Inline form (success)

```json
{
  "content": [
    { "type": "image", "data": "<base64 PNG>", "mimeType": "image/png" }
  ]
}
```

### Auto-fell-back-to-file form

Same shape regardless of *why* the fallback fired (size cap exceeded, or the connected Skyline doesn't support the new JSON-RPC method yet — see [Version compatibility](#version-compatibility) below). The text message names the reason.

```json
{
  "content": [
    { "type": "text", "text": "Image exceeded inline cap (1.2 MB > 500 KB). Saved to: C:/.../skyline-form-PeakAreas-... .png\nUse the Read tool to view." }
  ]
}
```

### Explicit `inline` over cap (error)

Tool returns an error response so the caller knows to retry with `returnFormat="auto"` or `"file"`.

## Key behaviors

- **Server-side size cap on inline payload.** A 2000x1500 PNG can easily exceed 1 MB base64; LLM context budgets don't absorb that well. Cap at ~500 KB base64 (tune in implementation; same pattern as `MaxResponseChars` in the row tools — `internal static int` knob with `InternalsVisibleTo("TestFunctional")` for tests).
- **MIME type is always `image/png`.** No format negotiation in v1.
- **Capability discovery is out of scope.** v1 always honors the caller's `returnFormat`; clients that don't render images get a usable (if useless-looking) base64 blob when `returnFormat="inline"`. v2 can add client-capability-aware routing if a clear signal becomes available.
- **Permission flow unchanged for v1**, but interacts with the non-blocking permission prompt work (see related issue): whichever lands first should leave a clean seam for the other.

## Implementation notes

- `JsonUiService.GetFormImage` / `GetGraphImage` and `JsonTutorialCatalog.FetchTutorialImage` already produce a `Bitmap`. Refactor so the bitmap-producing core is separate from the file-writing tail; add a sibling method (or overload) that returns the PNG bytes.
- New POCO `ImageBytesMetadata { byte[] Data, string FilePath, string MimeType, int? OriginalBytes }` in `JsonToolModels.cs` (link-compiled into both NF472 and .NET 8). `FilePath` is set only when the inline path fell back; `OriginalBytes` is the pre-cap size so the LLM can see why the fallback happened.
- `IJsonToolService` gets three new methods returning `ImageBytesMetadata`. Existing methods unchanged.
- MCP tool wrappers (`SkylineTools.cs`) convert `ImageBytesMetadata.Data` to an `ImageContentBlock` and emit it via the tool's return; when `FilePath` is set, emit a `TextContentBlock` with the path message instead. Return type for the tool method changes from `string` to `IList<ContentBlock>` (or whichever shape the SDK exposes — confirm in implementation).
- Existing `string`-returning wrappers stay for `returnFormat="file"`. Or: a single wrapper that returns content blocks and handles both modes.

## Version compatibility

This change spans three boundaries; each has its own compatibility surface. The design is **additive at every layer that matters**, and the auto-fallback handles the one place where mixed versions can still collide.

### IJsonToolService (C# API used by other tools)

Purely additive: three new methods (`GetFormImageBytes`, `GetGraphImageBytes`, `GetTutorialImageBytes`) returning `ImageBytesMetadata`. No existing methods change signature. Third-party consumers that link against `IJsonToolService` keep working unchanged.

### MCP tool schema (LLM ↔ SkylineMcpServer)

The MCP `tools/list` schema gains one optional parameter (`returnFormat`) with a default on each of the three existing tools. Per MCP protocol, optional parameters with defaults are additive — clients that don't know to send the new field get the default, which matches today's behavior (file path on disk) for any tool whose schema they haven't re-fetched.

### JSON-RPC over named pipe (SkylineMcpServer ↔ Skyline)

This is the only layer where a mixed pairing can fail, and the failure mode is well-bounded:

- **New SkylineMcpServer + old Skyline**: the wrapper for `returnFormat="auto"` or `"inline"` calls the new `GetFormImageBytes` JSON-RPC method, which the old Skyline doesn't expose. Existing error-enrichment in `SkylineTools.Invoke` already maps this to `"This method is not available in {skylineId}. A newer version of Skyline may be required."` — so the failure is *visible and self-describing*, not silent.
- **Old SkylineMcpServer + new Skyline**: old MCP doesn't call the new methods; new Skyline serves them anyway and they go unused. No problem.

For `returnFormat="auto"` specifically, the wrapper **catches the "Unknown method" exception from the new bytes call and falls back to the existing `GetFormImage(formId, filePath)` JSON-RPC call**. The response is then the same shape as a size-cap fallback — text content with the file path and a one-line note ("Inline image not available on Skyline {ver}; saved to file."). Users on the skew pairing get a working tool; users on `returnFormat="inline"` see the explicit version-mismatch message so they know exactly what to retry with.

Net effect: a user can update SkylineMcpServer ahead of Skyline (or vice versa) and the default tool behavior keeps working.

## Tests

- Inline return: bitmap produced, base64-encoded, MIME type is `image/png`, byte length matches the file form for the same source.
- Size-cap fallback: lower the size cap (via test-only knob) to force the fallback; response carries `FilePath` and the file on disk has identical bytes to the in-memory bitmap.
- Version-skew fallback: simulate an "Unknown method" response on the JSON-RPC bytes call (mock the client or use a test-only override) and verify `returnFormat="auto"` falls back to the file path while `returnFormat="inline"` surfaces the version-mismatch error.
- Explicit `inline` over cap: returns an error response (not a silent fallback).
- Explicit `file`: matches today's behavior exactly (regression).
- Permission denial: same behavior across all `returnFormat` values.
- `SkylineMcpTest` end-to-end: verify the wire format actually emits `ImageContentBlock` JSON the MCP framework recognizes.

## Out of scope

- Removing or changing the existing file-based response when `returnFormat="file"` is requested.
- Multi-image responses in one tool call.
- Formats other than PNG.
- Client capability negotiation beyond a best-effort signal.

## Related

- Non-blocking screen-capture permission prompt (separate issue) — touches the same `GetFormImage` flow. Either order works, but the second one to land should adapt to the first's permission model.

## Getting Started

brendanx67:
## Summary

* New `returnFormat` parameter on `skyline_get_graph_image`, `skyline_get_form_image`, and `skyline_get_tutorial_image`: `auto` (default, inline with file fallback), `inline` (always inline, error if cap exceeded or Skyline too old), `file` (preserves existing behavior).
* Wrappers now return `CallToolResult` and emit `ImageContentBlock` for inline mode, so the model sees the PNG directly without a separate `Read` round-trip.
* New `ImageBytesMetadata` POCO and three new `IJsonToolService` methods (`GetGraphImageBytes`, `GetFormImageBytes`, `GetTutorialImageBytes`) carry raw PNG bytes plus a server-suggested fallback path that the wrapper writes when the inline payload exceeds the cap.
* Cap is `~500 KB` raw bytes by default, tunable via `SKYLINE_MCP_INLINE_IMAGE_CAP_BYTES` (same pattern as `MaxResponseChars` for the row tools).
* Version-skew safe: `auto` against an older Skyline catches "Unknown method:" and falls back to the existing file-based JSON-RPC call; `inline` surfaces the version error so the caller can decide. `IJsonToolService` and MCP tool schema changes are purely additive.

Fixes #4220

## Test plan

- [x] `TestSkylineMcp` extended with three new validation blocks across two subprocess scenarios:
  - file mode regression: `TextContentBlock` with file path; file on disk passes PNG signature check.
  - auto / inline (image fits cap): `ImageContentBlock` with `mimeType=image/png`; base64 bytes pass PNG signature check.
  - auto over cap (separate subprocess, `SKYLINE_MCP_INLINE_IMAGE_CAP_BYTES=100`): `TextContentBlock` containing "exceeded inline cap" plus a path that exists on disk and is a valid PNG.
  - inline over cap: `CallToolResult` with `isError=true` and an error message containing "exceeded inline cap".
  - Direct JSON-RPC pipe call to `GetGraphImageBytes` round-trips `ImageBytesMetadata` across the Newtonsoft (server) / System.Text.Json (client) boundary.
- [x] `CodeInspection` clean.
- [x] Skyline.sln + SkylineMcp.sln build clean.


brendanx67:
Addressed Copilot review in 175a802:

* **`SkylineTools.cs` file-mode denial** (#1043): file-mode now detects screen-capture denial responses (prefix "Screen capture") and passes them through verbatim instead of wrapping as "saved to: ...". `InvokeImage`'s `fileCall` was promoted from `Func<…,string>` to `Func<…,CallToolResult>` so each tool formats its own file-mode response.
* **Tutorial not-found regression** (#1043 second comment): the tutorial wrapper now returns a `TextContentBlock` (no `isError`) for `metadata == null`, restoring the prior contract.
* **`GetFormImageBytes` denial throws** (`JsonUiService.cs:355`): the bytes method now returns `ImageBytesMetadata { Message = denial }` instead of throwing. `InvokeImage` emits the `Message` as plain text (no `isError`) so `auto` mode behaves like the legacy file path on denial.
* **`ImageBytesMetadata` doc** (`JsonToolModels.cs:196`): documented that `Data` carries arbitrary image bytes and that `MimeType` is `image/png` for form/graph captures but follows the source filename extension for tutorials. Added doc for the new `Message` field.
* **English literal in tests** (`SkylineMcpTest.cs:595, 614`): extracted to `JsonToolConstants.MSG_INLINE_CAP_EXCEEDED` and updated both call sites + the production wrapper to reference the shared constant.

Tests + CodeInspection re-run green locally.

brendanx67:
Self-review (fresh-context Claude agent) returned six findings. All addressed or considered, in commit a7dd4e2:

**Addressed in code:**

* **`SkylineTools.cs:1056` (important)** - version-skew catch was matching on `ex.Message.Contains("Unknown method:")`, brittle to any future error whose message happens to include that substring. Threaded the JSON-RPC error code through `SkylineJsonToolClient.Call` via a new public `JsonRpcException` (derives from `InvalidOperationException` for back-compat). Detection in `SkylineTools.IsMethodNotFound` now matches on `code == ERROR_METHOD_NOT_FOUND` instead of the message text.
* **`SkylineTools.cs:1226` (important)** - `WriteBytesToDisk` in cap-fallback could throw `IOException` / `UnauthorizedAccessException`, which `InvokeContent` was mistranslating as "Skyline disconnected during the operation." Wrapped the disk write in a tight try/catch that surfaces a real error naming the actual write failure and the attempted target.
* **`SkylineMcpTest.cs:548,621` (nit)** - graph cleanup was the last line of the happy path, so an assertion failure would leave the graph open and trigger a GC-LEAK on top of the real failure. Wrapped each in `try { ... } finally { close graph }`.
* **Tutorial-image dead null-check (nit)** - the same structural fix above (typed `JsonRpcException`) resolves this implicitly. Server-side `IOException` now flows back as `JsonRpcException` via the JSON-RPC error envelope, not raw `IOException`, so the wrapper's `InvokeContent` no longer mistranslates a failed GitHub download as "Skyline disconnected".

**Considered and decided not to fix in this PR:**

* **`Tools/SkylineTools.cs:555,572,670` positional-arg breakage (important)** - verified no in-tree positional C# callers (`SkylineTools.GetGraphImage(...)` not used anywhere). MCP/JSON callers always use named arguments. Acceptable as-is.
* **`JsonUiService.cs:531` timestamp collision (nit)** - the second-resolution path can race on concurrent calls, but the fix touches the existing file-on-disk path that has always had this property. Out of scope for this PR; worth a separate issue if it ever bites.

**Tests added:**

* `TestVersionMismatchError` now asserts the typed-exception contract: calling a non-existent method must throw `JsonRpcException` with `Code == ERROR_METHOD_NOT_FOUND` and a message naming the unknown method. Belt-and-braces alongside the existing JSON-RPC envelope shape check.

Build + tests + CodeInspection all green locally.

**Author follow-up on the agent's question:** yes, threading the JSON-RPC `code` through the typed client was the right call - did it in this commit. Future call sites that need to distinguish method-not-found from `InvalidOperationException` can `catch (JsonRpcException ex) when (ex.Code == ...)` cleanly instead of substring sniffing.

PR #4245, opened 2026-05-28
Make the post-test GC leak check adaptive

bspratt:
## Summary

- `GarbageCollectionTracker.CheckAfterTest`'s Phase 2 retry loop now keeps forcing GC while the survivor count is still dropping, and declares a leak only after the count has been stable across `GC_STABLE_CYCLES_REQUIRED` consecutive cycles (or `GC_MAX_RETRY_CYCLES` total cycles elapse).
- Real-leak detection time is unchanged at a 1.5s minimum (3 stable cycles), but transient cleanup that completes within ~10s no longer false-positives.
- Addresses agent-categorical GC-LEAK reports on MacCoss TeamCity Agent 1 (Windows Server 2022 amd64) — 125 deterministic Pass2 tests with survivor counts capping around 9 (cleanup completing past the old 1.5s budget). AWS agents (Windows Server 2019 amd64) and master on the same MacCoss agent were both clean, ruling out a hard static root and pointing at the threshold itself.

## Test plan

- [ ] TeamCity nightly run on MacCoss Agent 1 (Windows Server 2022 — was the source of the false positives)
- [ ] TeamCity nightly run on AWS agents (Windows Server 2019 — regression check, should remain clean)
- [ ] Spot-check that a synthetic real leak still trips the detector within the original ~1.5s window

### DIFF
--- a/pwiz_tools/Skyline/Controls/FilesTree/FilesTreeForm.cs
+++ b/pwiz_tools/Skyline/Controls/FilesTree/FilesTreeForm.cs
@@ -111,6 +111,11 @@ namespace pwiz.Skyline.Controls.FilesTree
             filesTree.LostFocus -= FilesTree_LostFocus;
 
             base.OnHandleDestroyed(e);
+
+            // Break back-pointer to SkylineWindow so accessibility/layout-cache
+            // pinning of this form (e.g. via DockZone.cachedLayoutEventArgs) does
+            // not transitively root SkylineWindow and its undo stack.
+            SkylineWindow = null;
         }
 
         [Browsable(false)]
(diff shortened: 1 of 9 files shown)


bspratt:
Superseded by #4248 -- a clean, single-commit PR with just the UiaDisconnectAllProviders mitigation. Most of the commits on this branch were exploratory probes (adaptive threshold, FinalCheck attribution, per-form Receiver/back-pointer null-on-handle-destroyed) testing wrong hypotheses about the SkylineWindow leak root cause. The actual root was Windows UI Automation pinning AccessibleObjects on Control instances, addressed in #4248.

PR #4246, opened 2026-05-29
OspreySharp Stage 5 perf: SIMD SVM + parallel LOESS + parallel KDE

brendanx67:
## Summary

* Fix #1 vectorized the `LinearSvmClassifier.Train` dual coordinate descent inner loops via `System.Numerics.Vector<double>`. AVX2/AVX-512 lowering of the dot product + weight update on the SVM hot path. Astral 3-file end-to-end C# wall: 22:02 → 19:48 (-10%).
* Fix #5b parallelized `LoessRegression.LoessFitInternal`'s outer per-x-point loop via `Parallel.For`. Each point's local regression is independent. At Astral 3-file scale, reconciliation planning runs 9 LOESS fits (3 files × initial + 2 robustness iters) that previously consumed ~130s of single-core CPU; the parallel form collapses to a few seconds wall. Astral 3-file end-to-end C# wall: 19:48 → 15:17 (-23% on top of Fix #1).
* Fix #6 parallelized `PepEstimator.Fit`'s outer KDE binning loop via `Parallel.For`. Each bin's two `Pdf` evaluations are independent and write to their own `bins[i]` slot; the inner `Pdf` sum stays serial so PDF values are bit-for-bit identical to the prior implementation. Astral 3-file end-to-end C# wall: 15:17 → 15:08.
* `Measure-Pipeline.ps1` aligns the stage5/stage6 boundary cross-impl by deriving the reconciliation portion of C# stage5 from the existing `[TIMING] Percolator/Simple FDR` and `[TIMING] First-pass protein FDR` lines and shifting it into stage6 (where Rust labels the same work). This resolves the historical "Astral stage5 4x C#/Rust" figure as a combination of the now-fixed LOESS gap and the now-fixed label mismatch.
* `Osprey-workflow.html` refreshed with round-4 medians (3-rep) across Windows native + WSL /mnt/c + WSL /home. C# beats Rust end-to-end on every cell: Windows native 0.79x Stellar / 0.66x Astral; WSL /mnt/c 0.77x / 0.79x; WSL /home 0.93x / 0.96x.

## Test plan

- [x] `Compare-EndToEnd-Crossimpl.ps1 -Dataset Stellar -Files Single` — PASS at 1e-9 (Stage 7 protein FDR + blib SQL row+col)
- [x] `Compare-EndToEnd-Crossimpl.ps1 -Dataset Astral -Files All` — PASS at 1e-9 (Stage 7 protein FDR + blib SQL row+col)
- [x] Round-4 perf table refresh: 3-rep median across Windows native + WSL /mnt/c + WSL /home; both impls, both datasets; variance under 7s on Stellar / 25s on Astral per cell
- [x] All `OspreySharp.Test` tests pass (build with -RunTests green)

### DIFF
--- a/pwiz_tools/OspreySharp/Osprey-workflow.html
+++ b/pwiz_tools/OspreySharp/Osprey-workflow.html
@@ -504,13 +504,16 @@
   <p>
     <strong>End-to-end performance (3-file regression, Windows
     native, C: SSD source data, AV-scan excluded; all 8 cells
-    median-of-3 refreshed 2026-05-24):</strong>
-    Measured on the <code>test/integration-bucket3</code> branch
-    (PRs #38-#45 stacked, including the decoy gap-fill exclusion fix
-    AND the Welford → sum/n revert that recovered ~21 s on Stellar
-    Rust stage1to4) immediately after 3-file Astral straight-through
-    cross-impl bit-equality landed at 1e-9. Variance is &lt;5 s on
-    Stellar and &lt;90 s on Astral across the three repeats.
+    median-of-3 refreshed 2026-05-28 — round 3):</strong>
+    Measured on the <code>Skyline/work/20260527_svm_stage5_perf</code>
+    branch with Fix #1 landed (SIMD inner loops in
+    <code>LinearSvmClassifier.Train</code>). Round 3 walls run
+    ~20% slower than round 2 (2026-05-26 night) for both impls
+    uniformly — environmental drift, not a runtime change.  The
+    C#/Rust ratios are the load-bearing comparison; absolute walls
+    should be interpreted relative to the round-3 column on its own.
+    Variance is &lt;10 s on Stellar and &lt;90 s on Astral across
+    the three repeats.
   </p>
   <table style="margin: 8px 0; font-family: ui-monospace, SFMono-Regular, monospace; font-size: 11px; border-collapse: collapse;">
     <thead>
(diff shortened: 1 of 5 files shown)


brendanx67:
Following up on the self-review's question about the SIMD SVM parity drift profile:

The `Compare-EndToEnd-Crossimpl.ps1` gate that PASSed on Stellar 1-file and Astral 3-file (1e-9 per-column on Stage 7 protein FDR + 1e-9 per row+col on blib SQL contents) only reports PASS/FAIL at the tolerance, not the actual max-absolute-delta per column. So I can confirm the SVM SIMD change consumes a sub-1e-9 fraction of the gate's headroom, but I don't have a precise number.

If we want to characterize how much SIMD-induced drift the gate can absorb before tightening it, the right follow-up is a small change to `Compare-Stage7-Crossimpl.ps1` and `Compare-Blib-Crossimpl.ps1` to also report the observed max-absolute-delta alongside PASS/FAIL. That would give us a quantitative budget headed into future SIMD ports (LDA, peak scoring). Filing as a follow-up for the next sprint rather than blocking this PR on it.

Commits 9a503c6 address the documentation and test-coverage findings (lane-stride comment in `LinearSvmClassifier.cs`, clarified `PepEstimator.cs` comment, new `TestSvmHighDimensional` exercising p=21 features through the SIMD path).

PR #4260, opened 2026-05-31
Diagnose vendor reader test file-lock failures: retry + RestartManager

bspratt:
## Summary

`VendorReaderTestHarness` does a rename round-trip on the test `.d`
directory after each vendor reader test to verify all handles were
released. On TeamCity it occasionally fails ("Cannot rename ...: there
are unreleased file locks!") even though the same test passes most
other runs -- consistent with a transient external holder (Windows
Defender, Search indexer, OS file-handle release lag) rather than a
reader-side handle leak (which would fail deterministically).

Two changes:

- **Retry the rename** up to 5 times with backoff (0/250/500/750/1000 ms,
  total 2.5s budget on failure paths only). Transient holders typically
  clear in well under a second; a real reader bug remains pinned no
  matter how long we wait, so the final failure path still surfaces
  those.

- **On final failure, use the Windows Restart Manager API**
  (`RmStartSession` / `RmRegisterResources` / `RmGetList`) to enumerate
  the processes holding handles on the path, and include the result
  plus the underlying OS error in the thrown message. Replaces the
  generic "unreleased file locks!" string with concrete diagnostic
  data, e.g. `Lock holders: MsMpEng.exe (PID 1234)`.

Restart Manager support is MSVC-only; on other toolsets the helper
returns a stub string. New library dependency: `rstrtmgr.lib` on MSVC.

## What this does NOT do

- It does not silence the failure in CI -- a real reader-side handle
  leak still fails the test, just with better diagnostics.
- It does not change reader behavior -- this is purely test-harness
  diagnostic code.

## Test plan

- [ ] TC nightly -- the intermittent ImsSynthCCS.d failure should
      either stop occurring (transient AV) or, if it still fires,
      now name the locking process in the failure message
- [ ] Linux/macOS CI -- builds with the stub helper, no Restart
      Manager linkage

### DIFF
--- a/pwiz/utility/misc/Filesystem.cpp
+++ b/pwiz/utility/misc/Filesystem.cpp
@@ -527,7 +527,12 @@ PWIZ_API_DECL void force_close_handles_to_filepath(const std::string& filepath, 
 
 PWIZ_API_DECL std::string find_locking_processes(const std::string& path)
 {
-#ifdef _MSC_VER
+    // Gate on PWIZ_HAS_RESTART_MANAGER (set by the Jamfile under
+    // <toolset>msvc:) rather than _MSC_VER -- clang-cl also defines
+    // _MSC_VER but the Jamfile does not link rstrtmgr.lib for non-msvc
+    // toolsets, so an _MSC_VER-only gate would compile the body but
+    // fail at link.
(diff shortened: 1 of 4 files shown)


chambm:
This is a good addition but the find locking process function should go
into filesystem.cpp.

PR #4265, opened 2026-06-03
Fixed spurious GC-LEAK reports by opting the test host into the latest accessibility level

bspratt:
## Summary

* The post-test `GarbageCollectionTracker` intermittently reported `SkylineWindow, SrmDocument xN` for 100+ functional tests, but only on the Windows Server 2022 agent (TCA1). Root cause: at the framework-default WinForms accessibility level (TestRunner targets .NET 4.7.2), closing a window does not release the UI Automation accessible-object provider handles, which transitively pin `SkylineWindow`/`SrmDocument` through each dialog's back-pointer.
* `TestRunner.Main` now explicitly sets all four `Switch.UseLegacyAccessibilityFeatures*` to `false` before any control is created, opting the test host into the latest accessibility level. Empirically, at that level the providers are released on window close and the leak does not occur.
* Test-host only -- this lives in `TestRunner`, so shipping Skyline accessibility behavior is unchanged. It is OS-independent and harmless on agents that were already clean.
* Removes the ineffective `UiaDisconnectAllProviders` call (and its P/Invoke) added by #4248 -- see below.

## Why the previous fix (#4248) didn't take

#4248 had the right diagnosis -- the same dotMemory retention path we re-confirmed:

    RefCounted handle (Windows accessibility) -> ControlAccessibleObject -> Control
      -> events (EventHandlerList) -> EventHandler -> dialog -> SkylineWindow

...but the wrong mechanism:

* It called `UiaDisconnectAllProviders()` to release the providers at runtime. On this OS / framework patch level that call is inert. This was actually raised in review -- "I was hopeful that `UiaDisconnectAllProviders` would fix things, but ... this API call does not do anything" -- but it was attributed to a possibly-different cause, and the test-plan item "TC nightly leak count should drop to 0" was never confirmed before merge. The next nightly still showed the leaks.
* The #4248 repro was a debug build under dotMemory, and dotMemory's own hooks activate UIA (noted in #4248 itself). Under the profiler the leak is always present and a disconnect cannot be seen to "turn it off"; without the profiler (CI) the result was never confirmed. So the local repro could not validate the CI fix.

The real lever is not a runtime cleanup call -- it is the accessibility level, set once at startup, which governs whether the framework retains these handles at all. Forcing the switches is a deterministic on/off (default -> leaks; all-`false` -> clean; all-`true` -> clean); `UiaDisconnectAllProviders` never was.

## Why explicit `false`, not "leave it at the default"

`AppContext.TryGetSwitch` reads the same (all `false`) whether a switch is unset or explicitly set to `false`, yet the two behave oppositely: the framework default for a 4.7.2 target leaks, while explicitly opting into the latest level is clean. So the switches must be set explicitly; leaving them unset does nothing. (Setting all four to `true` -- full legacy -- also stops the leak by disabling the modern UIA provider path entirely; that is a larger change, kept in a code comment as a fallback if a future framework regresses the lifetime behavior at the latest level.)

## Test plan

Validated on the Windows Server 2022 agent (TCA1) that produced the failures -- debug build, no profiler attached:

- [x] Default level (switches unset) -- reproduces the leak (`TestManagingSearchTools x2`, `TestEncyclopeDiaPeptidesWithNoSignal x3`)
- [x] Forced all-`false` (this fix) -- both offenders clean
- [x] Forced all-`true` (legacy) -- also clean, confirming the accessibility level is the lever
- [x] Full Pass2 with the fix in place -- the agent's GC-LEAK false positives are gone and the accessibility-level change introduced no collateral failures (one unrelated failure, `TestDiaUmpireWiffFile`, is a pre-existing persistent-data-dir hygiene issue where DiaUmpire writes `-diaumpire.mz5`/`.mzid.gz` into a persistent dir -- independent of this change)

### DIFF
--- a/pwiz_tools/Skyline/TestRunner/Program.cs
+++ b/pwiz_tools/Skyline/TestRunner/Program.cs
@@ -272,6 +272,22 @@ namespace TestRunner
         static int Main(string[] args)
         {
             Console.OutputEncoding = Encoding.UTF8;
+
+            // EXPERIMENT: force a deterministic WinForms accessibility level before any control is
+            // created, overriding config/framework-quirk defaults, so the UseLegacyAccessibilityFeatures
+            // (#2) test isn't confounded by an unknown switch state. SetSwitch wins over config when it
+            // runs before the level is first read. Flag A below logs the resulting state.
+            //   SKYLINE_LEGACY_ACCESSIBILITY=1 -> full legacy (Level0); =0 -> latest (Level4); unset -> defaults.
+            var legacyAccessibility = Environment.GetEnvironmentVariable("SKYLINE_LEGACY_ACCESSIBILITY");
+            if (legacyAccessibility == "1" || legacyAccessibility == "0")
+            {
+                var useLegacy = legacyAccessibility == "1";
+                AppContext.SetSwitch("Switch.UseLegacyAccessibilityFeatures", useLegacy);
+                AppContext.SetSwitch("Switch.UseLegacyAccessibilityFeatures.2", useLegacy);
+                AppContext.SetSwitch("Switch.UseLegacyAccessibilityFeatures.3", useLegacy);
+                AppContext.SetSwitch("Switch.UseLegacyAccessibilityFeatures.4", useLegacy);
+            }
+
             Application.SetUnhandledExceptionMode(UnhandledExceptionMode.CatchException);
             Application.ThreadException += ThreadExceptionEventHandler;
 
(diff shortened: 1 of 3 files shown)


nickshulman (approved):
Nice. This fixes the leaks I was seeing on macs2 when I had Remote Desktop connected.

I imagine we should do this in app.config for Skyline so that we always run with the latest accessibility stuff. I can't imagine that we are actually depending on any sort of old behavior.

bspratt:
> Nice. This fixes the leaks I was seeing on macs2 when I had Remote Desktop connected.

Excellent!

> I imagine we should do this in app.config for Skyline so that we always run with the latest accessibility stuff. I can't imagine that we are actually depending on any sort of old behavior.

I have no idea what folks are expecting of Skyline accessibility in the wild, I was just going to leave it alone since there have been no complaints. The leak is minor and really doesn't happen at human scale.

PR #4272, opened 2026-06-05
Guarded Thermo analyzer lookup against empty analyzer list

bspratt:
## Summary

* Guard the Thermo analyzer lookup in `findInstrumentConfiguration` with `analyzerCount > 0` before accessing `analyzer(analyzerCount - 1)`. When an instrument configuration has no analyzer components, `analyzerCount - 1` underflows (size_t) and throws `out_of_range`; the guard avoids the exception path entirely.
* Map `BrowseOnly|Win32` to the `BrowseOnly|x64` build config in `pwiz.sln` and drop the Win32 build entry, so pressing F5 (run/debug) does not trigger a Win32 build that cannot succeed.

## Test plan

- [x] Built and tested locally by the author.

### DIFF
--- a/pwiz.sln
+++ b/pwiz.sln
@@ -11,7 +11,8 @@ Global
 		BrowseOnly|x64 = BrowseOnly|x64
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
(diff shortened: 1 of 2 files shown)


chambm:
The win32 config lets you look at the 32bit code branches (msfilereader
only maybe?). Not sure we need to fix this until the c++ goes away
entirely. The analyzer guard is good though.

PR #4311, opened 2026-06-17
pwiz-sharp: added mzPeak format support (read, write, msconvert)

chambm:
## Summary

* Adds **mzPeak** (Apache Parquet tables packed in a ZIP) as a first-class MSData format in pwiz-sharp — reader, writer, `IReader`/format plumbing, and `msconvert --mzpeak` output.
* Full-document round-trip parity with mzMLb (metadata + binary) across all 8 vendor readers, gated in the vendor reader test harness (`TestMzPeakRoundTrip`).
* **Cross-stack read support** for mzPeak files written by the independent mzPeak.NET (Apache-Arrow) stack: type/width-adaptive column reads (uint8 ms_level, float32 scalars, int32/uint32/uint64 ints), Parquet-statistics row-group discovery with multi-group merge, and the separate `wavelength_spectra_*` (UV/DAD) entries appended after the MS spectra.
* Lazy per-row-group reads with bounded memory; parquet read in place from the ZIP (no extraction) when entries are Stored.

## Test plan

- [x] MsData.Tests - 97/97 (mzPeak column/IReader/round-trip + cross-stack read tests against committed mzPeak.NET fixtures)
- [x] Thermo/Agilent/Bruker/Sciex/Waters/Mobilion/UIMF/UNIFI vendor reader tests - 112/112 (each includes the mzPeak round-trip)

See ai/todos/active/TODO-20260616_pwiz_sharp_mzpeak.md


chambm:
@mobiusklein I vibe coded this partly based on MzPeak.NET. Want to try it out?

mobiusklein:
And I had just reinstalled Visual Studio too.

This looks structurally correct, but I haven't run it yet. I know you're
only prototyping so the null marking, chunked layout and so on aren't there
yet, which is where a lot of the complexity lives. I saw the comments say
something about no async runtime. I had to absorb the async dependency from
ParquetSharp's Arrow interop layer. Is this a no-go scenario for you?

chambm:
> And I had just reinstalled Visual Studio too. This looks structurally correct, but I haven't run it yet. I know you're only prototyping so the null marking, chunked layout and so on aren't there yet, which is where a lot of the complexity lives. I saw the comments say something about no async runtime. I had to absorb the async dependency from ParquetSharp's Arrow interop layer. Is this a no-go scenario for you?

Right, we don't use async/await in Skyline as a rule due to its tendency to use the system thread pool which makes non-instrumented handle and leak detection annoyingly tricky. I did defer the chunked implementation but I'll kick it off now. This format seems really complicated to me: it'll be interesting to see how much traction it gets in the community.

mobiusklein:
Thank you for explaining. If the async was isolated to be entirely within a `SpectrumList` implementation and/or run on a same thread-`TaskScheduler`, would that be acceptable? The same thread scheduler seems like it would be safer for leak avoidance, but I have not dealt with this issue myself.

The only reason the original code used `async` anywhere was because ParquetSharp _only_ lets you read Arrow arrays via an `async` enumerator interface they implement from the .NET Apache Arrow library, and that leaks into everything else. My attempts to write nesting depth-independent column readers against ParquetSharp weren't successful, but it was comparatively easy using Arrow's Visitor API. The Arrow API doesn't directly use the .NET Arrow library to do the conversion, that gets done in the C++ library bindings, but getting the Arrow arrays out is more complicated.

Which part is most complicated seeming to you?

mobiusklein:
I realized something silly when I was writing my implementation in C#. `Task.Wait` != `await Task`, and blocks the calling thread with no interaction with the runtime's global thread pool. If I provided a pure synchronous implementation would that be acceptable instead of re-writing the whole format here?

chambm:
What do you mean rewriting the whole format? But yes, a pure synchronous implementation should be fine.

mobiusklein:
Here you are re-implementing reading every aspect of the format using ParquetSharp's raw Parquet-only API, which has to work from the root of the schema instead of walking it like a tree, whether because that's how Claude chose to do it or because it ran into a quirk of how ParquetSharp's `LogicalVisitor` API worked and had to backtrack. Additionally, it hard-coded the columns it would load, and that prevents it from handling arbitrary CV params promoted to columns and won't load any array that isn't the m/z or intensity.

I've recently added a pure synchronous version of each API in https://github.com/HUPO-PSI/mzPeak.NET that was previously asynchronous, and did what I could to minimize code duplication. Those APIs only ever call one async function, the Arrow ParquetReader's `ReadNextRecordBatchAsync` method, which they call `.Result` on to force the calling thread to block. I also made it considerably faster by being less pessimistic about incorrectly indexed files, but I can implement a single pass validation rather than repeatedly doing $\Theta(n)$ per spectrum where $n$ is the read record batch size.

PR #4347, opened 2026-07-01
Fixed reconciliation to pair library-supplied decoys by base_id

maccoss:
## Summary

Reconciliation's inter-replicate peak improvement was silently excluding
**library-supplied decoys** (Carafe / FDRBench-manifest, i.e. `--decoys-in-library`
runs), which biased second-pass FDR optimistic. This ports the planner half of the
Rust fix (maccoss/osprey `0abe0ff`) to C#.

**The issue.** Reconciliation lines up each peptide's peak across replicate runs and
improves its boundaries — and it must do this for a target *and its paired decoy*
symmetrically, because decoys drive the second-pass FDR estimate. To find each
decoy's target, `ReconciliationPlanner` matched on the modified sequence: it assumed
every decoy was named `DECOY_<peptide>` and stripped the `DECOY_` prefix. That works
for Osprey-generated (reverse) decoys, but **library-supplied decoys have no `DECOY_`
prefix** (their modseq is either reversed or cycled until obtaining a unique sequence not 
present in the targets), so they never matched a target and were skipped by the planner.

**Impact.** Targets got their peaks reconciled but their library decoys did not — an
asymmetric target/decoy pool that biases the second-pass SVM and makes reported FDR
optimistic (you think you're at 1% when the true FDP is higher). Only affects
`--decoys-in-library` runs; Osprey-generated reverse-decoy runs (e.g. Stellar) were
unaffected.

**The fix.** Pair decoys to targets by the **`base_id` (`EntryId & 0x7FFFFFFF`)** that
Osprey already assigns to every target/decoy pair at library load, instead of by
prefix-stripping the modseq. `ReconciliationPlanner` now keys its passing-precursor
set on `(base_id, charge)` and looks entries up by `base_id` — the same linkage
`ConsensusRts.Compute` already used (the consensus-RT half of `0abe0ff` had been
ported; the planner half had not). Reverse-decoy output is byte-identical.

**Validation (FDRBench).** With the fix, entrapment-based true-FDP tracks at/below the
unity line through the 1% operating point (combined 0.98%, paired 0.88% at q=0.01):

stellar-FDRBench-post-fix

*FDRBench estimated FDP vs FDR threshold, Stellar HeLa entrapment data (27,479
discoveries): both the combined and paired-method curves sit on/below the unity line
through 1% — FDR is properly controlled.*

## Test plan

- [x] `TestPlanLibraryDecoyReconciledByBaseId` — no-prefix library decoy of a passing target is reconciled (fails under the old prefix-strip)
- [x] `TestComputeLibraryDecoyPairedByBaseId` — same at the `ConsensusRts` pairing site
- [x] `TestPlanLibraryDecoyNotReconciledWhenTargetFails` — decoy of a *failing* target is not reconciled (locks the passing-set semantics)
- [x] `TestReconciliationPairsDecoysByBaseIdNotPrefix` — CodeInspection guard forbidding a `"DECOY_"` literal in reconciliation pairing code, so prefix pairing can't return
- [x] Osprey.Test: 443 passed, 0 failed (net472 + net8.0)
- [x] Stellar `regression.ps1`: golden + resume + HPC chain all PASS at 1e-9 (reverse-decoy output unchanged)

Cross-reference: maccoss/osprey `0abe0ff` (fixed both `compute_consensus_rts` and `plan_reconciliation` on the Rust side).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/pwiz_tools/Osprey/Osprey.FDR/Reconciliation/ReconciliationPlanner.cs
+++ b/pwiz_tools/Osprey/Osprey.FDR/Reconciliation/ReconciliationPlanner.cs
@@ -36,8 +36,6 @@ namespace pwiz.Osprey.FDR.Reconciliation
     /// </summary>
     public static class ReconciliationPlanner
     {
-        private const string DECOY_PREFIX = @"DECOY_";
-
         // Minimum RT tolerance floor (minutes) accommodates scan-resolution rounding.
         private const double MIN_RT_TOLERANCE = 0.1;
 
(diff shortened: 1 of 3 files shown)


maccoss:
Self-review note (gap-fill path checked): the same target/decoy pairing bug does **not** exist in gap-fill. `GapFillTargetIdentifier` resolves the paired decoy through the authoritative library manifest linkage `libLookup: (mod_seq, charge) -> (targetEntryId, decoyEntryId)`, reading `DecoyEntryId` directly rather than stripping a `DECOY_` prefix. That mirrors Rust `identify_gap_fill_targets` (which uses `lib_lookup`), which is why `0abe0ff` correctly left it untouched. So the two pairing sites affected by the prefix bug -- `plan_reconciliation` (this fix) and `compute_consensus_rts` (already on base_id via `ConsensusRts`) -- are the complete set; no third instance remains.

PR #4353, opened 2026-07-02
Carried the join-wide first-pass base_id set in reconciliation.json (v3); removed the pass-2 protein rescue

brendanx67:
## Summary

* Fixed a latent HPC per-file compaction divergence (mode3: HPC chain != straight-through). The per-file rescore worker recomputed a PER-FILE first-pass base_id set instead of the straight-through GLOBAL set; the old `|| protein-q` rescue clause had been masking the gap. FirstJoin now persists the join-wide first-pass base_id set in `reconciliation.json` (bumped to v3, required `first_pass_base_ids`); `RescoreCompaction` compacts to exactly that set and hard-fails if absent.
* Removed the pass-2 protein-FDR rescue (admitting a peptide that failed peptide/precursor FDR because its parent protein passed) from all three survival gates. Matches Mike's intent; `--protein-fdr` becomes reporting-only. Protein-FDR reporting (`proteins.csv` + q-values) is kept.

Scope note: the rescue removal is ~a no-op on final output (golden delta net -3). The anti-conservative `--protein-fdr` behavior is caused by the 2nd-pass Percolator RECALIBRATION on a decoy-depleted null, NOT the rescue; that fix is tracked separately (TODO-osprey_pass2_recalibration_fix). This PR is HPC hardening + cleanup, not the calibration fix.

Cross-impl parity: pairs with maccoss/osprey#48 (`reconciliation.json` v3, byte-identical).

## Test plan

- [x] Build-Osprey -RunTests -RunInspection: 446 tests, 0 warnings
- [x] regression.ps1 Stellar mode3 (HPC chain == straight): PASS on the reconciliation-v3 fix
- [ ] regression.ps1 Stellar + Astral: mode1 golden re-captured after auditing the output diff; mode2/mode3 green
- [ ] /pw-self-review clean; un-draft; land with maccoss/osprey#48

### DIFF
--- a/pwiz_tools/Osprey/Osprey.FDR/Reconciliation/ConsensusRts.cs
+++ b/pwiz_tools/Osprey/Osprey.FDR/Reconciliation/ConsensusRts.cs
@@ -53,13 +53,6 @@ namespace pwiz.Osprey.FDR.Reconciliation
         /// <param name="consensusFdr">
         /// FDR threshold for selecting consensus peptides (typically 0.01).
         /// </param>
-        /// <param name="proteinFdrThreshold">
-        /// If &gt; 0, rescue borderline peptides whose first-pass protein
-        /// q-value is &lt;= this threshold. Lets peptides from strong proteins
-        /// contribute to consensus RT computation even if their own peptide
-        /// q-value is borderline. Typically set to <c>config.ProteinFdr</c>.
-        /// Pass 0.0 to disable.
-        /// </param>
         /// <param name="invPredictTrace">
         /// If non-null, populated with one <see cref="InvPredictRecord"/> per
         /// detection contributing to a consensus computation, capturing the
@@ -71,7 +64,6 @@ namespace pwiz.Osprey.FDR.Reconciliation
             IReadOnlyList<KeyValuePair<string, IReadOnlyList<FdrEntry>>> perFileEntries,
             IReadOnlyDictionary<string, RTCalibration> perFileCalibrations,
             double consensusFdr,
-            double proteinFdrThreshold,
             IList<InvPredictRecord> invPredictTrace = null)
         {
             if (perFileEntries == null)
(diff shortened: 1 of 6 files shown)


brendanx67:
## ⚠️ Not ready to merge — mode3 (HPC==straight) regression found

Autonomous night-session validation ran `regression.ps1 -Dataset Stellar`:

- mode2 (resume==straight): **PASS**
- mode1 (vs golden): FAIL — expected-shaped (golden captured *with* the rescue; the 2nd-pass Percolator retrains on the smaller set). Would just need a golden re-capture — but only after mode3 is fixed.
- **mode3 (HPC-chain == straight-through self-consistency): FAIL** — clean master passes all three modes, so this is a regression this change introduces.

### Root cause (confirmed via `-KeepOutput` compaction logs)
Straight-through compacts to a **global** first-pass base_id set (a base_id passing peptide-q in *any* file is kept in *all* files). The HPC `PerFileRescore` worker sees only one file, so `RescoreCompaction.Apply` recomputes a **per-file** set. On master the `|| protein-q ≤ ProteinFdr` clause was **load-bearing** for HPC cross-file coverage — it inflated the per-file worker set to ≈ the global set (66749 ≈ 66857). Removing it drops the workers to per-file peptide-q only (62991 vs the global 66793), so the workers now drop ~3800 cross-file entries that straight keeps; the chaotic 2nd-pass Percolator amplifies that into a wholesale blib divergence.

### Fix direction (architectural — needs review, not a quick patch)
The worker must compact to the **global** first-pass base_id set FirstJoin computed (plumbed through the reconciliation bundle), not a per-file recomputation. This is a latent HPC production-parity issue the rescue removal merely *exposed*; it deserves its own scoping. Reverting the `RescoreCompaction` edit is NOT a fix (it reintroduces the anti-conservative rescue in the HPC output and still diverges from straight).

Full write-up: `ai/.tmp/handoff-20260702-morning.md`. The unit changes + rescue lock-out tests are solid; the blocker is the HPC compaction plumbing.

brendanx67:
Superseded by #4358, which carries only the reconciliation.json v3 change (the join-wide first-pass base_id set) and is a clean C#/Rust bit-parity pair with maccoss/osprey#48 (verified end-to-end at 1e-9 on Stellar).

The other two pieces from this branch are split out:
- Pass-2 protein-FDR rescue removal: the entrapment oracle showed it is ~a no-op on final output (net -3 discoveries); the real anti-conservative cause of `--protein-fdr` is the 2nd-pass Percolator recalibration on a decoy-depleted null, not the rescue. Both tracked in backlog TODO-osprey_pass2_recalibration_fix.
- Keeping the rescue in C# preserves parity with Rust (which still carries it); removing it is a deliberate divergence to do as a dual C#+Rust change with Mike's sign-off.

Closing in favor of #4358.

PR #4400, opened 2026-07-09
Streamed the first-pass FDR projection from parquet, dropping the 53 GB stub buffer

### LINKED ISSUE #4397: Osprey: Stage-5 rematerializes 191M fat FdrEntry stubs (53 GB) only to convert them to 32 B projections
## Problem

On the 82-file Astral fit test, the peak live memory is the **Stage-5 `FdrEntry` stub buffer: 53.13 GB**.

`PerFileScoringTask.cs:334-346` — after per-file scoring releases its transients — loops every scored file and calls `ParquetScoreCache.LoadFdrStubsFromParquet`, **rematerializing all 191,227,519 `FdrEntry` objects** (~280 B each: 16 B object header + 15 doubles + 7 heap pointers, with the six heavy arrays left null), publishing them as `ScoredEntries`.

`FirstJoinTask.cs:1361` then immediately calls `FdrProjectionSet.BuildFromEntries(perFileEntries, releaseStubs: true)`, converting each into a **32 B `FdrProjection`** and dropping the stubs per file.

**We allocate ~53 GB of objects purely to produce ~6 GB of structs.**

Measured anchors (`OSPREY_LOG_MEMORY=1`):
```
[MEM Stage 5 start: 82 files loaded (stubs)]  managed_heap=53.13 GB, peak_paged=63.01 GB
[MEM projection built: 191227519 rows ...]    managed_heap=59.06 GB
```
Note the 59.06 GB is **not** live coexistence — `releaseStubs: true` already frees stubs incrementally; the probe reads total managed heap, so that figure is live projection (~6 GB) plus not-yet-collected stub garbage. The real live peak is the 53.13 GB stub buffer.

## Consumers

On the straight-through path `ScoredEntries` has exactly **one** consumer: `FirstJoinTask`. `PerFileRescoreTask.cs:411` reads it only on the HPC merge path (after an early return). `needsResidentFirstPassPool` (`FirstJoinTask.cs:258`) still needs fat stubs for `--model-diagnostics` / FDRBench-pass-1.

## Fix

Load the lean projection **directly from parquet**; never allocate `FdrEntry`.

Byte-parity crux: `BuildFromEntries` assigns `PeptideId` as the **global Ordinal rank** of the modified sequence across all files. A streaming loader must assign insertion-order ids and remap them to the ordinal rank at the end (remap table ~9 MB at 2.3 M peptides).

**Expected: 53.13 GB → ~6 GB** (191 M × 32 B).

## Progress

Foundation on `Skyline/work/20260708_osprey_lean_fdr_stub` (off #4378), commit `9b2073426`:
- `FdrProjectionSet.Builder` — streaming rows, insertion-order ids, ordinal remap in `Build()`
- `ParquetScoreCache.ReadFdrStubScalars` — 5 columns, no `FdrEntry`; callback-based because Osprey.IO must not depend on Osprey.FDR
- `TestFdrProjectionBuilderMatchesBuildFromEntries` pins element-for-element parity; build + 474 tests green

Remaining: wire `PerFileScoringTask` + `FirstJoinTask`; keep fat-stub loads lazy for model-diagnostics/FDRBench and the merge path; then `regression.ps1 -Dataset Stellar`.

Requested by Mike. Tracked in `ai/todos/active/TODO-20260708_osprey_lean_fdrentry_memory.md`.

maccoss:
## Summary

* `PerFileScoringTask` rematerialized **191,227,519 fat `FdrEntry` stubs (~53.13 GB** on an 82-file Astral run) from the per-file `.scores.parquet`, only for `FirstJoinTask` to immediately convert each ~280 B object into a 32 B `FdrProjection` and drop it.
* It now streams `FdrProjection` rows straight out of parquet -- **no `FdrEntry` is ever allocated** on that path -- and publishes them; `FirstJoinTask` consumes the prebuilt set instead of calling `BuildFromEntries`.
* Fat stubs are still produced on the rehydrate/merge paths and for `--model-diagnostics` / FDRBench pass 1. The lean gate mirrors `FirstJoinTask`'s dispatch exactly (`UseFdrProjection && FdrMethod == Percolator && !needsResidentFirstPassPool`), so a non-Percolator method or `OSPREY_FDR_PROJECTION=0` still gets a populated stub buffer.
* Reads five parquet columns instead of eleven (`FdrProjection` carries no scan/RT/bounds).

**Byte-parity mechanism:** `BuildFromEntries` assigns `PeptideId` as the global Ordinal rank of the modified sequence across *all* files, which a streaming loader cannot know up front. `FdrProjectionSet.Builder` assigns insertion-order ids while rows arrive and remaps them to the Ordinal rank in `Build()` (remap table ~9 MB at 2.3M peptides). Row order, file order and `ParquetIndex` are unchanged, so the result is element-for-element identical.

**Layering:** `Osprey.IO` must not depend on `Osprey.FDR`, so `ParquetScoreCache.ReadFdrStubScalars` streams scalars through a callback and `Osprey.Tasks` assembles the rows -- the same pattern as the existing `loadFileFeatures` delegate.

Expected: `[MEM Stage 5 start]` drops from **53.13 GB to ~6 GB** (191M x 32 B). Not yet measured at scale -- see test plan.

Requested by Mike.

Fixes #4397

## Test plan

- [x] `TestFdrProjectionBuilderMatchesBuildFromEntries` - pins `Builder` element-for-element against `BuildFromEntries`
- [x] Build + 474 unit tests green
- [x] `regression.ps1 -Dataset Stellar` - **mode1/2/3 PASS, byte-identical**. mode1 exercises the new lean path; mode2 (resume) and mode3 (HPC chain) exercise the fat-stub rehydrate fallback
- [x] `Test-PerfGate.ps1 -Dataset Stellar` - **PASSED**, total +0.1% median. `stage5` +2.3% median (below the 5% WARN) is consistent with the ordinal remap pass; Stellar is only ~7M rows
- [ ] 82-file Astral run with `OSPREY_LOG_MEMORY=1` to confirm 53 -> ~6 GB, and to check the remap cost at 191M rows
- [x] `regression.ps1 -Dataset All` before merge

## Notes

Based on #4378 (needs `FdrProjection`, which lives there), so this is stacked -- retarget to master when #4378 merges.

### DIFF
--- a/pwiz_tools/Osprey/Osprey.FDR/FdrProjection.cs
+++ b/pwiz_tools/Osprey/Osprey.FDR/FdrProjection.cs
@@ -297,5 +297,105 @@ namespace pwiz.Osprey.FDR
 
             return new FdrProjectionSet(perFile, peptideById);
         }
+
+        /// <summary>
+        /// Streaming builder that produces a byte-identical <see cref="FdrProjectionSet"/>
+        /// without ever materializing the fat <see cref="FdrEntry"/> stub buffer. The
+        /// 191M-row stub rematerialize (PerFileScoringTask) cost ~53 GB on an 82-file
+        /// Astral run purely to be converted here into 32 B rows; feeding parquet scalars
+        /// straight in drops that to ~6 GB.
+        ///
+        /// Parity invariant: <see cref="BuildFromEntries"/> assigns
+        /// <see cref="FdrProjection.PeptideId"/> as the Ordinal rank of the modified
+        /// sequence among the distinct set across ALL files. A streaming pass cannot know
+        /// that rank up front, so ids are assigned in INSERTION order while rows arrive and
+        /// remapped to the ordinal rank in <see cref="Build"/>. Row order, file order and
+        /// <see cref="FdrProjection.ParquetIndex"/> are unchanged, so the result is
+        /// element-for-element identical to <see cref="BuildFromEntries"/>.
+        /// </summary>
+        public sealed class Builder
+        {
+            private readonly Dictionary<string, int> _insertionIdByPeptide =
+                new Dictionary<string, int>(StringComparer.Ordinal);
+            private readonly List<string> _distinctByInsertion = new List<string>();
+            private readonly List<KeyValuePair<string, List<FdrProjection>>> _perFile =
+                new List<KeyValuePair<string, List<FdrProjection>>>();
+            private List<FdrProjection> _rows;
+            private ushort _fileIdx;
+
+            /// <summary>
+            /// Open a file's row list. Files must be added in the same order the join
+            /// expects (the scoring order), which is what fixes <see cref="FdrProjection.FileIdx"/>.
+            /// </summary>
+            public void BeginFile(string fileName, int capacityHint = 0)
+            {
+                _rows = capacityHint > 0
+                    ? new List<FdrProjection>(capacityHint)
+                    : new List<FdrProjection>();
+                _perFile.Add(new KeyValuePair<string, List<FdrProjection>>(fileName, _rows));
+            }
+
+            /// <summary>
+            /// Append one parquet row of the open file. <see cref="FdrProjection.ParquetIndex"/>
+            /// is the running per-file row ordinal, matching
+            /// <c>LoadFdrStubsFromParquet</c>'s <c>ParquetIndex = stubs.Count</c>.
+            /// <see cref="FdrProjection.Score"/> stays 0 -- the FDR pass writes it back.
+            /// </summary>
+            public void AddRow(uint entryId, byte charge, bool isDecoy, double coelutionSum,
+                string modifiedSequence)
+            {
+                string modseq = modifiedSequence ?? string.Empty;
+                if (!_insertionIdByPeptide.TryGetValue(modseq, out int insertionId))
+                {
+                    insertionId = _distinctByInsertion.Count;
+                    _insertionIdByPeptide[modseq] = insertionId;
+                    _distinctByInsertion.Add(modseq);
+                }
+                _rows.Add(new FdrProjection(
+                    entryId, (uint)_rows.Count, insertionId, _fileIdx, charge, isDecoy,
+                    coelutionSum, 0.0));
+            }
+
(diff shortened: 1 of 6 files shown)


maccoss:
## Self-review (`/pw-self-review`) — findings addressed in `52052d989`

**Medium — `FdrProjections` published but not declared in `Publishes`.** Real defect, fixed. `PipelineContext` builds `_producerByByproduct` solely from `Publishes`, so the undeclared byproduct had no registered producer: `ctx.Get<FdrProjections>()` could not lazily materialize it and would throw `UnknownByproductException` on a cache miss. It resolved only because `FirstJoinTask` reads the declared `ScoredEntries` first, materializing `PerFileScoringTask` and co-publishing both in the same `FinalizeAndCheck`. That is incidental read-ordering coupling, not a contract — reorder those reads, or add any task that reads `FdrProjections` standalone, and it throws. Neither byte-parity nor the perf gate would have caught it.

Answering the reviewer's question directly: the `Publishes` registration was **missed, not intentional**. Added `typeof(FdrProjections)`. It is not a `PerFileEntries` subtype, so it does not participate in the ScoredEntries → CompactedEntries → RescoredEntries milestone chain and does not disturb that ordering assertion.

**Low — `Builder` single-threaded precondition undocumented.** Documented. `FileIdx` and the per-file `ParquetIndex` running counts are only meaningful under a deterministic sequence; `PerFileScoringTask` drives it after its scoring fan-out has joined.

**Low — `ReadFdrStubScalars` row-group-skip parity untested.** Acknowledged, not yet fixed. The two readers are copy-identical (`ParquetScoreCache.cs:746-747` vs `:808-809`), but a multi-row-group parquet with a degenerate (null `entry_id`/`is_decoy`) group is untested, and a divergence there would silently shift `ParquetIndex` and misalign every downstream feature lookup. Tracking as a follow-up test; Stellar's goldens would not necessarily expose it.

Reviewer confirmed clean: gate equivalence (algebraically identical to `FirstJoinTask`'s dispatch across all four flag combinations), Builder byte-parity, publish ordering across all three producer paths, the empty-`ScoredEntries` blast radius, and cross-impl neutrality.

Build + 474 unit tests green after the fix.

PR #4406, opened 2026-07-10
Released the first-pass FdrProjections once Stage 5 consumes them

### LINKED ISSUE #4405: Osprey: first-pass FdrProjections stay pinned in PipelineContext through Stages 6-7 (~6 GB)
## Summary

`PerFileScoringTask` publishes the first-pass `FdrProjections` byproduct (`PerFileScoringTask.cs:618`). `FirstJoinTask` consumes it once, during first-pass FDR (`FirstJoinTask.cs:266`). Nothing ever drops it, because **`PipelineContext` has no release/evict mechanism** -- `_byproducts` is only ever added to and read (`PipelineContext.cs:79`, `:312`, `:324`).

Result: the full projection set -- **191,227,519 rows x 32 B `FdrProjection` = ~5.7 GiB**, plus the 2,322,560-entry interned peptide table -- stays reachable through reconciliation (Stage 6), second-pass FDR, protein FDR, and the blib write, long after first-pass FDR is finished with it.

## Evidence

82-file Astral run, `OSPREY_LOG_MEMORY=1`. The `reconciliation-floor` probe forces a full blocking GC before reporting, so this is **live data, not slack or garbage**:

| | before #4400 | after #4400 |
|---|---|---|
| `[MEM reconciliation-floor]` (post-GC) | **8.40 GB** | **18.30 GB** |

This also explains why the run's *peak* working set did not fall despite `[MEM Stage 5 start]` dropping 53.13 -> 22.74 GB: the 53 GB stub buffer was freed, and then ~6 GB of projections were pinned for the remainder of the run.

## Root cause

Introduced by #4400 (merged as `c9526f0e6`). The pre-#4400 code built the projections as a local inside `RunFirstPassProjection` via `BuildFromEntries(releaseStubs: true)`, so they became collectible when Stage 5 ended. The lean path publishes them into the byproduct cache instead, and the cache is process-lifetime.

Not a defect in the lean loader itself -- the streaming build is correct and is a large win (Stage-5 heap -30 GB, `stage5` -23.9%). The defect is lifetime.

## Fix

`Get<TInfo>()` already lazily re-materializes a missing byproduct through its registered producer's `Rehydrate`, so eviction is safe: a later consumer would rebuild rather than observe a null. And `FdrProjections` has exactly one consumer.

Proposed:
* Add `PipelineContext.Release<TInfo>()` that removes the entry from `_byproducts`, documented as "drop a byproduct once its consumer is done; `Get<TInfo>()` will rematerialize via the producer if anyone asks again."
* Call it in `FirstJoinTask` immediately after `RunFirstPassProjection` returns.
* Regression test: after `FirstJoinTask` runs the lean path, `ctx.TryGet<FdrProjections>(out _)` is `false`.

## Impact

~6 GB of live memory returned across Stages 6-7 on an 82-file run. Scales linearly with total scored entries across all files, so it grows with file count -- the exact property #4400 was written to remove.

Blocks the 32 GB target (see #4404): the ordering there is (1) this fix, (2) reduce the first-pass Percolator live peak, (3) GC tuning.

Reported by Mike.

maccoss:
## Summary

* `PipelineContext`'s byproduct cache is process-lifetime -- `Publish` only ever adds, and nothing ever removes. `PerFileScoringTask` publishes the first-pass `FdrProjections`; `FirstJoinTask` is its only consumer and uses it once, during first-pass FDR. The set then stayed reachable through reconciliation, second-pass FDR, protein FDR and the blib write.
* At 82 Astral files that pinned **~5.7 GiB** (191,227,519 rows x 32 B `FdrProjection`) plus the 2,322,560-entry interned peptide table -- memory that scales with *total scored entries across all files*, which is precisely what streaming the projection out of parquet (#4400) was written to stop paying for.
* Added `PipelineContext.Release<TInfo>()`, called from `FirstJoinTask` as soon as `RunFirstPassProjection` returns (before the `StopAfterStage5` early-return, so that path drops it too).
* Drive-by: fixed two pre-existing `RedundantNameQualifier` warnings in `PerFileScoringTask` (they arrived with #4381) so the zero-warning inspection gate is clean on this branch. Nine `SystemMemory.cs` warnings remain and are also pre-existing on master -- untouched here.

## Evidence

`[MEM reconciliation-floor]` forces a full blocking GC before reporting, so it is **live data**, not slack or uncollected garbage:

| | before #4400 | after #4400 |
|---|---|---|
| `[MEM reconciliation-floor]` (post-GC, 82 files) | **8.40 GB** | **18.30 GB** |

This also explains why the 82-file run's *peak* working set never fell despite `[MEM Stage 5 start]` dropping 53.13 -> 22.74 GB: the stub buffer was freed and then ~6 GB of projections were pinned for the rest of the run.

## Release semantics -- deliberately final

`Release` evicts, it does not arm a rebuild. A producer materializes at most once per context (`_materialized` in `DemandByType`), so a later `Get<TInfo>()` finds no byproduct, skips the already-materialized producer, and throws `UnknownByproductException`.

The tempting alternative -- also evict from `_materialized` so `Get` re-demands -- is **wrong**: a producer's `Rehydrate` publishes *all* of its byproducts, and `Publish` uses `_byproducts.Add`, which throws on the siblings still cached. My first draft documented `Release` as "costs recomputation, never correctness"; `TestGetRematerializesAfterRelease` failed and proved that false. The contract, and the test, now say what the code does: release only when every consumer has run; the failure mode is a loud exception, never a null.

Reported by Mike.

Fixes #4405

## Test plan

- [x] `TestReleaseDropsByproduct` - Release removes the entry, returns true, and is a no-op on a second call
- [x] `TestGetAfterReleaseThrowsRatherThanRebuilding` - pins the finality contract and that the producer is not re-demanded
- [x] Build + 487 unit tests green (Debug)
- [x] ReSharper inspection clean for files touched by this PR
- [ ] `regression.ps1 -Dataset All` - deferred: an 82-file memory run currently owns the Release binaries on this box. TeamCity's Osprey Perf/Regression config covers Stellar + Astral mode1/2/3.
- [ ] 82-file Astral rerun with `OSPREY_LOG_MEMORY=1` to confirm `[MEM reconciliation-floor]` returns to ~8.4 GB


maccoss:
Self-review (fresh-context agent) run on `46f75adcb`. It confirmed the two things that would have invalidated this PR:

* **Sole consumer.** `FdrProjections` is read only at `FirstJoinTask.cs:267`; the `else` branch (`OSPREY_FDR_PROJECTION=0`, `--fdr-method simple`, `--model-diagnostics`, FDRBench pass 1) never reads it, and `Rehydrate` (worker/resume) never touches it.
* **The release actually frees memory.** `survivors` come back from `ReloadFirstPassSurvivors` -> `LoadFdrStubsFromParquet` as a fresh `List<FdrEntry>` and do **not** alias `projections.PerFile` or `PeptideById`. Nothing else roots the `FdrProjectionSet`.

It also noted Rust scopes the equivalent buffers by RAII (`drop(per_file_entries)`, `pipeline.rs:5507`), so this moves C# toward Rust's memory behavior rather than away.

**Medium finding addressed in `863ad0b3d`.** Nothing asserted that `FirstJoinTask` *calls* `Release`, so a refactor could drop the call, re-pin 5.7 GiB, and leave every test green. Rather than add an integration assertion for a call site, I made the retention structurally hard to reintroduce: added `PipelineContext.Consume<TInfo>()` (Get + Release) and used it at the call site. The byproduct is now dropped *before* Stage-5 FDR runs, and re-pinning would take a deliberate change back to `Get`. Pinned by `TestConsumeReturnsValueAndDropsIt`.

**Low finding addressed.** `Release` now documents that it is not for the `PerFileEntries` milestone family, since the DEBUG republish guard (`_consumedByproducts` / `AssertMilestoneConsumedBeforeRepublish`) is not updated on release. Inert today -- `FdrProjections` is not a milestone type -- but a latent trap.

**Answering the reviewer's follow-up** ("should `Release` reject milestone types?"): documented rather than enforced, for now. Enforcing would mean teaching `PipelineContext` which types are milestones at the `Release` call, which duplicates knowledge that currently lives in `_milestoneByBuffer`. If a second caller ever wants `Release` on a shared-buffer milestone, that is the moment to make it a hard guard.

488 unit tests green.

PR #4411, opened 2026-07-11
Fixed intermittent floating GraphFullScan teardown hang

bspratt:
## Summary

* The Full Scan graph is always shown as a floating window; DigitalRune's floating-pane teardown intermittently wedged the UI thread in a native `SetWindowRgn` call when it ran under a held `DockPanelLayoutLock` (SuspendLayout + CoverControl), hanging the whole run. Seen on the 2026-07-07 trunk nightly (`TestRunner` exit -1); also reachable by users who remove all results (or load a no-results document / saved layout) with a floating Full Scan graph open.
* Destroy the graph *before* the layout lock in `UpdateGraphUI` and `LoadLayout`, instead of under it. Empirically, closing it with the layout unlocked is safe; only the held lock triggers the native wedge.
* Added a deterministic regression guard to `MixedPolarityFullScanTest`: with the graph open, remove results and assert `SkylineWindow.GraphFullScan == null`.

## Test plan

- [x] TestMixedPolarityFullScan - 2500 iterations x 5 languages, no hang (unfixed reproduced a hang ~1/700, at iteration 709)
- [x] FullScanGraphTest - 5 languages
- [x] CodeInspection test + full-solution ReSharper inspection - clean
- [x] Fresh-context self-review - no blocking findings

### DIFF
--- a/pwiz_tools/Skyline/SkylineGraphs.cs
+++ b/pwiz_tools/Skyline/SkylineGraphs.cs
@@ -301,11 +301,24 @@ namespace pwiz.Skyline
                     listUpdateGraphs.Add(_graphSpectrum);
             }
 
+            // The Full Scan graph is always shown as a floating window, and DigitalRune's
+            // floating-pane teardown intermittently wedges the UI thread in a native SetWindowRgn
+            // call when it runs while the dock panel layout is locked (SuspendLayout + CoverControl).
+            // When the graph is about to be torn down anyway - because results were removed, or
+            // because a saved layout is about to replace it - close it here, before locking the
+            // layout below, where closing a floating window is safe.
+            string layoutFile = GetViewFile(DocumentFilePath);
+            bool willLoadLayout = docIdChanged && File.Exists(layoutFile);
+            if (_graphFullScan != null &&
+                (willLoadLayout || (_graphFullScan.Visible && !settingsNew.HasResults)))
+            {
+                DestroyGraphFullScan();
+            }
+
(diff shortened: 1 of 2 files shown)


nickshulman (commented):
It seems this bug should maybe affect all windows that get automatically closed when there are no results, including LiveResultsGrid. I think the hang only happens if a SetWindowRgn happens while the window is in the middle of painting itself, and GraphFullScan probably paints slower than other windows which is why we've only seen this with GraphFullScan. I will do some investigating and see whether I can get this to happen with other windows if I add a "Thread.Sleep" to their paint methods, and see if we need a more general fix.

nickshulman (commented):
This does not look like the right fix. In the Nightly run, the hang is actually coming from the "Invoke" call in GraphFullScan.LoadingTextIfNoChange.

When we were working on the FilesTree earlier this year, we learned that there is no safe way to call "Control.Invoke" or "Control.BeginInvoke" on a control whose handle might have been destroyed. There is always a window of time where a hang is possible because it will try to re-create the handle at a time when that's not possible.

The only safe thing to do is either to call "Invoke" on a Control which is guaranteed to be longer lived such as Program.MainWindow. Or, you do the thing that we do in Receiver.cs where we use a WindowsFormsSynchronizationContext which also lives long enough to never hit this hang.

I'll try to come up with a different fix for this one.

bspratt:
@nickshulman OK, I'll take down the PR

PR #4422, opened 2026-07-15
Filter chromatogram extraction on uninterpreted mzML CV/user parameters

bspratt:
## Summary

* Adds spectrum filtering on the uninterpreted mzML controlled-vocabulary (CV) and user parameters that Skyline reads through the ProteoWizard API but does not interpret into its own typed fields (e.g. base peak intensity MS:1000505, the Thermo filter string MS:1000512, vendor userParams). They become filterable in chromatogram extraction and selectable in the spectrum filter editor.
image

Help button offers this:
image

* Dynamic CV columns are evaluated directly against each spectrum's parameters (bypassing the   SpectrumClass projection); numeric vs. string is inferred from the operator and operand. The   ordered comparisons (`>`, `<`, `>=`, `<=`) are numeric and report a clear error on a present non-numeric value; equals/not-equals compare per value — numeric where the value is a number  (so `1.0e03` equals `1000`), string otherwise — so they never abort on a text value.
* Terms persist in the per-file metadata cache (a self-describing protobuf field, so no cache-format bump) and are captured during extraction only when a document actually filters on them, so the cost is paid only when needed.
* The filter editor offers CV terms from the PSI-MS ontology catalog (available with no import) plus terms discovered from imported data.
* Is Blank / Is Not Blank extended to handle flag terms (e.g. MS:1000497  "zoom scan") as present / not present.
* Text/command-line authoring: a filter string can name a CV term by accession (`MS:1000505`) or a userParam by the `userParam:` marker, and use friendly operator words (`isblank`, `greaterthan`); an invalid filter's error lists the available operators.

Follows #4349 (merged): Phase 1 surfaces the uninterpreted mzML terms in the full-scan viewer;
this PR lets you filter chromatogram extraction on them.

## Test plan

- [x] TestSpectrumClassFilter — CV predicate, numeric/string & per-value equals, round-trip, Is(Not)Blank, operator scoping
- [x] TestCvSpectrumFilter — end-to-end extraction: capture, cache round-trip, editor, catalog
- [x] SpectrumClassFilterParserTest — filter-string parsing: CV accession & userParam references, operator aliases, nonsense inputs, error vocabulary
- [x] ChromatogramGroupIdsTest — proto round-trip for every filter operator (cache persistence)
- [x] TestResultFileMetaDataOtherParams — cache (de)serialization of the term bag, including unit accession and definition
- [x] EditSpectrumFilterTest, SpectrumFilterTransitionListTest, SpectrumFilterImportErrorTest — editor and transition-list authoring
- [x] CodeInspection + full ReSharper inspection — clean on changed code


pwiz_tools/Skyline/Model/Results/Spectra/SpectrumClassColumn.cs
```diff
@@ -249,6 +500,87 @@ public override string GetAbbreviatedColumnName()
...
+                {
+                    return _accession;
+                }
+                // A controlled-vocabulary term shows its friendly name with the accession as the precise
+                // cue, e.g. "base peak intensity (MS:1000505)"; a userParam (name == accession) shows just
+                // its name.
+                return Equals(_name, _accession)
+                    ? _name
+                    : string.Format(CultureInfo.CurrentCulture, @"{0} ({1})", _name, _accession);
+            }
```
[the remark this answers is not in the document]
bspratt: Addressed in 8d9df0e85: GetLocalizedColumnName now formats through the passed cultureInfo (both the resource lookup and string.Format), not CurrentCulture. Resolving as outdated.

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumClassFilter.cs
```diff
@@ -410,12 +604,240 @@ public static SpectrumClassFilter ParseFilterString(string filterString)
...
+        {
+            return char.IsWhiteSpace(c) || @"""'()=<>".IndexOf(c) >= 0;
+        }
+
+        // The canonical encoded column name for a CV/userParam identity (a CV accession encodes as
+        // "cvid...", any other name as the "cvup..." userParam form; see SpectrumClassColumn).
+        private static string EncodeCvColumnToken(string identity)
+        {
+            return SpectrumClassColumn.CvParam(identity, null, false).ColumnName;
+        }
```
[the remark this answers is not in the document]
bspratt: Wont fix - accepted by design. This is a documented edge: a userParam literally named like a PSI-MS accession (userParam:MS:1000505) encodes to the same cvid... token as the real CV term.

The suggested fix (force cvup... for userParam: references) would actually break matching: capture/discovery encode identities the same way (a captured userParam is stored with Accession = param.name and DiscoverCvColumns runs it through the same accession-detecting encoder), so such a userParam is stored/discovered as cvid... too. Forcing the filter string to cvup... would then match nothing.

A correct fix needs a userParam-vs-CV distinction threaded through the whole identity model (capture, discovery, cache, filter), not just the filter encoder - disproportionate for a case that essentially never occurs (a vendor naming a userParam exactly like an accession). Keeping the lenient behavior; resolving.

pwiz_tools/Shared/Common/DataBinding/FilterOperation.cs
```diff
@@ -79,7 +83,9 @@ public static class FilterOperations
             OP_CONTAINS,
             OP_NOT_CONTAINS,
             OP_STARTS_WITH,
-            OP_NOT_STARTS_WITH
+            OP_NOT_STARTS_WITH,
+            OP_IS_DECLARED,
```
nickshulman: We should make "OP_IS_BLANK" and "OP_IS_NOT_BLANK" behave the way we want instead of adding "OP_IS_DECLARED" and "OP_IS_NOT_DECLARED".

We can decide what "is blank" means for these new values, and we should make it so if the value is present and is equal to the empty string, it is specifically not blank, and they can use "OP_EQUALS" with the empty string.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -1583,6 +1583,102 @@ private SpectrumMetadata GetSpectrumMetadata(Spectrum spectrum)
...
+        // can act on (numeric or string terms by value, flag terms by presence), so all are offered;
+        // only the ontology's grouping/category nodes are excluded.
+        private static readonly HashSet<CVID> SPECTRUM_LEVEL_CV_ROOTS = new HashSet<CVID>
+        {
+            CVID.MS_spectrum_property,   // MS:1003058
+            CVID.MS_spectrum_attribute,  // MS:1000499
+            CVID.MS_scan_attribute       // MS:1000503
+        };
+
+        private static readonly object _spectrumCvTermCatalogLock = new object();
```
nickshulman: Is this lock so that we can lazily initialize "_spectrumCvTermCatalog" the first time we need it?
The usual way to do that would be to have that be an ordinary static field in some other class, and the regular C# class loading would initialize it the first time anyone used that other class.

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumClassColumn.cs
```diff
@@ -145,7 +147,256 @@ public override string ToString()
...
+        /// </summary>
+        public static IList<SpectrumClassColumn> DiscoverCvColumns(SrmDocument document)
+        {
+            var measuredResults = document.Settings.MeasuredResults;
+            if (measuredResults == null)
+            {
+                return Array.Empty<SpectrumClassColumn>();
+            }
+
+            var spectra = measuredResults.Chromatograms.SelectMany(chromatogramSet => chromatogramSet.MSDataFilePaths)
```
nickshulman: You should just use "measuredResults.GetResultFileMetadatas()" to get all the ResultFileMetadata objects without having to specify each one.

pwiz_tools/Skyline/ProtocolBuffers/ChromatogramGroupData.proto
```diff
@@ -48,8 +48,18 @@ message ResultFileMetaDataProto {
...
-		reserved 15; // was google.protobuf.DoubleValue sourceOffsetVoltage 
+		reserved 15; // was google.protobuf.DoubleValue sourceOffsetVoltage
 		double sourceOffsetVoltage = 16;
+		repeated int32 otherParamIndex = 17; // one-based indexes into "otherParams"
+	}
+
+	message OtherParam {
+		string accession = 1;
+		string name = 2;
+		string value = 3;
```
nickshulman, line 59: We should probably use `google.protobuf.Value` here so we don't have to convert everything to a string.

pwiz_tools/Shared/Common/DataBinding/FilterOperation.cs
bspratt: Yeah, I thought about this one a lot. To my ear "Is Blank" / "Is Not Blank" expects that the column has a value that might happen to be empty. But many CV terms are pure flags (e.g. zoom scan, profile spectrum) that can never carry a value, and for those asking "is it blank?" doesn't make sense. The only thing you can ask is whether the term is present, and "Is Declared" / "Is Not Declared" does that. But maybe there's a better name for that?

pwiz_tools/Skyline/ProtocolBuffers/ChromatogramGroupData.proto
bspratt: I think this is actually OK as is - keep in mind that these are CVParams and userParams that Skyline doesn't interpret for its usual data structures, so the values are actually received as strings*. No type conversion is attempted until a comparison is performed, and only then depending on what's being compared to.

*You can probably make the case that pwiz converting raw number values to strings for passing through the API is inefficient, but that's how it works. It might be interesting sometime to try a side project where that doesn't happen, to see whether that turns out to have been expensive or not. I wonder if that's part of Matt's C# port.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
bspratt: Right, thanks.

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumClassColumn.cs
bspratt: You're right!

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
bspratt: Done in 7ad581b52: replaced the explicit lock with a nested static holder type, so the CLR handles thread-safe first-use initialization. Kept the read-only result.

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumClassColumn.cs
bspratt: Done in 7ad581b52: now uses MeasuredResults.GetResultFileMetadatas() instead of the manual per-file lookup.

pwiz_tools/Skyline/ProtocolBuffers/ChromatogramGroupData.proto
bspratt: @nickshulman what say you?

pwiz_tools/Shared/Common/DataBinding/FilterOperation.cs
bspratt: @nickshulman what say you?

nickshulman: It would be bad to say "Zoom scan is always blank because it does not have a value".

It is much better to declare the semantics "CV params are always blank when they are absent, and never blank when they are present."

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumClassColumn.cs
```diff
@@ -145,7 +147,277 @@ public override string ToString()
...
+        /// <summary>
+        /// The CV/user-parameter columns to offer in the filter editor for <paramref name="document"/>:
+        /// the ontology catalog (always available) plus any discovered from imported data (vendor
+        /// userParams, and CV terms typed from their observed values, which take precedence over the
+        /// typeless catalog entry with the same accession).
+        /// </summary>
+        public static IList<SpectrumClassColumn> GetEditorCvColumns(SrmDocument document)
+        {
+            var byColumnName = new Dictionary<string, SpectrumClassColumn>();
+            foreach (var column in GetCvColumnCatalog())
```
nickshulman, line 320: I wonder if we should make it so that if the document has at least one Replicate, we only populate the dropdown with the CV params that we actually found in the document. If they are defining a spectrum filter in a blank document, they get the complete list, but, if they have at least one replicate it might be better for us to guide them towards the CV params that actually exist.

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumClassColumn.cs
```diff
@@ -145,7 +147,277 @@ public override string ToString()
...
+
+        /// <summary>
+        /// The catalog of uninterpreted mzML CV terms that can appear on a spectrum, as dynamic columns,
+        /// so the filter editor can offer them before any data is imported. Typeless: the editor offers
+        /// every operator and the predicate infers numeric vs. string from the operator and operand.
+        /// </summary>
+        public static IList<SpectrumClassColumn> GetCvColumnCatalog()
+        {
+            return MsDataFileImpl.GetSpectrumCvTermCatalog()
+                .Select(term => (SpectrumClassColumn)new CvParamColumn(term.Accession, term.Name, false))
```
nickshulman: I think we really do know whether these CV params are numeric so we do not need to pass "false" for "isNumeric". I am not sure where that information is in the source code.

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumClassColumn.cs
bspratt: I did think about this - it's worth noting that the FullScan graph now provides a way to know what CVParams are actually in use. I think we can easily get into trouble NOT showing things based on document context (thinking of ongoing issues with report types), but I could see offering a visual indicator in the list for things that are known to be available. Make the text blue maybe.

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumClassColumn.cs
bspratt: I'm not sure we do carry that info at the moment but I'm sure it could be readily done.

pwiz_tools/Shared/Common/DataBinding/FilterOperation.cs
bspratt: Fair enough. I don't find it intuitive but it's at least easily explained, which is what matters.

nickshulman (commented):
I wonder if we should add "units" and "operand_type" attributes to "FilterSpec" in "ColumnSpec.cs".
Do you think we need to remember whether the value they typed in there was in seconds or minutes in case a different raw file used different units?

bspratt:
> I wonder if we should add "units" and "operand_type" attributes to "FilterSpec" in "ColumnSpec.cs". Do you think we need to remember whether the value they typed in there was in seconds or minutes in case a different raw file used different units?

I suppose so - but then we're drawn into dealing with units conversion, probably. Not sure how large that can of worms might be.

pwiz_tools/Shared/Common/DataBinding/FilterOperation.cs
bspratt: Done in 89548ee77. The Declared operators are gone - registry, resource strings, the .skyd proto enum values and their map entries - and for a CV/userParam column blank now means absent: a term is blank when the spectrum does not carry it, and is never blank when it does. So Is Not Blank is how a flag like zoom scan gets matched, and Equals with an empty operand is how "present but empty" gets asked about specifically. It lives in `SpectrumClassFilter.CompileCvSpec`, where the Declared special case used to be. No custom `IFilterHandler` was needed: `EditSpectrumFilterDlg` lists every operation ungated, so `NumericFilterHandler.CanBeBlank == false` does not keep Is Blank off a numeric CV column.

One gap you should know about, which I left alone. The readable filter-string syntax cannot express Equals "". `FilterClauseSerializer` (from #4115) renders Is Blank as `= ''` and maps `= ''` back to Is Blank, with a comment that a literal equals-empty is "not separately expressible (or needed) in this syntax". That was true when blank meant null-or-empty; under these semantics the two are opposites for a flag term, so a filter authored in the editor as Equals "" and routed through a transition list comes back as Is Blank. Documents and the .skyd cache persist the operator itself, so they are unaffected.

I did not change it because `= ''` for Is Blank shipped in 26.1, and reinterpreting it would change the meaning of filter strings already authored against ordinary columns - and scoping the new reading to CV columns would mean a per-column hook in the shared serializer. The exposure is narrow in practice: for a term the ontology gives no value type (a pure flag, which can only ever be present with an empty value), Equals "" says exactly what Is Not Blank says, so the escape hatch only matters for a term declared to carry a value that some file wrote out empty anyway. A test pins the current lossy behavior so any change to it has to be deliberate. Happy to make it lossless if you would rather have it that way.

Related: now that each catalog term carries its declared value type (see the other thread), the editor could stop offering the value comparisons on terms that cannot carry a value, which is where this corner comes from in the first place.

nickshulman: FilterClauseSerializer is brand new this version.
We should change the parser and serializer so it uses "is null" and "is not null" for "OP_IS_BLANK" and "OP_IS_NOT_BLANK".
We don't use this filter syntax in .sky or .skyd files, so we don't really need to worry about backwards compatibility. There might be some transition lists that use the old syntax, but they'll just have to be updated.

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumClassColumn.cs
bspratt: Also note that we'd have to re-scan the raw data, we do not actually carry a scan's extra CVParam metadata in memory. Not a huge deal, but not free either. We can probably just scan the first 100 spectra and call it good.

nickshulman: Aren't you storing these values in memory? Isn't that what the ChromatogramGroupData.proto change does? The ResultFileMetaData objects all get read into memory when the .skyd file is loaded.

bspratt: Yes, but such metadata are only read and preserved when they are needed due to CVParam-based spectrum filtering at chromatogram extraction time. It's a waste of pwiz access time and skyd space otherwise.

bspratt: @nickshulman  We now indicate (text in color System.HotTrack) the filters will actually have an effect, based on what's in the .skyd and scanning for what's in the data files if they're available:
image

pwiz_tools/Skyline/EditUI/EditSpectrumFilterDlg.cs
```diff
@@ -181,6 +199,45 @@ public void SetProperty(SpectrumClassColumn spectrumClassColumn)
             }
         }
 
+        /// <summary>
+        /// A filterable column the dialog offers, reduced to what the dialog needs (caption, path, type).
+        /// Backed either by a <see cref="ColumnDescriptor"/> (a property of the databound row type) or by
+        /// a dynamic <see cref="SpectrumClassColumn"/> (a discovered mzML CV/user parameter, which has no
+        /// such property).
+        /// </summary>
+        private class FilterColumn
```
nickshulman, line 226: Do you need this class? Could you use "SpectrumClassColumn" instead? Is there anything on FilterColumn which is not available on SpectrumClassColumn?

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumClassColumn.cs
nickshulman: I think you haven't pushed the commits for highlighting the filters.

bspratt:
@nickshulman check out the revised description and images in the PR description at the top of this page. I think you'll be pleased.

pwiz_tools/Skyline/EditUI/EditSpectrumFilterDlg.cs
```diff
@@ -37,12 +38,47 @@ public partial class EditSpectrumFilterDlg : CommonFormEx
...
+        {
+            FilterOperations.OP_IS_BLANK, FilterOperations.OP_IS_NOT_BLANK
+        };
+
+        private IList<SpectrumClassColumn> _extraColumns;
+        private SpectrumColumnScanner.Availability _columnAvailability = SpectrumColumnScanner.Availability.UNKNOWN;
+        private Font _unanswerableFont;
         private bool _updating;
 
         public EditSpectrumFilterDlg(ColumnDescriptor rootColumn, FilterPages filterPages)
```
nickshulman, line 62: If you bring up the spectrum filter dialog from SpectrumFilterDataGridViewColumn it uses this constructor which does not pass in "extraColumns".
Is there a reason we only pass in extraColumns when launched from the "Edit Spectrum Filter" menu item? SpectrumFilterDataGridViewColumn does have the SrmDocument.

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumColumnScanner.cs
```diff
@@ -0,0 +1,357 @@
...
+    /// the document whose author is about to write their first one - has nothing stored, and the files
+    /// themselves have to be read.
+    ///
+    /// Two approximations keep that read affordable. Only the head of a file is examined, on the assumption
+    /// that a run does not start emitting a new term partway through. And only one file per distinct
+    /// <see cref="FileSignature"/> is opened, because which terms appear is a property of what wrote the
+    /// file - the vendor format, the instrument, the conversion that produced it - not of the sample in it.
+    /// A hundred replicates off one instrument therefore cost one file open, not a hundred.
+    /// </summary>
+    public static class SpectrumColumnScanner
```
nickshulman: This would probably be better being a class with a constructor that takes a MeasuredResults. That way, it can hang onto the answers to some questions.
When you have a static class with a lot of static methods that all take the same first parameter, it's a good indication that they should be instance methods on an object whose constructor took that parameter.

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumColumnScanner.cs
```diff
@@ -0,0 +1,357 @@
...
+            return true;
+        }
+
+        /// <summary>
+        /// The accessions in the head of one file, or null if it could not be read. Missing or unreadable
+        /// files are ordinary here - documents get shared without their raw data - so this reports the
+        /// failure as "nothing learned" rather than raising it at the user, who did not ask for a file to
+        /// be opened and cannot act on the news that one is gone.
+        /// </summary>
+        private static HashSet<string> ReadAccessions(MsDataFileUri filePath, string documentFilePath,
```
nickshulman: Why are we reading these values from the MsDataFile, instead of using what we have in the ResultFileMetadata? Is it because the ResultFileMetadata only contains the CV params for the spectra that were actually included in a chromatogram? Or is it because if an older version of Skyline extracted the chromatograms there will be no CV parameters in there?

bspratt: Under normal circumstances there's no reason to process and preserve CVParams (other than the ones we convert such as MSLevel). It's a performance and resource hit so we only do it when spectrum filters are in play.

bspratt:
@nickshulman Thanks for looking at this - as you observed, committed code is stale, I just wanted feedback on the visuals for the moment.

pwiz_tools/Shared/Common/DataBinding/FilterOperation.cs
bspratt: **Posted by Claude Code from Brian's account.** The analysis and wording below are Claude's, not Brian's.

Done - `FilterClauseSerializer` writes `is null` / `is not null`, and the parser reads them. `= ''` no longer maps back to a blank test, so equality with the empty string is expressible and round-trips as itself. Both tests that pinned the old behaviour are inverted rather than deleted.

One deliberate departure from your call that old transition lists "will just have to be updated". On a column that cannot hold an empty value, `= ''` is unsatisfiable - nothing can equal the empty string there - so it can only ever have meant the blank test, and it is read as one. `CollisionEnergy = ''` therefore still imports, where it would otherwise now fail to parse at all rather than merely change meaning. On a text column, where an empty value is real, the comparison is taken at face value and the meaning does change. The decision is made by asking the column's own filter handler whether an empty operand parses, not by inspecting its type.

Flagging it because it is narrower than what you approved, and because the remaining break - text columns only - is the one worth a release note.

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumClassColumn.cs
bspratt: **Posted by Claude Code from Brian's account.** The analysis and wording below are Claude's, not Brian's.

You were right that the metadata is already in memory - the earlier reply about needing a rescan was wrong about the mechanism for the interpreted columns, which the `.skyd` answers with no file access at all.

Restricting the list itself would have been circular, though. Capture is gated on the document already referencing a CV filter (`SpectraChromDataProvider`), so for a document that has never carried one the discovered set is empty - and that is exactly the document whose author needs the guidance. So the terms are all still offered, and the ones this data can answer are marked instead. For CV terms with nothing captured, the head of one file per distinct writer is read to find out.

That marking is three states, not two: the accent means the data has values, italic means it was examined and has none, and plain means nothing was determined. The third state matters - a document with no results, or one whose files are missing, must not be told its terms are absent when they were never looked for.

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumClassColumn.cs
bspratt: **Posted by Claude Code from Brian's account.** The analysis and wording below are Claude's, not Brian's.

We do, and it comes from the ontology's `has_value_type`, so `isNumeric` is no longer guessed for catalog terms. The same declaration answers a second question: a term with **no** declared value type is a pure flag, and 16 of the 59 catalog terms are - `zoom scan`, `demultiplexed spectrum`, `consensus spectrum`, the ion-mobility frame types. Those columns now offer only Is Blank and Is Not Blank, since no comparison against them could ever match.

A vendor userParam still has no ontology entry, so its type is inferred from the values seen and its operators are left unrestricted - guessing wrong there would withhold a filter that works.

pwiz_tools/Skyline/EditUI/EditSpectrumFilterDlg.cs
bspratt: **Posted by Claude Code from Brian's account.** The analysis and wording below are Claude's, not Brian's.

`FilterColumn` holds two things `SpectrumClassColumn` does not.

A `Caption` that can differ from the column's own name. The dropdown and its reverse lookup are keyed on it, and a vendor userParam whose name matches a built-in property's caption is now offered under the `userParam:` marker - previously it was dropped, which meant `ResolveFilterColumn` found nothing, the row never appeared, and confirming the dialog wrote the clause back without the criterion it was opened to edit. I would keep that out of `SpectrumClassColumn`: it disambiguates two entries colliding *in this list*, and those columns are shared statics.

And `PropertyType` from the `ColumnDescriptor`, which is what the operand parsing is given. I checked whether that could just be `ValueType` instead: 12 of the 13 interpreted columns agree, and `SourceOffsetVoltage` does not - the descriptor gives `double?`, `ValueType` gives `double`, because the column reads `SpectrumMetadata.SourceOffsetVoltage` (non-nullable) while `SpectrumClass.SourceOffsetVoltage` is still `double?`. That looks like a leftover from making the field non-nullable in the cache - worth reconciling on its own account, but until it is, merging the two types would quietly change that column's operand typing, and nullability is what decides whether the blank tests apply.

So: happy to do it, after that one is settled.

pwiz_tools/Skyline/EditUI/EditSpectrumFilterDlg.cs
bspratt: **Posted by Claude Code from Brian's account.** The analysis and wording below are Claude's, not Brian's.

No good reason - the menu path was simply built first. Fixed in 95d7da629: the Document Grid column, Transition Settings and the peptide-search flow all pass the CV columns now, each from a document it already had.

Column availability is still left undetermined at those three, so the terms are offered but not styled. Determining it means reading files behind a wait dialog, which seemed wrong to add to a modal settings dialog, and in the peptide-search flow there are usually no results to judge from. Undetermined renders plain and claims nothing, so this is honest rather than merely absent - but say the word if you want the scan there too.

pwiz_tools/Skyline/Model/Results/Spectra/SpectrumColumnScanner.cs
bspratt: **Posted by Claude Code from Brian's account.** The analysis and wording below are Claude's, not Brian's.

Agreed and done in aae5001eb. It is constructed with the `MeasuredResults` and the document path, and the work that does not depend on which columns are being asked about - choosing one file per distinct writer, and sampling their stored metadata - is now computed once per scanner instead of on every call.

The learned accessions stay static, deliberately: they are keyed by file format plus instrument configuration, so a file already read for one document answers for another.

PR #4453, opened 2026-07-24
Fix nightly heap-leak reports for native-dialog tests (mute under Terminal Services)

nickshulman:
## Status: DRAFT — best-guess fix, pending broader nightly confirmation (~8am 2026-07-25)

Nightly flags a native (Win32) **heap** leak on the native-dialog/connector test family
(`TestNativeMessageBox`, `TestNativeFileDialog`, `TestMcpConnectorBackgroundDialog`,
`TestPrmMcpConnector`) on every machine that has the tests. Managed and user/GDI deltas are ~0.

## Root cause (confirmed)

The Windows **common file dialog** grows the process heap on every show **only when the process
runs in a Terminal Services (RDP) session** — including a *disconnected* one, which is exactly
the state nightly agents sit in (logged in over RDP, then disconnected). Established by a
same-machine A/B (console vs RDP, identical binary) and a bare-dialog probe that uses **no
Skyline code**:

* Physical console: no leak; the probe plateaus.
* RDP session: the identical bare `SaveFileDialog`/`OpenFileDialog` loop climbs and does not
  settle. The **legacy comdlg32** path is a genuine unbounded leak (~27 KB/dialog, dead-linear to
  1000+ dialogs, no plateau); the **modern IFileDialog** path is a saturating cache (~4 KB/iter
  decaying to <1 KB by ~iter 500). A bare `MessageBox` never leaks — isolating it to the common
  *file* dialog, not "RDP in general."

Because the offending (legacy) path never plateaus, a warm-up cannot help; and it is an
OS/remoted-display artifact with no Skyline code involved. So the fix is test-infrastructure.

## Best-guess fix (this PR)

* **Revert the diagnostic test split** — back to the original `TestNativeMessageBox` and
  `TestMcpConnectorBackgroundDialog`.
* **Mute the native-dialog family from the HEAP check only**, gated on
  `SystemInformation.TerminalServerSession` so the physical console keeps full heap-leak
  detection. Managed-memory and handle leak checks stay active for these tests everywhere.
* **Keep the `HeapProbe` DevTool** (the diagnostic that isolated the cause). Removed the
  no-Skyline `TestNativeOpenFileDialogLeak` — it used the modern dialog (~5.7 KB/iter, under
  threshold) so it did not reproduce the failure.

The mute list is **provisional** pending tomorrow's broader fleet results — in particular
`TestMcpConnectorBackgroundDialog`, whose dialog is a managed `BackgroundThreadLongWaitDlg`, not
a native file dialog; it may be a separate/noise effect and could be dropped from the list.

## Verification

* Builds; reverted `TestNativeMessageBox` passes on the console.
* The mute's *effect* is only active under Terminal Services, so it is not exercised on a console
  session — to be confirmed on the RDP-session box and by tomorrow's nightly.

Details and the full investigation log: `ai/todos/active/TODO-20260723_native_dialog_leak_iterations.md`.


nickshulman:
Superseded by #4459, which keeps the heap check enabled for these tests instead of muting it.

The measurements behind that PR: the growth is a bounded Windows shell cache rather than a leak (a bare SaveFileDialog loop plateaus near 1.4 MB after ~150 dialogs and hands memory back), and disconnecting the session with tsdiscon reproduces the nightly state, where TestPrmMcpConnector needs pass 27 and TestNativeFileDialog pass 28 -- both beyond the old 24-pass ceiling. Extra iterations get them there; muting is not needed.

Note the HeapProbe DevTool added here is NOT in #4459 and is worth keeping -- happy to bring it over in a separate PR if you want it.

PR #4459, opened 2026-07-25
Gave the three native-file-dialog tests extra leak iterations instead of muting the heap check

nickshulman:
## Summary

Gives `TestNativeFileDialog`, `TestNativeMessageBox` and `TestPrmMcpConnector` four times the
default leak-check iteration ceiling. `MutedHeapMemoryLeakTestNames` stays empty -- the heap
check remains enforced for these tests.

Reopened and rebased on master now that #4458 and #4474 are merged. Changes since it was closed:

* Dropped the `TestMcpConnectorBackgroundDialog` entry. #4458 reworked that test to create one
  window instead of ~100,000, and it now converges in 8 passes on all four agents that have
  reported (UW7 -16.8 KB, DEV6 -33.3 KB, DEV4 -24 KB), so it does not need an override.
* Removed the Terminal Services framing. The session headers added by #4458 falsify it:
  BRENDANX-DT1 reports `TerminalServerSession=False` and still leaks all three tests.
* Named the growth instead of inferring it. Diffing the process heap's block-size histogram
  across a pass-1 run shows the accumulating blocks hold Explorer's cached DirectUI view markup
  (`<duixml>`, `ViewHost`, `ProperTreeModule`, `DetailsContainer`) plus the navigated path's own
  strings, one set per dialog shown -- the shell view of the modern Explorer-hosted `IFileDialog`,
  cached per instance on a UI thread that lives for the whole process.

## Why extra iterations rather than muting

The growth saturates. On a long-lived thread the marginal cost decays to ~2 KB by the 40th dialog.
Legacy comdlg32 is genuinely dead-linear at ~29 KB/dialog, but Skyline never takes that path:
`AutoUpgradeEnabled` is never set false and `ShowHelp`/`ShowReadOnly` are never set at all.

It is the dialog and not the connector. `TestSetItemMcpConnector` drives the same connector with
the same 30 ms polling and the same startup but opens no native dialog; it converges and passes,
and shares none of the block sizes above.

## The objection this PR was closed on

The closing comment gave two reasons. The first -- that the growth is dead linear with no plateau,
so extra iterations only delay the same failure -- rested on the "unbounded" reading of 25 samples
that CORRECTION 3 in the TODO later withdrew as an over-extrapolation.

The second still stands and is worth stating plainly: a test given 96 passes has more chances to
exit on a lucky trailing window, which can turn a real leak into a reported pass. The per-run delta
here is a wide distribution straddling the 20 KB threshold, so that is a real risk, not a
theoretical one. It is the price of keeping the heap check enabled for these tests at all; muting
them would give up the detection outright.

A per-test heap *threshold* override would be the better instrument -- it would keep a fixed
iteration count while still catching a gross leak -- but TestRunner has only mute lists and
iteration overrides today. Worth considering as a follow-up if these tests keep flagging.

## Alternative that was tested and rejected

Pre-saturating the shell cache inside the test (showing the dialog N extra times per run) was
measured on `TestNativeMessageBox`: 2 dialogs/run failed 2/2 (28.9, 36.0 KB), 8 dialogs/run passed
2/2 (11.7, 18.4 KB), 16 dialogs/run failed 2/3 (8.5, 20.5 KB heap, and one run that passed the heap
check at 0.0 KB but tripped the total-memory threshold at 179.6 KB). It shifts the mean without
removing the variance, and trades the heap threshold for the memory one. It also takes the test
from 1-2 s to 8 s per execution, on every execution rather than only during pass 1.

## Test plan

- [x] Solution builds
- [x] CodeInspection passes
- [x] Verified on this machine with the override in place: TestNativeFileDialog -39.9 KB (25
      iterations), TestNativeMessageBox -62.2 KB (24), TestPrmMcpConnector +14.8 KB (25). PRM
      converged on pass 25, one past the old 24 ceiling that would have failed it.
- [x] Three further trials of each, plus a disconnected-session run (`tsdiscon`) where PRM needed
      pass 27 and TestNativeFileDialog pass 28 -- both of which the old ceiling would have failed

### DIFF
--- a/pwiz_tools/Skyline/TestRunner/Program.cs
+++ b/pwiz_tools/Skyline/TestRunner/Program.cs
@@ -114,32 +114,15 @@ namespace TestRunner
             // with no potential for infinite looping on a detected leak.
             {"TestLibraryExplorer", new ExpandedLeakCheck()},
             {"TestLibraryExplorerAsSmallMolecules", new ExpandedLeakCheck()},
-            // Every one of these shows the native common file dialog, and the growth they report is
-            // Windows', not Skyline's. Diffing the process heap's block-size histogram across a
-            // pass-1 run names it: the blocks that accumulate hold Explorer's cached DirectUI view
-            // markup ("<duixml>", "ViewHost", "ProperTreeModule", "DetailsContainer") plus the
-            // navigated path's own strings, one set per dialog shown. That is the shell view of the
-            // modern Explorer-hosted IFileDialog, cached per instance on a UI thread that lives for
-            // the whole process, so nothing gives it back until CoUninitialize at exit.
-            // The growth SATURATES rather than leaking without bound: on a long-lived thread the
-            // marginal cost decays to ~2 KB by the 40th dialog. (Legacy comdlg32 really is
-            // dead-linear at ~29 KB/dialog, but Skyline never takes that path -- AutoUpgradeEnabled
-            // is never set false, and ShowHelp/ShowReadOnly are never set at all.)
-            // It is the dialog and not the connector: TestSetItemMcpConnector drives the same
-            // connector with the same polling but opens no native dialog, and it converges and
-            // passes while sharing none of the block sizes above.
-            // So these tests do settle -- they just need more runs than the default 24 to get
-            // there. TestPrmMcpConnector sits right at the 20 KB threshold at 24 iterations
-            // (26.6 KB one run, 13.8 KB the next) and falls to -20.5 KB by 48. x4 rather than the
-            // default x2 because the nightly machines are further up the curve than the box this
-            // was measured on: on 2026-07-23/24 they reported 26-71 KB where this machine reported
-            // 13-27 KB. Extra iterations are only CONSUMED when a test has not settled (the loop
-            // exits as soon as the trailing deltas pass), so a generous ceiling costs nothing on a
-            // machine that converges early.
-            // Tradeoff to know about: because the per-run delta is a wide distribution straddling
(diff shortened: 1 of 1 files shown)


nickshulman:
Closing without merging. #4458 is merged and is the right basis for this work.

To be precise about what this PR did, since it matters for the record: it did NOT mute anything -- MutedHeapMemoryLeakTestNames stayed empty and the heap check stayed enforced. It only raised the iteration ceiling. But that is still the wrong move now, for a stronger reason than muting:

* On the agents that actually fail, the growth is dead linear with no plateau (61-212 KB/run per the log analysis in #4458), so extra iterations never converge -- they only delay the same failure.
* Worse, a test given 96 passes can find a lucky trailing window and report a pass. That deletes the failing data point the new diagnostics are meant to explain. When the goal is more signal, loosening the gate is counterproductive.
* It also conflicts with #4458 in TestRunner/Program.cs, and #4458 rework of TestMcpConnectorBackgroundDialog (one window instead of ~100,000) removes the reason for that test entry here.

What is worth keeping from this branch is the evidence, not the change. Both sessions independently established that Skyline only ever takes the modern IFileDialog path (AutoUpgradeEnabled defaults true and is never set false; ShowHelp/ShowReadOnly never set; UseVistaDialogInternal=True verified by reflection), and that the unbounded ~27 KB/dialog figure was a HeapProbe artifact of setting AutoUpgradeEnabled=false. Those were the premises #4453 rested on. Detail is in ai/todos/active/TODO-20260725_native_dialog_leak_fix.md.

Issue #4486, opened 2026-07-28
Osprey Stage 7: SecondPassFDR is the whole-run memory peak (~45 GB at 82 files) -- characterize live vs GC-gray before choosing a lever


brendanx67:
## Summary

Stage 7 (`SecondPassFDR`) is the whole-run memory HIGH POINT: **~45 GB private** on the
82-file SEA-AD Astral pass2ab run. Unlike the Stage-6 slope in #4472, **the root cause here
is uncharacterized** — it is not yet known whether that 45 GB is a live survivor pool or
Server-GC committed-but-free "gray" (#4404). That distinction decides the lever, so it has
to be settled before any fix is designed.

Split out of #4472 so that issue stays scoped to the two O(files) structures whose root
cause IS known (Stage-6 `_perFileEntries`, and the HPC-resume pre-compaction pool) and can
be closed by a PR with a crisp byte-parity gate.

## What was measured

2026-07-19, 82-file SEA-AD Astral pass2ab run, perfviz `--memstamp`:

* **peak 44.1 GB private / 29.1 GB managed at SecondPassFDR** — the run's high point.
* There are **no reporting gaps** at this stage; progress coverage is fine. It is only the
  memory peak.

## Suspected, NOT confirmed

Likely the 2nd-pass Percolator retrain plus scoring on the reconciled survivor pool, holding
the survivor `FdrEntry` and features resident — i.e. an O(survivors) projection
(`FdrStreamingSink` / the 2nd-pass projection). This is the residual left over from #4374,
carried forward here so it is not lost.

Treat that as a hypothesis. The whole point of this issue is that it has not been verified.

## First step (before choosing any lever)

1. `[MEM]` probes at the SecondPassFDR peak. Use the **post-GC** probe
   (`ProfilerHooks.LogManagedHeapAfterGcIfEnabled`, enabled by `OSPREY_LOG_MEMORY`), not the
   `--memstamp` trace: `--memstamp` is `GC.GetTotalMemory(false)` and includes uncollected
   garbage, so it cannot distinguish a live pool from GC gray — which is exactly the
   question. ProfilerHooks' own docs note two identical runs can differ by tens of GB on
   whether a gen-2 landed before the sample.
2. dotMemory at that same boundary (`Profile-Osprey.ps1 -MemoryProfile` captures a retention
   snapshot at the post-GC probe) to see WHICH objects are retained.
3. Only then choose a lever.

## Reproducing at small scale

[...]

Issue #4472, opened 2026-07-25
Osprey Stage 6: PerFileRescore holds every file's survivor FdrEntry buffer resident (O(files)) -- stream per file


brendanx67:
## Summary

`PerFileRescoreTask` holds **every file's post-compaction survivor `FdrEntry` lists resident at once**, so Stage-6 memory is O(files). This is the remaining O(files) resident structure between us and a 500-file run on a 64 GB box, now that Stage-5 FirstPassFDR is memory-flat (PR #4435).

The fix is to hold only the bounded cross-file reconciliation state resident and reload each file's survivors on demand -- the same treatment Stage 5 received.

**Scope**: `pwiz_tools/Osprey/Osprey.Tasks/PerFileRescoreTask.cs` (the `_perFileEntries` buffer and `RescoreAllFiles`), `Osprey.FDR/Reconciliation/*` (the cross-file reconciliation that requires the all-files view).

## What we observed (2026-07-19, 82-file SEA-AD Astral pass2ab run, perfviz --memstamp)

PerFileRescoring's private-bytes band is far lower than before the 5-day memory push (~50-60+ GB down to ~25-30 GB), BUT it has a **positive slope in file count** -- it still grows O(files), unlike the now-flat FirstPassFDR band. At 500 files that slope re-crosses the wall.

## Root cause (code-verified)

Verified against master on 2026-07-25 (originally found at df1084f93):

* **`PerFileRescoreTask.cs:113`** -- `private List<KeyValuePair<string, List<FdrEntry>>> _perFileEntries;`
* Set from the compacted survivor set for ALL files at **`:201`** -- `_perFileEntries = ctx.Get<CompactedEntries>().Value;` -- then published at `:208` (`ctx.Publish(new RescoredEntries(_perFileEntries))`), overlaid in place, and rescored across `nTotalFiles = perFileEntries.Count`.
* The code labels it itself, at ~`:590`: *"clean PERSISTENT managed heap (all files' rescored FdrEntry buffer + ...)"*, probed as `reconciliation-resident`.

So the resident cost is O(total survivors) = O(files). At 82 files the survivor pool is ~12.4M entries (~5.39M of which pass to the blib); at 500 files it is 5-6x that.

This is **genuine live O(files) growth** -- distinct from the Server-GC committed-but-free "gray" seen elsewhere (see #4404). It will not decommit away.

[...]

PR #4488, opened 2026-07-28
Released the Stage 6 rescore payload after writing its reconciled parquet

brendanx67:
## Summary

* Stage-6 rescore memory no longer grows with file count: the 82-file SEA-AD Astral set went from a projected **~197 GB** to a measured **32.2 GB peak** with a level band
* The resume/HPC rehydrate now compacts one file at a time instead of materializing every file's pre-compaction pool, and stops loading PIN features it never reads
* `--model-diagnostics` works on that bounded path - it no longer forces the resident pool, and the report is byte-identical
* The `--input-scores` path had NO unbounded-memory check at all; it now **warns and names the consumer** that put the run on the resident pool, so an operator meeting an O(files) peak at scale knows which knob caused it

Addresses the Stage-6 half of #4472. Stage 7 (`SecondPassFDR` / `ExpectReconciledInput`) is deliberately out of scope - see below.

## What was actually wrong

The issue assumed the resident cost was the survivor `FdrEntry` lists and that consensus/reconciliation state was O(distinct peptides). Neither held:

* Stage-6 survivors arrive **lean** (~184 B; `LoadFdrStubsFromParquet` reads 10 scalar columns).
* `PerFileConsensusTargets` / `ReconciliationActions` / `PerFileGapFillForRescore` are keyed **by file** - O(files x targets), not O(distinct).
* Two separate O(files) terms were in play, and they needed different fixes:
  1. **During the rescore loop** - `OverlayRescoredEntries` replaces each lean stub with a rescored entry carrying `Features` / `CwtCandidates` / `Fragment*` / `ReferenceXic*` (~1.28 KB/entry measured), and gap fill appends more. Nothing reads those after the per-file parquet write.
  2. **Before the loop even starts** - the rehydrate materialized every file's FULL pre-compaction pool (~4.25M rows/file, 1.19 GB/file) and only compacted afterwards.

## The fixes

**Release the payload after the write.** Six arrays nulled once the file's reconciled parquet is on disk. **Ordering is load-bearing**: `ReconciledParquetWriter.BuildOverlay` selects overlay rows by `Features != null`, so releasing before the write would silently emit parquets with no overlaid rows. Documented at the null site.

**Skip PIN features on the reconciled-bundle rehydrate.** `FirstJoinTask` nulls them before compaction anyway and `RescoreHydration` never reads them - ~800 MB/file loaded only to be discarded. Gated on `NeedsResidentPool` being false, which is exactly the flag saying no resident 2nd-pass Percolator will want them.

**Stream the compaction.** `RescoreCompaction.Apply` retains `GlobalFirstPassBaseIds` UNION the action-target base_ids - both from the small `reconciliation.json` envelopes, neither from the loaded pool. So a pre-pass over the envelopes builds the retain set, then each file loads -> overlays -> compacts -> releases in turn. Compaction discards ~52x, so peak went from N x 1.19 GB to 1.19 GB + N x ~23 MB.

**Byte-identity mechanism:** `FirstJoinTask` still calls `RescoreCompaction.Apply`, unchanged. On an already-compacted buffer it re-derives the identical retain set, removes nothing, and rebuilds the action map by walking the same lists in the same order - so `Apply` remains the sole authority on both the survivor set and the `vec_idx` remap, and streaming is a strictly conservative pre-filter.

**Bound `--model-diagnostics`.** The report is built from PRE-compaction entries (decoys + entrapment still present), so handing it a compacted pool would produce a plausible WRONG report. Instead the streaming hydrate feeds `ModelDiagnosticsData.Accumulator` every pre-compaction entry as each file loads - the accumulator's `Add` takes only scalars (identity, decoy flag, score, q-values), all present on a rehydrated stub. Same entries in, same report out, bounded.

## Measured

82-file SEA-AD Astral, `--task PerFileRescoring`, identical Stage 1-5 artifacts:

Four runs, same 82 files and same Stage 1-5 artifacts. "pre" = before the review fixes, "final" = review fixes + `master` merge:

| | pre, plain | pre, mdiag | **final, mdiag** | **final, plain (control)** |
|---|---|---|---|---|
| peak private | 32.2 GB | 35.2 GB | **33.9 GB** | **33.5 GB** |
| floor drift (private) | +19 MB/file | +36 MB/file | **+21 MB/file** | **+7 MB/file** |
| floor drift (managed) | +18 MB/file | +26 MB/file | **+1 MB/file** | **+23 MB/file** |
| reporting gaps >= 30s | **0** (max 12s) | **0** (max 11s) | **0** (max 11s) | **0** |
| rescored | 6,954,057 | 6,954,057 | **6,954,057** | **6,954,057** |
| wall | 2:45:12 | 2:47:27 | **2:47:39** | **2:45:22** |

Against the original projection of **~197 GB** (1.19 GB/file x 82) for the unbounded path, every column above is bounded.

The last column is the final build (review fixes + `master` merge).

### Read the per-file drift numbers with caution

A non-mdiag **control** was then run on the same final build, and it re-measured the non-mdiag path at **+7 MB/file private** where the earlier non-mdiag run measured **+19 MB/file** - a 3x difference on a path **nothing in this PR touches**. That is run-to-run variance in the floor fit, not a change.

So the per-file drift metric is too noisy at n=1 to support a precise improvement claim, and the honest summary is narrower than the table alone suggests:

* **Solid:** no O(files) growth on any run; **0 reporting gaps >= 30 s** in all four; **6,954,057 entries rescored identically** in all four; report byte-identical; no wall-clock cost.
* **Directional:** releasing the accumulator is clearly visible mid-run - an elapsed-matched A/B (`perfviz.py a.log b.log`) shows the baseline climbing monotonically 26 -> 33 GB through the rescore loop while the fixed build stays flat - but the two memory axes disagree on magnitude (private says mdiag still costs ~14 MB/file vs ~17 before; managed says mdiag now sits *below* non-mdiag), so no single figure is quoted.
* **Not claimed:** that the `--model-diagnostics` memory penalty is eliminated. It is reduced; the measurement cannot resolve by how much.

The **report is byte-identical except its embedded `generatedUtc`** (verified by `cmp`/`diff`: one differing line, the `osprey-data` JSON header; `ospreyVersion` / `fileCount` / `featureCount` / `runFdr` all match). That is what proves the accumulator is released strictly *after* the report is built.

Peak moved only 35.2 -> 33.9 GB, which is expected rather than disappointing: peak is set during the hydrate, while the accumulator is necessarily resident. Releasing it can only buy floor, not peak.

Read with `ai/scripts/perfviz.py` from the run's `--timestamp --memstamp` log; method and traps in `ai/docs/memory-band-guide.md`.

## What this does NOT fix

**A residual ~19 MB/file** (36 with mdiag) - the all-files lean survivor list (~184 B/entry), which is what #4472's L2 targeted. At 500 files that is ~10-18 GB of rise, inside 64 GB, so it is no longer the scaling blocker. Remaining work on #4472.

**Stage 7 (`SecondPassFDR`).** `ExpectReconciledInput` still takes the resident pool. That is a bounded-height bulge (~51 GB peak on a 64 GB box in a full run), not a growth curve, so it does not prevent an 82-file run - and Percolator is being removed from that stage, so hardening it now would harden code about to be replaced. Tracked as #4486. It now emits the named warning rather than being silently unchecked, and `regression.ps1` no longer needs `OSPREY_ALLOW_UNBOUNDED_MEMORY=1` to get past it.

## Why the guard warns instead of throwing

An earlier revision of this PR threw on the `--input-scores` resident-pool path, matching the `--input-files` paths. That was wrong, and worth stating explicitly because it is the kind of change that looks strictly safer than it is: **every configuration reaching that loader worked before this PR** - `--task SecondPassFDR`, a `--task FirstPassFDR` re-run over parquets whose sidecars are on disk, and the `OSPREY_DUMP_PERCOLATOR` bisection dump. Throwing would have turned a memory characteristic into three hard failures.

It was invisible locally because `regression.ps1` set `OSPREY_ALLOW_UNBOUNDED_MEMORY=1` around the **entire** HPC chain, so mode 3 could not have caught it - or any future guard regression on an `--input-scores` worker, which is precisely what mode 3 exists to exercise. **That opt-in is removed** and mode 3 now runs with the guard armed. Mode 2 keeps a scoped one for a genuinely unfixed O(files) path - see "Known boundary" below.

## Test plan

- [x] `regression.ps1 -Dataset All` - **18/18 PASS**, byte-identical at 1e-9, all four datasets, modes 1/1b/2/3. Re-run after the review fixes AND the `master` merge (#4487), **with the guard armed** - the mode 3 opt-in that used to wrap the whole HPC chain is gone
- [x] `Build-Osprey.ps1 -RunTests -RunInspection` - 551/551, zero warnings
- [x] The new warning is **verified firing, not asserted**: captured from Astral's mode 3 merge node - `[WARN] The reconciled-input merge (--task SecondPassFDR, tracked in #4486) requires the RESIDENT pre-compaction first-pass pool...` - while mode 3 still PASSED, which is the intended behaviour (warn, do not break)
- [x] `OSPREY_FDR_PROJECTION=0` on Stellar - PASS. **Not covered by the normal gate**, and the one path where the feature-skip could change behaviour invisibly (`ComputePass2Resident` lets an unmapped identity keep stale `Features`).
- [x] 82-file Stage-6 run, with and without `--model-diagnostics`, identical rescored counts
- [x] New unit test: streaming vs resident+`Apply` equivalence (survivors, action map incl. remap, `Stats`)

## Notes for review

**`regression.ps1` mode 3 runs one rescore worker per stem, so it never exercises the multi-file Stage-6 loop.** Modes 1 and 2 are the real gates for this change.

**The `ShouldStreamCompaction` gate is the fragile part.** Streaming must never hand a pre-compacted pool to a consumer expecting pre-compaction. It is now a single predicate, `PreCompactionPoolReason(...) == null`, which is also what produces the warning text - so the decision and the explanation can never disagree about why. Any future consumer inserted between `PerFileScoring` and FirstJoin's compaction re-opens that silently, and no gate we have would catch it: **adding a consumer means adding a clause there.**

`regression.ps1` now runs every Osprey invocation with `--timestamp --memstamp`, so any gate log doubles as a memory-band trace.

See ai/todos/active/TODO-20260727_osprey_stage6_rescore_streaming.md

## Review pass

A `/code-review max` pass over this branch found 15 verified findings, all in the wrappers around the streaming core - two independent angles verified the compaction/ordering mechanism itself as clean. The substantive ones, now fixed:

* The resident-pool **throw** would have hard-failed `--task SecondPassFDR`, a `--task FirstPassFDR` re-run, and `OSPREY_DUMP_PERCOLATOR` - all of which worked before this PR. Now a named warning (see above).
* `ReleaseRescoredPayload` missed **both resume arms** (`TryResumeRescoredFile` and the whole-task `Rehydrate`), so O(files) retention was fully intact on any resume.
* The release was unconditional after a `WriteReconciledAndStamp` that can no-op; when it no-ops those arrays are the only copy of the rescore. Now gated on the write having persisted.
* The mdiag accumulator was pinned on a process-lifetime byproduct slot with no reader left (~1-2 GB at 82 files). Released once its report is written.
* `TotalPreCompactionStubs` was an `int` that overflows past ~505 files.
* `PreCompactionTallies` was keyed by file stem; two `--input-scores` paths in different directories can share a stem, silently overwriting a tally. Now positional.

Two further defects were caught by `-RunInspection` afterwards: a condition tested twice across an `&&` (dead code in the helper), and an extracted `TryHydrateRescoreBundle` that returned success on a `null` hydrate result, leaving callers dereferencing a field whose non-nullness could only be established inside the callee.

One finding was **rejected after measurement**: caching the `Process` object behind `--memstamp` measured 1.01 -> 0.95 ms/line, which at one log line every 2-5 seconds is nothing, and the cached form silently reports a *frozen, flat memory curve* if its required `Refresh()` is ever dropped - a false "bounded memory" result on the very instrument used to prove bounded memory. Left as-is.

## Known boundary: `--model-diagnostics` on a full resume

Removing both harness opt-ins surfaced a pre-existing gap the env var had been hiding: a **full resume** with `--model-diagnostics` still forces the resident pool, because FirstJoin skips the first-pass score pass and reports through the batch `ModelDiagnosticsReport.Write`, which reads resident per-file entries. The guard is right to throw there - it genuinely is O(files).

`regression.ps1` mode 2 therefore keeps a **scoped** opt-in, commented with the reason. Mode 3 keeps none.

This is not the scale case. `--model-diagnostics` over `--input-scores` - what an 82-file run uses - streams off `ModelDiagnosticsData.Accumulator` and needs no opt-in at any file count. A `-LinkFrom` resume is likewise unaffected. The fix is ~40-60 lines reusing the accumulator plumbing this PR already built; scoped as a follow-up rather than grown into this PR.


brendanx67:
## 82-file Stage-6 run: memory is level, no reporting gaps

Full 82-file SEA-AD Astral `--task PerFileRescoring` run on the three commits here, measured with `--timestamp --memstamp` and read by `ai/scripts/perfviz.py`:

```
duration      : 2:45:12
reporting gap : max 12s   median 2s   p95 6s
gaps >= 30s   : 0   OK
managed MB    : peak 22.3 GB   floor  6.9 ->  8.3 GB   drift +1.42 GB   +18 MB/file
total MB      : peak 32.2 GB   floor 16.3 -> 17.8 GB   drift +1.49 GB   +19 MB/file
rescored      : 6,954,057
```

### Before / after

| | before (16 files) | after (82 files) |
|---|---|---|
| peak private | 48.0 GB | **32.2 GB** at 5x the file count |
| gaps >= 30s | 4 (68s, 33s, 59s, 56s) | **0**, max 12s |
| rehydrate | climbed monotonically to a 49 GB spike | one bounded pre-pass, then flat |

The old path projected **~197 GB** for 82 files (1.19 GB/file x 82) and would not have completed on a 128 GB box. This run finished at 32.2 GB peak with ~96 GB free.

### What is NOT fixed

A residual **+19 MB/file** floor drift remains - the all-files lean survivor list (~184 B/entry), which is what the original issue's L2 targeted. At 500 files that is ~9.5 GB of rise, comfortably inside 64 GB, so it is no longer the scaling blocker. Tracked as remaining work on #4472.

Note the 20-file measurement suggested +41-48 MB/file; the per-file cost falls as the run lengthens, so that extrapolation was pessimistic.

### Reproducing

```
ai/scripts/Osprey/SEA-AD/Measure-Stage6Rescore.ps1 -FileCounts 82 -PhaseDir <stage-1-5 artifacts>
python ai/scripts/perfviz.py <run>/stage6-82f.log --files 82 --png out.png
```

Method and traps: `ai/docs/memory-band-guide.md`.

PR #4495, opened 2026-07-28
Fixed sparse and marker-overlapping labels on Volcano and Relative Abundance plots

### LINKED ISSUE #4330: Volcano and Relative Abundance plots do not show enough labels
It looks like the label sampling filter is too aggressive and there is too much space left blank on the plots.

rita-gwen:
## Summary

* Changed the label sampler cap from a product of two density caps to a min, so each cap acts as a hard limit instead of compounding (the product collapsed to ~0.04 at high candidate counts, starving the plot).
* Added a post-annealer pass (`LabelLayout.PruneOverlappingLabels`) that hides a label if it overlaps a kept label or covers the center of another point's marker; a label's own marker is excluded and selected labels are never pruned. Deterministic order, spatial-hashed.
* Fixed `GraphObjList.FindPoint` to skip invisible objects, so hidden (sampled-out or pruned) labels no longer intercept mouse-over and screen the visible label underneath.

Fixes #4330

## Test plan

- [x] TestLabelLayoutDeterminism - re-snapshotted; two runs byte-identical (determinism holds)
- [x] TestVolcanoPlotFormatting - green (formatting and remainingObjs invariant intact)
- [x] TestVolcanoPlotLayout - green (saved-layout path unaffected)
- [x] TestPeakAreaRelativeAbundanceGraph - green
- [x] CodeInspection + QuickInspection - 0 warnings / 0 errors
- [x] Manual repro (dense volcano): label wall reduced to well-separated labels off the marker cloud


pwiz_tools/Shared/zedgraph/ZedGraph/LabelLayout.cs
```diff
@@ -707,6 +716,203 @@ public bool ApplyPlacements(LayoutResult result, Graphics g)
...
+            // survivors do not depend on alphabetical ordering and two runs over the same data agree.
+            entries.Sort((a, b) =>
+            {
+                if (a.Point.IsSelected != b.Point.IsSelected)
+                    return a.Point.IsSelected ? -1 : 1;
+                var hashCompare = a.Hash.CompareTo(b.Hash);
+                if (hashCompare != 0)
+                    return hashCompare;
+                return string.CompareOrdinal(a.Text, b.Text);
+            });
```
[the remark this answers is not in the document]
rita-gwen: Addressed in 57bb572 (pushed before this review posted): the sort now falls through to a stable `UniqueID`-based tiebreaker (`LabelRect.TieBreak = point.UniqueID?.ToString()`), so distinct labels with identical text never compare equal. Verified determinism holds (TestLabelLayoutDeterminism, two runs byte-identical).

pwiz_tools/Skyline/TestFunctional/LabelLayoutTest.cs
```diff
@@ -35,14 +35,14 @@ namespace pwiz.SkylineTestFunctional
     public class LabelLayoutTest : AbstractFunctionalTestEx
```
nickshulman, line 36: This test fails in French, Chinese and Japanese.
It passes if you run it offscreen, but fails when run onscreen.

nickshulman (commented):
I see there are a bunch of "catch (InvalidOperationException)" in LabelLayout.cs.
It looks like those are there because you are iterating over curve items on a background thread while the UI might be updating.
It would probably be better to fix that by changing LabelLayout so that it's operating on an immutable snapshot of the shapes that it is supposed to be fitting. Then, after it has done its work on a background thread, the UI thread can move everything to where they need to go.

Catching exceptions related to collections being modified will probably result in strange behavior depending on timing.

pwiz_tools/Skyline/TestFunctional/LabelLayoutTest.cs
rita-gwen: Fixed in 2bc19d2. The pinned snapshot encodes English chart geometry - localized axis titles/fonts shift the chart rect, which legitimately changes which labels the sampler/pruner keep - so the pins are now checked only in English. The run-to-run determinism comparison still runs in every language, and a new VerifyLayoutInvariants() verifies the pruner's actual guarantees (no non-selected label overlaps another or covers a foreign marker center) under any geometry. The test also clears the static saved layout before each capture, which fixes a related failure when all languages run as one SkylineTester batch. Verified onscreen in en/ja/fr/zh, single and batch.

rita-gwen:
Re the catch (InvalidOperationException) blocks in LabelLayout.cs (review comment from @nickshulman): refactored as suggested in 2bc19d2. The LabelLayout constructor now runs on the UI thread and captures an immutable snapshot of the point-marker geometry (screen rectangle from LineItem.GetCoords plus the transformed point center); the density grid, the annealer's marker lookups, and the overlap pruner all read only that snapshot, so the background thread never touches the live CurveList. All the InvalidOperationException/ArgumentOutOfRangeException catches are removed, along with the worker-thread LabelLayout construction fallback in LabelLayoutRunner. The snapshot preserves the previous rectangle values and match semantics exactly, so layout output is unchanged (the English pinned-snapshot test still passes untouched).

rita-gwen:
New since the last review (7ddfbd86), plus a master merge to bring the branch up to date.

The sampler was still starving dense plots. On `ProteoBugs/NotEnoughLabels/ExtracellularVesicalMagNet.sky`, the Relative Abundance panel showed 3-4 labels with both formatted series marked as labeled, while most of the plot was empty.

Root cause: the per-cell cap in `SamplePointsByDensityGrid` divided by `GridCell.PointCount`, which `LabelLayout.FillDensityGrid` increments once per **data point marker**, not per label candidate. So `MAX_LABELS_PER_CELL = 4` actually meant "4 divided by the marker density under this label". Where the labeled points sit on a ribbon of ~1600 markers, a cell covering 30 markers gives a keep probability of 4/30, no matter how much free space the chart has.

It was also double counting, which is why the cap can simply drop: the markers under a label are already respected downstream twice, by the annealer's `EvaluateLabelBaseCost` density term and by `PruneOverlappingLabels`, which hides a label covering a foreign marker center.

The fix counts label candidates per cell (a histogram over the cells already computed for the candidates) and caps on that, 13 lines in `LabelLayoutRunner.cs`. `MAX_LABEL_AREA_RATIO` stays at 0.3; the plot looks right without raising it.

Verification: the developer confirmed the repro plot visually, and `LabelLayoutTest`, `VolcanoPlotFormattingTest`, `VolcanoPlotLayoutTest` and `PeakAreaRelativeAbundanceGraphTest` all pass. Worth noting that those four pass identically with and without the fix, so none of them covers this behavior - reproducing it needs a document whose labeled points sit on a large marker cloud, which `Rat_Plasma.sky` is not. We decided against adding one for now.

PR #4498, opened 2026-07-29
skyline: Stopped inferring a Waters lockmass function from waters_connect mzML

bspratt:
## Summary

* Skyline discarded every spectrum in Waters mzML written by DATA Convert - all 57,164 in
  the reported DDA file. Skyline decides a file is Waters from `MS:1000526 Waters raw
  format` on the source file, not from its nativeIDs, so it recognized these correctly and
  then applied MassLynx assumptions to them. The ids are the waters_connect dialect,
  `channel=2 process=0 spectrum=1 scan=1`, where MS:1000769 defines `function=N process=M
  scan=K`; the function number was parsed by position from the abbreviated id, so the
  leading MS1 at channel 2 read as function 2, was taken for the lockspray function, and
  everything at channel 2 or above was dropped
* This surfaced now because DATA Convert 4.0.0.2619 began correctly declaring the data's
  Waters provenance. An older converter omitted `MS:1000526`, so Skyline never treated
  those files as Waters and they imported fine - the same dialect, working, purely because
  the metadata was worse. Waters improving it is what exposed the bug
* Stopped inferring a lockmass function from a dialect that does not carry one. Per Waters
  (Ian Morns), DATA Convert lockmass corrects and drops the reference scans rather than
  exporting them, so there is nothing to infer
* Waters function numbers are now parsed by nativeID layout rather than by position, which
  also fixes the merged DDA form `merged=N function=M process=P scans=A-B`
* Skyline honors `MS:1000928 calibration spectrum` wherever it appears, since the term is
  standard and the UIMF reader already emits it. Nothing here writes it - the existing
  heuristic remains for the files that carry no such label, which is all of them today
* Global TIC no longer sums the lockmass function when `ignoreCalibrationScans` excludes
  its scans. Skyline sets that flag on every open, so the reported TIC maximum for a
  Waters MSe file was the lockspray scan's own signal
* `SpectrumListWrapper` forwards `calibrationSpectraAreOmitted`, which the base
  implementation answered `false` for any wrapped list - peak picking or lockmass
  refining, both of which Skyline applies
* The msLevel filter now accepts a declared ms level whether or not the spectrum is a
  "mass spectrum". `MS:1000928 calibration spectrum` is a child of "spectrum type" but not
  of "mass spectrum", and the UIMF reader writes it as the sole type on frames that carry a
  real ms level, so the predicate discarded that level and treated them as ms level 0 -
  `msconvert --filter "msLevel 1-"` dropped UIMF calibration frames. **Reviewers: this is a
  pwiz-module change to shared analysis code that every vendor passes through, deliberately
  included here rather than split out. It is pre-existing and not caused by the Waters work;
  it is included because honoring MS:1000928 on read and ignoring it in the filter is
  incoherent.** No checked-in baseline moves - no vendor reader test uses an msLevel filter
  and the UIMF reference has no calibration frames - so the new unit test is the only thing
  behind it, and it is mutation-verified
* `SpectrumList_Waters` resolves the lockmass function in its constructor rather than
  lazily. The lazy version raced (`spectrum()` holds `readMutex`,
  `calibrationSpectraAreOmitted()` does not) and returned a wrong `false` in DDA mode,
  where `createDDAIndex()` never primed the cache

Reported by Hans.

## Test plan

- [x] Reader_Waters_Test - 64 configs, including a new `ignoreCalibrationScans` case
      against MSe_Short.raw, which unlike the existing one has a real lockspray function
- [x] SpectrumList_LockmassRefinerTest, SpectrumList_FilterTest, and the rest of the
      spectrum_processing suite
- [x] TestWatersCalibrationSpectrum - new; three mzML fixtures (with and without the
      calibration label, and one in the waters_connect dialect), plus every nativeID layout
      the Waters reader emits. The waters_connect fixture is the reported shape - a leading
      MS1 on channel 2 followed by channel 3, and no lockspray scan, as DATA Convert writes
      it. Mutation-verified: reinstating the channel-as-function misreading fails it
- [x] TestTicChromatogram, TestInstrumentInfo, TestScanDescription, WatersImsMse* (2),
      WatersFileTypeTest, WatersCacheTest, WatersMultiReplicateTest
- [x] CodeInspection, and ReSharper whole-solution inspection - zero errors and zero
      warnings across the entire solution, re-run on the final head
- [x] Customer files: DDA 0 to 30 non-zero peaks; MSe 4 to 5 peaks, matching a direct
      import of the source .raw to within 3% at the same retention time. Measured before
      the calibration-spectrum writer was withdrawn; that withdrawal only removed
      msconvert-side code which Skyline never executes, and the msLevel predicate is not
      reachable from Skyline either, so these numbers are unaffected


pwiz_tools/Skyline/Model/Results/SpectraChromDataProvider.cs
```diff
@@ -1228,13 +1228,13 @@ private SpectrumInfo ReadSpectrum(ref int i)
                             // For Waters msE skip any lockspray data
                             if (_filter.IsWatersMse)
                             {
-                                // looking for the 3 in 3.0.1 (or the 10 in 10.0.1) or the 2 in 1.2.3 if it's combined ion mobility'
-                                if (MsDataSpectrum.WatersFunctionNumberFromId(nextSpectrum.Id, _dataFile.HasCombinedIonMobilitySpectra && nextSpectrum.IonMobilities != null) > 2)
+                                // Functions above 2 are lockspray, not data. Null when the file carries no
+                                // function number, in which case only an explicit MS:1000928 identifies lockspray.
+                                if (nextSpectrum.WatersFunctionNumber > 2 || nextSpectrum.IsCalibrationSpectrum)
                                     continue;
```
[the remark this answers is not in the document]
bspratt: Not a compile error - lifted relational operators return `bool`, not `bool?`. The C# spec gives `<` `>` `<=` `>=` a lifted form whose result type stays `bool`, evaluating to `false` when either operand is null; only lifted equality and arithmetic produce nullable results. The solution builds clean, which is the stronger proof.

The null-is-false behavior here is deliberate rather than incidental: a waters_connect spectrum carries no function number, and `null > 2` being false is what stops it from being discarded as lockspray. Substituting 0 for "missing" would give the same answer today but hard-codes a sentinel where the code means "unknown", which is the distinction this PR exists to draw.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -817,8 +832,7 @@ protected virtual SpectrumList SpectrumList
                             var msLevel = GetMsLevel(spectrum);
                             if (msLevel == 1)
                             {
-                                var function = MsDataSpectrum.WatersFunctionNumberFromId(id.abbreviate(spectrum.id), 
-                                    HasCombinedIonMobilitySpectra && spectrum.id.Contains(MERGED_TAG));
+                                var function = MsDataSpectrum.WatersFunctionNumberFromNativeId(spectrum.id);
                                 if (function > 1)
                                     _lockmassFunction = function; // Ignore all scans in this function for chromatogram extraction purposes
                             }
```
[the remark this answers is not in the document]
bspratt: Same as the other thread: `function > 1` where `function` is `int?` compiles and is `false` when null, because lifted relational operators return `bool` rather than `bool?`. Verified by the build being clean.

Relying on that is the point of the change. `WatersFunctionNumberFromNativeId` returns null for a dialect that carries no function number - waters_connect `channel=` ids - and `null > 1` being false is exactly what makes the heuristic decline to infer a lockmass function for those files. That is the customer-facing fix: previously the channel number was read positionally as a function number, so every spectrum at channel 2 or above was discarded. An explicit null branch here could only restate what the language already guarantees.

pwiz/analysis/spectrum_processing/SpectrumList_Filter.cpp
```diff
@@ -318,11 +318,15 @@ PWIZ_API_DECL SpectrumList_FilterPredicate_MSLevelSet::SpectrumList_FilterPredic
 
 PWIZ_API_DECL boost::logic::tribool SpectrumList_FilterPredicate_MSLevelSet::accept(const msdata::Spectrum& spectrum) const
 {
-    CVParam param = spectrum.cvParamChild(MS_spectrum_type);
-    if (param.cvid == CVID_Unknown) return boost::logic::indeterminate;
-    if (!cvIsA(param.cvid, MS_mass_spectrum))
+    // A spectrum may carry more than one child of "spectrum type" - a Waters lockspray scan is both an
```
chambm: This is incorrect. A spectrum should only have one spectrum type: the mzML mapping file (not schema) specifies that. Is there a problem replacing MS1 spectrum with calibration spectrum (there will still be ms level = 1)?

pwiz/data/vendor_readers/Waters/Reader_Waters.cpp
```diff
@@ -122,15 +122,35 @@ void fillInMetadata(const string& rawpath, RawDataPtr rawdata, MSData& msd)
...
+    // 160109_Mix1_calcurve_070.mzML announces "SRM spectrum" while holding none. Driving the loop from
+    // the functions actually indexed would fix all of them at once, and is worth doing separately.
+    bool hasCalibrationSpectra = sl != NULL && sl->hasCalibrationSpectra();
+
     for (int function : rawdata->FunctionIndexList())
     {
         try
         {
+            bool isLockMass = sl != NULL && sl->isLockMassFunction(function);
+            if (isLockMass && !hasCalibrationSpectra)
```
chambm: Promote the !hasCalibrationSpectra to the above boolean to avoid calling isLockMassFunction when hasCalibrationSpectra is true (rename variable to ignoreLockMassSpectrum or something like that).

pwiz/data/vendor_readers/Waters/Reader_Waters.cpp
```diff
@@ -122,15 +122,35 @@ void fillInMetadata(const string& rawpath, RawDataPtr rawdata, MSData& msd)
...
+                continue;
+
             int msLevel;
             CVID spectrumType;
             translateFunctionType(WatersToPwizFunctionType(rawdata->Info.GetFunctionType(function)), msLevel, spectrumType);
             if (spectrumType != CVID_Unknown)
                 msd.fileDescription.fileContent.set(spectrumType);
+            if (isLockMass)
+                msd.fileDescription.fileContent.set(MS_calibration_spectrum);
```
chambm: translateFunctionType may need to be refactored so it returns calibration spectrum for the lockmass function.

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
```diff
@@ -185,6 +193,11 @@ PWIZ_API_DECL SpectrumPtr SpectrumList_Waters::spectrum(size_t index, DetailLeve
     }
 
     result->set(spectrumType);
+    // Mark lockmass (lockspray) scans so consumers need not guess which function is the reference.
+    // Additive: the underlying spectrum type is left in place, since "calibration spectrum" is not a
+    // child of "mass spectrum" and replacing the type would strip m/z units, ms level, polarity etc.
```
chambm: Hmm. It's been a long time but I expect calibration spectrum got parented under spectrum because other kinds of spectra can have calibration. What does this mean that replacing the type would "strip m/z units, ms level, polarity"? `isMS` can still be true if `spectrumType` is refactored to be calibration spectrum.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -2486,9 +2503,64 @@ public ImmutableList<MsPrecursor> Precursors
         public string InstrumentSerialNumber { get; set; }
         public string InstrumentVendor { get; set; }
 
-        public static int WatersFunctionNumberFromId(string id, bool isCombinedIonMobility)
+        /// <summary>
+        /// True when this spectrum is explicitly labeled as a calibration (Waters lockmass) scan with
+        /// MS:1000928. Authoritative wherever it appears - no inference needed.
+        /// </summary>
+        public bool IsCalibrationSpectrum { get; set; }
```
chambm, line 2527: Can't Skyline just skip calibration spectra without allocating an MsDataSpectrum?

chambm (changes requested):
Thanks for tackling this Brian. It does need some changes due to a misunderstanding about mzML spectrum type.

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
bspratt: But you wouldn't know it was MS - you could make a good guess based on the presence of those things, but it would be a guess.

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
bspratt: Normally those never get passed in the first place - Skyline asks the API to leave them out. But you never know.

pwiz/analysis/spectrum_processing/SpectrumList_Filter.cpp
bspratt: On closer inspection of the ontology I see that you are correct - this is a terrible design flaw in the ontology, though. In practice only a semantic validator would care - I think we're better off being technically wrong. There are many kinds of spectrum and any of them could be calibration data, but you still would want to know what kind of spectrum it is.

pwiz/data/vendor_readers/Waters/Reader_Waters.cpp
bspratt: Done in df90ce8ba. The per-file test is now `ignoreLockMassFunction` and gates the call in the skip check, so `isLockMassFunction` is not consulted there for files whose lockmass scans are present.

I kept a single boolean rather than adding its complement: reaching the declaration line for the lockmass function already implies its spectra are in the output, since otherwise the guard above would have skipped it, so that call needs no second flag.

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
chambm: **Only** a mass spectrum would have an ms level AFAIK, so you don't have to guess for our case. But for other types of spectrum I suppose you'd have to guess or know the ontology (currently there's only 2, EMR spectrum and mass spectrum).

pwiz/analysis/spectrum_processing/SpectrumList_Filter.cpp
chambm: I think that's a risky assumption about only semantic validators caring. Either way this is a potentially breaking change. I think I'd rather break on the side of semantic correctness than not. Unless we want to propose a change in the CV, which would certainly delay things.

bspratt: We should definitely propose a change, but yeah, we can't wait for it either. 

I guess the real answer is "the existing calibration-indicating term is unusable, just drop the whole thing until the ontology is fixed and carry on with our heuristic for now".

pwiz/data/vendor_readers/Waters/SpectrumList_Waters.cpp
bspratt: Good point, we can just look for an MS level declaration instead of checking the spectrum type.

bspratt: "strip m/z units, ms level, polarity" is just incorrect thinking, ignore

pwiz/analysis/spectrum_processing/SpectrumList_Filter.cpp
bspratt: @chambm OK, it is now using that calibration term properly. The reader understands that a spectrum of type "calibration" may implement a mass spectrum in practice. So we now key on the presence of an MS level attribute instead of looking at the spectrum type. This properly supports UIMF mzML output which is the only known source of explicitly declared calibration spectra.

I backed out any change to our writer behavior as regards lockmass data, as that would probably cause trouble for other readers which presumably have already figured out how to deal with the situation.

pwiz/data/vendor_readers/Waters/Reader_Waters.cpp
bspratt: I'm just going to back out these complications.

chambm:
They already did the change for calibration spectrum from spectrum type to spectrum attribute. Should we utilize it? We don't have to update the CV even because we're not semantic validating.

bspratt:
Ah, I see you reported the problem, thanks. So they will keep the CVID but change the parenting? That seems likely to mess up some readers. Although I don't see any updated ontology yet so maybe it's still in discussion.

However it lands we will need to handle current and future versions of calibration data declaration. But for the moment I think this PR is good to go?

chambm:
Yes, same CVID, different parent. I doubt it would affect any readers but it is possible. They've already agreed to the change, but I guess it would require some more changes to the PR, so if you don't think it's worth bundling in this PR, that's ok.

pwiz/analysis/spectrum_processing/SpectrumList_Filter.cpp
```diff
@@ -318,12 +318,25 @@ PWIZ_API_DECL SpectrumList_FilterPredicate_MSLevelSet::SpectrumList_FilterPredic
 
 PWIZ_API_DECL boost::logic::tribool SpectrumList_FilterPredicate_MSLevelSet::accept(const msdata::Spectrum& spectrum) const
 {
-    CVParam param = spectrum.cvParamChild(MS_spectrum_type);
-    if (param.cvid == CVID_Unknown) return boost::logic::indeterminate;
-    if (!cvIsA(param.cvid, MS_mass_spectrum))
-        return msLevelSet_.contains(0); // non-MS spectra are considered ms level 0
-    param = spectrum.cvParam(MS_ms_level);
-    if (param.cvid == CVID_Unknown) return boost::logic::indeterminate;
+    if (!spectrum.hasCVParamChild(MS_spectrum_type))
```
chambm, line 321: Shouldn't this be reverted since it was only need for multiple spectrum types?

pwiz_tools/BiblioSpec/src/PwizReader.cpp
```diff
@@ -426,11 +428,28 @@ void PwizReader::addCharges(BiblioSpec::Spectrum& returnSpectrum,
...
+/**
+ * Put a spectrum's peaks in ascending m/z order if its writer did not. Ascending m/z is nowhere
+ * required of a writer, but everything downstream assumes it - and this has to happen here, on the
+ * SpectrumInfo, because it is the only point both transfer paths share. BlibBuild never builds a
+ * BiblioSpec::Spectrum at all: it goes BuildParser -> getSpectrum(int, SpecData&) -> transferSpec,
+ * and hands the raw arrays to insertPeaks, so a library built from a writer that presented some
+ * other order would be stored in that order and stay that way for every consumer of the .blib.
+ */
+void PwizReader::ensureMzAscending(SpectrumInfo& specInfo)
+{
```
chambm: This seems potentially expensive for an extremely rare case. How about we do it for the first spectrum and then assume the rest follow suit?

pwiz_tools/Skyline/TestData/PwizFileInfoTest.cs
```diff
@@ -102,11 +104,205 @@ public void TestTicChromatogram()
...
+            }
+        }
+
+        /// <summary>
+        /// Read the first m/z array straight out of the mzML text and report whether it ascends.
+        /// Deliberately does not go through <see cref="MsDataFileImpl"/>, which now sorts on read -
+        /// the point is to see the order the file is actually stored in. The fixture is written
+        /// uncompressed 64-bit precisely so this stays a few lines.
+        /// </summary>
+        private static bool IsMzArrayAscendingInFile(string path)
```
chambm: This seems a bit excessive. It's testing that the so-named unsorted file hasn't been sneakily sorted later?

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
```diff
@@ -2426,6 +2443,34 @@ public void SetArrays(double[] mzs, double[] intensities, double[] ionMobilities
...
+        /// binary searches the m/z axis, so on a spectrum ordered any other way the search lands
+        /// nowhere useful and the chromatogram comes out empty with no error at all. Writers that
+        /// present some other order do exist - one shipped peaks in ascending intensity - so the
+        /// order is checked rather than trusted.
+        /// The check is a single pass and costs nothing next to decoding the arrays; the sort only
+        /// runs for a writer that did not present m/z order.
+        /// </summary>
+        private void EnsureMzAscending()
+        {
+            if (Mzs == null || Intensities == null || Intensities.Length != Mzs.Length)
```
chambm: I don't imagine Brendan would be ok with a definitely slowdown in 99.9% of Skyline data import so that the 0.1% works. Again I think this needs to do the first spectrum and cache whether sorting is needed. It's possible to it like every 1000th spectrum instead, and if a later sort fails, assume the import is compromised and restart the import with the cached flag set to sort. More complicated but much more defensible for the general case.

chambm (changes requested):

pwiz_tools/Shared/ProteowizardWrapper/MsDataFileImpl.cs
bspratt: Fair - but maybe find the first one with more than, say, 10 m/z values in case the first scans are before the sample arrives?

You can probably make a case for just throwing an exception on unsorted m/z with a note to the user that the data needs to be reprocessed, but if we make itcheap enough to set the initial flag then no harm in being perfectly general, I guess.

Hmmm... actually sortedness only really matters to us in the context of binary search, maybe it's cheap to detect unsortedness during the search?

bspratt: Done in 5f88ea972, using your first suggestion plus the >10 m/z refinement.

`MzOrderVerdict` holds the per-file answer. The two verdicts are deliberately not symmetric:

* **Out of order -> condemned**, at any peak count. Two peaks in the wrong order is already proof the writer does not sort, so no size bar applies there; every spectrum in the file is sorted from then on.
* **In order -> trusted**, but only from a spectrum with more than 10 m/z values. Three peaks ascend by chance, which is exactly how a first-spectrum-only probe gets fooled. Until such a spectrum arrives the checking continues, so a short leading scan costs a few extra passes rather than the correctness of the whole file.

So the 99.9% case pays a handful of `IsSorted` passes over the first spectra and nothing after that. I did not take the sample-every-1000th-and-restart option: it only pays for itself on a file that changes its mind partway through, and the order is a property of the writer, not of the scan - none of the eight Waters files we hold mixes the two. Restart plumbing for a case never observed seemed the worse trade, but say the word if you would rather have it.

Not taken either: detecting unsortedness inside the binary search. It is free on the 99.9% path, but a binary search can land in a locally ordered neighbourhood and return a plausible wrong answer without ever stepping on the disorder, so the guarantee gets weaker rather than just cheaper.

Combined ion mobility spectra are excluded upstream of the verdict (`IonMobilities == null`), which matters more now than before - they are legitimately only m/z sorted within each mobility bin, so letting them vote would condemn every IMS file on its first spectrum.

pwiz_tools/BiblioSpec/src/PwizReader.cpp
bspratt: Same fix in 5f88ea972 - `PwizReader` now holds a `MzOrderVerdict` (`MzOrderVerdict.h`, mirroring the Skyline-side class) and resets it in `openFile`, so the question is settled from the first few spectra rather than re-asked for every one. Same asymmetry: out of order condemns the file at any peak count, in order is only believed from a spectrum with more than 10 peaks.

The state machine is unit tested in `SpectrumTest`, which is why it is a small class rather than a pair of member bools - `ensureMzAscending` itself needs a file to reach, but the rule it encodes is the part worth pinning, including that no later spectrum can talk a condemned file back round.

One caveat specific to this side: `openFile` sets `combineIonMobilitySpectra = true`, and a combined spectrum is only m/z ordered within each mobility bin, so it fails the check and condemns the file immediately. That is the same work Phase 9 already did and it is correct here - `SpectrumInfo::data` is a flat `vector<MZIntensityPair>` with no mobility axis to shred, and `binPeaks` merges only *adjacent* equal bins, so flattening to m/z order is what makes the binning right. The latch cannot spare that cost, but for IMS input the sort is doing real work rather than looking for something that is not there.

`Spectrum::setRawPeaks` is left checking every time: it serves the `.blib` re-read and copy paths, where there is no file being streamed and so nothing to form a verdict about.

pwiz_tools/Skyline/TestData/PwizFileInfoTest.cs
bspratt: Yes, exactly that - and it earns its keep, because every other assertion in the test passes on an already-sorted file. Without the pin, an msconvert round trip of the fixture would leave the test green while it covered nothing at all. A review pass caught that the first time round, which is why it is there.

It has grown a second job in 5f88ea972. The fixture now leads with a short spectrum that *is* in m/z order, so the test also pins the shape the >10-peak rule exists for: reorder the file so the unsorted spectrum comes first and the too-few-peaks-to-mean-anything rule stops being covered. Hence three assertions now - first ascending, the two after it not.

Mutation-verified in both directions: dropping the peak-count rule (settle on the first ordered spectrum, whatever its size) fails the test at `CollectionAssert.AreEqual ... spectrum 1`, with the second spectrum coming back in writer order.

Happy to drop it if you would still rather not have the file's own bytes asserted on, but it is the only thing standing between this test and silently covering nothing.

pwiz/analysis/spectrum_processing/SpectrumList_Filter.cpp
bspratt: I think this is a different change than the one you have in mind - the multiple-spectrum-types one *was* reverted, back in Phase 6 when the writing half of this PR was withdrawn. Nothing in pwiz emits two children of `spectrum type` any more, so the order-independence fix went out with it.

What is here now fixes a calibration-spectrum-**only** spectrum, which is what `SpectrumList_UIMF.cpp:106` writes today on master, with no help from this branch:

```cpp
CVID spectrumType = rawIndexEntry.frameType == FrameType_Calibration ? MS_calibration_spectrum : ...
result->set(MS_ms_level, msLevel);
result->set(spectrumType);
```

`MS_calibration_spectrum`'s only CV parent is `MS_spectrum_type` (`cv.cpp:7055`) - it is not under `MS_mass_spectrum`. So master's `!cvIsA(param.cvid, MS_mass_spectrum)` returned `msLevelSet_.contains(0)` and threw away a perfectly good declared ms level: `msconvert --filter "msLevel 1-"` drops UIMF calibration frames, and `msLevel 0` keeps them.

Worth noting that restoring the reverted hunk would *not* have fixed this - it asked `hasCVParamChild(MS_mass_spectrum)`, which is equally false for a calibration-only spectrum. It only ever helped when both terms were present, which was the situation our own additive writer created. Asking for the declared ms level first is the superset: it fixes the calibration case and gets order-independence for free, without needing an additive writer to exist.

It does become moot, but only once psi-ms #539 lands *and* UIMF is updated to write `MS1 spectrum` plus the attribute. Until then this is the only thing keeping a UIMF calibration frame's declared ms level from being discarded.

Mutation-verified: restoring the type-first ordering fails `testMSLevelSetCalibrationSpectrum` at `SpectrumList_FilterTest.cpp:561`, `expected "2" but got "1"`. Happy to pull it into its own PR if you would rather it not ride along with the Waters work - it is a pwiz change in a skyline-prefixed PR either way.

bspratt: oof, AI slop. I shouldn't let Claude handle the responses directly

chambm: This is going to break with the new CV, when calibration spectrum is no longer a spectrum type. (Also the comment is wrong here about multiple spectrum types.) Fixing UIMF in this PR is the right path I think, which doesn't fix existing UIMF files, but we could fix those broken UIMF files in the mzML reader. Which I'm inclined say we should do so since they are OUR broken files. In untested code:
if (spectrum.hasCVParam(MS_calibration_spectrum) && !spectrum.hasCVParamChild(MS_spectrum_type))
{
int msLevel = spectrum.cvParamValueOrDefault(MS_ms_level, 0);
if (msLevel > 0) spectrum.set(msLevel == 1 ? MS_MS1_spectrum : MS_MSn_spectrum);
}

chambm: Although I guess it's not just mzML. Someone could have converted to any of the open mzML formats like mz5, mzMLb, perhaps even mzXML. Tricky. Maybe a centralized workaround location (I don't think we've had a case like this before) and then call it in mzML, mz5, and mzMLb right before they return a spectrum.

chambm: Which is possibly more effort than the UIMF format warrants, as I haven't heard of anybody using it recently (a similar fate to mz5 and mzMLb, sigh). I'm sure multi-parquet-in-a-zip mzPeak will change things though. /s

We could put some opt-in telemetry into pwiz that tracks what formats people use, input and output.

bspratt:
As I think about this more, I see a philosophical issue - the current proposed fix takes care to deal with the m/z sort issue where we care in Skyline and BiblioSpec, but running such data through msconvert faithfully preserves the unsortedness.

It seems more useful to the community to deal with this at a lower level such that anything at all using pwiz is spared the m/z sort issue - so msconvert, mscat etc would all emit sorted m/z. But it's arguably a data integrity issue silently rejiggering things like that.

I doubt anyone out there really wants to preserve funky m/z order but it's worth careful consideration.

bspratt:
**Split: the m/z sort work has moved out of this PR.**

Brian's call, and it makes both stories reviewable on their own - the DATA Convert 5.x peak-order regression has nothing to do with lockmass/calibration scans beyond having been reported by the same customer in the same week.

This PR is now the calibration/waters_connect story only: 14 files, no BiblioSpec, no `CommonUtil`, no `ArrayUtil` move. `591445b58` reverts the two sort commits; the body above has been trimmed to match.

The sort work is now on `Skyline/work/20260804_mzml_mz_sort_order`, and its PR will also carry a **general pwiz-level fix**, which is the more important outcome of your review. Two things prompted that:

1. Your perf objection is addressed - the check is settled per file, not per spectrum. Details are on the threads below, and they travel with the code to the new PR.
2. Measuring for that reply turned up something worse than we thought. **msconvert passes the bad order straight through**, byte for byte, no filters involved. So converting one of these files produces another broken file, and every consumer downstream breaks the same silent way - not just Skyline. That undercuts the "fix only in code we control" scope note that used to be in this description, which you'd already questioned.

Filters repair it only incidentally and only when they happen to do work: `mzWindow` fixes it, `threshold count 3` fixes it, `threshold count 100` does **not** - `ThresholdFilter::operator()` returns early at line 264 when the count exceeds the peak count, before reaching the m/z re-sort at line 423. `sortByScanTime` reorders spectra, not peaks; nothing in pwiz reorders peaks within a spectrum today.

So the new PR proposes correcting on read at the `ReaderList::read` seam, with a warning when it fires. Happy to argue that one there rather than here.

**Three threads below now discuss code that is no longer in this PR** - `MsDataFileImpl.cs:2471`, `PwizReader.cpp:440` and `PwizFileInfoTest.cs:206`. Replies are already posted on each; they carry over to the new PR unchanged. The fourth, on `SpectrumList_Filter.cpp:321`, stays here - it is the Phase 8 msLevel/calibration change and belongs with this story.

bspratt:
UIMF now skips calibration frames at index time (also the TIC and the  retention-time map, which were still reaching them), and mzML/mzMLb/mz5 go through a new SpectrumList_IgnoreCalibrationScans wrapper in pwiz/data/msdata/. It is gated on fileContent declaring MS:1000928, because Skyline sets ignoreCalibrationScans for every file it opens and an ungated version would put a full metadata pass on every mzML import; a file that does not declare the term comes back untouched. Not wired for mzXML — no CV term to  carry the label.

On the msLevel reorder you questioned: pwiz-sharp's MsLevelPredicate (Analysis/Predicates.cs:119) already reads the   declared ms level and nothing else, with no spectrum-type gate. So this moves the C++ toward what your port does, and   costs the port nothing.

 One gap, not covered: the UIMF skip itself. BSA_10ugml_CID.UIMF holds only frameType 1 and 2, so no test reaches the new branch.

pwiz/analysis/spectrum_processing/SpectrumList_FilterTest.cpp
```diff
@@ -519,9 +519,16 @@ void testMSLevelSet(SpectrumListPtr sl)
...
+// scan as both an MS1 spectrum and a calibration spectrum. Both declare an ms level, so both are
+// filtered on the level they declare rather than being taken for level 0. Dropping calibration data
+// altogether is Reader::Config::ignoreCalibrationScans' job, not this filter's.
+//
+// Because a declared level is read before the type is consulted at all, these assertions hold
+// however the CV happens to attach "calibration spectrum" to "spectrum type" - is_a here, has-a in
+// a later release, which cvIsA would not follow.
+//
 // Uses its own list, since the shared one is pinned by exact sizes and ids throughout this file.
 void testMSLevelSetCalibrationSpectrum()
```
chambm: I think we can drop this test entirely as it's not this filter's job to do anything with calibration spectra anymore.

pwiz/data/msdata/SpectrumList_IgnoreCalibrationScans.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+PWIZ_API_DECL size_t SpectrumList_IgnoreCalibrationScans::findAbbreviated(const string& abbreviatedId, char delimiter) const
+{
+    return fromInnerIndex(inner_->findAbbreviated(abbreviatedId, delimiter));
+}
+
+
+PWIZ_API_DECL bool SpectrumList_IgnoreCalibrationScans::calibrationSpectraAreOmitted() const
+{
+    return indexMap_.size() != inner_->size() || inner_->calibrationSpectraAreOmitted();
```
chambm: Seems like this would return false if the file happened to have no calibration spectra, but it should be true because if the file DID have calibration spectra, they would have been dropped. So it should just always return true, right?

pwiz/data/msdata/SpectrumList_IgnoreCalibrationScans.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+
+SpectrumList_IgnoreCalibrationScans::SpectrumList_IgnoreCalibrationScans(const SpectrumListPtr& inner)
+:   SpectrumListWrapper(inner)
+{
+    size_t innerSize = inner_->size();
+    indexMap_.reserve(innerSize);
+    identities_.reserve(innerSize);
+
+    for (size_t i = 0; i < innerSize; ++i)
```
chambm: There's a utility function that you can give a lambda function which will return the lowest detail level that populates a given CV term. Definitely don't want to use full metadata here if instant metadata will work.

pwiz_tools/commandline/msconvert.cpp
```diff
@@ -446,7 +446,7 @@ Config parseCommandLine(int argc, char** argv)
             ": if supported by vendor; combine MS2 spectra referring to the same precursor and survey scan, and calculate accurate precursor masses.")
         ("ignoreCalibrationScans",
             po::value<bool>(&config.ignoreCalibrationScans)->zero_tokens(),
-            ": do not process calibration scans (currently only applies to Waters lockmass function)")
+            ": do not process calibration scans (Waters lockmass function, UIMF calibration frames, and spectra labeled MS:1000928 in mzML/mzMLb/mz5)")
```
chambm: I suggest nixing "in mzML/mzMLb/mz5" and including the term name "calibration spectrum".

chambm (changes requested):

Issue #4496, opened 2026-07-29
Osprey: read vendor raw directly via pwiz_data_cli in the net472 configuration - drop the msconvert step


brendanx67:
## Summary

Read vendor raw files directly in Osprey's **net472** configuration by integrating
`pwiz_data_cli.dll` - the same build Skyline uses on master - so the test and development
workflow drops its msconvert step entirely.

**Near-term goal:** run `--task PerFileScoring` with `.raw` files as input, producing the
`.spectra.bin` caches and the Pass-1 `.scores.parquet` directly. Nothing downstream changes.

## Why now

Today Osprey reads mzML only (`Osprey.IO` has exactly one reader,
`MzmlReader.cs`, 874 lines, called at `PerFileScoringTask.cs:2709`). Every large dataset
therefore costs a three-step staging dance:

1. download `.raw`
2. **msconvert to mzML** - hours per dataset, and roughly 0.73x the raw size again on disk
3. run Osprey to build `.spectra.bin`, then delete the `.raw` to reclaim space

Step 2 and the subsequent delete are pure overhead. Measured on the 82-file SEA-AD Astral
set: 484 GB raw -> 324.5 GB mzML -> 345.9 GB caches. The 164-file TDP-43 set now staged
locally is 784 GB of raw awaiting the same treatment.

Removing step 2 matters more than the disk: **we are iterating heavily on the FDR tasks**,
and every new dataset currently costs a conversion pass before the first Osprey run. Getting
a dataset to Pass-1 parquet as fast as possible is what enables cyclic development on the
later stages.

## Scope

* Add a vendor-raw reader to `Osprey.IO` behind the same contract `MzmlReader` provides -
  `LoadAllSpectra(path) -> MzmlResult { Ms2Spectra, Ms1Spectra, UnsortedSpectrumCount }` -
  so `PerFileScoringTask` selects by extension and nothing downstream is aware.
* Wire `pwiz_data_cli.dll` into the **net472** configuration only.
  `pwiz_tools/Shared/ProteowizardWrapper` already references it
  (`ProteowizardWrapper.csproj:91`) and is itself `v4.7.2`, so it is the natural integration
  point rather than binding the CLI assembly directly.
* Normal PR against `master`. Mergeable and releasable through the existing path.

### The interface surface is small - `ProteowizardWrapper` already covers all of it

Osprey's requirement is fully determined by what `SpectraCache` serializes (v4 format,
[...]

PR #4500, opened 2026-07-29
pwiz: Preserved full double precision in cvParam text

brendanx67:
## Summary

**Experimental / not ready for review** - posted to get TeamCity coverage and for information. See the note to Matt at the bottom; declining this in favour of #4178 is a reasonable outcome.

* `pwiz::util::toString(double)` truncates every double cvParam to **12 fractional digits** (`double12_policy`, `String.cpp`), so a Thermo scan start time of `1.8119944333330003` minutes is written as `1.811994433333` and reloads one ULP away. An mzML therefore cannot reproduce the value read from the raw file.
* The 12-digit policy exists to avoid `lexical_cast` noise like `123.00000000007`. Shortest round-trip formatting satisfies that aim and is also exact, so this keeps the karma output whenever it already reloads bit-exact and escalates only where it would lose information. **Existing output and file sizes are unchanged except where today's output is silently lossy** (+0.09% on a 2.41 GB Astral file).
* Deliberately not `std::to_chars`: floating-point `to_chars` is C++17 but library support is not universal across the toolsets pwiz builds with (`Jamroot.jam:393-396` targets msvc, gcc, darwin and clang; libstdc++ needs GCC 11). Deliberately not `snprintf` either: that follows the global locale and would emit a comma decimal point into XML under e.g. `de_DE`. Uses a classic-locale `ostringstream`, which is portable and locale-safe, on a path reached only for the rare values the fast path cannot represent.
* New `StringTest` asserts the property directly: everything `toString` writes must parse back equal, over the motivating Thermo retention times, range extremes, and 20,000 generated values across 40 decades - plus that ordinary values keep their exact existing text.

## Test plan

- [x] `StringTest` (new) - passes
- [x] `pwiz/data/msdata` - `Serializer_mzML_Test`, `IOTest`, `DiffTest`, `MSDataTest`, `SpectrumList_mzML_Test`, `ChromatogramList_mzML_Test`, `Serializer_mzXML_Test` all pass
- [x] Vendor readers - Thermo, Waters, Agilent, ABI, UNIFI, Shimadzu, UIMF, Mobilion all pass
- [x] No baseline regeneration needed: `VendorReaderTestHarness.cpp:382` compares parsed `MSData` via `Diff<MSData, DiffConfig>` with a precision tolerance, not mzML text, so extra digits do not disturb baselines
- [ ] `ReaderTest` and 4 of 52 `Reader_Bruker_Test` cases fail locally, but **identically on master**: both are `--without-compassxtract` artifacts (`Bruker API was built with only BAF and TDF support; YEP and FID files not supported`). Left for TeamCity to confirm on a full build.

## Note for @chambm

This is informational, not a request. Two things worth knowing:

1. **Nothing of ours is blocked on it.** Our downstream work reads mzML correctly without this fix, verified byte-for-byte on three datasets whose mzML was written by unpatched msconvert.
2. **#4178 may make it moot.** If ProteoWizard core moves to .NET 8, the conversion becomes .NET's own, which has been correctly-rounded shortest-round-trip since .NET Core 3.0, and the 12-digit policy disappears on its own.

One consequence if it does land: every mzML already archived was written by unpatched msconvert, so reading those raw files directly would no longer byte-match the stored mzML. That is a one-time discontinuity to land deliberately, not an argument against.

Related: #4496

### DIFF
--- a/pwiz/utility/misc/Jamfile.jam
+++ b/pwiz/utility/misc/Jamfile.jam
@@ -122,6 +122,7 @@ unit-test-if-exists almost_equal_test : almost_equal_test.cpp pwiz_utility_misc 
 unit-test-if-exists endian_test : endian_test.cpp pwiz_utility_misc Std ;
 unit-test-if-exists BinaryDataTestNative : BinaryDataTest.cpp pwiz_utility_misc Std ;
 unit-test-if-exists Base64Test : Base64Test.cpp pwiz_utility_misc Std ;
+unit-test-if-exists StringTest : StringTest.cpp pwiz_utility_misc Std ;
 unit-test-if-exists IntegerSetTest : IntegerSetTest.cpp pwiz_utility_misc Std ;
 unit-test-if-exists IterationListenerTest : IterationListenerTest.cpp pwiz_utility_misc Std ;
 unit-test-if-exists DateTimeTest : DateTimeTest.cpp pwiz_utility_misc Std ;
--- a/pwiz/utility/misc/String.cpp
+++ b/pwiz/utility/misc/String.cpp
@@ -21,6 +21,9 @@
 #include "String.hpp"
 #include <limits>
 #include <algorithm>
+#include <iomanip>
+#include <locale>
+#include <sstream>
 #include <boost/spirit/include/karma.hpp>
 
 using boost::spirit::karma::real_policies;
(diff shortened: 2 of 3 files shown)


chambm:
I agree this may be obsolete by the .NET8 port, but since we switched to Karma in the first place for speed I was suspicious of the performance impact of this change, so I had Claude profile with and without. It's a 1.5% hit on conversion time. So I'm curious about the motivation for this. Did a scan time difference of a single ULP actually affect Osprey's output, or was it just a test comparing the raw scan time output from directly reading RAW vs. mzML?

brendanx67:
Closing this rather than shipping or narrowing it. The analysis is worth keeping, so here is the reasoning.

## What the change buys, measured

The values that could reach an Osprey `.spectra.bin` through `toString` are only the scalar cvParams a vendor reader constructs - scan start time, isolation window target and offsets, selected ion m/z. The binary m/z and intensity arrays never touch it, and `toString` is never on the mzML **read** path at all: `IO.cpp:208` copies the `value` attribute verbatim as a string. So the narrowest useful version of this PR is "apply round-tripping to `MS_scan_start_time`".

At 12 fractional digits on a value in minutes, the truncation is about 60 picoseconds. Nothing downstream resolves it. The reason it looked significant was bit parity between the raw and mzML paths - and both paths truncate identically, so parity holds without this change.

That is now measured rather than inferred: a raw-vs-mzML `.spectra.bin` comparison run against a `pwiz_data_cli` built deliberately WITHOUT this change returned `PARITY: 12,784 bytes identical` on the committed Thermo fixture. master is self-consistent as it stands.

|  | with this PR | without |
|---|---|---|
| raw-vs-mzML parity | proven | **proven** |
| Skyline saved documents | breaks a stored `equals` filter operand | fine |
| msconvert comparisons (Wine) | 7 failing | green |
| precision gained downstream | ~60 ps of RT | - |

## Two things that outlive this PR

**The truncation is a genuine fidelity limitation, not a non-issue.** `value="0.5903117"` in an mzML is what pwiz stored, not what the instrument recorded, so an mzML cannot reproduce the retention time of the raw it was converted from. This is closed as "not worth the blast radius now", not as "not a defect". `ParamContainer::set` routes every double cvParam through `toString`, so any fix has to be evaluated against every consumer of pwiz - a wide review for precision no current consumer measures.

**The Agilent finding is the durable part**, and it is a Skyline fragility rather than a pwiz one: a saved document with a spectrum-class filter doing `equals` on a double operand captured from pwiz output breaks if that output ever gains digits. Filed as #4506 so it is not rediscovered the next time someone touches precision anywhere in this chain.

## Preserved, not discarded

The branch `Skyline/work/20260729_pwiz_tostring_roundtrip` stays on the remote, including `pwiz/utility/misc/StringTest.cpp` - a property test asserting that everything `toString` writes reloads bit-exact, over the motivating Thermo retention times, range extremes and 20,000 generated values. If full-precision cvParams are ever wanted, the work and its test are a checkout away.

brendanx67:
Indeed. It was just comparing .raw with .mzML that brought it up. In the end other fixes were able to mitigate the issue. So, even for Osprey round-trip testing this is no longer necessary.

PR #4551, opened 2026-08-08
pwiz: Fixed off-by-one in spectrum and chromatogram list index bounds checks

bspratt:
## Summary

* Index bounds guards for several varieties of SpectrumList and ChromatogramList tested `index > size` instead of `>=`, but of course index[size()] is just as wrong as index[size()+1]. In practice this hasn't been a runtime issue (that we have noticed?), but code inspection is right in flagging it.  Seen in 29 places across 18 files.
* Completes 18e8ba470 (2009), which applied this same fix, but to only four instances.
* MGF and MSn `spectrumIdentity()` had no guard at all — added.

While we were in there, I decided to normalize line endings for the files we were already touching - tired of being warned about that.  Review with `git diff --ignore-cr-at-eol`; the logic change is 31 one-character edits.

## Test plan

- [x] SpectrumList_mzXML_Test, _MGF_Test, _MSn_Test, ProteinList_DecoyGeneratorTest - new out-of-bounds assertions, verified red before the fix
- [x] pwiz/data and pwiz/analysis suites - green

### DIFF
--- a/pwiz/analysis/proteome_processing/ProteinList_DecoyGenerator.cpp
+++ b/pwiz/analysis/proteome_processing/ProteinList_DecoyGenerator.cpp
@@ -82,7 +82,7 @@ PWIZ_API_DECL size_t ProteinList_DecoyGenerator::find(const string& id) const
 
 PWIZ_API_DECL ProteinPtr ProteinList_DecoyGenerator::protein(size_t index, bool getSequence) const
 {
-    if (index > size())
+    if (index >= size())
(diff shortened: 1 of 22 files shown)


chambm (approved):
Funny how long this has been around. What triggered you to discover it?

bspratt:
> Funny how long this has been around. What triggered you to discover it?

It came up in a Claude "/code-review max" while I was looking into handling unordered m/z values.

PR #4552, opened 2026-08-08
pwiz: Corrected disordered m/z arrays on read

bspratt:
## Summary

Waters DATA Convert 5.0.0.2900 writes each spectrum''s peak arrays sorted by ascending
**intensity**. Extraction binary searches the m/z axis, so on such a file the search lands at an
arbitrary index and nothing accumulates: `TotalArea=0` at every point, no error, no warning. In
metabolomics that reads as "compound not detected". 4.0.0.2619 is correct, so it is a converter
regression. Ascending m/z is nowhere in the mzML spec - it is a de facto standard everything
assumes. msconvert propagated the bad order byte for byte, so converting one of these files
produced another broken file.

* Sorted a spectrum''s peaks by m/z in `SpectrumListBase::ensureMzAscending()` when its writer did
  not, called by each list that reads a format some other tool wrote - mzML, mzXML, MGF, MSn, mz5,
  BTDX
* Carried every array holding one value per peak through the same permutation, identified by
  length the way `ThresholdFilter::getExtraArrays` does it, `integerDataArrayPtrs` included -
  reordering m/z and intensity while leaving a third array behind would put every value against
  the wrong peak, which is worse than the disorder being repaired
* Left spectra ordered on some other axis exactly as written, identified by CV term: combined ion
  mobility and scanning quadrupole position, where m/z ascends only within each block, and
  wavelength, since `Spectrum::getMZArray()` returns a diode-array trace too

Why `SpectrumListBase` rather than a wrapper over `run.spectrumListPtr`: a wrapper is invisible
until something reaches past it. `SpectrumList_PeakPicker`, `SpectrumList_LockmassRefiner`,
`SpectrumList_3D` and `SpectrumList_PrecursorRefine` all `dynamic_cast` the outermost list to a
concrete vendor list, and `SpectrumListWrapper` does not forward `calibrationSpectraAreOmitted()`
or the `spectrum(seed, bool)` overload. Every one of those degrades silently. The readers already
share a base class with a single funnel each, so the correction goes there instead and nothing is
interposed.

The vendor lists do not call it - peaks arriving through a vendor API are already ascending. Note
that Skyline still sorts combined ion mobility spectra itself, in
`SpectraChromDataProvider.EnsureSortedMzs` and `ScanProvider`; those are not redundant with this,
they cover exactly the case pwiz must leave alone.

Reported by Hans.

## Test plan

- [x] `SpectrumListBaseTest` - 9 cases: reordering with intensities still paired, sorted input untouched, a short ordered leader not settling the file, a condemned file staying condemned, combined IMS left alone, combined IMS not vouching for the writer, a wavelength trace neither sorted nor treated as evidence, extra per-peak arrays travelling with their peaks while a non-per-peak array is left alone, and a metadata-only spectrum settling nothing
- [x] `SpectrumList_mzML_Test` - writes a disordered mzML and reads it back through the real reader, indexed and non-indexed. Mutation-verified: removing the call from `SpectrumList_mzML.cpp` fails it
- [x] `TestUnsortedMzArrays` - Skyline side, through `pwiz_data_cli`, with a fixture whose stored order is asserted first so it cannot be silently regenerated in m/z order
- [x] Whole `pwiz/data/msdata` and `pwiz/analysis/spectrum_processing` suites
- [x] End to end: the customer-shaped fixture through msconvert goes from 3 and 2 m/z descents to 0, filtered and unfiltered, and prints a warning naming the first offending spectrum
- [x] Clean build and test


pwiz/data/msdata/SpectrumListBase.cpp
```diff
@@ -42,6 +89,96 @@ PWIZ_API_DECL void pwiz::msdata::ListBase::warn_once(const char * msg) const
...
+        // Seems fine - but a short list can be in order by chance, so it does not settle anything
+        MzOrderVerdict expected = MzOrderVerdict::unsettled;
+        if (mzs.size() > MIN_PEAK_COUNT_FOR_MZ_SORT_CHECK)
+            mzOrderVerdict_.compare_exchange_strong(expected, MzOrderVerdict::writerSortsByMz);
+        return;
+    }
+
+    // One spectrum out of order condemns the file, and always wins: a writer cannot be talked back
+    // into good standing by a later spectrum that happens to ascend. The exchange also tells us
+    // whether this is the first such spectrum, which is what the warning below is keyed on.
```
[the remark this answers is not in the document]
bspratt: Good catch - the comment was accurate only against a later ascending spectrum, not absolutely. Once the verdict latches to writerSortsByMz the function returns at the top, so a later out-of-order spectrum is never examined.

Reworded in 53a134f8b to state both halves, and to name the trade-off explicitly: condemnation is permanent, but a writer that sorts some spectra and not others is only corrected up to the first long ascending one. That is the deliberate cost of settling the question once per file rather than re-asking it for every spectrum; the converter this was written for got every spectrum wrong.

pwiz/utility/misc/VendorReaderTestHarness.cpp
```diff
@@ -281,6 +281,46 @@ string headStream(istream& is, size_t maxLength)
...
+    for (size_t i = 0, end = msd.run.spectrumListPtr->size(); i < end; ++i)
+    {
+        SpectrumPtr s = msd.run.spectrumListPtr->spectrum(i, true);
+        if (!s.get() || !s->getMZArray().get() || hasNonMzOrderingAxis(*s))
+            continue;
+        const BinaryData<double>& mzs = s->getMZArray()->data;
+        if (!std::is_sorted(mzs.begin(), mzs.end()))
+            throw runtime_error(unit_assert_message(__FILE__, __LINE__,
+                (format + " spectrum " + lexical_cast<string>(i) + " (\"" + s->id +
+                 "\") is not in ascending m/z order").c_str()));
```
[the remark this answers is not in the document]
bspratt: Fixed in 873a128c3 - added an explicit `#include <algorithm>`.

Confirmed the concern was real rather than theoretical: nothing in this file's include list pulls in `<algorithm>` directly, and `pwiz/utility/misc/Std.hpp` does not include it either, so `std::is_sorted` was resolving purely through a transitive include that happens to be there on this toolchain.

pwiz_tools/Skyline/TestData/PwizFileInfoTest.cs
```diff
@@ -108,6 +110,126 @@ public void TestTicChromatogram()
...
+                case 1:
+                    byMz = new Dictionary<double, double> { { 500.5, 10 }, { 300.3, 20 }, { 700.7, 30 }, { 200.2, 40 }, { 600.6, 50 }, { 400.4, 60 } };
+                    break;
+                default:
+                    byMz = new Dictionary<double, double> { { 400.4, 15 }, { 700.7, 25 }, { 200.2, 35 }, { 600.6, 45 }, { 300.3, 55 }, { 500.5, 65 } };
+                    break;
+            }
+            return byMz[mz];
         /// The open dialogs match source types by string equality, so the type names in
         /// <see cref="DataSourceUtil"/> have to agree with the reader's own names wherever it
```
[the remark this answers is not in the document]
bspratt: Fixed in 873a128c3 - restored the closing brace and the `/// <summary>` opening tag.

Worth recording where this came from, since it was not in the original change: it was a merge artifact. `git blame` puts the end of `ExpectedIntensity` at this branch's 3271d39c7 and the following doc comment at master's 656bd2b28, which added an unrelated test method at that same point in the file. The line-based merge interleaved the two additions without producing a conflict, so it merged "cleanly" into code that does not parse.

You were right that it should have been a compile error - and it was. The remote tip genuinely did not build the Skyline TestData project until this fix went up. Verified after the fix with a direct `TestData` project build.

I checked the blast radius rather than assuming it was isolated: diffing both merge parents against their common ancestor, this file is the only one touched on both sides (18 files changed on the branch, 68 on master's, exactly one in the intersection), so no other file could have been damaged this way.

bspratt:
Following up on the three suppressed comments from the Copilot review, since two of them are worth an explicit answer.

**Tab and trailing whitespace in the `ensureMzAscending` comment block (`SpectrumListBase.cpp`) - fixed** in 746fb3e92. I initially dismissed this as pre-existing and was wrong: `git blame` puts that line at 24a87014f5, a commit on this branch, so the branch had in fact introduced a tab. Verified afterwards that no tabs or trailing whitespace remain in any file this branch touches.

**Matching CV param `name` strings rather than accessions in `PwizFileInfoTest.cs` - not changing.** The stated reason was locale-independence, and that premise does not hold for mzML: `IO.cpp:176` writes the attribute as `cvTermInfo(cvParam.cvid).name`, which comes from the compiled-in PSI-MS CV table. There is no localization path, so `name="m/z array"` is a fixed ASCII string in every mzML pwiz writes, regardless of locale.

The residual risk is different from the one raised - a future PSI-MS CV rewording, not a locale - and it is benign here: `IsMzArrayAscendingInFile` asserts on `arrayAt > 0`, so a renamed term makes the test fail loudly rather than silently pass. That is the right failure direction for a fixture-shape guard, so accessions would trade readability for no real safety.

bspratt:
@chambm can I get the blessing? Waters would like to have this.

pwiz/data/msdata/SpectrumListBase.hpp
```diff
@@ -80,6 +118,14 @@ class PWIZ_API_DECL SpectrumListBase : public SpectrumList
...
     ListBase impl_;
     size_t spectrum_id_mismatch_hash_;
+
+    /// what this file has shown so far about the way its writer orders peaks
+    enum class MzOrderVerdict { unsettled, writerSortsByMz, writerDoesNotSortByMz };
+
+    // mutable and atomic because spectrum() is const and is called from worker threads;
+    // condemnation is an unconditional store while the good verdict is a compare-exchange from
+    // unsettled, so no late-arriving good spectrum can un-condemn a file
+    mutable std::atomic<MzOrderVerdict> mzOrderVerdict_;
```
chambm: Long ago I decided to use Boost.Atomic instead of Std.Atomic because of a reason I can't remember. :( Might have something to do with C++/CLI or Wine. Let's stick with it for consistency (although all C++ is going away soon anyway).

pwiz/data/msdata/SpectrumListBase.cpp
```diff
@@ -42,6 +106,93 @@ PWIZ_API_DECL void pwiz::msdata::ListBase::warn_once(const char * msg) const
...
+
+    // Nothing to reorder, and nothing to learn, until the binary data is actually here. Tested
+    // first because a metadata-only read still carries the array objects with their cvParams - IO
+    // builds those and skips only the base64 decode - so without this every metadata pass would pay
+    // for the scans below and, having no data, could never settle the verdict to stop paying.
+    // Sizes come off the arrays rather than from defaultArrayLength, which is only guaranteed from
+    // DetailLevel_FullMetadata up.
+    bool hasPeakData = false;
+    for (size_t i = 0; i < spectrum->binaryDataArrayPtrs.size() && !hasPeakData; ++i)
+        hasPeakData = spectrum->binaryDataArrayPtrs[i].get() &&
```
chambm: Couldn't this check specifically for mzArray rather than iterating through all arrays?

pwiz/data/msdata/SpectrumListBase.cpp
```diff
@@ -42,6 +106,93 @@ PWIZ_API_DECL void pwiz::msdata::ListBase::warn_once(const char * msg) const
...
+    // The exchange also tells us whether this is the first such spectrum, which the warning below
+    // is keyed on.
+    bool isFirstFoundSpectrumOutOfOrder =
+        mzOrderVerdict_.exchange(MzOrderVerdict::writerDoesNotSortByMz) != MzOrderVerdict::writerDoesNotSortByMz;
+
+    // A spectrum may carry other values like signal-to-noise, baseline, resolution or charge array
+    // alongside m/z and intensity. Make sure those other arrays are permuted in the same way as m/z
+    // and intensity.
+    // Stable, so peaks sharing an m/z keep the order the writer gave them.
+    std::vector<size_t> order(mzs.size());
```
chambm: How about using pwiz::util::sort_together() here? As written though it might only work for binaryDataArrayPtrs or integerDataArrayPtrs: I'm not sure the template would allow passing both together. Fairly easy to fix though by adding a second container of arrays to co-sort. I don't think it has the exception safety this implementation has, but I'm not sure how necessary that is, either. And if we do care about the exception safety then it should be ported into sort_together anyway, as that one is a lot more likely to be used.

pwiz/data/msdata/SpectrumListBaseTest.cpp
```diff
@@ -50,13 +50,388 @@ void test()
...
+
+    unit_assert(mzsOf(list.spectrum(0, true)) == imsMzs);
+    unit_assert_equal(200.2, mzsOf(list.spectrum(1, true))[0], 1e-9);
+}
+
+
+// Spectrum::getMZArray() returns a wavelength array too, so a diode-array trace would otherwise be
+// judged as if its wavelength axis were m/z - and an ascending UV trace with enough points would
+// settle the verdict and disable the repair for every real spectrum in the file.
+void testWavelengthSpectrumIsNeitherSortedNorEvidence()
```
chambm, line 241: Actually wavelength spectra should probably be assumed to be sorted by wavelength. Not worth bothering about though.

pwiz/data/msdata/SpectrumListBase.cpp
bspratt: Good call - done. Asking `getMZArray()` directly replaces the loop entirely, and since that lookup was needed a few lines further down anyway, a metadata-only pass now costs one array lookup instead of the loop plus that lookup. The `< 2` guard folds in at the same time, so the later size check is just the m/z-vs-intensity length sanity check now.

One ordering detail worth noting: `getMZArray()` also returns a wavelength array for a diode-array trace, so it still has to run ahead of `hasNonMzOrderingAxis` rather than being merged into it - a UV trace reaches the lookup and is then correctly refused by the wavelength check.

`SpectrumListBaseTest` still green.

pwiz/data/msdata/SpectrumListBase.hpp
bspratt: Happy to switch, but before I do - I checked the tree and it looks like the convention has drifted the other way, so switching may cost consistency rather than buy it:

- `boost::atomic` / `boost/atomic.hpp`: no uses anywhere under `pwiz/` or `pwiz_aux/` (the Boost header is vendored, just unused).
- `std::atomic`: already in use, and in the same shape as here - `pwiz/analysis/dia_umpire/DiaUmpire.cpp` has `mutable std::atomic<bool> canceled_` and `mutable std::atomic<size_t> windowsProcessed`, plus several `std::atomic<size_t>` locals.

So as things stand `std::atomic` is the existing precedent and this would be the only `boost::atomic` in the codebase. If the original reason was C++/CLI or Wine it would presumably have bitten DiaUmpire too, though I have not tried to confirm that either way.

Your call - say the word and I will change it. I just did not want to make it a one-off in the other direction without flagging what is already there.

pwiz/data/msdata/SpectrumListBase.cpp
bspratt: I looked at `sort_together` for exactly this and hit the limit you suspected. `sort_together` declares its scratch space as `std::vector<ContainerT> tmpValuesRanges` (sort_together.hpp:52), so every co-sorted range has to be the same type as the sort container. Here the permutation has to cover both `BinaryData<double>` (m/z, intensity, and any per-peak doubles like signal-to-noise or baseline) and `BinaryData<IntegerDataArray::value_type>` for the charge array, which lives in `integerDataArrayPtrs` as a separate member of `Spectrum`. Two calls does not substitute, since the second would be sorting on an already-sorted key and would just apply the identity.

So it needs the extension you describe - a second container-of-arrays parameter - which I did not want to make unilaterally in a shared utility that already has callers (`SpectrumList_MZWindow.cpp:56-63`).

On exception safety, I do not think it is actually a differentiator: `sort_together` fills every `tmpValuesRanges` entry in the gather loop before it touches anything, then swaps, so a throw part way through cannot leave one array permuted against another - same property this has. The reason it matters here is specific: if m/z ended up sorted while intensity did not, the m/z axis would then ascend, this function would take the "seems fine" branch forever after, and nothing downstream could detect the mispairing.

Happy either way - extend `sort_together` and use it here, or leave it and do the extension separately. Given your note that the C++ is on its way out, my instinct is the latter, but it is your call.

pwiz/data/msdata/SpectrumListBaseTest.cpp
bspratt: Agreed, and noting for the record that the code already behaves that way - a wavelength array is exempted by `hasNonMzOrderingAxis`, so a diode-array trace is left exactly as written and is not treated as evidence about the writer either way. The test fixture is deliberately descending only to prove the guard fires rather than to suggest anything about how such data is ordered in practice.

Not changing anything here.

pwiz/data/msdata/SpectrumListBase.hpp
bspratt: ugh, that was Claude, not me. It looks like it is used in Bumbershoot, but not elsewhere.

pwiz/data/msdata/SpectrumListBase.cpp
bspratt: claude again, but at least not wrong

pwiz/data/msdata/SpectrumListBase.cpp
bspratt: Claude again. But yes, better to improve the existing sort_together.

pwiz/data/msdata/SpectrumListBaseTest.cpp
bspratt: Claude again. I've asked it not to run ahead of me like this any more.

chambm: > Claude again. I've asked it not to run ahead of me like this any more.

Haha, been there. It's hard to know how proactive Claude is going to be when you ask it to handle some PR feedback.

pwiz/data/msdata/SpectrumListBase.cpp
chambm: Agreed. Even though C++ will go away soon, might as well keep it tidy.

chambm (approved):

PR #4574, opened 2026-08-13
Proteo wizard master

kpnovoselov:


chambm:
Sorry you(r AI) did all this work. :) We're not switching to CMake as a build system. But we're also moving away from Boost.Build, because I've been working with Claude to port all of our C++ code to C#. That's #4178 . I think that will actually involve a few CMake subprojects to make P/Invokable C shims for C++ dependencies (Mascot DAT reader, Mobilion MBI reader), but no vcpkg.

Can you separate all these CMake/vcpkg changes from the imzML ones? I'm happy to look at a contribution that adds imzML support.

kpnovoselov:
Hello Matt,

I made a mistake and selected your repository as a pull request target
instead of my fork. Sorry about that.

 We have a C++ application which uses the ProteoWizard for uniform access
to raw files from different MS vendors and this app uses vcpkg in order to
manage dependencies. So, I made a vcpkg port which builds the pwiz library
with Boost.Build, copies all lib files into your build folder, and
generates CMake files required for CMake find_package. Unfortunately, for
now, it is a manual process with some python scripts which parses your
build folders and generates templates for CMake files. I can share what I
did, but CMake or vcpkg port generation should be a part of your build
system and my experience with Boost.Build is limited.

As for imzml support, I can make a separate pull request. I just added
imzml related onologies and made pwiz::msdata::ScanSettings inherited from
ParamContainer. But because my requirement was only to write the imzml, I
just re-implement your serializer and never implemented the reader.

PR #4598, opened 2026-08-20
Sciex ZT Scan: report each bin's collision energy instead of the ramp midpoint

chambm:
## What

ZT Scan is a new Sciex acquisition mode on the ZenoTOF 8600. Q1 scans **continuously** across the precursor range rather than stepping it — the method stores a single MS/MS experiment whose accumulation time is the entire sweep (0.86 s on the example file) — and the instrument ramps the collision energy in hardware from one end of the precursor range to the other as the quadrupole moves. The sweep is then digitized into hundreds of narrow "encoded" bins, and both SDKs surface each bin as its own experiment.

**Neither SDK records a per-bin collision energy.** Both report the ramp *midpoint* on every bin. On the example file that means all 429 bins came out as `collision energy 30.0` for a ramp that actually runs 18 → 43 eV — wrong by up to 12 eV at either end of the sweep.

This reconstructs the per-bin value.

## How

`ZtScanBin` (`WiffFile.hpp`) carries the ramp endpoints and interpolates on the bin's **ordinal** within the sweep.

Ordinal rather than quadrupole m/z because the quadrupole scans at a constant Da/s, which makes the two axes equivalent, while the ordinal needs no extra method parameters (the precursor start/stop masses are exposed by only one of the two APIs) and lands exactly on the stated endpoints at the first and last bin.

Detection and endpoint sourcing necessarily differ per container:

| | `.wiff` (`WiffFileImpl`) | `.wiff2` (`WiffFile2Impl`) |
|---|---|---|
| Detect | sample custom field `Is ZT Scan` | method group name `ZTScan` |
| Endpoints | `CE ∓ CES` | `CERampStart` / `CERampStop` |

The legacy container only encodes the ramp as midpoint ∓ half-range, so an 18 → 43 eV ramp round-trips as `30 / 12` = 18 → 42 and the top end lands up to **1 eV low**. `.wiff2` stores the true endpoints. The two containers therefore disagree by ≤1 eV at the top of the sweep for the same acquisition; that is a limitation of what the legacy container records, not an implementation divergence. `.wiff2` is what SCIEX OS writes as the primary file.

The wiff2 detection uses programmatic method keys rather than rendered labels, so it is not sensitive to the acquisition software's language. The legacy `Is ZT Scan` field *is* a rendered label, so a localized acquisition degrades to the old flat value rather than misbehaving.

Anything that is not a ZT Scan — and any failure to read the method, too few bins, missing CE/CES — falls back to the SDK's value unchanged.

## Also: ZenoTOF 8600 instrument model

`MS:1003443` already existed in the CV but the reader had no enum entry, so these files could not be converted at all without `--ignoreUnknownInstrumentError`, and got the generic `MS:1000495` term. Note the model-name ladder is duplicated between `WiffFileImpl` and `WiffFile2Impl`, so both needed the new entry — `Reader_ABI_Test` does not catch that, because it validates the enum→CV mapping but never exercises name matching against a real instrument string.

## Verification

Against a 300,570-spectrum ZT Scan acquisition (430 experiments × 699 cycles, quad sweeping 392.76–899.78 in 429 bins of 1.1819 Da):

- `.wiff2` ramps **18.0 → 43.0**, `.wiff` ramps **18.0 → 42.0**, 0.0584 eV per bin, resetting each cycle, MS1 untouched.
- Both containers now emit `MS:1003443 ZenoTOF 8600` with the full Q-q-TOF component list, with no `--ignoreUnknownInstrumentError`.
- `Reader_ABI_Test` passes (the checked-in corpus has no ZT Scan file, so this is a regression check on ordinary Sciex acquisitions rather than a check of the new path).

## Open questions

- **The linear-ramp assumption is unconfirmed.** The files state only the endpoints, never the shape. Linear is the natural reading of a start/stop pair over a constant-speed sweep, and it reproduces the reported midpoint (0.05 × 650 − 2 = 30.5 → 30), but it is worth confirming with Sciex. `ZtScanBin::collisionEnergy()` is the only place that would need to change.
- **No test fixture guards this.** The example acquisition is 8.6 GB; a trimmed one would be needed.
- The equivalent change for the managed pwiz-sharp reader is not in this PR.

https://claude.ai/code/session_0182oRXbeKQGEpF1kTkdQAsr

### DIFF
--- a/pwiz/data/vendor_readers/ABI/Reader_ABI_Detail.cpp
+++ b/pwiz/data/vendor_readers/ABI/Reader_ABI_Detail.cpp
@@ -102,6 +102,7 @@ InstrumentConfigurationPtr translateAsInstrumentConfiguration(InstrumentModel in
         case X500QTOF:
         case NlxTof:
         case ZenoTOF7600:
+        case ZenoTOF8600:
             ic.componentList.push_back(source);
             ic.componentList.push_back(Component(MS_quadrupole, 2));
             ic.componentList.push_back(Component(MS_quadrupole, 3));
@@ -154,6 +155,7 @@ PWIZ_API_DECL CVID translateAsInstrumentModel(InstrumentModel instrumentModel)
         case API5600TripleTOF:  return MS_TripleTOF_5600;
         case API6600TripleTOF:  return MS_TripleTOF_6600;
         case ZenoTOF7600:       return MS_ZenoTOF_7600;
+        case ZenoTOF8600:       return MS_ZenoTOF_8600;
         case QStar:             return MS_QSTAR;
         case QStarPulsarI:      return MS_QSTAR_Pulsar;
         case QStarXL:           return MS_QSTAR_XL;
--- a/pwiz_aux/msrc/utility/vendor_api/ABI/WiffFile.cpp
+++ b/pwiz_aux/msrc/utility/vendor_api/ABI/WiffFile.cpp
@@ -105,9 +105,36 @@ class WiffFileImpl : public WiffFile
     void setExperiment(int sample, int period, int experiment) const;
     void setCycle(int sample, int period, int experiment, int cycle) const;
 
+    /// The ZT Scan sweep bin for a 1-based experiment, or NULL when the acquisition is not a
+    /// ZT Scan (in which case the SDK's collision energy is used unchanged). See ZtScanBin.
+    const ZtScanBin* getZtScanBin(int sample, int period, int experiment) const;
+
     mutable int currentSample, currentPeriod, currentExperiment, currentCycle;
 
     private:
+    /// Detects a ZT Scan acquisition and maps each Product experiment to its bin within the
+    /// quadrupole sweep. The result (empty for a non-ZT Scan) is cached per (sample, period) so
+    /// an ordinary acquisition is probed only once.
+    ///
+    /// A legacy WIFF flags the mode with the sample-level custom field "Is ZT Scan", and carries
+    /// the CE ramp only as the legacy CE / CES pair, which encodes it as midpoint -/+ half-range.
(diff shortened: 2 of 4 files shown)


dtabb73:
I have a triplicate of HeLa WIFFs from ZT Scan on a ZenoTOF8600, provided by SCIEX.  They have titles like "20260715_Whisper20-JCcol_2p4DaZTscan_350ngHeLa_Robust-LDR_1."  Each LC-MS/MS amounts to about **100 GB**.  I have processed the WIFFs to mzMLs in the SCIEX Data Converter.  The isolation window target m/zs are written, but the high and low offsets. are not written to the isolation windows.  Could I run any tests to help you?

chambm:
Each experiment within a cycle should be a separate point on the quadrupole ramp. For my ZTScan file, there are 430 "bins" for each cycle, and each bin corresponds to a single quadrupole position along the sweep.
| ms level |	Experiment 	| Data Points |	Scan Time (s) 	| Isolation m/z |
|--------|--------|--------|--------|--------|
| 1 |	1 |	8244 |	0.000833333333 |	n/a |
| 2 |	2 |	1767 |	0.00137721458 |	393.35063 |
| 2 |	3 |	3100 |	0.001410664988 |	394.5325 |
| 2 |	4 |	4342 |	0.001444115396 |	395.71436 |
| 2 |	5 |	5409 |	0.001477565805 |	396.89623 |
| 2 |	6 |	6756 |	0.001511016213 |	398.0781 |
| 2 |	7 |	7892 |	0.001544466621 |	399.25996 |
| 2 |	8 |	7826 |	0.001577917029 |	400.44183 |
| 2 |	9 | 	7742 |	0.001611367437 |	401.62369 |
| 2 |	10 |	7805 |	0.001644817845 |	402.80556 |

But this is how ProteoWizard is presenting the data. You processed your WIFFs with Sciex's data converter. I don't know if they're doing anything different.

PR #4610, opened 2026-08-24
skyline: Fixed test stability defects that are not specific to the .NET 8 port

brendanx67:
## Summary

Test stability work done on the .NET 8 branches that applies to master just as much. Brought over
so master and the nightly get it now rather than after the port lands. The net8-specific half of
that work (net472 removal, the staging rework, MascotShim) is not here and stays on #4605.

* Fixed the MIDAS library loader recalculating results against chromatogram caches the importer
  was still joining and deleting. `LibraryManager` is the only background loader that watches
  `MeasuredResults`, and it does so only for MIDAS, so it woke on every partial-cache join. The
  work does not need the cache at all - MIDAS spectra are read from the raw files - so it now
  waits for the results to be loaded
* Fixed unattended-dialog timeouts cascading into a `ThreadExceptionDialog`. A dialog nobody
  dismisses now fails the test naming the dialog and its message, instead of being reported
  through UI that times out in turn and is then reported as a hang
* Fixed the parallel worker teardown never killing the host worker, and stopped it from killing
  a concurrent run's containers when it launched none of its own
* Added the locking process to tool-extraction failures, so "access to the path is denied" names
  what is holding the file, and stopped re-extracting archive entries that already match
* Fixed two tests reading state that had not caught up yet, and one that could write two audit
  log entries in either order
* Added file-size and missing-file detail to `PooledFileStream`'s modified-file reporting, and a
  diff explainer so an equality failure says which member differs rather than printing two type
  names

Measured on the branch these came from: `TestMidas` went from 14 failures in 200 executions to 0
in 400, at 8 workers across 5 languages, verified in both directions against the same
configuration.

## Test plan

- [x] Solution builds clean (Debug|x64)
- [x] ReSharper inspectcode - 0 errors, 0 warnings from this branch
- [x] CodeInspection
- [x] TestUnattendedDialogTimeout - the new verifier for the dialog watchdog
- [x] TestMidas, TestMidasModifications
- [x] PeakAreaDotpGraphTest, TestAuditLogTutorial
- [x] FilesDirTest, SafeDeleteTest, DsvHelperTest
- [x] /code-review max, findings verified and folded in before opening


pwiz_tools/Skyline/TestUtil/EqualityExplainer.cs
```diff
@@ -0,0 +1,170 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using pwiz.Common.SystemUtil;
+using pwiz.Skyline.Util.Extensions;
```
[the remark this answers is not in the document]
brendanx67: ReSharper reported nothing. This is for TextUtil.

pwiz_tools/Skyline/Util/UtilInstall.cs
```diff
@@ -351,6 +364,57 @@ public static string GetCachedDownloadsDirectory()
         {
             return Path.Combine(ToolDescriptionHelpers.GetSkylineInstallationPath(), @"CachedDownloadsForTests");
         }
+
+        /// <summary>
+        /// Extracts an archive, leaving alone any file that is already the size the archive says it
+        /// should be.
+        /// <para>These archives hold version-pinned tool executables and the MSVC runtime DLLs beside
```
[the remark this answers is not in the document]
brendanx67: Fixed in 7508e47e5d — the summary now says the file is left alone when it matches the archive's size **and** CRC.

pwiz_tools/Skyline/TestRunnerLib/RunTests.cs
```diff
@@ -1461,22 +1461,106 @@ public static void KillParallelWorkers(int hostWorkerPid, string workerNames = n
...
                 if (hostWorkerPid > 0)
-                    Process.GetProcessById(hostWorkerPid).Kill();
+                {
+                    var hostWorker = Process.GetProcessById(hostWorkerPid);
+                    // Check the name before killing. On the normal-exit path the host worker has
+                    // usually exited already, which is exactly when Windows is free to hand its pid
+                    // to something else, and killing a stranger is worse than leaking a process.
+                    if (!hostWorker.HasExited &&
+                        Equals(hostWorker.ProcessName, Process.GetCurrentProcess().ProcessName))
+                    {
```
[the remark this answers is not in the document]
brendanx67: Good catch, and fixed in 7508e47e5d. Rather than special-case the dotCover name, the caller now holds the host worker's `Process` instead of its pid. That keeps a handle open so the pid cannot be recycled, which removes the reason the name check existed at all — so coverage runs tear the host worker down like any other.

pwiz_tools/Skyline/Model/Lib/Library.cs
```diff
@@ -139,7 +172,12 @@ protected override bool LoadBackground(IDocumentContainer container, SrmDocument
                     }
                 }
 
-                var missingMidasFiles = MidasLibrary.GetMissingFiles(document, libraries.Libraries);
+                // Leave the MIDAS work until the results are loaded. Starting earlier gains nothing,
+                // because the spectra come from the raw files, and it is repeated and thrown away for
+                // every partial cache the importer joins.
+                var missingMidasFiles = IsReadyForMidasWork(docCurrent)
+                    ? MidasLibrary.GetMissingFiles(document, libraries.Libraries)
+                    : Array.Empty<string>();
```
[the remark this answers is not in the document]
brendanx67: Agreed and fixed in 7508e47e5d. `EqualsId` compares `ReferenceEquals(Id, node.Id)` — identity, not instance — so `document` can be an older copy of the same document with correspondingly stale `MSDataFileInfos`. `GetMissingFiles` now reads `docCurrent`, matching both the readiness check and the library list.

pwiz_tools/Skyline/Util/UtilIO.cs
```diff
@@ -595,11 +611,51 @@ public string ModifiedExplanation
...
+                }
+            }
+        }
+
         public bool IsOpen
         {
             get { return ConnectionPool.IsInPool(this); }
         }
 
+        private static long FileSizeOrMissing(string filePath)
```
nickshulman: I would recommend making this return `Nullable<long>` because that implies that it might be missing, which would let you rename this method to "FileSize".

pwiz_tools/Skyline/Model/Results/DocNodeChromInfo.cs
```diff
@@ -467,6 +468,54 @@ public TransitionGroupChromInfo ChangeLibraryDotProduct(float? prop)
...
+                (@"ReintegratedPeak", ReintegratedPeak, info.ReintegratedPeak)
+            };
+            var differences = members
+                .Where(m => !Equals(m.Mine, m.Theirs))
+                .Select(m => string.Format(@"{0} {1} vs {2}",
+                    m.Name, m.Mine ?? @"(null)", m.Theirs ?? @"(null)"))
+                .ToList();
+            return differences.Count == 0 ? null : TextUtil.LineSeparate(differences);
+        }
+
```
bspratt, line 518: It would be cool if we could use reflection to automate maintenance of interface providers when members are added, but that's probably not feasible. But perhaps we could tell Claude's code-review to watch out for that?

pwiz_tools/Skyline/SkylineTester/Main.cs
```diff
@@ -88,6 +90,64 @@ private void Run(TabBase fromTab)
...
+                return;
+
+            var message = string.Join(Environment.NewLine,
+                string.Format("There are {0} worker containers still running that are not associated with this run.",
+                    leftovers.Count),
+                string.Empty,
+                string.Join(Environment.NewLine, leftovers),
+                string.Empty,
+                "They are left over from an earlier run that was stopped before it could clean up. They hold",
+                "the source directory open and can make a later build fail with a file lock.",
```
bspratt: I might add "probably" left over from etc

pwiz_tools/Skyline/TestFunctional/JsonToolServerTest.cs
```diff
@@ -235,7 +235,7 @@ private static string DescribeWhatIsOnScreen()
 
             // Shows whether a UI thread is parked in a modal loop (a message box, Form.ShowDialog, a shell common
             // dialog) and what put it there -- which names the dialog even when its own window says nothing.
-            Collect(sb, @"managed call stacks", () => sb.AppendLine(GetCallStacks()));
+            Collect(sb, @"managed call stacks", () => sb.AppendLine(HangDetection.TryGetThreadDump()));
```
bspratt: is @"managed call stacks" still an accurate description?

pwiz_tools/Skyline/TestUtil/HangDetection.cs
```diff
@@ -242,10 +252,137 @@ public void Dispose()
...
+        /// <param name="reason">Why the full dump was unavailable, reported above the stack so a
+        /// degraded diagnostic is never mistaken for the real one.</param>
+        /// <param name="describeRuntime">Whether to add what ClrMD can see of this machine's runtime.
+        /// It costs another attach, so the caller decides: worth it when the dump failed quickly,
+        /// never when it failed by running out of time.</param>
+        public static string GetCallingThreadStack(string reason, bool describeRuntime = true)
+        {
+            var stackLines = new List<string>
+            {
+                string.Format(@"*** Thread dump unavailable: {0}", reason)
```
nickshulman: A method named "GetCallingThreadStack" should really only include the callstack.

"Thread dump unavailable" happens to be the reason that the one caller falls back to using the callstack, but it's not part of the callstack itself, so it should be the caller's responsibility to write out that line.

Similarly, instead of passing in a boolean for whether to describe the runtime, it would make more sense for the caller to describe the runtime as a separate operation.

pwiz_tools/Skyline/TestUtil/EqualityExplainer.cs
```diff
@@ -0,0 +1,170 @@
...
+    /// <summary>
+    /// Turns "these two objects are not equal" into a statement of WHICH member differs, for
+    /// assertion messages.
+    ///
+    /// <para>Prefers the type's own <see cref="IExplainDiff"/>, which is declared next to the
+    /// Equals it explains and so can reach private fields. Falls back to a best-effort walk of
+    /// public properties for types that have not adopted the interface, so an unfamiliar failure
+    /// still says something useful without anyone having to add code first.</para>
+    /// </summary>
+    public static class EqualityExplainer
```
bspratt, line 37: I might have called it InequalityExplainer, but I like how it does address the maintenance issue (so far as possible) I wondered about in IExplainDiff

bspratt (approved):
No substantial issues, just a few possibly clarifying points

bspratt:
> No substantial issues, just a few possibly clarifying points

Though my local ReSharper does complain:
 [WARNING] TestUtil\HangDetection.cs:312 - The parameter 'describeRuntime' has the same default value

nickshulman (approved):
Looks good to me.

bspratt:
FWIW, here's what my Claude says about /code-review max:
------------------

  Reviewed at 2133fdbca. Ran the full local gates against a clean checkout of the branch, then read the diff with
  surrounding context. The diagnostic work here is genuinely good — the comments explain mechanism rather than restating
  code, and several of them (the reentrant-WndProc analysis, the DAC measurements, the docCurrent vs document note in
  LoadBackground) are the kind of thing that saves the next person days. Almost everything below is in the
  test-infrastructure half rather than in the fixes themselves, and the cluster around parallel worker teardown is where
  I'd concentrate.

  Gates

  ┌──────────────────────────────────────────────────────────┬───────────┐
  │                           Gate                           │  Result   │
  ├──────────────────────────────────────────────────────────┼───────────┤
  │ Build, Release|x64, full solution                        │ pass      │
  ├──────────────────────────────────────────────────────────┼───────────┤
  │ CodeInspectionTest                                       │ pass      │
  ├──────────────────────────────────────────────────────────┼───────────┤
  │ ReSharper inspectcode, full solution, --severity=WARNING │ 1 warning │
  └──────────────────────────────────────────────────────────┴───────────┘

  [WARNING] TestUtil\HangDetection.cs:312 - The parameter 'describeRuntime' has the same default value

  GetCallingThreadStack(failureReason, true) passes the value describeRuntime already defaults to at
  HangDetection.cs:325. Zero-tolerance gate, so this reddens CI on its own. My suggestion is to drop the default from
  the parameter rather than the argument from the call: the XML doc says "the caller decides", and with a default one of
  the two callers is inheriting rather than deciding. The other site at :308 already passes false explicitly, so
  removing the default makes the pair symmetric.

  Would fix before merge

  1. The new leftover-worker prompt can block the nightly. SkylineTester/Main.cs:59

  Run(TabBase) now opens with OfferToStopLeftoverWorkers(), which can show a modal Yes/No MessageBox. Three of the paths
  into Run are fully unattended:

  - SkylineTesterWindow.cs:478 — if (_openFile != null && Path.GetExtension(_openFile) == ".skytr") RunUI(Run);, which
    is how SkylineNightly starts a pass
  - Main.cs:34 — RunByTimer(TabBase fromTab) { RunUI(() => Run(fromTab)); }, driven by the 30-second run-again timer at
    TabNightly.cs:278
  - Main.cs:249 — if (_restart) { _restart = false; Run(); }

  Any container matching the worker image makes leftovers.Count > 0, and MessageBox.Show(this, ...) then blocks the UI
  thread with nobody to answer it. The pass is lost with no test output and no report. The catch (Exception) { return; }
  around the detection does not cover this, since a modal dialog is not an exception. Gating the prompt on an
  interactive run would keep the convenience without putting it in the nightly's path.

  2. Teardown kills the host worker on the success path, which breaks parallel coverage runs. TestRunnerLib/RunTests.cs
  + TestRunner/Program.cs

  Under coverage=on the host worker is the coverage profiler, and it writes coverage-{workerName}.dcvr only when the
  profiled process exits. Three things line up badly:

  - The server task returns as soon as it sends quit — Program.cs:1642, workerSender.TrySendFrame(WORKER_QUIT_MESSAGE);
    ... return; — so foreach (var task in tasks) task.Wait(); completing does not mean the host worker has exited.
  - There is no HostWorker.WaitForExit() anywhere in Program.cs; the two WaitForExit(1000) calls at :809 and :855 are
    launch-failure probes.
  - ParallelWorkerTeardown.Dispose → KillParallelWorkers → if (hostWorker != null && !hostWorker.HasExited)
    hostWorker.Kill();

  Master never killed the host worker on a clean finish — it exited on its own, and the report ran after the block.
  Moving GenerateCoverageReport inside the teardown scope fixed the symptom its comment names, but the kill it now sits
  in front of is what guarantees the snapshot is never written. A bounded HostWorker.WaitForExit(...) before the report
  looks like the missing piece.

  Parallel worker teardown

  3. The console-control handler is the one teardown call site without a run tag. Program.cs:1457

  RunTests.KillParallelWorkers(HostWorker, workerNames);   // no runTag

  :1438 and :1783 both pass GetTestRunTimeStamp(). This one does not, and the rewrite also dropped master's workerNames
  ??= string.Join(" ", GetDockerWorkerNames()) fallback — so with runTag == null, GetWorkersStillRunning skips the tag
  branch, returns empty, and KillParallelWorkers returns having killed nothing where master killed every worker.

  It matters most here because LaunchAndWaitForDockerWorker (:892) commits workerNames = currentWorkerNames only after
  the container connects (:902), and never on the "did not connect" throw. A container started but not yet connected is
  running and absent from workerNames. And this handler calls Process.GetCurrentProcess().Kill() immediately after, so
  Dispose/ProcessExit never run — line 1457 is the only teardown on that path, which is the SkylineTester-stop exit your
  own comments identify as the leak source.

  4. Teardown now shells out to an unbounded docker ps on the normal exit of every parallel run. RunTests.cs:1430

  RunCommand ends with p?.WaitForExit(); — no timeout. GetWorkersStillRunning → GetRunningWorkerNames →
  GetDockerWorkerNames → RunCommand("docker", "ps ..."), and KillParallelWorkers now calls it unconditionally. On master
  this ran only in the workerNames == null fallback, and only from the two termination paths; it was never on a normal
  finish, because nothing tore down on a normal finish.

  That defeats DOCKER_KILL_TIMEOUT_MILLIS sitting immediately above it. A daemon that is starting or wedged makes the
  CLI block rather than fail, so the run never exits and the 30-second kill bound is never reached — precisely what that
  constant's comment says must not happen ("a wedged daemon must not keep a finished run alive"). The same call is now
  also on SkylineTester's UI thread at the top of every Run (Main.cs:111), where the catch (Exception) { return; }
  cannot help, because a hang is not an exception. RunCommand needs a bounded overload, or GetDockerWorkerNames needs
  its own.

  5. The ProcessExit hook cannot finish inside the budget the runtime gives it. RunTests.cs, ParallelWorkerTeardown

  Both TestRunner and TestRunnerLib target v4.7.2, and .NET Framework caps the total execution time of all ProcessExit
  handlers at about two seconds. OnProcessExit → TearDown → KillParallelWorkers → the unbounded docker ps above, then
  KillWorkers waiting up to DOCKER_KILL_TIMEOUT_MILLIS (30 s). The runtime tears the process down mid-docker ps, docker
  kill is never issued, and the containers leak.

  It fails silently, too: TearDown sets _torndown via Interlocked.Exchange on entry, so a handler that is cut off has
  already marked itself done and nothing retries or reports. That leaves the Environment.Exit hole the constructor
  comment says it closes still open — those calls are at Program.cs 1505/1545/1571/1746/1833, all inside the scope.

  6. RunTests.GetOrphanedWorkerNames has no callers.

  It reads as though it was meant to be what OfferToStopLeftoverWorkers calls, but that uses GetRunningWorkerNames() —
  and has to, since SkylineTester has no run tag before its run starts, so GetOrphanedWorkerNames(null) returns empty by
  construction. Worth either deleting or documenting the intended caller.

  7. The leftover prompt is unscoped. SkylineTester/Main.cs:111

  GetRunningWorkerNames() filters by image only, so a concurrent run on another checkout on the same machine appears in
  the list, and Yes kills its containers. This is the hazard GetWorkersStillRunning's own doc argues against ("killing
  every worker on the machine reaches a concurrent run's containers - neither is acceptable"). The confirmation prompt
  mitigates it, but the dialog text asserts "not associated with this run", which is stronger than what the code
  actually knows.

  Diagnostics that can't diagnose

  8. TransitionGroupChromInfo.ExplainDiff cannot explain the one member that needs it. DocNodeChromInfo.cs:486

  PeakCountRatio is float (:248) and Equals compares it with == (:522) — the only member that does; every other uses
  .Equals. ExplainDiff boxes both sides into (string, object, object) and filters on !Equals(m.Mine, m.Theirs). For NaN,
  which PeakCountRatio is whenever TransitionCount is 0, == is false but boxed Single.Equals is true. So Equals says
  unequal, ExplainDiff finds nothing, and the reader gets "ExplainDiff found no difference - it may be out of date with
  Equals" — pointing at the member list instead of naming the value. The diagnostic misses exactly the case it exists
  for.

  base.Equals(other) is also outside the member list, so a ChromInfo.Equals mismatch (:1618, false when exactly one
  FileId is null) lands in the same fallback.

  9. The 60-second library wait sits inside a bare catch, so it can cost four minutes and then remove itself.
  AssertEx.cs:1651

  WaitForProcessing is not a soft wait — on expiry it calls Assert.Fail (ResultsUtil.cs:271), and GetWaitCycles there
  has no Debug multiplier, so 60 s is 60 s. DocumentClonedLoadable (:1664) wraps both ForceDocumentLoad calls in catch {
  retry++; } and runs up to two iterations, so a loader that never settles burns up to 4 × 60 s of silent sleeping
  before the final unguarded comparison. Previously nothing waited here at all, so this is pure addition rather than a
  lengthening.

  Your comment diagnoses this precisely — "a wait that expires there does not fail the test, it is swallowed, and the
  final attempt compares the documents WITHOUT forcing the load, which is the race this call exists to close" — and then
  answers it by making the timeout long. That leaves the fix conditional on the wait never expiring, on exactly the
  loaded machines where the race appears.

  There is a second edge in the same loop: target and actual are assigned one at a time (:1673-1674). If the first
  succeeds and the second throws, target is fully loaded and actual is not, which is the asymmetric load state this
  change exists to eliminate. The retry re-forces both, so it takes two failures to reach the final comparison in that
  state — but the bare catch means both are discarded and the comparison then produces the PeptideLibraries load-state
  inequality you were chasing.

  10. UnattendedDialogTimeoutTest reads and clears Program.TestExceptions without the lock the writer takes.
  UnattendedDialogTimeoutTest.cs:91

  Program.AddTestException does lock (TestExceptions) { TestExceptions.Add(exception); } (Program.cs:876), and it is
  reached from arbitrary background threads, since Program.ReportException is CommonActionUtil.ExceptionReporter
  (Program.cs:185) and the background loaders route failures through it. The test's ToArray() and Clear() take no lock.

  The quiet failure is the one I'd worry about more than the race: Clear() discards every exception the harness holds,
  not just the provoked one. A genuine unrelated failure recorded before or during this test is thrown away,
  AbstractFunctionalTest's later Assert.IsFalse(Program.TestExceptions.Any()) passes, and the real bug disappears.
  Conversely Assert.AreSame(firstTimeout, reported[0]) fails spuriously whenever anything else reported first. Removing
  just the provoked instance, under the same lock, would be the matching shape.

  The stale-status guard is one layer out from the reordering

  11. FileProgressControl.SetStatus is not the only consumer of the snapshot, and the rule it adds already exists in the
  model.

  - AllChromatogramsGraph.cs:457 hands the same status to graphChromatograms.UpdateStatus(loadingStatus) one line before
    progressControl.SetStatus(loadingStatus), and the overall progressBarTotal.Value a few lines above is computed from
    the same snapshot. Both are still dragged backwards while the per-file bar is held at 100, so the views disagree.
  - MultiProgressStatus.ChangeStatus (MultiProgressStatus.cs:251-259) is already the same rule — "Avoid overwriting a
    final status with a non-final status", keyed on ReferenceEquals(Id). So a per-file status cannot regress within a
    snapshot; what reorders is whole snapshots. At that level a file at 80% in the newer snapshot and 40% in the older
    is still rewound, because neither is final. The guard closes the terminal case that produced the stuck 99%, which
    matches the symptom, but not the general reordering.
  - SkylineWindow.UpdateProgress (Skyline.cs:3906-3930) has the sharper version of the same hole: a final status is
    removed from _listProgress, so a later stale non-final status with the same Id no longer matches and falls into else
    { i = _listProgress.Count; _listProgress.Add(status); }, resurrecting a finished import and leaving _timerProgress
    running.

  I'm not claiming the last one is caused by this PR — it isn't. But if snapshots really do arrive out of order, that
  path has the same premise and a worse outcome, and it's worth deciding whether the guard belongs where the snapshot is
  dispatched rather than in one control.

  Smaller things

  - CommonAlertDlg.cs:478 — _exception is DialogTimeoutException misses a wrapped timeout (TargetInvocationException,
    MultiException, new IOException(msg, x)), and the handlers that pass only x.Message to MessageDlg.Show still produce
    the original cascade. Walking InnerException would widen the fix cheaply.
  - AuditLogTutorialTest.cs:324 — the new WaitForConditionUI(() => ...Rows.Count > auditRowsBefore) uses the default
    budget (3 min, 12 min in Debug). This PR bounds two other waits explicitly for that exact reason, and a row count
    that never advances is the same "never coming" shape.
  - AssertEx.cs:1701 — the explainer is grafted onto one call site ahead of the existing Cloned(...). Wiring it into
    Cloned/AreEqual would give every type the better message. Deliberate scoping, or a first step?
  - EqualityExplainer.cs:~140 — if (differences.Count >= MAX_DIFFERENCES || ...) continue; keeps walking every remaining
    property after the cap; break on the count check would be clearer.
  - UtilInstall.cs:323 — TryHelper.Try<Exception>(..., 4, 1000) retries the base type, so a corrupt archive or a full
    disk gets three extra attempts (each re-opening the zip and re-CRCing everything already extracted) before falling
    through the when (x is UnauthorizedAccessException || x is IOException) filter written to explain it. TryTwice, used
    44 lines above in the same method, catches exactly those two.
  - UtilInstall.cs:398 — IsAlreadyExtracted derives the destination with Path.Combine(installPath,
    entry.FileName.Replace('/', Path.DirectorySeparatorChar)), while ZipEntry.Extract strips a drive prefix and a
    leading separator before combining. For ordinary relative entries the two agree, so this is latent rather than
    active — but where they diverge the check stats a different file from the one that will be written, and the result
    (skipped forever, or re-extracted every run) is silent either way. Reusing the same normalization would close it.
  - FileLockingProcessFinder now returns an IOException when it names a process, so UtilInstall.cs:332's throw described
    converts an UnauthorizedAccessException into an IOException. Nothing catches that type around this path today, so
    this is a note rather than a problem — but TestFunctional.cs:2787 pipes every functional-test failure through the
    same method, which widens the blast radius more than the tool-extraction case suggests.

  What I checked and found sound

  - ProgressStatus.Id is new object() per constructor and preserved through ChangeProp, so the SetStatus guard's
    ReferenceEquals really does distinguish chains. Reset() does not clear Status, and retry produces a new status, so
    the retry path still restarts. Only non-final statuses are blocked, so a late error still lands.
  - LibraryManager.StateChanged — MidasLibrary.GetMissingFiles returns non-empty only when some
    MSDataFileInfo.HasMidasSpectra, so the early-out cannot skip work that existed. Checking both documents covers the
    unflag case and LibrarySpecsUnloaded.Any() preserves the retry. I could not find a path where the MIDAS work fails
    to be re-triggered once results finish loading.
  - PeptideLibraries.ExplainDiff matches its Equals member for member, and RankId/PeptideCount are plain
    auto-properties, so it will not trip the Assume.IsTrue(IsLoaded) getters.
  - MeasuredResults.IsLoaded is _countUnloaded-based rather than IsNotLoadedExplained == null, so
    ExplainUnloadedAgainstCache() never runs on a hot path, and IsCachedFile is a HashSet lookup.
  - The French-message reasoning in ToFileLockingException holds: '([^']+)' over L'accès au chemin d'accès '<path>'
    yields ["accès au chemin d", "<path>"], and FirstOrDefault(IsRootedPath) picks the path.
  - GetWaitCycles really is x4 in Debug, so the "30s -> 2 min" comments are accurate, and
    Program.UnitTestTimeoutMultiplier still applies on top for multi-process runs.
  - TryHelper.Try rethrows on the final attempt, so the new catch in UtilInstall is reachable.
  - MessageBox.Show and the English literals in SkylineTester/Main.cs are sanctioned — CodeInspectionTest lists
    SkylineTester under NonSkylineDirectories().

pwiz_tools/Skyline/TestUtil/HangDetection.cs
brendanx67: Done in 946b8d382a. `GetCallingThreadStack()` now takes no arguments and returns only the stack; `TryGetThreadDump` composes the "unavailable" line, the runtime description, and the stack. That also removed the `describeRuntime` bool rather than tuning it, which is what the Code Inspection gate was failing on (`RedundantArgumentDefaultValue`) - so this fixed the red check at its root instead of the symptom I was about to patch.

pwiz_tools/Skyline/Util/UtilIO.cs
brendanx67: Done in eafb1f8ffa. It is now `long?`, the helper is named `FileSize`, and the property it fills is `FileSizeAtOpen` - so the "could not read it" case is in the type instead of encoded as -1, and the two names say which size is which.

pwiz_tools/Skyline/TestFunctional/JsonToolServerTest.cs
brendanx67: Good catch - fixed in eafb1f8ffa. It now reads "thread dump", since after the refactor that section returns the full dump where the attach works and only the calling thread's stack where it does not.

pwiz_tools/Skyline/SkylineTester/Main.cs
brendanx67: Added in eafb1f8ffa, and the review that followed made the case for it stronger. The prompt now says "test worker containers running that this run did not start", and warns that containers are listed by image - so a run from another checkout on this machine appears here too. That is all the code actually knows; the earlier wording claimed more.

pwiz_tools/Skyline/TestUtil/EqualityExplainer.cs
brendanx67: Leaving the name as is, but the maintenance point is worth recording: it explains why an equality assertion failed, so `EqualityExplainer` covers what it does, and renaming it touches several files for a debatable gain this close to merge. Happy to change it if you feel strongly.

pwiz_tools/Skyline/Model/Results/DocNodeChromInfo.cs
brendanx67: Your hypothetical is already real, in this exact type. `Equals` here compares `Annotations` twice and never compares `MassError` - which is the only state-carrying property it misses - and `ExplainDiff` mirrors that faithfully, so it cannot name the one member that drifted. Brendan is filing that separately; it is pre-existing, and the two-line fix changes document equality, so it wants its own test pass rather than a ride-along here.

On reflection as the default: `EqualityExplainer` already falls back to a public-property walk when a type has no `IExplainDiff`, so it is available - it just stays test-side, where it reads as best-effort. Making it a type's own `ExplainDiff` would make it authoritative, and it is wrong in ways that matter: `PeptideLibraries.Equals` compares `_rankIdName`, which has no public accessor, so a walk cannot see it; some getters `Assume.IsTrue(IsLoaded)` and would throw exactly when state is already bad; and it would report members `Equals` deliberately ignores. Worst for your concern, a type with a reflection default would LOOK covered while still missing every privately-compared field, which hides the drift instead of catching it.

The version that would catch it is a test rather than a default: vary one member, assert unequal AND that `ExplainDiff` names it. Note what that would have caught here - varying `MassError` produces objects `Equals` calls equal, so it fails on the real bug, not just the missing explanation.

PR #4620, opened 2026-08-26
skyline: Added .NET 8 ClickOnce publishing and official signing

nickshulman:
Stacked on #4619 (base branch `Skyline/work/20260612_net8_port`), which carries the .NET 8 port itself.

## Summary

* Added ClickOnce publishing for the net8 Skyline, driven from `pwiz_tools\Skyline\Publish-ClickOnce.ps1` and `Properties\PublishProfiles\ClickOnceProfile.pubxml` (.NET 5+ ClickOnce is publish-profile driven and cannot be produced by `dotnet publish`)
* Fixed the deployment payload so the installed app actually works: files reaching `$(OutDir)` by Copy task are invisible to publish, and the manifest named 103 files twice, which the sxs parser rejects outright at install
* Added setup.exe carrying the .NET 8 Desktop Runtime prerequisite, plus an install page
* Added `UserConfigMigrator`: .NET 8 dropped the ClickOnce awareness that kept user.config where `Settings.Default.Upgrade` could find it, so every update started from default settings
* Added `Publish-ClickOnce.ps1 -KeyLocker` to sign with the University of Washington certificate through the DigiCert Signing Manager KSP, using dotnet-mage rather than mage.exe (a .NET ClickOnce application manifest is `<name>.dll.manifest`, which mage.exe cannot update)

## Notes for the signing check

`-KeyLocker` preflights the certificate, dotnet-mage and signtool before building, so a missing piece fails in seconds rather than after a full build. Nothing uploads; the signed tree is left in `pwiz_tools\Skyline\publish`.

Already verified here against a self-signed certificate, so a failure isolates the KSP: dotnet-mage 10.0.0 signs both net8 manifests, preserves the deployment identity, and re-hashes the payload it lists (appending 10 bytes to Launcher.exe moved the recorded size 17848 -> 17858), which is why the PEs are signed before the manifests. Not verified: whether the KSP can be driven through `signtool /csp` + `/kc` and `dotnet-mage -CryptoProvider` + `-KeyContainer`.

Two cautions: a KeyLocker-signed build carries the real Skyline-daily deployment identity, so installing it upgrades a production Skyline-daily rather than sitting beside it; and KeyLocker meters per signing operation, so the script prints the count for the run before spending it.

## Test plan

- [x] TestUserConfigMigration - picks the highest previous user.config below the running version, from either the ClickOnce data directory or an earlier net8 settings folder, and never overwrites an existing one
- [x] CodeInspection
- [x] Release solution build (net8) on the merged base
- [ ] Official DigiCert signing - `-KeyLocker` against the real certificate
- [ ] net472 -> net8 in-place upgrade verified from an actually installed client, not just at unit level


nickshulman:
Closing this pull request because we will not be using ClickOnce next version.

PR #4629, opened 2026-08-31
skyline: Fixed the net10 nightly failures and the pass-1 NHibernate leak

brendanx67:
## Summary

* Fixed eight net10 nightly failures that were one defect: a benign ObjectDisposedException WinForms raises when signalling a marshaled call whose wait handle it already disposed (dotnet/winforms#14996, new in .NET 9). Filtered deliberately narrowly - any Skyline frame anywhere on the stack leaves it reported
* Fixed the pass-1 managed leak from undisposed NHibernate SessionFactories in IonMobilityDb and OptimizationDb. Nine of eleven leaking tests go clean; TestFilesTreeForm from 2,072.8 KB to 0
* Fixed TestAuditLogTutorial's intermittent NullReferenceException - FindOpenForm only sees a dockable form once its handle exists, so it now waits. Was 1 failure in 22 runs, 0 in 609 after
* Stopped a superseded loader's self-cancellation ending a results test's wait. RetentionTimeManager decides "cancelled" by reference comparison against the current document, so it means superseded, not stopped - and the loader restarts. Regression from the net8 port's loosened IsFinal; master is unaffected
* Added a loader lifecycle trace to the "still not loaded" failures. The thread dump taken there is always empty of loader threads, so it cannot say whether the commit ran, was rejected, or never happened

## Test plan

- [x] Solution builds clean in Release/net10; CodeInspection 0 failures
- [x] TestData.dll in full - 358 instances, 0 failures
- [x] 9-hour parallel run, 8 workers, five languages: 47,222 instances. The two tests fixed here did not fail once
- [x] TestAuditLogTutorial 0 in 609 parallel executions (was 1 in 22)
- [x] WatersImsMse...AsSmallMolecules 0 in ~6,000 (was 2 in 291), with the fix's own trace confirming it rides out the race rather than the race having stopped
- [x] Instrumentation verified against a passing run via SKYLINE_LOADER_TRACE, not first trusted at the failure

Reviewed with /code-review max before opening; the seven findings it raised that I could confirm against the source are fixed in the last three commits.

### DIFF
--- a/pwiz_tools/Shared/Common/Database/NHibernate/SessionWithLock.cs
+++ b/pwiz_tools/Shared/Common/Database/NHibernate/SessionWithLock.cs
@@ -35,23 +35,33 @@ namespace pwiz.Common.Database.NHibernate
     public class SessionWithLock : AbstractSessionWithLock, ISession
     {
         readonly ISession _session;
+        // A factory this session OWNS and disposes with itself. NHibernate registers every
+        // SessionFactory in the static SessionFactoryObjectFactory.Instances dictionary and
+        // only Dispose() removes it, so an undisposed factory is rooted for the life of the
+        // process along with its persisters, dialect and emitted proxies. Tying the factory
+        // to a session that callers always dispose means no factory can outlive its scope.
+        readonly ISessionFactory _ownedSessionFactory;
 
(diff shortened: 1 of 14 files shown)


chambm:
@brendanx67 This ready to merge?

PR #4631, opened 2026-09-01
skyline: Fixed the small-molecule conversion tests racing the library loader

chambm:
## The failure

`RefineConvertToSmallMoleculesTest` and `RefineConvertToSmallMoleculeMassesAndNamesTest` fail whichever one runs **first** in a TestRunner process, and pass on every later run in that process:

```
TransitionGroupChromInfo results differ:
  YLGAYLLATLGGNASPSAQDVLK Charge 2  vs  pep_YLGAYLLATLGGNASPSAQDVLK Charge [M+2H]
  [0] [0] LibraryDotProduct 0.72051907 vs (null)
```

Reversing the order moves the failure to the other test, so it is ordering, not mode:

| run order | converted `.blib` written | failed |
|---|---|---|
| `formulas`, then `masses_and_names` | only `masses_and_names` | `formulas` |
| `masses_and_names`, then `formulas` | only `formulas` | `masses_and_names` |

## Cause

`RefinementSettings.ConvertToSmallMolecules` converts libraries only when the document reports them loaded:

```csharp
var canConvertLibraries =
    invertChargesMode == ConvertToSmallMoleculesChargesMode.none &&
    mode != ConvertToSmallMoleculesMode.masses_only &&
    document.Settings.PeptideSettings.Libraries.IsLoaded;
```

That is the right call for a document that never loaded a library. The problem is on the test side: `AsSmallMoleculeTestUtil.ConvertToSmallMolecules` takes its document from `ResultsTestDocumentContainer.AssertComplete()`, which only inspects the **chromatogram** loader's progress and returns while the library manager is still working.

So on a cold cache the conversion reads `IsLoaded` as false, skips the library branch, and yields a converted document with no library dot product while the original has one. On a warm cache the library is already present and the two agree.

## This was a coverage hole, not only a flake

In the runs that failed, no `worm.1.1.converted_to_small_molecules.blib` was written at all and the converted precursors carried no `<bibliospec_spectrum_info>`. Library conversion was not being exercised in those runs — the test failed only because the two sides then disagreed about a dot product.

## Fix

Wait for the libraries before converting. A document with no libraries reports itself loaded, so this is a no-op everywhere else.

## Verification

Reproduced the order dependence, then re-ran both orderings after the change: all orderings pass, and both `formulas` and `masses_and_names` now write a converted `.blib` regardless of order. `masses_only` still writes none, which is correct — that mode is excluded by the `mode != masses_only` term.

Built and run on the .NET 10 port branch (#4619), where all three touched files are byte-identical to `master`; this PR exists to get it exercised on `master` itself. Every API it uses is already used in the same file on `master`.

https://claude.ai/code/session_0182oRXbeKQGEpF1kTkdQAsr

### DIFF
--- a/pwiz_tools/Skyline/TestUtil/AsSmallMoleculeTestUtil.cs
+++ b/pwiz_tools/Skyline/TestUtil/AsSmallMoleculeTestUtil.cs
@@ -226,6 +226,13 @@ namespace pwiz.SkylineTestUtil
                     {
                         docContainer.SetDocument(docLoad, null, true);
                         docContainer.AssertComplete();
+                        // ConvertToSmallMolecules skips library conversion entirely unless
+                        // Libraries.IsLoaded, so the conversion under test depends on winning a
+                        // race with the library loader. Whichever conversion test ran first in a
+                        // process lost it: no converted .blib was written, the converted document
+                        // had no library dot product where the original did, and the comparison
+                        // below failed. Later tests in the same process passed on a warm cache.
+                        docContainer.WaitForLibrariesLoaded();
                         doc = docContainer.Document;
                     }
                 }
@@ -280,6 +287,7 @@ namespace pwiz.SkylineTestUtil
             {
                 docContainer.SetDocument(docResults, null, true);
                 docContainer.AssertComplete();
+                docContainer.WaitForLibrariesLoaded();
                 doc = docContainer.Document;
             }
             AssertEx.ConvertedSmallMoleculeDocumentIsSimilar(docOriginal, doc, Path.GetDirectoryName(docPath), mode);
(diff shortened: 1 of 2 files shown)


chambm:
Addressed both suppressed suggestions from the Copilot review in 214fc5ff2a.

**`WaitForLibrariesLoaded` never re-checks after the final sleep** — fixed. The loop sleeps *after* its last check, so libraries finishing during that final interval produced a spurious failure. That is the same shape of race this method exists to close, so it was worth getting right.

**`GetWaitCycles()` can return 0 when `millis < SLEEP_INTERVAL`** — covered by the same change rather than separately. The new post-loop check evaluates the condition once regardless of cycle count, so a sub-interval timeout can no longer fail without ever looking.

I deliberately did **not** harden `GetWaitCycles` itself. It is shared with the pre-existing `WaitForProcessing`, so a `Math.Max(1, ...)` there would change behaviour beyond this PR for a case no caller passes today. Happy to do it if a reviewer prefers the helper fixed at the source.

Re-verified both orderings after the change — `formulas`-first and `masses_and_names`-first both pass, and both write a converted `.blib`.

chambm:
Not needed with #4629 fixes.

chambm:
It seems this race still exists even with 4629. This fixes it according to my testing.

Issue #4484, opened 2026-07-28
Osprey: retire OSPREY_PASS2_QVALUE=percolator as the default; choose the pass-2 FDR default (transfer vs transfer-compete + protein-compact)


brendanx67:
## Summary

Mike and Brendan have decided the current default 2nd-pass FDR mode -- `OSPREY_PASS2_QVALUE=percolator` (retrain the Percolator SVM + recompute a target/decoy null on the reconciled **+ compacted** pool) -- is **not a valid approach**. The compacted pool is decoy-depleted, so the retrained null is thin/biased and the reported q is anti-conservative (Stellar libdecoy entrapment: **1.57% true FDP @ nominal 1%** vs 0.92% for the 1st-pass q -- see #4363). This issue is the **decision + default-flip**: pick the replacement default among the frozen-model modes and resolve the specific open question of whether **`transfer-compete` is actually valid and/or better than plain `transfer`**.

Nothing changes until we decide -- all frozen modes shipped **off by default** in #4446 (C#), with the Rust port tracked in maccoss/osprey#57. This issue is the umbrella "flip the default" decision that consumes #4363 (calibration root cause) and #4436 (sensitivity recovery) as inputs.

## The four modes (`OspreyEnvironment.cs`, `docs/12-second-pass-fdr.md`)

| Mode | Retrain? | Null population | Level |
|---|---|---|---|
| `percolator` (default today) | **yes** | reconciled + **compacted** (decoy-depleted -> anti-conservative) | precursor + peptide |
| `transfer` | no | pass-1 q carried; only reconciliation-**moved** peaks re-mapped via each file's full-pool score->q table (quantile bins + PAVA isotonic); co-monotone confidence transfer (Rost 2016 TRIC) | precursor + peptide |
| `transfer-compete` | no (frozen model) | **fresh full-population** target-decoy competition on frozen 1st-pass scores | precursor |
| `protein-compact` | no (frozen model) | competition **constrained to the protein stratum** (base_ids of proteins with >=2 peptides passing 1st-pass protein FDR, admitted as target+decoy pairs); off-stratum survivors keep pass-1 q so report = pass-1 union stratum-passers; independent filtering (Bourgon 2010) | precursor |

Plus the diagnostic A/B lever `OSPREY_PROTEIN_COMPACT_RETRAIN` (with `protein-compact`, retrain instead of frozen, to isolate the frozen-vs-retrain calibration difference for the entrapment oracle).

[...]

PR #4642, opened 2026-09-07
osprey: Made SecondPassFDR fold over the runs instead of holding them

brendanx67:
## Summary

* `SecondPassFDR` held every run's survivors at once. On a 446-run CHS cohort that peaked at
  **68.0 GB managed / 70.5 GB private** and was killed at run 381 of 446 with 0.34 GB free,
  having computed nothing. This makes the join fold over the runs one at a time: each run is
  rebuilt from its own artifacts, folded, and dropped.
* The run now COMPLETES - exit 0 in 71 minutes - at **23.5 GB managed / 36.4 GB private**, and
  the memory floor went from **RISING +71 MB/run** to **FALLING -5 MB/run**. A rising floor is
  `O(files)` accumulation; it is gone.
* `SecondPassFDR` now peaks below `FirstPassFDR` (32.5 GB / 40.1 GB on the same cohort) on both
  axes, which was the acceptance criterion - the second pass works on a 4.6x smaller bounded
  set, so it should never have cost more.
* Boundary 3 -> 4 is held and asserted per run: a run carrying a current
  `.2nd-pass.fdr_scores.bin` is rebuilt without opening any first-pass file.
* `OSPREY_STAGE7_STREAM=0` keeps the resident arm as the byte-identity oracle, with a
  validity-key term so an in-place A/B cannot adopt the other arm's `.blib`.

Part of #4486. Not `Fixes`: `--model-diagnostics` still needs the resident pool, and that is
the next piece of work (see "Not done" below).

## The wall was not where it was thought to be

The 446-run failure had been attributed to Stage 7's survivor-pool rebuild
(`BuildRescoredPool`). It is not. On the `--task SecondPassFDR` leg the run dies inside

```
--input-scores: loading 446 per-file score parquet(s)
    Loading file 381/446: ... (from ....scores-reconciled.parquet)
```

which is `PerFileScoringTask`'s merge building `ScoredEntries`, **before Stage 7 computes
anything**. `BuildRescoredPool` is never reached there - `PerFileRescoreTask.Rehydrate`'s
`ExpectReconciledInput` branch publishes the resident `RescoredEntries`.

That matters because the previously planned fix - shrinking the row from 274 B to 88 B - would
not have moved this number. A narrower row is still 446 runs of rows, held before the first
fold starts.

## What changed

Every step in Stage 7 was already a fold to `O(distinct)` - the fragment release, the pass-2
competition, protein parsimony, the experiment-q re-clamp, all three `.blib` gates. None of
them could bound anything, because the stage was *handed* the whole pool before the first one
ran. **A fold does not bound anything unless its source is per-run too.**

* `ScoringTaskShared.CanStreamStage7Join` admits the reconciled-input leg to the per-run shape
  `CanHydratePerRun` already gives the rescore: the merge publishes one empty list per run and
  reads no rows.
* `RescoreHydration.RefillOneRunSurvivors` rebuilds one run in place - load its reconciled
  stubs, overlay, compact - and the fold drops it again.
* The two facts a resident pool carried *between* passes - the second-pass sidecar overlay and
  the experiment-q floors - are re-applied per run through `RescoredEntries.AddPostMaterialize`,
  in the order the stage computed them. A floor raises a value the overlay has just written, so
  the composition order is the correctness argument; a test pins it.

The run log states the shape rather than leaving it to be assumed, because a resident Stage 7
and a streamed one produce identical bytes by design:

```
446 of 446 run(s) carry a current 2nd-pass sidecar and are rebuilt without opening any 1st-pass file.
Second-pass fold reading the worker's written answer for all 446 file(s); no 1st-pass sidecar is opened.
```

## Measured, 446 runs, same bed and recipe

| task | managed peak | private peak |
|---|---|---|
| `FirstPassFDR` (same cohort) | 32.5 GB | 40.1 GB |
| `SecondPassFDR` **before** | 68.0 GB | 70.5 GB |
| `SecondPassFDR` **after** | **23.5 GB** | **36.4 GB** |

The post-GC probes say the same thing about live data - `stage7-inherited` 5.12 GB,
`stage7-pool` 5.12 GB (nothing built), `stage7-pass2-scored` 2.97 GB, `stage7-protein-fdr`
2.98 GB - so what remains in the peak is transient and reclaimable, not a pool.

Stage 7 trades wall clock for memory: each fold pass rebuilds every run from disk. That is the
intended direction - the target is stated as memory, not wall clock - and every fold now
reports per-run progress, because an unreported one is a multi-minute silence.

## Also fixed on the way

* `TransferOneFile`'s missing-features branch was a `return` where the loop it was extracted
  from had a `continue`, so one entry without reconciled features abandoned the rest of that
  run's survivors and skipped the run's completion count.
* The pass-2 competition reported progress from `ReadFile`, which runs only where a file must
  be *recomputed* - so on the path this work exists to produce, every file answered by the
  worker, the counter never moved: 654 s of silence at 446 runs.
* `FdrScoresSidecar.TryWalkRecords` returned `false` after handing records to its caller.
  Chunking the read moved the file access inside the walk and took away the guarantee the
  OOM-filtered catch rested on, so a mid-walk IO fault left half a file's values applied while
  the caller treated `false` as non-fatal. It throws now, naming the file and the count.
* Mode 10 deleted `OSPREY_PASS2_QVALUE` / `OSPREY_EXPERIMENT_AGG` instead of restoring them,
  and it runs before modes 4, 2, 5-9 and every later dataset - so a developer's exported arm
  was silently dropped while the summary reported it as passing.

## Gate coverage this adds

* Mode 3 asserts the marker line for the per-run fold. Output cannot distinguish the two arms,
  so the gate has to assert the path. The summary now says which shape ran, and where it could
  not: `Stellar` PASS, and SKIP with the reason on the three datasets that set
  `--model-diagnostics`. That narrowness is the finding, not a workaround.
* Mode 10 is cut from two arms to one. `protein-compact` refuses a mean(best-N) first pass, so
  the `meanbest2` arm runs `transfer` for pass 2 regardless - one leg exercises both ideas, and
  the leg drops from 446.8 s to 216.9 s.
* Mode 10 now asserts the arm ENGAGED before asserting what it produced. It did not, and every
  file check in it passes on a run with both variables ignored, because the default mode writes
  the same files. Negative control: neither marker appears in a default run's log.

## Not done, and named rather than left to be found

`--model-diagnostics` is the one Stage 7 consumer still handed the whole pool.
`ModelDiagnosticsData`'s pass-2 builders index their runs by position and revisit a run across
two loops, so they need a list. `CanStreamStage7Join` declines the report leg outright rather
than letting it stream and then pull the pool back through `.Value` - the same peak by a longer
route, with nothing in the log to say so. The fix is the accumulator the pass-1 report already
uses, and it closes the gate-coverage gap above at the same time.

## Test plan

- [x] `regression.ps1 -Dataset All` - 79 PASS / 0 FAIL on this head
- [x] `regression.ps1 -Dataset Stellar` - 16 legs, including `mode3 (HPC chain==straight)`
      (byte-identity on exactly the leg this changes) and the new `mode3 (streamed join)`
- [x] `regression.ps1 -Dataset StellarLibDecoy` - 22 legs, including mode 10 with both markers
- [x] `Build-Osprey.ps1 -RunTests -RunInspection` - 606 tests, inspection zero-warning
- [x] `TestPostMaterializeOverlaysRunInOrderAndOnlyWithAPerFileSource` - new; pins the per-run
      overlay chain's order and its refusal on a source-less milestone
- [x] 446-run CHS `--task SecondPassFDR` - exit 0 in 71 min; `.blib` carries 45,643 `RefSpectra`,
      20,351,447 `RetentionTimes` and 446 `SpectrumSourceFiles` (the count that catches a
      `.blib` silently missing runs), 5,502 protein groups at 1% protein FDR
- [x] TeamCity Perf/Regression on `pull/4642` - 78 PASS / 0 FAIL / 0 SKIP in 01:19:30. That
      build ran on `1cc6fcc1`, so it predates the review fixes and the mode-10 cut; the cut
      removes 223.1 s from the wall

---

## Added 2026-09-08: the pass-2 retrain removal, and P16

Two further pieces landed on this branch. Net **-65 lines** (691 added, 756 deleted) - the
removal takes out more than the new capability adds.

### The pass-2 retrain is gone, not disabled

Pass-2 SVM retraining was cut in `ad4ef8d106`, but a fallback kept it reachable: when the
frozen 1st-pass model byproduct was absent, `ComputePass2Resident` logged a WARNING and fell
through into the 2nd-pass Percolator retrain. That shipped output computed by the method the
project removed for a correctness reason - compaction leaves the pool decoy-depleted, so
retraining mis-estimates the null (#4484, closed "do not re-open the retrain") - on a
warning, where the standing rule is hard-fail.

* The fallback is now an `InvalidOperationException` naming the issue and the remedy.
* The projection 2nd-pass branch, unreachable since the retrain went, is deleted - with the
  `!config.ModelDiagnostics` term in its condition, which did nothing at runtime and read as
  though the flag reroutes the analysis.
* `pass2Contributions` is consequently null in every configuration and deletes end to end.
  `BuildPass2` KEEPS its parameter: those cards are re-sourced from the frozen pass-1 model
  next, and the non-null contract stays under unit test.

### P16: a report is a derived view over the artifacts

`SecondPassFDR` now folds its diagnostics product from a completed second pass, symmetric
with pass 1's arm, placed **ahead of the marker wipe** - the wipe clears the very stamps that
entitle the fold to adopt that pass. `--task ModelDiagnostics` invokes whichever folds are
missing instead of refusing and naming a producer, which is P15's ordinary resume applied to
the diagnostics outputs.

### New gate: `regression.ps1` mode 11

Deletes both diagnostics products from a completed run and asks for the report again - the
pay-later scenario no other leg presented, because every other leg passes the flag up front.
It asserts BOTH halves of P16's test: each pass logs its fold marker AND the join's markers
are absent (a re-analysis produces the RIGHT report, so only the log distinguishes them), and
both products are byte-compared against what the flag-up-front run wrote.

**It found a real defect on its first run.** The pass-2 fold reported the FIRST pass's
q-values under a pass-2 heading whenever the survivor pool was resident: the streamed overlay
is a no-op there, and the join never needed a resident equivalent because `ComputeAndPersist`
stamps the pool as it computes. Proof: the folded `coAssignment.experimentCutoff` was 0.6691,
which the log shows is the pass-1 boundary, against the up-front run's -0.1516. Exit code,
markers, memory band and file-level checks all passed through it. Fixed by
`OverlayPass2OntoResidentPool`.

Mode 11 excludes, **by name and in its summary line on every run**, the pass-1 views no
pay-later path can rebuild today (`cal`, `model`, `featureHistEdges`, `featureCount`,
`modelComposite`). All but `cal` have one cause - the model is not retrained on a resumed run.
That skip-list is a known gap, not an accepted difference, and removing it is the acceptance
test for moving the pass-1 diagnostics work out of `PerFileScoring`.

### The pay-later report is CHEAPER than the join it replaces

Measured at 446 files on a completed CHS cohort (every artifact current, pass-1 diagnostics
product present, pass-2 product absent), `--task SecondPassFDR --model-diagnostics`:

| | pay-later fold | the same work inside the join |
|---|---|---|
| peak | **14.0 GB managed / 22.1 GB private** | 18.2 GB / 29.0 GB |
| wall | **18 min** | 1:08:40 |
| exit | 0 | 0 |

with the streamed marker (`folding over 446 run(s) ... no all-runs survivor pool`) and ZERO
occurrences of `[STAGE-WALL] second-pass-fdr`, `Running protein-level FDR`, `Re-scoring file`,
`Scoring file`, `Running First-pass Percolator` or `Rebuilding first-pass survivors`. The
`.blib` is untouched at its original timestamp.

Reaching that fold needed the pass-2 product to be a DECLARED output. On
`--task SecondPassFDR` it was not, so a completed cohort had nothing outstanding and the
driver skipped the task entirely (`SecondPassFDR: skipping (outputs valid)`) - the fold could
never run. It was left undeclared because declaring a diagnostics output used to force a
pass-2 re-join on every resume; the fold arm is what retires that objection.

### Known limitations, stated rather than discovered later

* **`--task ModelDiagnostics` over `--input-scores` fails** (`HydrateReconciliationOverlay`),
  reproducibly at 10 files in 33 s - not a memory problem. It is the two-seams defect already
  written up in the `--input-scores` retirement proposal: the input kind and `--task` are two
  opinions about where the pipeline starts, and `PerFileRescoreTask.IsIncluded` believes the
  older one, joins the graph and demands state a diagnostics fold never publishes. The fix is
  that retirement, which is scoped to its own branch. The working entry points are the
  ordinary command plus the flag, and the per-task ones.
* **The ordinary `-i` path still builds the resident Stage-7 pool at 446** (~91 GB private),
  because `CanStreamStage7Join` requires `ExpectReconciledInput`, which only
  `--task SecondPassFDR` sets. Pre-existing and unchanged by this PR - the streamed arm this
  PR added is on the `--task SecondPassFDR` leg, which is where the acceptance criterion was
  measured.
* **Pass-1 completeness on the pay-later path**: `cal`, `model`, `featureHistEdges`,
  `featureCount` and `modelComposite` cannot be rebuilt, because the model is not retrained on
  a resumed run and the CAL view is captured in `PerFileScoringTask` memory and never read
  back. Mode 11 excludes exactly those five by name and says so in its summary line every run.

### Test plan

- [x] `regression.ps1 -Dataset All` from the tip - **84 legs, 0 FAIL**. `mode1 (vs golden)`
      green on all four datasets, so the removal moved nothing; `mode10` (the `transfer` arm,
      the only leg executing the changed `ComputePass2Resident`) green; `mode11` green
      independently on StellarLibDecoy, StellarGenDecoyEntrap and Astral
- [x] `Build-Osprey.ps1 -RunTests -RunInspection` - 604 tests, inspection zero-warning
- [x] 446-file pay-later pass-2 fold - exit 0 in 18 min at 14.0 GB managed / 22.1 GB private
- [x] merged current `master`


brendanx67:
Both of Copilot's findings are real, and both are fixed in `cb1d67369a`, which landed after the review ran (it reviewed `1cc6fcc1`).

**1. `transfer` reading `rescored.Value` after the stream started.** Exactly right, and it was the most serious thing on the branch: `ReleaseUnscorableLibraryFragments` streams first and drops the pool, then `ComputeAndPersist` takes the non-frozen path and throws hours into Stage 7. `CanStreamStage7Join` now also requires `protein-compact` — the same predicate `ComputeAndPersist` itself branches on for `frozenCompetition`, so the two move together when `transfer`'s per-run half moves into `Pass2PerFileWorker`. Every other pass-2 mode still computes its per-file half in the join, over the whole pool, and streaming underneath those does not make them per-run — it takes their input away.

**2. mode 10 deleting `OSPREY_PASS2_QVALUE` / `OSPREY_EXPERIMENT_AGG` instead of restoring them.** Also right, and worse than it looks from the diff: mode 10 runs *before* modes 4, 2, 5, 6, 7, 8 and 9 and before every later dataset, so a developer who exported an arm had it silently deleted and the rest of the suite ran the default while the summary reported the arm as passing. It now saves the prior value (including "was not set") and restores it, matching the file's existing `$priorVerifyWorker` / `$script:priorAllowResident` convention. The same commit also records a non-zero exit from an alt arm as a mode-10 failure instead of discarding it.

An independent review pass found the same two, plus five more of the same shape — a source-less milestone that folded 446 empty runs and exited 0 with an empty `.blib`; a write loop that would have overwritten every worker-owned sidecar with `Score = 0.0`; a format-probe-vs-validity-stamp divergence; an unresolved parquet path that gave the streamed arm pre-reconciliation rows; and a `_streamed`-after-`yield` window. All in `cb1d67369a`.

Gates on the fixed code: `regression.ps1 -Dataset All` **78 PASS / 0 FAIL** locally, and TeamCity Perf/Regression **78 PASS / 0 FAIL / 0 SKIP in 01:19:30** on `pull/4642` (that build was on `1cc6fcc1`, so it predates these fixes).

brendanx67:
Thanks - all three verified against the code rather than taken at face value. Two are real
and pre-date this PR's diagnostics work; one is already answered by a deliberate design
choice. None are changed here, and the reasons differ per item.

**1. `PerFileRescoreTask.cs:2114` / `WorkerOwnedPass2Sidecars` staleness — not changing.**

Checking presence of the producer's stamp rather than recomputing the producer's key is
deliberate, and the comment at `Pass2FdrSidecar.cs:842-857` records why: recomputing
`PerFileRescoring`'s key from a different leg was wrong and broke exactly where it mattered.
`PerFileRescoreTask.ValidityKey` folds in `LibraryFragmentRelease.ValidityKeySuffix`, which
asks `RunsOnThisLeg(ctx)` -> `ExpectReconciledInput`, a PER-LEG flag - so a
`--task SecondPassFDR` process and a `--task PerFileRescoring` process compute different
keys for the same task. In an HPC chain that made `IsValid` say "not valid", Stage 7
conclude no worker had run, and recompute and rewrite every sidecar.

Staleness is covered by the producing task's own validity, not by this check. The residual
case you name - a *truncated* binary whose stamp survives - is precluded by the write
discipline rather than by luck: every sidecar is committed through `FileSaver`, an atomic
temp + rename, so a crash mid-write leaves the temp file and the target is absent or
complete (`FdrScoresSidecar.cs:458-474`). An `IsCurrentFormat` filter here would re-open
every worker file on every leg to defend against a state the atomic write already prevents.

**2. `FdrScoresSidecar.cs:805` / `TryWalkRecords` partial overlay — agreed, and already
tracked.**

Correct, and it is recorded as an open item on this branch's TODO: the OOM-filtered catch
was widened (in cherry-picked `4b9df2a836`) to enclose the whole walk, so a mid-body IO
fault returns `false` with records already overlaid, and the caller treats `false` as
non-fatal. The sibling `RestorePass1Scalars` stages into a buffer and applies only `if (ok)`
for precisely this reason.

Not fixed in this PR: it is a shared IO path with several callers, it pre-dates this work,
and the PR is at its merge decision. It wants its own change with its own gate rather than a
late edit here.

**3. `Pass2FdrSidecar.cs:695-697` / experiment records resolved per run — diagnosis correct,
but the naive fix is a correctness bug.**

The per-call resolution is deliberate and documented at `Pass2FdrSidecar.cs:682-687`:
`ResolvePass2ExperimentRecords` answers from the in-memory accumulator once a pass-2 path
has published one and from the on-disk sidecar otherwise, and **Stage 7 crosses that
boundary partway through** - protein FDR writes the sidecar. Hoisting the call out of the
lambda would freeze the pre-competition values into every later fold. Your own suggestion
(cache per sidecar generation, refresh after protein FDR rewrites it) is the right shape;
a plain hoist is not.

Scope check on this PR specifically: the two stream passes this PR adds run at the END of
Stage 7, after the pass-2 competition has published the scope, so they take the in-memory
branch - a `TryGet` per run, not a deserialize. Confirmed on the 446-run cohort, which wrote
`out.2nd-pass.fdr_experiment.bin` during the competition and completed in 69 min at 18.2 GB
managed. The O(runs x sidecar) case you describe is the resume path where the competition is
skipped and no scope is ever published; real, but not reached by this change and not
introduced by it.

Items 2 and 3 are recorded for follow-up rather than dropped.

PR #4655, opened 2026-09-10
skyline: Fixed intermittent TestNativeMessageBox failure on TeamCity

nickshulman:
## Summary

Fixes the intermittent failure of TestNativeMessageBox on TeamCity (build #21983): "Setting values is not supported for native dialog Dialog:Save As".

* The modern Save dialog is built in stages on the classic common-dialog template. It carries the Open dialog's classic file-name combo (control id 1148) for its first ~50 ms, destroys it, and creates its own file-name Edit (control id 1001) some 150 ms later. In that gap the dialog has no file-name field and no folder tree, so NativeDialog.Classify fell through to the generic message box, which was ready the moment its window was shown and refuses to take a file name. Whether the shell shows the window in that state depends on the machine; it did on the TeamCity agent.
* Changed NativeDialog.Classify to recognize a common file dialog first, by the classic file list (lst1, control id 1120) it carries from milliseconds after its window is created until the window is destroyed, and to report null until its file-name field says whether it is Open or Save.
* Made the generic dialog ready only once it has a visible button, so a dialog the shell has only just created is never reported as a message box.
* Reduced the length of the comments in the native dialog classes, removing those that only said where a member is used or restated the code, and removed two unused helpers.

## Test plan

- [x] TestNativeMessageBox - 3 loops
- [x] TestNativeFileDialog, TestLayoutExportImport, TestLibraryBuild, TestPrmMcpConnector, TestJsonToolServer, TestAlertWatch, TestMcpConnectorBackgroundDialog, TestMcpConnectorDisconnect, TestMcpConnectorFormThreading
- [x] CodeInspection test and quick ReSharper inspection on Skyline and TestFunctional (0 warnings)
- [x] A temporary probe that logged the child windows of Save, Open, multiselect Open, folder browser and message box dialogs every millisecond, with what NativeDialog.Create reported: with the fix, every dialog reports null until it is fully built, then its own kind

https://claude.ai/code/session_01DgaZyWLHomFnU9EdeWbmWh


nickshulman:
Closing this one. Its only CI run failed TestNativeMessageBox in a new way: reporting null for a file dialog until it is fully built drops the dialog from the dialog watcher's window snapshot, so the same dialog is then reported as a new blocking modal ("Namespace Tree Control"). We are going to try fixing the original intermittent failure in another way, on a branch that also keeps the watcher snapshot independent of classification.

PR #4680, opened 2026-09-17
osprey: Fixed parallel parquet writes corrupting the byte-typed charge column

brendanx67:
## Summary

* Staged a rebuilt `ParquetNet.dll` fixing a use-after-return in Parquet.Net's narrow-integer
  encoders: the widening buffer was returned to the `ArrayPool` in a `finally` and then read,
  so once #4652 made columns compress concurrently another column could overwrite it mid-encode
* Added `TestParquetRoundTripScalarStress` as the regression guard - it iterates in-process,
  because a 0.1% round-trip defect is invisible to a single-shot test
* Library-side fix is `2ca33b1` on the `maccoss-developers` `parquet-parallel-compression`
  branch, on the PR already awaiting review with the bool-encoder fix

## What was wrong

```csharp
int[] ints = ArrayPool<int>.Shared.Rent(data.Length);
try {
    for(int i = 0; i < data.Length; i++) ints[i] = data[i];
} finally {
    ArrayPool<int>.Shared.Return(ints);          // returned here...
}
Encode(ints.AsSpan(0, data.Length), destination); // ...then READ after return
```

Benign single-threaded - nothing rents in between, so the contents survive, which is why it was
latent for years. `sbyte`, `short` and `ushort` carry the identical copy-paste and are fixed too.

## Blast radius

* **Osprey only.** `WriteColumnsAsync` (the parallel API) has one caller in pwiz,
  `ParquetScoreCache.cs:770`. Skyline's `ParquetReportExporter.cs:50` uses the stock singular
  `WriteColumnAsync` and is sequential.
* **The `charge` column only.** Corruption is confined to `byte`/`sbyte`/`short`/`ushort`;
  Osprey's schema has exactly one such column. Everything else is uint, double, string, bool or
  a `byte[]` blob.
* Present since 2026-09-10 (#4652).

## Measured

20,000 write/read round-trips of one 7-row fixture per arm:

| arm | corrupted round-trips |
|---|---|
| parallel (default DOP) | **27 / 20,000** |
| serialized (`OSPREY_PARQUET_WRITE_THREADS=1`) | 0 / 20,000 |
| parallel, after the fix | 0 / 20,000, and **0 / 100,000** |

It surfaced as a ~2% intermittent failure across four Osprey unit tests. Because `charge` is part
of the `(entry_id, charge, scan_number)` identity, a corrupted value does not just lose a field -
the row fails identity matching downstream and is dropped.

## Upstream

Present in upstream `master` **and** `6.1.0` (`ParquetPlainEncoder.cs:580`), verified by fetching
the source; there is no `EncodeHwx` bypass for these types (that exists only for `bool`). Unlike
the bool-encoder byte count, this one is worth reporting upstream - and it corrects the fork's
note that 6.1.0 makes parallel writing safe.

## Test plan

- [x] `TestParquetRoundTripScalarStress` - new regression guard
- [x] Osprey pre-commit gate: 596 tests, zero inspection warnings
- [x] 100,000-iteration A/B, parallel and serialized, both clean after the fix
- [x] Full `regression-parallel.ps1 -Dataset All`: **70 PASS / 0 FAIL / 0 SKIP** (52:04). Goldens unchanged, and the new guard did not fire on any of the four beds


chambm:
> it was latent for years

🤣 

What version of ParquetNet? Looks like the parquet Spectronaut just sent to us requires 5.2. Can't open on 4.25. I could go back to them and tell them to fix that though.

